PR #1, opened 2021-06-11
Add first draft of a build pipeline

jfversluis:
Adds a crude version of a build pipeline. Right now builds on PRs and just builds the libraries on Windows and Mac and publishes unsigned NuGets.

TODO after this and/or if someone has solutions I missed at this time:
- Make unit tests work, I couldn't seem to be able to
- Build sample apps
- We will need to add publishing signed NuGets for actual releases


michael-hawker:
Going to try and re-trigger with close/open and see what might be going on, as I can't see connection from DevOps side even for me.

azure-pipelines.yml
```diff
@@ -0,0 +1,166 @@
...
+      - powershell: |
+          $buildSourceBranch = "$(Build.SourceBranch)"
+          $tagVersion = $buildSourceBranch.Substring($buildSourceBranch.LastIndexOf("/") + 1)
+          Write-Host("Branch = $buildSourceBranch, Version = $tagVersion");
+          Write-Host ("##vso[task.setvariable variable=NugetPackageVersion;]$tagVersion")
+        displayName: Set NuGet Version to Tag Number
+        condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/')
+      # restore, build and pack the packages
+      - task: CmdLine@2
+        displayName: Build CommunityToolkit.Maui.csproj
```
michael-hawker, line 86:
```suggestion
        displayName: 'Build Community Toolkit'
```
Making consistent with the lower job, also just needed to trigger the CI... 😋

michael-hawker:
Going to try committing an empty pipeline file into main branch if that at least makes this show up on the DevOps side.

michael-hawker:
@jfversluis looks like there's still references in the pipeline config to the `xamarin` org, so you'll need to fix those:

image

azure-pipelines.yml
```diff
@@ -1,19 +1,127 @@
-# Starter pipeline
-# Start with a minimal pipeline that you can customize to build and deploy your code.
-# Add steps that build, run tests, deploy, and more:
-# https://aka.ms/yaml
+variables:
+  CurrentSemanticVersionBase: '7.0.0'
```
jfversluis, line 2: I think we wanted to match the current Windows Community Toolkit versioning and thus start at 7?

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="MSBuild.Sdk.Extras/2.1.2">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0-ios;net6.0-android;net6.0-maccatalyst</TargetFrameworks>
+    <TargetFrameworks>net6.0-ios14.0;net6.0-android;net6.0-maccatalyst13.5</TargetFrameworks>
```
jfversluis, line 5: No sure about these version numbers here?

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -23,7 +23,8 @@
     <UseFullSemVerForNuGet>false</UseFullSemVerForNuGet>
     <Title>CommunityToolkit.Maui</Title>
     <Description>The .NET MAUI Community Toolkit is a collection of Animations, Behaviors, Converters, and Effects for development with .NET MAUI. It simplifies and demonstrates common developer tasks building iOS, Android, macOS and Windows apps with .NET MAUI.</Description>
-    <PackageIcon>icon.png</PackageIcon>
+    <!-- TODO: Get a new amazing icon! -->
+    <!--<PackageIcon>icon.png</PackageIcon>-->
```
jfversluis, line 27: WE NEED AN ICON!

jfversluis:
Unit tests seem to be unable to run at this time. Added to an issue here: https://github.com/microsoft/vstest/issues/2985

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
michael-hawker: We have our general Toolkit icon from the WCT repo, feel free to just copy that 😉 https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/main/build/nuget.png

It's the B/W Toolbox image that we use as the 'Toolkit' icon and that I made the organization icon.

azure-pipelines.yml
michael-hawker: Up to you folks, I know we're probably going to be on different shipping cycles, so not sure how feasible it's going to be to always ship in-sync?

We're currently on 7.0 released and moving towards our 7.1 release for this month. Probably plan a 7.2 in time for .NET 6/WinUI 2.7/WinUI 3 1.0 though no idea if all those milestones align 🤷‍♂️? Lots of moving parts and dependencies now! 😋

I know after the 7.1 for WCT we'll start splitting out the .NET Community Toolkit bits to their own repo in September-ish, so that'd be part of that 7.2 drop as well is that being independent.

Love to know your thoughts. If you think you'll be snapping to WinUI 3 releases, then that does make it a bit easier assuming WinUI 2.x and WinUI 3 releases are relatively synced in terms of scheduling in the future (as we'll be updating the WCT more often to match the WinUI 2.x releases moving forward now that we take them as a dependency).

michael-hawker:
Looks like the pipeline is working now! 🎉🎉🎉 That's awesome!

@jfversluis did you figure out how to setup the Release pipeline too? (At least push to the `MainLatest` feed?)

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: Do we need the specific iOS / macOS OS version numbers?

If not, let's leave them off for now since .NET MAUI only doesn't specify an OS:
```csharp
net6.0-ios;net6.0-maccatalyst;net6.0-android
```
https://github.com/dotnet/maui/blob/cadf5472f5d3d4a1201928b54c05a81a614fa3ba/Directory.Build.props#L8

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
jfversluis: Added that for now!

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
jfversluis: Yeah that's the funny thing. They don't specify it, but when I don't it starts complaining that I have to 🤷

azure-pipelines.yml
jfversluis: Definitely true. We might want to discuss this a bit further. @brminnick any thoughts on this as well? Maybe we should follow up on Teams

jfversluis:
> Looks like the pipeline is working now! 🎉🎉🎉 That's awesome!
> 
> @jfversluis did you figure out how to setup the Release pipeline too? (At least push to the `MainLatest` feed?)

image

Created the release, can't enable the continuous one though. I have been able to push a NuGet to that feed manually :) I guess we want to hold off on that daily until we figure out the version numbers

azure-pipelines.yml
brminnick: I think it's best to keep our versioning separate. 

Since the MAUI Toolkit doesn't have a dependency on the Windows Toolkit, there's not much need for them to keep their versions in lock-step. Keeping the version numbers the same would just add more overhead for both toolkits because each time we release a new version of the Maui Toolkit, the Windows Toolkit would need to release a new version and vice-versa.

We kept the Xamarin.CommunityToolkit NuGet Version in lock-step with the Xamarin.CommunityToolkit.Markdown NuGet and we ended up releasing "new versions" of the Markup NuGet that were exactly the same as the previous version (no code in the Markup library had changed).

I vote that we start both `CommunityToolkit.Maui` `CommunityToolkit.Maui.Markup` at v1.0.0-pre1 and let them increase their version numbers independently from each other and independently from Windows Toolkit's versioning.

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: Ah - if we're getting build errors, then let's keep the OS numbers 💯 

According to the docs, .NET MAUI supports iOS 10+ and macOS 10+: https://docs.microsoft.com/dotnet/maui/supported-platforms

image

Supporting iOS 10+ is surprising to me. I didn't expect it to go that far back.

jfversluis: I think we might need to dig into this a bit deeper though, because I think the things I set now are the _target_ versions, not the minimum versions.

azure-pipelines.yml
jfversluis: Makes sense to me! Reset it 1 for now, we can always decide otherwise before we merge 😬

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: I [confirmed with the engineering team](https://teams.microsoft.com/l/message/19:989ffa44998147aca4ceaf7482967668@thread.skype/1628880426523?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=0056f60b-301f-43ac-bbcf-f356d3c42c92&parentMessageId=1628875764413&teamName=VS%20Client%20Experiences&channelName=MAUI%20%F0%9F%8C%BA&createdTime=1628880426523) that these are the current targets for Maui:
- iOS 10+
- macOS 10.13+

brminnick: > I think the things I set now are the target versions, not the minimum versions

These look to be the minimum versions. 

👇 Here's what I got when I ran `dotnet pack` to create a NuGet Package using `<TargetFrameworks>net6.0-ios14.0;net6.0-android;net6.0-maccatalyst13.5</TargetFrameworks>`:
image

jfversluis: That's the thing, I don't think that's the case. I think it's more like the Android way: there target version specifies which APIs you can use, but it's not the minimum version that you can run it on. Does that make sense?

brminnick: And here's the errors I got when I tried to add this NuGet Package to a new Maui App that was using lower OS versions:

### Maui Sample App Versions:
```xml
<TargetFrameworks>net6.0-ios13.0;net6.0-android29.0;net6.0-maccatalyst13.4</TargetFrameworks>
```

### NuGet Errors
image

brminnick: ## tl;dr

Here's what we need to set our `TargetFrameworks` to:

```xml
<TargetFrameworks>net6.0-ios10.0;net6.0-android21.0;net6.0-maccatalyst10.13</TargetFrameworks>
```

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="MSBuild.Sdk.Extras/2.1.2">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0-ios;net6.0-android;net6.0-maccatalyst</TargetFrameworks>
+    <TargetFrameworks>net6.0-ios14.0;net6.0-android30.0;net6.0-maccatalyst13.5</TargetFrameworks>
```
brminnick, line 5: We need these to adhere with .NET MAUI:

```xml
<TargetFrameworks>net6.0-ios10.0;net6.0-android21.0;net6.0-maccatalyst10.13</TargetFrameworks>
```

brminnick (changes requested):

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: See this conversation for more details: https://github.com/CommunityToolkit/Maui/pull/1#discussion_r688727400

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: > target version specifies which APIs you can use, but it's not the minimum version

It's both, right? Like targeting .NET Standard 2.0 means that I cannot use the `IAsyncEnumerable` because that is only available in .NET Standard 2.1. It also means that anybody targeting .NET Standard 2.0+ can use the library.

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
jfversluis: For iOS and Android that might work, for maccatalyst it does not: https://dev.azure.com/dotnet/CommunityToolkit/_build/results?buildId=56105&view=logs&j=c44e3ce6-80b8-5090-9435-9627042472dc&t=01dd27be-fb26-56b1-ffa1-33467f7cc411

jfversluis: I think the catalyst version number != mac OS version number, to make it even more confusing :D

brminnick: Ah - maccatalyst version matches the iOS version, not the macOS version

https://stackoverflow.com/a/63598566/5953643

brminnick: 👇 Let's try this instead
```xml
<TargetFrameworks>net6.0-ios10.0;net6.0-android21.0;net6.0-maccatalyst10.0</TargetFrameworks>
```

brminnick: Wait - what? This says that `net6.0-android30.0` is the _only valid version_?

image

brminnick: Update from the Engineering team: "Don't use version numbers"

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: What errors are we getting when we omit the iOS/Android/MacCatalyst version numbers?

After speaking with the engineering team, they recommend that we don't include version numbers.

brminnick: Got it! 

We need to remove both the OS version numbers AND `RuntimeIdentifier`, `InvariantGlobalization` and `UseInterpreter` from the csproj. I'll do that now 👍

brminnick (approved):

michael-hawker:
@jfversluis I've added you as a Project Administrator in DevOps, that should hopefully allow you to fix the settings in the Release Pipeline now?

PR #5, opened 2021-07-29
SnackBar

### LINKED ISSUE #118: [Proposal] Snackbar
# Snackbar

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample

## Summary
[summary]: #summary

Adds a snackbar.
It has to use Native Android Snackbar (`Google.Android.Material.Snackbar`) internally.
All other platforms implement a default popup container with text and a button.

## Detailed Design
[design]: #detailed-design

### Snackbar.shared.cs

```cs
namespace CommunityToolkit.Maui.Controls.Snackbar;

public class Snackbar : ISnackbar
{
	Action Action { get; set; }
	string ActionButtonText { get; set; }
	IView? Anchor { get; set; }
	TimeSpan Duration { get; set; }
	bool IsShown { get; }
	string Text { get; set; }
	SnackbarOptions VisualOptions { get; set; }

	event EventHandler? Dismissed;
	event EventHandler<ShownEventArgs>? Shown;

	Task Dismiss();
	Task Show();

    public static ISnackBar Make();
}

public interface Snackbar
{
	Action Action { get; }
	string ActionButtonText { get; }
	IView? Anchor { get;  }
	TimeSpan Duration { get; }
	bool IsShown { get; }
	string Text { get; }
	SnackbarOptions VisualOptions { get; }

	event EventHandler? Dismissed;
	event EventHandler<ShownEventArgs>? Shown;

	Task Dismiss();
	Task Show();
}

public class ShownEventArgs : EventArgs
{
	public ShownEventArgs(bool isShown)
	{
		IsShown = isShown;
	}

	public bool IsShown { get; }
}

public class SnackbarOptions : ITextStyle
{
	public double CharacterSpacing { get; set; } = 0.0d;

	public Font Font { get; set; } = Font.SystemFontOfSize(14);

	public Color TextColor { get; set; } = Colors.Black;

	public Color ActionTextColor { get; set; } = Colors.Black;

	public Color BackgroundColor { get; set; } = Colors.LightGray;

	public CornerRadius CornerRadius { get; set; } = new CornerRadius(4, 4, 4, 4);
}
```

## Visual with default settings

Android:
image

VladislavAntonyuk:
Fix #118 
|     Platform    |  Default setings  |  Custom visual options  |
|---------|---|---|
| Android | image | image |
| iOS/MacCatalyst     |  image  |   image |


src/CommunityToolkit.Maui/Platforms/Android/Views/SnackBar.cs
```diff
@@ -0,0 +1,156 @@
+﻿#if ANDROID
```
maxkoshevoi, line 1: 1) Since this file is under `Platforms/Android`, it shouldn't need `#if ANDROID` (if I understand MAUI correctly)
2) Is XCT gonna use `Platforms` folder to separate code for different platforms instead of doing that based on a file name? I liked the original approach where all code for a feature was in the same place.

CC: @jfversluis

pictos:
Hey there, please wait until we define what we will migrate to this repo and how. If I'm not wrong this repo isn't even with the XCT

src/CommunityToolkit.Maui/Platforms/Android/Views/SnackBar.cs
VladislavAntonyuk: 1. I expected the same, but it doesn't work on my device. Maybe because this is a Class library project and not a Console application.
2. I don't think we should change it in XCT. As for Maui separation by platforms folders out of the box it make sense here.

VladislavAntonyuk:
@pictos, @jfversluis, @brminnick please review.

maxkoshevoi:
Hi @jfversluis, sorry about that! 😅
Will use it from now on.

VladislavAntonyuk:
@brminnick Could you please update the task requirements according to the discussion (Use native Snackbar/toast for Android and popup for other platforms)

brminnick:
Thanks for the reminder! I'll update the [detailed design](https://github.com/CommunityToolkit/Maui/issues/118) this weekend.

Just curious - would you then be closing this PR and splitting it into two separate PRs; one for SnackBar and one for Toast?

VladislavAntonyuk:
actually, snack bar and toast only differ with the action button. maybe I will update this PR with the snack bar implementation only and then add toast.

src/CommunityToolkit.Maui/Extensions/VisualElementExtension.shared.cs
```diff
@@ -0,0 +1,140 @@
...
+
+	/// <summary>
+	/// Display toast with the custom options
+	/// </summary>
+	/// <param name="visualElement">Toast anchor</param>
+	/// <param name="toastOptions">Toast options</param>
+	/// <returns>Task</returns>
+	public static async Task DisplayToastAsync(this VisualElement visualElement, ToastOptions toastOptions)
+	{
+		_ = visualElement ?? throw new ArgumentNullException(nameof(visualElement));
```
brminnick, line 47: Let's use `ArgumentNullException.ThrowIfNull` 💯

src/CommunityToolkit.Maui/Extensions/VisualElementExtension.shared.cs
```diff
@@ -0,0 +1,140 @@
...
+	/*/// <summary>
+	/// Display toast with the default visual configuration
+	/// </summary>
+	/// <param name="visualElement">Toast anchor</param>
+	/// <param name="message">Toast text</param>
+	/// <param name="durationMilliseconds">Toast duration (milliseconds)</param>
+	/// <returns>Task</returns>
+	public static async Task DisplayToastAsync(this VisualElement visualElement, string message, int durationMilliseconds = 3000)
+	{
+		_ = visualElement ?? throw new ArgumentNullException(nameof(visualElement));
```
brminnick, line 25: Let's use `ArgumentNullException.ThrowIfNull` 💯

src/CommunityToolkit.Maui/Extensions/VisualElementExtension.shared.cs
```diff
@@ -0,0 +1,140 @@
...
+	/// </summary>
+	/// <param name="visualElement">Anchor element</param>
+	/// <param name="message">Text of the snackbar</param>
+	/// <param name="actionButtonText">Text of the snackbar button</param>
+	/// <param name="action">Action of the snackbar button</param>
+	/// <param name="duration">Snackbar duration</param>
+	/// <returns>True if snackbar action is executed. False if snackbar is closed by timeout</returns>
+	public static async Task<bool> DisplaySnackBarAsync(this VisualElement visualElement, string message, string actionButtonText, Func<Task> action, TimeSpan? duration = null)
+	{
+		_ = visualElement ?? throw new ArgumentNullException(nameof(visualElement));
```
brminnick, line 75: Let's use `ArgumentNullException.ThrowIfNull` 💯

src/CommunityToolkit.Maui/Extensions/VisualElementExtension.shared.cs
```diff
@@ -0,0 +1,140 @@
...
+
+	/// <summary>
+	/// Display snackbar with custom configuration
+	/// </summary>
+	/// <param name="visualElement">Snackbar anchor</param>
+	/// <param name="snackBarOptions">Snackbar options</param>
+	/// <returns>True if snackbar action is executed. False if snackbar is closed by timeout</returns>
+	public static async Task<bool> DisplaySnackBarAsync(this VisualElement visualElement, SnackBarOptions snackBarOptions)
+	{
+		_ = visualElement ?? throw new ArgumentNullException(nameof(visualElement));
```
brminnick, line 107: Let's use `ArgumentNullException.ThrowIfNull` 💯

src/CommunityToolkit.Maui/Views/Snackbar/Helpers/NativeRoundedStackView.ios.macos.cs
```diff
@@ -0,0 +1,56 @@
...
+			Right = right;
+			Bottom = bottom;
+		}
+
+		public override void Draw(CGRect rect)
+		{
+			ClipsToBounds = true;
+			var path = GetRoundedPath(rect, Left, Top, Right, Bottom);
+			var maskLayer = new CAShapeLayer { Frame = rect, Path = path };
+			Layer!.Mask = maskLayer;
```
brminnick, line 31: Let's avoid using the Null Forgiving Operator, `!.`. 

It opens the door to future bugs in future code changes.

brminnick (changes requested):
Thanks Vlad!

Just some quick semantic feedback since it looks like you're still working on it 🙌

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -67,4 +67,8 @@
     <RuntimeIdentifier>win-x64</RuntimeIdentifier>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$(TargetFramework.Contains('-android'))">
```
VladislavAntonyuk, line 70: My android device is pretty old, and the app doesn't start without it.

samples/CommunityToolkit.Maui.Sample/Pages/Views/SnackBarPage.xaml.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace CommunityToolkit.Maui.Sample.Pages.Views;
+
+public partial class SnackBarPage : BasePage
+{
+	private Snackbar? snackbarWithAnchor;
+	public SnackBarPage()
+	{
+		InitializeComponent();
+		this.Anchor1 ??= new Microsoft.Maui.Controls.Button();
```
VladislavAntonyuk, line 21: I don't like that we have to add it for each control, because we know that after InitializeComponent() the control cannot be null. are there any hacks to avoid it?

brminnick: It's the only option we have right now. 

The MAUI team is aware of it: https://github.com/dotnet/maui/issues/1796#issuecomment-946348892

brminnick: You can just use `= new()` to save some keystrokes

VladislavAntonyuk:
@brminnick what do you think if we split this feature on parts based on Platforms? for example this PR is only for Android. In that case we can deliver the feature earlier. possibly receive some feedback.

src/CommunityToolkit.Maui/Views/BasePopupView.shared.cs
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	public event EventHandler<ShownEventArgs>? Shown;
+	public event EventHandler? Dismissed;
+
+	public abstract Task Show();
+	public async Task Dismiss()
+	{
+		await DismissInternal();
+		Dismissed?.Invoke(this, new EventArgs());
```
pictos, line 39:
```suggestion
		Dismissed?.Invoke(this,  EventArgs.Empty);
```

src/CommunityToolkit.Maui/Views/BasePopupView.shared.cs
```diff
@@ -0,0 +1,58 @@
...
+	internal abstract Task DismissInternal();
+
+	private static void OnIsShownChanged(BindableObject bindable, object oldValue, object newValue)
+	{
+		var popup = (BasePopupView)bindable;
+		popup.Shown?.Invoke(popup, new ShownEventArgs(popup.IsShown));
+	}
+}
+
+public class ShownEventArgs : EventArgs
```
pictos, line 50: Can we move this to his own file?

src/CommunityToolkit.Maui/Views/Snackbar/SnackBar.shared.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Threading.Tasks;
+using CommunityToolkit.Maui.Views;
+using Microsoft.Maui;
+using Microsoft.Maui.Controls;
+using Microsoft.Maui.Graphics;
+
+
+namespace CommunityToolkit.Maui.Controls.Snackbar;
+
+public class Snackbar : BasePopupView, IText
```
pictos, line 11: Since we're building a library, I think we should implement our views on top of interfaces, like .net Maui does.

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -49,7 +49,7 @@
     <UseInterpreter Condition="'$(Configuration)' == 'Debug'">True</UseInterpreter>
 
     <SupportedOSPlatformVersion Condition="'$(TargetFramework)' == 'net6.0-ios'">14.2</SupportedOSPlatformVersion>
-    <SupportedOSPlatformVersion Condition="'$(TargetFramework)' == 'net6.0-maccatalyst'">14.0</SupportedOSPlatformVersion>
+    <SupportedOSPlatformVersion Condition="'$(TargetFramework)' == 'net6.0-maccatalyst'">14.2</SupportedOSPlatformVersion>
```
pictos, line 52: Just out of curiosity, why did you change this value?

pictos (commented):
@VladislavAntonyuk Do you plan to implement it using the handler approach? Using property mapper and all that jazz?

brminnick:
> @brminnick what do you think if we split this feature on parts based on Platforms? for example this PR is only for Android.

I think it's important to include all platforms in the pull request. Excluding platforms adds complexity and confusion for the users. 

> Do you plan to implement it using the handler approach?

💯 All custom controls in the .NET MAUI toolkit must be done using Handlers. We made a commitment to the community that we would always use Handlers and no Renderers.

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
VladislavAntonyuk: just played with warnings. I will revert it

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -6,7 +6,7 @@
     <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) and '$(MSBuildRuntimeType)' == 'Full'">$(TargetFrameworks);net6.0-windows10.0.17763.0</TargetFrameworks>
     <UseMaui>true</UseMaui>
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
```
VladislavAntonyuk, line 9: because this PR is a draft for now, I disabled it

src/CommunityToolkit.Maui/Views/Popup/SnackBar/Platforms/PlatformPopupExtensions.windows.cs
```diff
@@ -1,14 +1,39 @@
...
+	public ToastNotification Show(Snackbar snackBar)
 	{
-		return new object();
+		var toastContentBuilder = new ToastContentBuilder()
+			.AddText(snackBar.Text)
+			.AddButton(new ToastButton() { ActivationType = ToastActivationType.Foreground }.SetContent(snackBar.ActionButtonText));
+		var toastContent = toastContentBuilder.GetToastContent();
+		toastContent.ActivationType = ToastActivationType.Background;
+		var toast = new ToastNotification(toastContent.GetXml());
+		toast.Activated += delegate (ToastNotification sender, object args)
```
pictos, line 27: do we want to unsubscribe from this event?

src/CommunityToolkit.Maui/Views/Popup/SnackBar/Platforms/PlatformPopupExtensions.windows.cs
```diff
@@ -1,14 +1,39 @@
...
+		{
+			Microsoft.Maui.Controls.Device.BeginInvokeOnMainThread(snackBar.Action);
+		};
+		toast.Dismissed += delegate (ToastNotification sender, ToastDismissedEventArgs args)
```
pictos, line 31: do we want to unsubscribe from this event?

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -69,6 +69,7 @@
   <ItemGroup Condition=" $(TargetFramework.Contains('-windows')) ">
     <Compile Include="**\*.windows.cs" />
     <Compile Include="**\*.windows.*.cs" />
+    <PackageReference Include="Microsoft.Toolkit.Uwp.Notifications" Version="7.1.2" />
```
pictos, line 72: I see `UWP` on this package, do you know if this is compatible with MAUI. AFAIK MAUI will not be compatible with anything that has `UWP` just `winUI`

pictos (changes requested):
Hey @VladislavAntonyuk some comments in the last commit

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
VladislavAntonyuk: starting from version 7 it supports WinUI

src/CommunityToolkit.Maui/Views/Popup/SnackBar/SnackBar.shared.cs
```diff
@@ -58,21 +63,25 @@ public static bool IsShown
...
+
+			if (value)
+				Shown?.Invoke(null, new ShownEventArgs(isShown));
 		}
 	}
 
 	/// <inheritdoc/>
 	public IView? Anchor { get; set; }
 
+	internal NativeSnackbar? NativeSnackbar { get; set; }
```
VladislavAntonyuk, line 75: it should be private set

brminnick:
Thanks @VladislavAntonyuk! I'm going to take a crack at rearchitecting this to remove the need for `IPlatformPopupExtensions`, and instead use partial classes for the platform-specific implementation. This'll should allow us to consolidate some code and remove a layer of abstraction.

If I'm able to do it, I think it'll be a good model for future controls to follow. I'll keep you posted!

pictos:
@brminnick I'll wait for your push before I move forward with the Popup implementation. From that I believe that we will end up in a solid base for implement new controls

VladislavAntonyuk:
I also kept that in mind, but thought about IPlatformPopupExtensions like IButtonHandler. If you think partial classes should be enough I am OK with that.

samples/CommunityToolkit.Maui.Sample/Platforms/Android/MainActivity.cs
```diff
@@ -1,10 +1,17 @@
...
 using Android.Content.PM;
+using Android.OS;
 using Microsoft.Maui;
 
 namespace CommunityToolkit.Maui.Sample;
 
 [Activity(Theme = "@style/Maui.SplashTheme", MainLauncher = true, ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation | ConfigChanges.UiMode | ConfigChanges.ScreenLayout | ConfigChanges.SmallestScreenSize)]
 public class MainActivity : MauiAppCompatActivity
 {
+	// https://github.com/dotnet/maui/pull/3345
```
brminnick, line 11: Appreciate the update, but let's remove the comment/link from the source code

src/CommunityToolkit.Maui.UnitTests/Alerts/SnackBar_Tests.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CommunityToolkit.Maui.Alerts.Snackbar;
+using Xunit;
+
+namespace CommunityToolkit.Maui.UnitTests.Alerts;
+
+public class Snackbar_Tests : BaseTest
```
brminnick, line 9: Let's make sure we cover the entire `ISnackbar` and `Snackbar` APIs.

Here's the list of APIs that we still need tests for:
- `ISnackbar.Action`
- `ISnackbar.ActionButtonText`
- `ISnackbar.Anchor`
- `ISnackbar.Duration`
- `ISnackbar.Text`
- `ISnackbar.VisualOptions`
- `Snackbar.Make`
- `DisplaySnackbar()`

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Toolkit.Uwp.Notifications;
+using Windows.UI.Notifications;
+
+namespace CommunityToolkit.Maui.Alerts.Snackbar;
+
+public partial class Snackbar
```
brminnick, line 8: Can you ensure this code works on Windows?

When I run this in the Sample App on Windows, I don't get a toast or a snackbar. Nothing happens.

ScreenFlow

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -136,4 +143,28 @@ void OnDismissed()
 		IsShown = false;
 		_weakEventManager.HandleEvent(this, EventArgs.Empty, nameof(Dismissed));
 	}
+}
+
+/// <summary>
+/// Extension methods for <see cref="VisualElement"/>.
+/// </summary>
+public static class VisualElementExtension
```
VladislavAntonyuk, line 151: let's move it back to VisualElementExtensions. We may call Toast and other Anchored Controls from it

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -91,14 +91,22 @@ public Snackbar()
 	/// </summary>
 	/// <returns></returns>
 	/// <exception cref="NotSupportedException"></exception>
```
VladislavAntonyuk, line 93: exception is not thrown here

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -91,14 +91,22 @@ public Snackbar()
...
+	{
+		OnShown();
+		return Task.CompletedTask;
+	}
 
 	/// <summary>
 	/// Dismiss Snackbar
 	/// </summary>
 	/// <returns></returns>
 	/// <exception cref="NotSupportedException"></exception>
```
VladislavAntonyuk, line 104: exception is not thrown here

src/CommunityToolkit.Maui/Views/Popup/Popup.ios.macos.cs
```diff
@@ -0,0 +1,53 @@
...
+
+	NSTimer? timer;
+
+	public TimeSpan Duration { get; set; }
+
+	public UIView? Anchor { get; set; }
+
+	public CGRect CornerRadius { get; set; }
+
+	public UIColor BackgroundColor { get; set; } = UIColor.Gray;
```
brminnick, line 26: Let's move`BackgroundColor` to `PopupView`.

I'd prefer for `Popup` to contain the functionality (e.g. Anchor, Timer, etc.) and `PopupView` to contain the UI elements .

src/CommunityToolkit.Maui/Views/Popup/Popup.ios.macos.cs
```diff
@@ -0,0 +1,53 @@
...
+		PopupView.ParentView?.BringSubviewToFront(PopupView);
+	}
+
+	NSTimer? timer;
+
+	public TimeSpan Duration { get; set; }
+
+	public UIView? Anchor { get; set; }
+
+	public CGRect CornerRadius { get; set; }
```
brminnick, line 24: Let's move`CornerRadius` to `PopupView`.

I'd prefer for `Popup` to contain the functionality (e.g. Anchor, Timer, etc.) and `PopupView` to contain the UI elements .

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -136,4 +143,28 @@ void OnDismissed()
...
+	/// Display snackbar with the anchor
+	/// </summary>
+	/// <param name="visualElement">Anchor element</param>
+	/// <param name="message">Text of the snackbar</param>
+	/// <param name="actionButtonText">Text of the snackbar button</param>
+	/// <param name="action">Action of the snackbar button</param>
+	/// <param name="duration">Snackbar duration</param>
+	/// <param name="visualOptions">Snackbar visual options</param>
+	/// <returns><see cref="Snackbar"/></returns>
+	public static Task DisplaySnackbar(
```
VladislavAntonyuk, line 163: it is useful to return snackbar instance here. we can later dismiss it, subscribe to events, etc.

brminnick (changes requested):
Thanks Vlad! 

The Windows Sample didn't work on my machine. Can you double check that?

We also need more unit tests to ensure we've covered every public API on `Snackbar` and `ISnackbar`.

I also recommend moving the UI elements in `Popup.macos.ios.cs` to `PopupView.macos.ios.cs`. This'll ensure that `Popup` contains the functionality for the popup (eg timer, duration, show/dismiss), and that `PopupView` contains all of the UI elements.

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.ios.macos.cs
```diff
@@ -77,7 +75,7 @@ static CGRect GetCornerRadius(CornerRadius cornerRadius)
 		return new CGRect(cornerRadius.BottomLeft, cornerRadius.TopLeft, cornerRadius.TopRight, cornerRadius.BottomRight);
 	}
 
-	sealed class AppleSnackbar : NativeToast, IDisposable
+	sealed class AppleSnackbar : Views.Toast.Toast, IDisposable
```
VladislavAntonyuk, line 78: let's rename it to AppleToast

src/CommunityToolkit.Maui/Views/Popup/SnackBar/ISnackbar.shared.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Maui;
+
+namespace CommunityToolkit.Maui.Views.Popup.SnackBar;
```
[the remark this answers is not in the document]
brminnick: FYI - I went ahead and moved this to `CommunityToolkit.Maui.Alerts.Snackbar`.

This better conveys to the users that `Snackbar` is to be used as an alert, like `DisplayAlert`, and not, for example, added to a Layout in XAML.

src/CommunityToolkit.Maui/Views/Popup/Popup.ios.macos.cs
```diff
@@ -5,9 +5,9 @@
 
 namespace CommunityToolkit.Maui.Views.Popup;
 
-class NativePopup
+class Popup
```
VladislavAntonyuk, line 8: Let's rename it to ApplePopup

src/CommunityToolkit.Maui/Views/Toast/Toast.ios.macos.cs
```diff
@@ -4,15 +4,15 @@
 using Foundation;
 using UIKit;
 
-namespace CommunityToolkit.Maui.Views.Popup.Snackbar;
+namespace CommunityToolkit.Maui.Views.Toast;
 
-class NativeToast : NativePopup
+class Toast : Popup.Popup
```
VladislavAntonyuk, line 9: AppleToast

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs
VladislavAntonyuk: in Windows, Toast is like a notification. Please disable silent mode

samples/CommunityToolkit.Maui.Sample/Platforms/Android/MainActivity.cs
VladislavAntonyuk: let's keep it while the issue exists. we can remove the whole method later

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs
brminnick: Ah - user error! 🤦‍♂️ 

I wish I would've asked you sooner before spending an hour debugging it 😆

brminnick: After disabling silent mode, I get a weird behavior:
When I click `OK`, a new instance of the app appears. 

Do you know why this is happening? 

ScreenFlow

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
brminnick: I prefer to keep it in `Snackbar` for two reasons:
- It  keeps the extension method close to the `Snackbar` code
  - In XCT, we had a few times where a bug was fixed in the base code, but the extension method was forgotten about.
- `VisualElementExtensions` becomes a dumping ground
  - Because `VisualElement` encompasses so many things in MAUI, it quickly became a huge file of 100s of lines of code

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
brminnick: No, let's continue to treat `DisplaySnackbar` like `DisplayAlert`.

If a user wants to do more with a Snackbar, they can use `new Snackbar()`.

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs
VladislavAntonyuk: replied in Discord, but it works like this: https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=desktop#step-3-handling-activation

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.ios.macos.cs
brminnick: My preference to keep it as `Toast` in case we decide to add a `Toast` control to Maui Community Toolkit in the future. 

Then, similar to the Snackbar architecture, we'd have `Toast.shared.cs` `Toast.ios.macos.cs`, `Toast.android.cs`, and `Toast.windows.cs`.

The filename is already `Toast.ios.macos.cs` which means it can only run on iOS + MacCatalyst anyways.

src/CommunityToolkit.Maui/Views/Popup/Popup.ios.macos.cs
brminnick: I prefer to keep it as `Popup` for a couple reasons:
- Pedro is currently working on the Popup control, and he'll be able to leverage this code
- The file name is `Popup.ios.macos.cs` which already implies that it can only run on iOS + MacCatalyst

src/CommunityToolkit.Maui/Views/Toast/Toast.ios.macos.cs
brminnick: My preference to keep it as `Toast` in case we decide to add a `Toast` control to Maui Community Toolkit in the future. 

Then, similar to the Snackbar architecture, we'd have `Toast.shared.cs` `Toast.ios.macos.cs`, `Toast.android.cs`, and `Toast.windows.cs`.

The filename is already `Toast.ios.macos.cs` which means it can only run on iOS + MacCatalyst anyways.

src/CommunityToolkit.Maui/Views/Popup/Popup.ios.macos.cs
VladislavAntonyuk: we have AppleSnackbar, as for Popup, it is our own Apple implementation.
I think Pedro will create Popup.shared.cs, where it will use this native popup implementation. Let's renamre this to ApplePopup or rename AppleSnackbar to just Snackbar

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs
brminnick: Got it.

This behavior on Windows is different from iOS, MacCatalyst and Android then. I'm not a Windows dev or a heavy Windows user, so I'm not sure what Windows developers/users would expect. 

This'll be important to note in the documentation.

Could you also update the text on the `SnackbarPage` to inform Windows users to disable silent mode so dummies like me don't think it's broken 😃

src/CommunityToolkit.Maui/Views/Popup/Popup.ios.macos.cs
brminnick: I prefer to change `AppleSnackbar` to `Snackbar` 👍

brminnick: Now that I think about it, we may have to keep it as `AppleSnackbar`. 

I purposefully made `AppleSnackbar` a nested class inside of `Snackbar.ios.macos.cs` since `AppleSnackbar` is the UIView for the iOS + MacCatalyst `Snackbar`.

Maybe `SnackbarUIView` or `SnackbarUI` is a better name for this class?

VladislavAntonyuk: SnackbarUIView sounds good. I will check it's necessary to  rename other classes to be consistent

pictos:
@brminnickordões this PR implements the pop-up control from XCT or will be a pop-up version of snackbar?

VladislavAntonyuk:
@pictos, it is a based popup. Just RoundedUIView.
It allows adding any Children. It can be used for Popup like Container.

pictos:
@VladislavAntonyuk but this will be limited to one action, and a text message right?

brminnick:
Yup! That `Popup` class is a base class for Snackbar (which is basically just a popup that only contains Text and a Button).

@pictos I recommend leveraging `Popup.ios.macos.cs` for your implementation of `Popup.shared.cs`. You're obviously welcome to tweak it for your implementation.

We'll be sure to check that the `Snackbar` still works properly when reviewing the `Popup` PR 💯

VladislavAntonyuk:
Popup class here is just a container. it doesn't contain any code specific to snackbar. PopupView class can be used as a container for any content

pictos:
@brminnick I'm working on porting our pop-up to handler. That will take me a few more days, and I'll share a draft PR when I have the android part done. That way we can talk as a team and talk how that should like.

I was just concerned if one will overlap other, but looks like that isn't true.

I'll take a depper look on this PR asap to be more familiar with the control and architecture

brminnick (approved):

brminnick:
Looks good to me! Excellent job, @VladislavAntonyuk!!

@pictos Do you want to give this one more review before we merge it?

pictos:
@brminnick no, we can merge it. And I can rebate against my branch and take a look there. 

:shipit:

emorell96:
> Thanks Vlad!
> 
> ~The Windows Sample didn't work on my machine. Can you double check that?~ **Edit:** User error
> 
> We also need more unit tests to ensure we've covered every public API on `Snackbar` and `ISnackbar`.
> 
> I also recommend moving the UI elements in `Popup.macos.ios.cs` to `PopupView.macos.ios.cs`. This'll ensure that `Popup` contains the functionality for the popup (eg timer, duration, show/dismiss), and that `PopupView` contains all of the UI elements.
> 
> ### Video of Snackbar on Windows Sample App
> Here's an update video of the `Snackbar` on Windows.
> 
> FYI - clicking the Action Button opens a new instance of the app:
> 
> ScreenFlow

Was this ever figured out? I am using snackbar source code as inspiration for a custom implementation but I have the same issue. The MAUI app will just create a new window.

pictos:
@emorell96 if I'm not wrong the fix for this is comming (:

emorell96:
@pictos is it this? https://github.com/CommunityToolkit/Maui/blob/a6d46f4633d27ce38101c17e56ecb584fc3dbff0/samples/CommunityToolkit.Maui.Sample/Platforms/Windows/App.xaml.cs#L24

So basically if a new process is started, the idea is to kill it as it starts?

Do you know if the events available in UWP are available on MAUI? Usually on UWP you can just override the OnActivated and hook up there in order to handle the activation as shown here: https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=uwp#step-3-handling-activation

And if it's different with MAUI, is it possible to use `ToastNotificationManagerCompat.OnActivated` as shown in the docs?

pictos:
@VladislavAntonyuk do you know the answers for this?

emorell96:
> @VladislavAntonyuk do you know the answers for this?

I just wanted to share a little what I've discovered after quite a bit of experimentation. I thought it might be useful for your nuget.

On Windows there's two kind of activations for UWP apps: foreground and background. The former will use the current app but you need to modify the package.appxmanifext file if you want to use `ToastNotificationManagerCompat.OnActivated`. You must both modify the appxmanifest as indicated [here](https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=uwp#step-3-handling-activation) in order to make this work without opening a new window. The background activation uses background tasks instead but they require the background tasks to be either registered as an In-Process background task or Out-of-Process background task as shown [here](https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=uwp#handling-background-activation).

For my use case the background task sounded ideal, but I was unable to make it work with MAUI (at least for the In-Process Background tasks). I kept getting argument out of range exception.

I might try the out-of-process background tasks later.

In any case, one way I found to solve this issue using the foreground activation (the one currently used [here](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs)), you need to make this change to the `package.appxmanifest`:

in the Package element this needs to be added:
```xml
xmlns:desktop="http://schemas.microsoft.com/appx/manifest/desktop/windows10" 
xmlns:com="http://schemas.microsoft.com/appx/manifest/com/windows10"
IgnorableNamespaces="uap rescap com desktop"
```

Inside of the applications:
```xml
<Applications>
    <Application Id="App" Executable="$targetnametoken$.exe" EntryPoint="$targetentrypoint$">
      <uap:VisualElements />
     <Extensions>

        <!-- Specify which CLSID to activate when toast clicked -->
        <desktop:Extension Category="windows.toastNotificationActivation">
          <desktop:ToastNotificationActivation ToastActivatorCLSID="guid-used-in-maui-csproj" /> 
        </desktop:Extension>

        <!--Register COM CLSID LocalServer32 registry key-->
        <com:Extension Category="windows.comServer">
          <com:ComServer>
            <com:ExeServer Executable="filename-used-for-the-exe-usually-the-assembly-name.exe" Arguments="-ToastActivated" DisplayName="Toast activator">
              <com:Class Id="guid-used-in-maui-csproj" DisplayName="Toast activator"/>
            </com:ExeServer>
          </com:ComServer>
        </com:Extension>

      </Extensions>
    </Application>
  </Applications>
```
Using `$targetnametoken$.exe` for the executable in com:ExeServer didn't work. I had to manually enter the assembly name.

Finally I register in the MauiProgram.cs the event:

```csharp
using Microsoft.Toolkit.Uwp.Notifications;
builder.
... (every other build action)
.ConfigureLifecycleEvents(configure =>
            {
				configure.AddWindows(c =>
				{
					c.OnLaunched((window, args) =>
					{
						ToastNotificationManagerCompat.OnActivated += WindowsNotificationService.OnActivated;
					});
                });
            });
```

I made this minimal repo that works:
https://github.com/emorell96/MauiWithWindowsToasts

Hope it helps.

Thanks for the replies!

pictos:
@emorell96 do you mind to open a issue and move the last post into it?

emorell96:
@pictos done: https://github.com/CommunityToolkit/Maui/issues/353

Issue #6, opened 2021-07-29
Localization classes migration


maxkoshevoi:
Migrate XCT localization classes to Maui CT.

Classes list:
- LocalizationResourceManager
- TranslateExtension
- LocalizedString

I am willing to take this on myself

maxkoshevoi:
@brminnick @pictos Hi, didn't found `[Proposal]` issues for these classes. Would they be migrated to a different package?

angelru:
@brminnick  and where will it be included? thanks

VladislavAntonyuk:
@brminnick I see it may be useful. We can discuss it at the Monthly meeting. I will prepare a small demo.

PR #7, opened 2021-07-30
Localization

maxkoshevoi:
**Related issues**

- Fixes: #6

**What was done**

- Migrated localization classes and classes they depend on, along with their test.

**Notes**

- There may be some overlap with windows toolkit (I'm sure that `ObservableObject` is already there). Those classes will need to be removed from here when common package is created.


pictos:
Hey there, please wait until we define what we will migrate to this repo and how. If I'm not wrong this repo isn't even with the XCT

brminnick:
Hi Max! We are not bringing `WeakEventManager` nor `ObservableObject` to the Maui Community Toolkit.

We will be leveraging + recommending the [Microsoft.Toolkit.Mvvm](https://www.nuget.org/packages/Microsoft.Toolkit.Mvvm/) NuGet package (soon to be renamed `CommunityToolkit.Mvvm`) for all MVVM tooling.

maxkoshevoi:
@brminnick Got it, thanks! I'll remove `WeakEventManager` and `ObservableObject` when this repo will begin to reference `Microsoft.Toolkit.Mvvm/CommunityToolkit.Mvvm` (or can I add this reference in this PR?)

maxkoshevoi:
@brminnick Added reference to `Microsoft.Toolkit.Mvvm` and removed duplicated classes (that I've added).
Or should this PR be moved to MVVM alltogated? Not everything here is MVVM related. Two classes only need `ObservableObject` (which is also useful outside MVVM) and third in markup extension.

PR #10, opened 2021-08-05
Add  from Xamarin.Essentials: `Contacts`, `FilePicker`, `MediaPicker` and `WebAuthenticator`

jamesmontemagno:
The following APIs are included in Xamarin.Essentials, but will not be included in the first release of .NET MAUI Essentials:
- Contacts
- FilePicker
- MediaPicker
- WebAuthenticator

To avoid breaking changes to developers using these Xamarin.Essentials APIs, let's bring them into the .NET MAUI Toolkit. This will also allow us to improve upon the existing APIs, making them more robust, to one day add them into .NET MAUI Essentials.

PR #12, opened 2021-08-06
Enable Platform Specific Partial Classes

brminnick:
This PR allows us to use partial classes to access platform-specific APIs.

Platform-specific partial classes is required for #10


src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -1,50 +1,79 @@
...
+
+  <ItemGroup Condition=" $(TargetFramework.Contains('-windows')) ">
+    <Compile Include="**\*.uwp.cs" />
+    <Compile Include="**\*.uwp.*.cs" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" $(TargetFramework.Contains('-maccatalyst')) ">
+    <Compile Include="**\*.macos.cs" />
+    <Compile Include="**\*.macos.*.cs" />
+  </ItemGroup>
```
pictos, line 77: Just out of curiosity. Keeping those props in `Directory.Build.props` causes issues?

brminnick: Yea, I put them here instead of `Directory.Build.props` so that they wouldn't affect CommunityToolkit.Maui.UnitTests. 

Although - didn't you mention that we could do device-specific tests? Maybe we _should_ put it in `Directory.Build.props`?

pictos: Yes, I mentioned that... Well, I'm ok with moving them out of `Directory.Build.props` and bring them back if needed.

PR #13, opened 2021-08-12
Removed IPage interface

pictos:
For Maui preview 7 we don't have the `IPage` interface anymore.

you can see more info [here](https://github.com/dotnet/maui/pull/1875)


brminnick (approved):
Good catch!

PR #16, opened 2021-08-19
Add NuGet signing step to the build pipeline

jfversluis:
I promise it works, look:

image

This is only triggered when the build is not triggered for a PR, we might fine tune that more for later, but this seems to be also what the WCT is doing.

Also removes some things that aren't used (poliCheck and nightly schedule)


jfversluis:
I'm going to be a rebel here and merge this with my superpowers

PR #19, opened 2021-08-22
Add Effects to the .NET MAUI Community Toolkit

jsuarezruiz:
Added:
- Effects from Xamarin Community Toolkit
- AppHostBuilderExtensions to allow configure the Toolkit (register handlers, effects etc)


src/CommunityToolkit.Maui/Hosting/AppHostBuilderExtensions.shared.cs
```diff
@@ -0,0 +1,20 @@
+﻿using CommunityToolkit.Maui.Effects;
+using Microsoft.Maui.Controls.Hosting;
+using Microsoft.Maui.Hosting;
+
+namespace CommunityToolkit.Maui.Hosting
+{
+    public static class AppHostBuilderExtensions
+    {
+        public static IAppHostBuilder ConfigureCommunityToolkit(this IAppHostBuilder builder)
```
brminnick, line 9: Love this idea! 

Two thoughts:
- Could we rename this to `UseMauiToolkit`? That way, it'll match `UseMaui`.
- Is there any downside to adding all of the Toolkit's effects here?
  - I personally can't think of anything detrimental, but wanted to at least ask the question!

pictos: We can create an overload that accepts an `Action<IEffectsBuilder>` and let the user configure by himself.
```csharp
public static IAppHostBuilder ConfigureCommunityToolkit(this IAppHostBuilder builder, Action<IEffectsBuilder> action)
{
    builder.ConfigureEffects(action);
    return builder;
}
```
Or a more general one
```csharp
 public static IAppHostBuilder AnotherConfig(this IAppHostBuilder builder, Func<IAppHostBuilder, IAppHostBuilder> func)
{
    return func(builder);
}
```

pictos: If we keep our lib `LinkerSafe` them we don't need to care about it, right? Because linker should remove all unused features

brminnick:
Closing #124

Issue #109, opened 2021-09-28
[Proposal] Expander


brminnick:
# Expander

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation: https://github.com/MicrosoftDocs/CommunityToolkit/pull/165#pullrequestreview-1172896030

## Summary
[summary]: #summary

The Xamarin.CommunityToolkit Expander control provides an expandable container to host any content. The control has a header and content, and the content is shown or hidden by tapping the Expander header. When only the Expander header is shown, the Expander is collapsed. When the Expander content is visible the Expander is expanded.

## Detailed Design
[design]: #detailed-design

### Expander.shared.cs

```cs
[ContentProperty(nameof(Content))]
public class Expander : BaseTemplatedView<StackLayout>
{
  public static readonly BindableProperty HeaderProperty;
  public static readonly BindableProperty ContentProperty;
  public static readonly BindableProperty ContentTemplateProperty;
  public static readonly BindableProperty IsExpandedProperty;
  public static readonly BindableProperty DirectionProperty;
  public static readonly BindableProperty TouchCaptureViewProperty;
  public static readonly BindableProperty AnimationLengthProperty;
  public static readonly BindableProperty ExpandAnimationLengthProperty;
  public static readonly BindableProperty CollapseAnimationLengthProperty;
  public static readonly BindableProperty AnimationEasingProperty;
  public static readonly BindableProperty ExpandAnimationEasingProperty;
  public static readonly BindableProperty CollapseAnimationEasingProperty;
  public static readonly BindableProperty StateProperty;
  public static readonly BindableProperty CommandParameterProperty;
  public static readonly BindableProperty CommandProperty;
  public static readonly BindableProperty ForceUpdateSizeCommandProperty;
  
  public event EventHandler Tapped { get; set; }
  
  public View? Header { get; set; } 
  public View? Content { get; set; }
  public DataTemplate? ContentTemplate { get; set; }
  public bool IsExpanded { get; set; }
  public ExpandDirection Direction { get; set; }
  public View? TouchCaptureView { get; set; }
  public uint AnimationLength { get; set; }
  public uint ExpandAnimationLength { get; set; }
  public uint CollapseAnimationLength { get; set; }
  public Easing AnimationEasing { get; set; }
[...]

PR #138, opened 2021-10-12
Expander

VladislavAntonyuk:
## Fixes

#109 

## Description

Add expander

## Detailed Solution

Port expander from XCT to MCT


emorell96:
So will this ever be revived? Or the whole XCT won't be fully ported unless it goes through all the bureaucracy even it was actually implemented?

VladislavAntonyuk:
@emorell96 it will after the proposal is approved. We will back to it later after GA.

Issue #59, opened 2021-09-25
[Proposal] `enum`-To-`int` Converter


brminnick:
# EnumToIntConverter

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation: Started at https://github.com/MicrosoftDocs/CommunityToolkit/pull/57

## Summary
[summary]: #summary

The `EnumToIntConverter` is a converter that allows you to convert a standard `Enum` (extending int) to its underlying primitive `int` type. It is useful when binding a collection of values representing an enumeration type with default numbering to a control such as a `Picker`.

For localization purposes or due to other requirements, the `enum` values often need to be converted to a human-readable string. In this case, when the user selects a value, the resulting `SelectedIndex` can easily be converted to the underlying enum value without requiring additional work in the associated ViewModel.

## Detailed Design
[design]: #detailed-design

### EnumToIntConverter.shared.cs

```cs
public class EnumToIntConverter : ValueConverterExtension, IValueConverter
{
  public object Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture) =>value is Enum ? System.Convert.ToInt32(value) : throw new ArgumentException($"{value?.GetType().Name} is not a valid enumeration type");

  public object ConvertBack(object? value, Type targetType, object? parameter, CultureInfo? culture) => value is int enumIntVal && Enum.IsDefined(targetType, enumIntVal) ? Enum.ToObject(targetType, enumIntVal) : throw new ArgumentException($"{value} is not valid for {targetType.Name}");
}
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<ContentPage
  x:Class="Xamarin.CommunityToolkit.Sample.Pages.Converters.EnumToIntConverterPage"
  xmlns="http://xamarin.com/schemas/2014/forms"
  xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
  xmlns:vm="clr-namespace:Xamarin.CommunityToolkit.Sample.ViewModels.Converters"
  xmlns:xct="http://xamarin.com/schemas/2020/toolkit">
  <ContentPage.BindingContext>
    <vm:EnumToIntConverterViewModel />
  </ContentPage.BindingContext>
  
[...]

PR #154, opened 2021-10-17
Added EnumToIntConverter Sample

steven62f:
## Description
#59 enum-To-int Converter Sample

## Detailed Solution
I created the EnumToIntConverterViewModel and update the associated page (EnumToIntConverterPage).

I also maked a code review on converters's ViewModels :
- Use File-Scoped Namespace
- Remove unnecessary namespaces
- Rename privates fields with _


StevenHanquez:
@pictos Yes, you are right. In the haste, I forgot to ask for permission to make the ticket. I would be more careful :)

brminnick:
Thanks @StevenHanquez!

The reason we ask folks to claim it first is just ensure that no one ends up doing duplicate work; we don't to end up with two PRs for the same task!

Issue #54, opened 2021-09-24
[Proposal] `Color`-To-`string` Converters


brminnick:
# ColorToStringConverter

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Documentation
* [x] Sample

## Summary
[summary]: #summary

Converts a `Color` to its associated `string`, e.g. RGB, RGBA, HEX, CMYK, etc.

## Detailed Design
[design]: #detailed-design

### ColorToStringConverter.shared.cs

```cs
public class ColorToRgbStringConverter : BaseConverterOneWay<Color, string>;
public class ColorToRgbaStringConverter : BaseConverterOneWay<Color, string>;
public class ColorToHexRgbStringConverter : BaseConverter<Color, string>;
public class ColorToHexRgbaStringConverter : BaseConverter<Color, string>;
public class ColorToCmykStringConverter : BaseConverterOneWay<Color, string>;
public class ColorToCmykaStringConverter : BaseConverterOneWay<Color, string>;
public class ColorToHslStringConverter : BaseConverterOneWay<Color, string>;
public class ColorToHslaStringConverter : BaseConverterOneWay<Color, string>;
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             x:Class="MyLittleApp.MainPage">
    <ContentPage.Resources>
         <ResourceDictionary>
             <toolkit: ColorToRgbStringConverter x:Key="ColorToRgbStringConverter" />
         </ResourceDictionary>
    </ContentPage.Resources>

    <StackLayout>

        <Label Text="{Binding LabelTextColor, Converter={StaticResource ColorToRgbStringConverter}}" />

    </StackLayout>
</ContentPage>
```

### C# Usage

```cs
class MyPage : ContentPage
{
  public MyPage()
  {
    Content = new Label().Bind(Image.TextProperty, named(ViewModel.LabelTextColor), converter: new ColorToRgbStringConverter());
  }
}
```

juansturla:
Hi! 
I will try to create the sample for this

stefanbogaard86:
[...]

Issue #53, opened 2021-09-24
[Proposal] `Color`-To-Component Converters


brminnick:
# ColorToComponentConverters

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Documentation
* [x] Sample

## Summary
[summary]: #summary

Converts the incoming value from `Color` and returns the object of a type `byte`.

## Detailed Design
[design]: #detailed-design

```cs
public class ColorToByteAlphaConverter : BaseConverterOneWay<Color, byte>;
public class ColorToByteRedConverter : BaseConverterOneWay<Color, byte>;
public class ColorToByteGreenConverter : BaseConverterOneWay<Color, byte>;
public class ColorToByteBlueConverter : BaseConverterOneWay<Color, byte>;
public class ColorToPercentCyanConverter : BaseConverterOneWay<Color, double>;
public class ColorToPercentMagentaConverter : BaseConverterOneWay<Color, double>;
public class ColorToPercentYellowConverter : BaseConverterOneWay<Color, double>;
public class ColorToBlackKeyConverter : BaseConverterOneWay<Color, double>;
public class ColorToDegreeHueConverter : BaseConverterOneWay<Color, double>;
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             x:Class="MyLittleApp.MainPage">
    <ContentPage.Resources>
         <ResourceDictionary>
             <toolkit: ColorToByteAlphaConverter x:Key="ColorToByteAlphaConverter" />
         </ResourceDictionary>
    </ContentPage.Resources>

    <StackLayout>

        <Label TextColor="{Binding MyTextColor, Converter={StaticResource ColorToByteAlphaConverter}}" />

    </StackLayout>
</ContentPage>
```

### C# Usage

```cs
class MyPage : ContentPage
{
  public MyPage()
  {
    Content = new Label().Bind(Image.TextColorProperty, named(ViewModel.LabelTextColor), converter: new ColorToByteAlphaConverter());
  }
}

```

Issue #52, opened 2021-09-24
[Proposal] `Color-To-Color` Converters


brminnick:
# ColorToBlackOrWhiteConverter

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Sample
* [x] Unit Tests
* [ ] Documentation: Not Started

## Summary
[summary]: #summary

A one-way converter that converts a `Color` to a monochrome value of black or white. If the incoming `Color` is light, it will be converted to `Colors.White`. If the incoming `Color` is dark, it will be converted to `Colors.Black`.

## Detailed Design
[design]: #detailed-design

### ColorExtension.shared.cs
```cs
public static Color ToBlackOrWhite(this Color baseColor) => baseColor.IsDark() ? Colors.Black : Colors.White;

public static bool IsDark(this Color c) => c.GetByteRed() + c.GetByteGreen() + c.GetByteBlue() <= 127 * 3;

public static Color ToBlackOrWhiteForText(this Color baseColor) => baseColor.IsDarkForTheEye() ? Colors.White : Colors.Black;

public static Color ToGrayScale(this Color baseColor)
{
    var avg = (baseColor.Red + baseColor.Blue + baseColor.Green) / 3;
    return Color.FromRgb(avg, avg, avg);
}

public static bool IsDarkForTheEye(this Color c) => (c.GetByteRed() * 0.299) + (c.GetByteGreen() * 0.587) + (c.GetByteBlue() * 0.114) <= 186;

public static Color ToInverseColor(this Color baseColor) => Color.FromRgb(1 - baseColor.Red, 1 - baseColor.Green, 1 - baseColor.Blue);
```

### ColorConverters.shared.cs

```cs
public class ColorToBlackOrWhiteConverter : BaseConverterOneWay<Color, Color>
{
    public override Color ConvertFrom(Color value) => value.ToBlackOrWhite();
}

public class ColorToColorForTextConverter : BaseConverterOneWay<Color, Color>
{
    public override Color ConvertFrom(Color value) => value.ToBlackOrWhiteForText();
}

public class ColorToGrayScaleColorConverter : BaseConverterOneWay<Color, Color>
{
    public override Color ConvertFrom(Color value) => value.ToGrayScale();
}

public class ColorToInverseColorConverter : BaseConverterOneWay<Color, Color>
{
[...]

PR #159, opened 2021-10-18
Added color converters samples

juansturla:
This is for a sample for
#52 Color to Colors converter
#53 Color to Color Component Converter
#54 Color to String converter

Based on MCT Sample, with some modifications on the code-behind:
- Labels with `FormattedStrings` did not work with `<span>`. Changed for a `Text `with `StringFormat`
- `BoxView` at the bottom of the page did not work either, had to move it to the top and set a default `BackgroundColor ="White"`
- There a warning at the constructor that says that the `picker` and the `boxView` might not have a value (I supressed the warning)
- I had to change the dictionary to be null-safe (if the `FieldInfo.Value` is null, returns `Colors.White`


pictos:
> There a warning at the constructor that says that the picker and the boxView might not have a value (I supressed the warning)

Probably that's caused by the source generator, maybe they fix it in the next release. For reference:

- https://github.com/dotnet/maui/issues/1796

For the other bugs I'll open issues on maui repo (feel free to do it, if you can and want to do)

juansturla:
Requested changes were solved

brminnick:
Approved. Thanks @juansturla!

PR #162, opened 2021-10-20
Add files which were missed from file scoped namespaces

VladislavAntonyuk:
1. Add files which were missed from file scoped namespaces
2. Fix warning for missed summary closed tag


VladislavAntonyuk:
🎆

PR #189, opened 2021-11-09
Added shared bug and feature PR template

bijington:
## Description

This provides a set of templates that can be used when creating a Pull Request.

I am not sure yet if I need to change things in order to actually use them.


jfversluis:
Awesome! No, this should just show up whenever it's merged. However, I would love it if we could make it more like this: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

bijington:
> Awesome! No, this should just show up whenever it's merged. However, I would love it if we could make it more like this: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

Issue forms look great! I will go off and begin the battle with tab indentation...

bijington:
So it appears that Issue Forms are not supported for PRs just yet.

jfversluis:
Aah my bad! I totally overlooked the fact that this was for PRs and not issues. Guess I was too obsessed with implementing the issue forms :D

bijington:
It looks like using multiple PR templates is only possible through the use of the query parameter. I think it makes this approach unusable. For now though the templates can be tested here:

### Bug

https://github.com/bijington/Maui/compare/main...bijington:Add-BaseGalleryPage?template=bug_template.md

### Feature

https://github.com/bijington/Maui/compare/main...bijington:Add-BaseGalleryPage?template=feature_template.md

They are quite similar so it should be relatively straightforward to use general language for a single template

brminnick:
With `PULL_REQUEST_TEMPLATE.md` now deleted, which template does GitHub choose if I open a PR, `bug_template.md` or `feature_template.md`?

bijington:
> With `PULL_REQUEST_TEMPLATE.md` now deleted, which template does GitHub choose if I open a PR, `bug_template.md` or `feature_template.md`?

Sadly it won't use either. I think we will need to either have a single generic template or one that would require the user to manually edit like this:

https://github.com/bijington/Maui/compare/main...bijington:Add-BaseGalleryPage

I suspect we probably want the former and just have a single generic template where we referred to `approved` bugs and `Championed` proposals. It is a shame this isn't more friendly

brminnick:
> Sadly it won't use either 

Dang. 

> I suspect we probably want the former and just have a single generic template where we referred to approved bugs and Championed proposals. It is a shame this isn't more friendly

Yea, let's go this route, then 👍

bijington:
In case it helps to visualise the resulting template changes:

Screenshot 2021-11-16 at 21 16 36

pictos (dismissed):

bijington:
@brminnick this is the PR we were discussing in case you wanted to have a quick peek. 

@pictos  thanks for the review :)

brminnick (dismissed):

brminnick:
Thanks @bijington!

Issue #88, opened 2021-09-26
[Proposal] BackgroundColorTo Animation Extension


brminnick:
# ColorTo

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation

## Summary
[summary]: #summary

Animation to change the Color of a VisualElement

## Detailed Design
[design]: #detailed-design

### ColorAnmiationExtensions.shared.cs

```cs
public static Task<bool> BackgroundColorTo(this VisualElement element, Color color, uint length = 250u, Easing? easing = null)
{
  ArgumentNullException.ThrowIfNull(element);
  
  var animationCompletionSource = new TaskCompletionSource<bool>();
  
  new Animation
  {
    { 0, 1, new Animation(v => element.BackgroundColor = new Color(v, element.BackgroundColor.G, element.BackgroundColor.B, element.BackgroundColor.A), element.BackgroundColor.R, color.R) },
    { 0, 1, new Animation(v => element.BackgroundColor = new Color(element.BackgroundColor.R, v, element.BackgroundColor.B, element.BackgroundColor.A), element.BackgroundColor.G, color.G) },
    { 0, 1, new Animation(v => element.BackgroundColor = new Color(element.BackgroundColor.R, element.BackgroundColor.G, v, element.BackgroundColor.A), element.BackgroundColor.B, color.B) },
    { 0, 1, new Animation(v => element.BackgroundColor = new Color(element.BackgroundColor.R, element.BackgroundColor.G, element.BackgroundColor.B, v), element.BackgroundColor.A, color.A) },
  }.Commit(element, nameof(ColorTo), 16, length, easing, (d, b) => animationCompletionSource.SetResult(true));
  
  return animationCompletionSource.Task;
}
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

N/A

### C# Usage

```cs
protected override async void OnAppearing()
{
  await myLabel.BackgroundColorTo(Colors.Red);
}
```

PieEatingNinjas:
Hi @brminnick. If nobody is working on this, I want to give it a try. Haven’t done anything on this project yet, this one seems to be a good one to start with. 🙂

brminnick:
Thanks @PieEatingNinjas! Assigned ✅

PieEatingNinjas:
[...]

PR #195, opened 2021-11-22
Add `BackgroundColorTo` Unit Tests

brminnick:
## Description

Adds Unit Tests for #88

## Detailed Solution

- Adds **ColorAnimationExtensions_Test.cs**
- Adds **AnimationExtensions.cs**
  - Introduces `EnableAnimations` extension method for `IView`
    - e.g. `myLabel.EnableAnimations()`
    - Allows us to set `IMauiContext` on an `IView` that can handle animations, which is required by `Microsoft.Maui.Controls.AnimationExtensions.Animate`
    - Otherwise, `Animation.Commit` will throw `ArgumentException`, causing `BackgroundColorTo` to return `false`
- Add `abstract` keyword to `BaseTest.cs`


src/CommunityToolkit.Maui.UnitTests/Extensions/ColorAnimationExtensions_Test.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Threading.Tasks;
+using CommunityToolkit.Maui.Extensions;
+using CommunityToolkit.Maui.UnitTests.Mocks;
+using Microsoft.Maui.Controls;
+using Microsoft.Maui.Graphics;
+using Xunit;
+
+namespace CommunityToolkit.Maui.UnitTests.Extensions;
+
+public class ColorAnimationExtensions_Test : BaseTest
```
PieEatingNinjas, line 14: Maybe add additional test where the `VisualElement` doens't have a `BackgroundColor` as I had to write logic to handle this scenario
```
[Fact]
public async Task BackgroundColorTo_VerifyColorChangedWhenInitialBackgroundNull()
{
	Color updatedBackgroundColor = Colors.Yellow;

	VisualElement element = MockAnimationHandler.Prepare(new Label());

	var isSuccessful = await element.BackgroundColorTo(updatedBackgroundColor);

	Assert.True(isSuccessful);
	Assert.Equal(updatedBackgroundColor, element.BackgroundColor);
}

```

brminnick: Great idea!

Done ✅

src/CommunityToolkit.Maui.UnitTests/Mocks/AnimationExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+			public override async void Start()
+			{
+				_cancellationTokenSource = new();
+
+				while (!_cancellationTokenSource.IsCancellationRequested)
+				{
+					Fire?.Invoke();
+
+					if (!_cancellationTokenSource.IsCancellationRequested)
+						await Task.Delay(16);
```
bijington, line 76: I assume you have actively chosen not to pass the `_cancellationTokenSource.Token` in to `Task.Delay` to avoid exceptions and given 16ms really is a short delay

bijington: That said would a `TimeSpan.FromMilliseconds` make this slightly more readable?

src/CommunityToolkit.Maui.UnitTests/Mocks/AnimationExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+using Microsoft.Maui.Handlers;
+
+namespace CommunityToolkit.Maui.UnitTests.Mocks;
+
+static class AnimationExtensions
+{
+	public static void EnableAnimations(this IView view) => MockAnimationHandler.Prepare(view);
+
+	// Inspired by Microsoft.Maui.Controls.Core.UnitTests.AnimationReadyHandler
+	class MockAnimationHandler : ViewHandler<IView, object>
```
bijington, line 19: I've been jumping back on forth on whether we should try to make our naming match the MAUI naming to potentially make it easier to migrate features across should we get the approval. 

I think I am now leaning towards not doing that given we aren't guaranteeing that our underlying mocks match the exact behaviour of the MAUI mocks.

I am not entirely sure why I am writing these thoughts down now but what do you think?

And for the record I like the inclusion of Mock to make it clear what it does

src/CommunityToolkit.Maui.UnitTests/Mocks/AnimationExtensions.cs
brminnick: > chosen not to pass the _cancellationTokenSource.Token in to Task.Delay to avoid exceptions
Yup! 💯 

> would a TimeSpan.FromMilliseconds make this slightly more readable

Also yup 👍

src/CommunityToolkit.Maui.UnitTests/Mocks/AnimationExtensions.cs
brminnick: > whether we should try to make our naming match the MAUI naming to potentially make it easier to migrate features across should we get the approval

For `CommunityToolkit.Maui`, I 100% agree.

For `CommunityToolkit.Maui.UnitTests`, I don't 100% disagree, but I think what we have here is better and easier for other community members to leverage when they add new unit tests in the future.

>  I like the inclusion of Mock to make it clear what it does
Haha - thanks! I know I'm biased, but I think I improved on how the MAUI team is using `Microsoft.Maui.Controls.Core.UnitTests.AnimationReadyHandler` to enable animations.

I really like how clean `myLabel.EnableAnimations()` is, compared to their implementation:
```cs
var myLabel = AnimationReadyHandler.Prepare(new Label());
```

This extension method should make it super easy to enable animations for the unit tests in the future 💯

bijington: Yes I agree this does feel like an improvement 😃 

I do like the extension method addition too. I have been experimenting with making mocks more descriptive by using creator/builder patterns such as:

```csharp
// creator - probably can find a better example
var myLabel = LabelCreator.WithAnimationsEnabled();

// builder
var myLabel = LabelBuilder
    .WithAnimationsEnabled()
    .WithBackgroundColor(Color.Red)
    .Build();
```

I do think what we have is descriptive though and easy to implement which is a really great thing!

src/CommunityToolkit.Maui.UnitTests/Extensions/ColorAnimationExtensions_Test.cs
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	[Fact]
+	public async Task BackgroundColorTo_VerifyFalseWhenAnimationContextNotSet()
+	{
+		VisualElement element = new Label();
+
+		var isSuccessful = await element.BackgroundColorTo(Colors.Red);
+
+		Assert.False(isSuccessful);
```
bijington, line 53: Sorry one final question... what should the `BackgroundColor` be if the animation doesn't start? I am starting to wonder if this is a silly question now but does it still apply the color change but just not animate it or does it bail out of making any change if it can't animate?

bijington: Looking through the code I expect it to not change the `BackgroundColor`, is it worth an `Assert` just to confirm that it doesn't change it?

brminnick: I like it!

Added ✅

bijington (dismissed):
Looking great!

bijington (approved):

Issue #61, opened 2021-09-25
[Proposal] ImageResourceConverter


brminnick:
# ImageResourceConverter

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation

## Summary
[summary]: #summary

Converts embedded image resource ID to it ImageSource

## Detailed Design
[design]: #detailed-design

### ImageResourceConverter.shared.cs
```cs
public class ImageResourceConverter : IValueConverter
{
  public object? Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
  {
	if (value == null)
		return null;

	if (value is not string imageId)
		throw new ArgumentException("Value is not a string", nameof(value));

	return ImageSource.FromResource(imageId, Application.Current.GetType().GetTypeInfo().Assembly);
  }

  public object ConvertBack(object? value, Type? targetType, object? parameter, CultureInfo? culture) => throw new NotImplementedException();
}
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             xmlns:myEnum="MyLittleApp.Models"
             x:Class="MyLittleApp.MainPage">

    <ContentPage.Resources>
        <ResourceDictionary>
            <xct: ImageResourceConverter x:Key="ImageResourceConverter" />
        </ResourceDictionary>
    </ContentPage.Resources>

    <StackLayout>

        <Image Source="{Binding ImageResource, Converter={StaticResource ImageResourceConverter}}" />

    </StackLayout>
</ContentPage>

```

### C# Usage

```cs
class MyPage : ContentPage
{
  public MyPage()
  {
    Content = new StackLayout
    {
      Children = 
      {
          new Image().Bind(Image.SourceProperty, nameof(ViewModel.ImageResource), converter: new ImageResourceConverter()),
      };
    }
  }
}

```

CalvinAllen:
![raw](https://user-images.githubusercontent.com/41448698/140981350-f07df71b-9518-487d-82e0-9e99e686d2f6.gif)

CalvinAllen:
[...]

Issue #51, opened 2021-09-24
[Proposal] `byte[]`-to-`ImageSource` Converter


brminnick:
# ByteArrayToImageSourceConverter

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation

## Summary
[summary]: #summary

The `ByteArrayToImageSourceConverter` is a `Converter` that allows the user to convert an incoming value from `byte[]` and returns an `ImageSource`. This object can then be used as the Source of an Image control.

## Detailed Design
[design]: #detailed-design

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             x:Class="MyLittleApp.MainPage">
    <ContentPage.Resources>
         <ResourceDictionary>
             <xct:ByteArrayToImageSourceConverter x:Key="ByteArrayToImageSourceConverter" />
         </ResourceDictionary>
    </ContentPage.Resources>

    <StackLayout>

        <Image Source="{Binding MyByteArray, Converter={StaticResource ByteArrayToImageSourceConverter}}" />

    </StackLayout>
</ContentPage>
```

### C# Usage

```cs
class MyPage : ContentPage
{
  public MyPage()
  {
    Content = new Image().Bind(Image.SourceProperty, named(ViewModel.MyByteArray), converter: new ByteArrayToImageSourceConverter());
  }
}
```

CalvinAllen:
I know we talked about the tests being completed, but I noticed there weren't any for `ConvertBack`, so I wrote a few and submitted PR #197

PR #196, opened 2021-11-22
#51 -- Adds tests for ImageResourceConverter

CalvinAllen:
## Fixes
Resolves the unit test portion of #61

## Description
[Proposal] ImageResourceConverter #61

## Detailed Solution
Unit tests to cover the ImageResouceConverter.  This required mocking out the Application.Current object to allow assembly discovery to function properly for the Test project, which now has an image as an embedded resource.,


CalvinAllen:
I can't tell if this is due to my changes or not - I don't think so?

brminnick (changes requested):
Looks pretty good! 

I recommend moving the initialization and disposal of `Application.Current` to the test class. Other tests may rely on `Application`, and we don't want to break them.

I also went ahead and added the `[NotNullIfNotNull] attribute to `ImageResourceConverter.Convert` to make your casting a bit easier + cleaner in `ImageResourceConverter`.

src/CommunityToolkit.Maui/Converters/ImageResourceConverter.shared.cs
```diff
@@ -9,24 +10,25 @@ namespace CommunityToolkit.Maui.Converters;
...
-            return null;
+	/// <summary>
+	/// Converts embedded image resource ID to it ImageSource.
+	/// </summary>
+	/// <param name="value">The value to convert.</param>
+	/// <param name="targetType">The type of the binding target property. This is not implemented.</param>
+	/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
+	/// <param name="culture">The culture to use in the converter. This is not implemented.</param>
+	/// <returns>The ImageSource related to the provided resource ID of the embedded image. If it's null it will returns null.</returns>
+	[return: NotNullIfNotNull("value")]
```
brminnick, line 21: Just FYI - `[return: NotNullIfNotNull("value")]` means that if `value is not null`, then whatever the method returns also won't be null.

Now that I've added it here, I should go through the other converters and clean them up in a future PR as well 😃

CalvinAllen: Ah, that makes sense.  I followed very closely to the `ByteArrayToImageSourceConverter_Tests` since this is my first foray into this project.

brminnick: Ah - I'm glad you pointed that out! 

Let's consolidate a couple duplicated methods, then! I'll leave the recommendation in another comment.

src/CommunityToolkit.Maui.UnitTests/BaseTest.cs
```diff
@@ -16,6 +16,7 @@ protected BaseTest()
 		defaultCulture = System.Threading.Thread.CurrentThread.CurrentCulture;
 		defaultUICulture = System.Threading.Thread.CurrentThread.CurrentUICulture;
 		Device.PlatformServices = new MockPlatformServices();
+		Application.Current = new MockApplication();
```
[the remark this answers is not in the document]
CalvinAllen: Changed and pushed

src/CommunityToolkit.Maui.UnitTests/BaseTest.cs
```diff
@@ -26,6 +27,7 @@ protected virtual void Dispose(bool isDisposing)
 			return;
 
 		Device.PlatformServices = null;
+		Application.Current = null;
```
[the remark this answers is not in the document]
CalvinAllen: Changed and pushed

src/CommunityToolkit.Maui/Converters/ImageResourceConverter.shared.cs
CalvinAllen: Yeah, I realized the duplicates, but just wasn't sure what pattern was in use for that kind of stuff.

src/CommunityToolkit.Maui.UnitTests/Converters/ImageResourceConverter_Tests.cs
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		for (var i = 0; i < a.Length; i++)
+		{
+			if (a.ReadByte() != b.ReadByte())
+				return false;
+		}
+
+		return true;
+	}
```
brminnick, line 85: Since these are also used by `ByteArrayToImageSourceConverter_Tests`, let's move the following methods to `BastTest` and increase their visibility slightly to `protected`:

- `protected static Stream GetStreamFromImageSource(ImageSource imageSource)`
- `protected static bool StreamEquals(Stream a, Stream b)`

And I'll add `[NotNullIfNotNull]` to `ByteArrayToImageSourceConverter` as well 👍

src/CommunityToolkit.Maui.UnitTests/Converters/ImageResourceConverter_Tests.cs
```diff
@@ -0,0 +1,83 @@
...
+	[Fact]
+	public void NullValueReturnsNull()
+	{
+		var imageResourceConverter = new ImageResourceConverter();
+
+		Assert.Null(imageResourceConverter.Convert(null, typeof(ImageResourceConverter), null, CultureInfo.CurrentCulture));
+	}
+
+	[Fact]
+	public void ThrowsIfNotAString()
```
CalvinAllen, line 48: @brminnick What is the preference for using multiple Asserts versus making this test a Theory with class data?  (just a question, cause I'm a n00b here)

src/CommunityToolkit.Maui.UnitTests/Converters/ImageResourceConverter_Tests.cs
CalvinAllen: Changed and pushed

src/CommunityToolkit.Maui.UnitTests/Converters/ImageResourceConverter_Tests.cs
brminnick: Mostly it was just me being lazy and copy/pasting your Assert statement 😄

You're welcome to refactor it into `Theory` + `MemberData` like we've done here for `ListIsNotNullOrEmptyConverter_Tests.cs`:
https://github.com/CommunityToolkit/Maui/blob/6767dd092da74344eec63c33dc706b2e11f05859/src/CommunityToolkit.Maui.UnitTests/Converters/ListIsNotNullOrEmptyConverter_Tests.cs#L12-L29

src/CommunityToolkit.Maui.UnitTests/BaseTest.cs
```diff
@@ -38,4 +39,34 @@ void IDisposable.Dispose()
...
+		return task.Result;
+	}
+
+	protected static bool StreamEquals(Stream a, Stream b)
+	{
+		if (a == b)
+			return true;
+
+		if (a == null
+			|| b == null
```
brminnick, line 58: You can get rid of the `null` checks since neither `Stream a`, nor `Stream b` are nullable

brminnick:
image
👀 Don't forget to click the **Update Branch** button here too

src/CommunityToolkit.Maui.UnitTests/BaseTest.cs
CalvinAllen: Got it - done

src/CommunityToolkit.Maui.UnitTests/Converters/ImageResourceConverter_Tests.cs
CalvinAllen: Gotcha - I think I will :)

CalvinAllen: Done

brminnick (dismissed):
Love it! 

Thanks @CalvinAllen!!

src/CommunityToolkit.Maui.UnitTests/xunit.runner.json
```diff
@@ -0,0 +1,4 @@
+{
+    "parallelizeAssembly": false,
+    "parallelizeTestCollections": false
+}
```
brminnick, line 4: Just FYI - I added this to ensure the unit tests run serially.

I noticed couple tests were randomly failing because we were setting `Application.Current` (which is a static and used by other tests) to null

brminnick (approved):

src/CommunityToolkit.Maui.UnitTests/xunit.runner.json
CalvinAllen: I was afraid of that, but wasn't sure how to solve it 😦

PR #200, opened 2021-11-23
Create Nullable `IValueConverter`s, and Add Nullable Static Analyzers

brminnick:
### ICommunityToolkitValueConverter

```cs
{
	new object? Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture);

	new object? ConvertBack(object? value, Type? targetType, object? parameter, CultureInfo? culture);

	object? IValueConverter.Convert(object value, Type targetType, object parameter, CultureInfo culture) =>
		Convert(value, targetType, parameter, culture);

	object? IValueConverter.ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) =>
		ConvertBack(value, targetType, parameter, culture);
}
```

### ICommunityToolkitIMultiValueConverter
```cs
public interface ICommunityToolkitIMultiValueConverter : IMultiValueConverter
{
	new object? Convert(object[]? values, Type? targetType, object? parameter, CultureInfo? culture);

	new object[]? ConvertBack(object? value, Type[]? targetTypes, object? parameter, CultureInfo? culture);

	object? IMultiValueConverter.Convert(object[] values, Type targetType, object parameter, CultureInfo culture) =>
		Convert(values, targetType, parameter, culture);

	object[]? IMultiValueConverter.ConvertBack(object value, Type[] targetTypes, object parameter, CultureInfo culture) =>
		ConvertBack(value, targetTypes, parameter, culture);
}
```

This PR creates `ICommunityToolkitValueConverter` and `ICommunityToolkitIMultiValueConverter`.

> Why?

The MAUI team has not yet enabled nullable (`#nullable enable`) for [`Microsoft.Maui.Controls.IValueConverter`](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/IValueConverter.cs). This means that, until they enable nullable, we have to assume that a nullable value can be passed into the methods and that a nullable value is required to return.

The best way to enforce this going forward is to create our own nullable implementations of IValueConverter that will be used by each Converter in the toolkit.

We can also leverage `[return: Null]` to let our users know that we are not returning `null` when we are 100% sure that `Convert` always returns a non-null value. For example, `BoolToObjectConverter.ConvertBack` will always return a `bool`.

Once the MAUI team enables nullable on `IValueConverter`, and they've decided which parameters should/shouldn't be null for `Convert` and `ConvertBack`, we can revisit `ICommunityToolkitValueConverter` and `ICommunityToolkitIMultiValueConverter`.

## Nullable Static Analyzers

This PR has a secondary goal of implementing [Nullable Static Analyzers](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#conditional-post-conditions-notnullwhen-maybenullwhen-and-notnullifnotnull).

> Why?

We have Nullable enabled by default, and, as responsible library maintainers, we owe a responsibility to our users to always handle `null` properly to make their lives easier.

## Detailed Solution

TBD


src/CommunityToolkit.Maui/Converters/BoolToObjectConverter.shared.cs
```diff
@@ -35,7 +36,7 @@ public class BoolToObjectConverter<TObject> : ValueConverterExtension, IValueCon
 	/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
 	/// <param name="culture">The culture to use in the converter.  This is not implemented.</param>
 	/// <returns>The object assigned to <see cref="TrueObject"/> if value equals True, otherwise the value assigned to <see cref="FalseObject"/>.</returns>
-	public object? Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
+	public object? Convert([NotNull] object? value, Type? targetType, object? parameter, CultureInfo? culture)
```
VladislavAntonyuk, line 39: If the value is not null, do we need nullable?

brminnick: Ha - kind of ironic, right?

`[NotNull]` here says that if this method successfully returns, then `value` is not null. It basically acts as a null-check for whomever is consuming this method, like so:

```cs
[Fact]
public void TestResult(bool? input)
{
    var boolToObjectConverter = new BoolToObjectConverter();
    var convertedResult = boolToObjectConverter.Convert(input, null, null, null);

    //No need to for if(input is not null) because the Roslyn Analyzers now know that `input` cannot be null because the `Convert` method successfully returned
    bool confirmedInput = input;

    Assert.NotNull(input);
}
```

We do still need nullable here because `ICommunityToolkitValueConverter` requires it. However, we can use these Nullable Static Analyzers to essentially work-around the interface requirements.

Does that make sense? I know this is a bit of an obscure topic 😅

VladislavAntonyuk: But we later throw exception if value is not bool (including null). We cannot guarantee that value is not null if user catch the exception

src/CommunityToolkit.Maui/Converters/BoolToObjectConverter.shared.cs
```diff
@@ -51,7 +52,8 @@ public class BoolToObjectConverter<TObject> : ValueConverterExtension, IValueCon
 	/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
 	/// <param name="culture">The culture to use in the converter.  This is not implemented.</param>
 	/// <returns>True if value equals <see cref="TrueObject"/>, otherwise False.</returns>
-	public object ConvertBack(object? value, Type? targetType, object? parameter, CultureInfo? culture)
+	[return: NotNull]
+	public object? ConvertBack(object? value, Type? targetType, object? parameter, CultureInfo? culture)
```
VladislavAntonyuk, line 56: I am just curious, do we have generic converters?
`IConverter<TInput,TOutput>`
Where we can specify constraints TInput:notnull

src/CommunityToolkit.Maui/Converters/BoolToObjectConverter.shared.cs
brminnick: > Or the Analyzer understands that it is not successful return 

Yup - the analyzer understands that it was not a successful return! It's super cool 💯

src/CommunityToolkit.Maui/Converters/BoolToObjectConverter.shared.cs
brminnick: We do not. 

I like that idea, but I'm hesitant to do that until the MAUI team enables nullable on `IValueConverter` because I honestly don't know which way the team will go: will all parameters be nullable?; will all parameters be not nullable?; will some parameters be nullable?; will the return types be nullable?; etc. 

If we guess wrong, we would end up with a lot of re-work and breaking changes on our APIs. 

With `ICommunityToolkitValueConverter`, because we assume everything in `IValueConverter` is nullable, there won't be any breaking changes to our users if/when the team enables nullable. For example, if the MAUI team decides that `Type targetType` doesn't need to be nullable, there's not breaking changes for our users since we're already handling the null check.

I love this for the future, though! Want to open that as a Discussion for a future Proposal?

VladislavAntonyuk: Yes, let's move it to discussion

src/CommunityToolkit.Maui.UnitTests/Converters/CompareConverter_Tests.cs
```diff
@@ -75,7 +75,7 @@ public void CompareConverterConvert(IComparable value, CompareConverter.Operator
 			ComparingValue = comparingValue
 		};
 
-		var result = compareConverter.Convert(value, typeof(BoolToObjectConverter_Tests), null!, CultureInfo.CurrentCulture);
+		object result = compareConverter.Convert(value, typeof(BoolToObjectConverter_Tests), null!, CultureInfo.CurrentCulture);
```
VladislavAntonyuk, line 78: to be consistent with other files, can you revert to var?

src/CommunityToolkit.Maui/Converters/ItemSelectedEventArgsConverter.shared.cs
```diff
@@ -18,6 +19,7 @@ public class ItemSelectedEventArgsConverter : ValueConverterExtension, IValueCon
 	/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
 	/// <param name="culture">The culture to use in the converter. This is not implemented.</param>
 	/// <returns>A <see cref="SelectedItemChangedEventArgs.SelectedItem"/> object from object of type <see cref="SelectedItemChangedEventArgs"/>.</returns>
+	[return: NotNullIfNotNull("value")]
```
VladislavAntonyuk, line 22: please use `nameof(value)`

src/CommunityToolkit.Maui/Converters/ItemTappedEventArgsConverter.shared.cs
```diff
@@ -18,6 +19,7 @@ public class ItemTappedEventArgsConverter : ValueConverterExtension, IValueConve
 	/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
 	/// <param name="culture">The culture to use in the converter. This is not implemented.</param>
 	/// <returns>A <see cref="ItemTappedEventArgs.Item"/> object from object of type <see cref="ItemTappedEventArgs"/>.</returns>
+	[return: NotNullIfNotNull("value")]
```
VladislavAntonyuk, line 22: please use `nameof(value)`

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui/Converters/ItemSelectedEventArgsConverter.shared.cs
brminnick: I wish I could! I hate hard-coded strings, but since `value` is a method parameter and the attribute is outside the method, we cannot use `nameof(value)` because the variable name `value` hasn't been defined yet.

src/CommunityToolkit.Maui/Converters/ItemTappedEventArgsConverter.shared.cs
brminnick: I wish I could! I hate hard-coded strings, but since `value` is a method parameter and the attribute is outside the method, we cannot use `nameof(value)` because the variable name `value` hasn't been defined yet.

PR #214, opened 2021-12-08
Project reorganization

### LINKED ISSUE #213: [Proposal] Project Reorganization
# Project Reorganization
* [x] Proposed
* [x] Prototype: Not Started
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample

## Link to Discussion

- https://github.com/CommunityToolkit/Maui/discussions/208

## Summary
[summary]: #summary

We will change the project structure to mimic the .NET MAUI structure, placing all code that doesn't use `Microsoft.Maui.Controls` into a new project, `CommunityToolkit.Maui.Core`.

This creates a new NuGet Package, `CommunityToolkit.Maui.Core` that new Toolkits like `CommunityToolkit.Comet` can leverage.

## Motivation
[motivation]: #motivation

Make our implementation as close as possible from .NET MAUI implementation, which will make it possible to easily support frameworks that are built on top of .NET MAUI and promote our features to the .NET MAUI.

## Detailed Design
[design]: #detailed-design

image

## CommunityToolkit.Maui.Core

In general, this project will have all the basement to develop our Toolkit, including some primitive types, interfaces and base classes, base views, and common code. This will be referenced by other Frameworks/Toolkit based on .NET MAUI that wants to have the same features that us.

Here we will have some:

- BaseViews, could be `Views` that will be used by other Views, like [`PaddingButton`](https://github.com/CommunityToolkit/Maui/blob/pj%2Freorganize-project/src/CommunityToolkit.Maui.Core/Platform/iOS/PaddedButton.ios.cs) (that's used by Snackbar) or the [`MCTPopup`](https://github.com/CommunityToolkit/Maui/blob/pj%2Fpopup-port/src/CommunityToolkit.Maui/Platform/Android/MCTPopup.android.cs) that will be a native control implemented in a way that can work with our handler. This same approach is used [here](https://github.com/dotnet/maui/blob/main/src/Core/src/Platform/iOS/MauiLabel.cs)

- Primitives, which will be base types that can be used by everyone, like our [`MathOperator`](https://github.com/CommunityToolkit/Maui/blob/pj%2Freorganize-project/src/CommunityToolkit.Maui.Core/Primitives/MathOperator.shared.cs). So other frameworks may not have the concept of `Behavior` or `Converter` but they can mimic them as helper classes/methods and use our primitives.

- Common Code, this will be all generic code (platform-specific or not) that can be used by other Frameworks/Toolkits

- Layout Managers, were introduced on .NET MAUI and [they live on `Microsoft.Maui.Core`](https://github.com/dotnet/Maui/blob/main/src/Core/src/Layouts/AbsoluteLayoutManager.cs) so makes sense to have our managers on `Core` as well.

- Handlers, on Core will be the most general Handler with the majority of features.

## CommunityToolkit.Maui:

This project has a reference to the `Core` project. Here will live the implementation of our Controls, Views, Behaviors, Animations, etc. In other words, this project will work with the .NET MAUI and will be MVVM friendly. Also, other Toolkits/Frameworks can reference this package if needed.

Here we will have some:

- View Implementation, with `BindableProperties`, support to attach effects, behaviors, triggers, and all that jazz. 

- Platform Configuration, that is Platform-specific features, that can relate to some control - like the `ArrowDirection` that is part of Popup and works just on iOS - or the application itself - like the `StatusBarColorEffect` from XCT.

- Handlers Implementation, We will add to our `PropertyMapper` and/or `CommandMapper` any Platform Configuration that some Handler/View may have. We also can implement here some features that we think will not be great to have on `Core`. [Here is a reference for this](https://github.com/dotnet/maui/tree/main/src/Controls/src/Core/HandlerImpl/Button)

- Layout, will be the implementation of ours custom layouts and will use the Layout Managers on `Core`

## Drawbacks
[drawbacks]: #drawbacks

- Adds an additional NuGet packages
  - Requires adjustments to the CI Pipeline to publish two Nuget Packages 
  - Additional dependencies increase compilation 
- Reorganizing the entire project may lead to unintentional breaking changes
  - This work requires touching every file; some files may be forgotten, some files may be miscopied

pictos:
## Description
This is a housekeeping task, where we will change the project structure to mimic the .NET MAUI structure.

## Detailed Solution

Split the old implementation into two new projects.

I just move all code and change some namespaces, all features and bugs should be the same as `main` branch


src/CommunityToolkit.Maui.Controls/src/CommunityToolkit.Maui.Controls/Alerts/Snackbar/Snackbar.android.cs
```diff
@@ -58,9 +59,10 @@ public async Task Show()
 	{
 		await Dismiss();
 
-		var rootView = Microsoft.Maui.Essentials.Platform.GetCurrentActivity(true).Window?.DecorView.FindViewById(Android.Resource.Id.Content);
+		var rootView = AppBuilderExtensions.Activity?.Window?.DecorView.FindViewById(Android.Resource.Id.Content);
```
pictos, line 62: in the future we can use the Essentials code again.

src/CommunityToolkit.Maui.Controls/src/CommunityToolkit.Maui.Controls/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Maui.Hosting;
+using Microsoft.Maui.LifecycleEvents;
+
+namespace CommunityToolkit.Maui;
+public static class AppBuilderExtensions
+{
+#if ANDROID
+	internal static Android.App.Activity? Activity { get; private set; }
```
pictos, line 13: if we need more code like this we can split it into other platform classes

src/CommunityToolkit.Maui.Controls/src/CommunityToolkit.Maui.Controls/CommunityToolkit.Maui.Controls.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <Compile Remove="**\android\**\*.cs" />
+    <None Include="**\android\**\*.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+  </ItemGroup>
+  <ItemGroup Condition="$(TargetFramework.Contains('-windows')) != true ">
+    <Compile Remove="**\*.windows.cs" />
+    <None Include="**\*.windows.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <Compile Remove="**\windows\**\*.cs" />
+    <None Include="**\windows\**\*.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+  </ItemGroup>
+
```
pictos, line 47: Not sure where I should add the mdoc target, if it was here or in Core csproj or both

src/CommunityToolkit.Maui.Core/AssemblyInfo.shared.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("CommunityToolkit.Maui")]
```
VladislavAntonyuk, line 3: Let's move it to csproj

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
```diff
@@ -12,7 +12,7 @@ public static MauiApp CreateMauiApp()
 	{
 		var builder = MauiApp.CreateBuilder();
 		builder.UseMauiApp<App>();
-
+		builder.UseCommunityToolkit();
```
VladislavAntonyuk, line 15: As far as I remember we want to avoid it.

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
```diff
@@ -0,0 +1,48 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net6.0-android;net6.0-ios;net6.0-maccatalyst</TargetFrameworks>
```
VladislavAntonyuk, line 4: Let's try to keep target framework net6.0 only

src/CommunityToolkit.Maui.Core/Core/ISnackbar.shared.cs
```diff
@@ -2,7 +2,7 @@
 using System.Threading.Tasks;
 using Microsoft.Maui;
 
-namespace CommunityToolkit.Maui.Alerts.Snackbar;
+namespace CommunityToolkit.Maui;
```
VladislavAntonyuk, line 5: Maybe CommunityToolkit.Maui.Core to have the namespace equals assembly name

src/CommunityToolkit.Maui.Core/Platform/iOS/PaddedButton.ios.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using UIKit;
+
+namespace Core.Platform.iOS;
+class PaddedButton : UIButton
```
VladislavAntonyuk, line 5: I think it shouldn't be in Core

src/CommunityToolkit.Maui.Core/Platform/iOS/PaddedLabel.ios.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using CoreGraphics;
+using UIKit;
+
+namespace CommunityToolkit.Maui.Platform;
```
VladislavAntonyuk, line 9: It shouldn't be in Core and should be macios

src/CommunityToolkit.Maui.Core/Primitives/MathOperator.shared.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace CommunityToolkit.Maui.Converters;
+namespace CommunityToolkit.Maui;
```
VladislavAntonyuk, line 3: Let's keep namespace equals to the folder structure

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -85,7 +85,7 @@ public Snackbar()
 		};
 	}
 
-#if NET6_0
+#if !(IOS || ANDROID || MACCATALYST || WINDOWS)
```
VladislavAntonyuk, line 88: Let's remove net6.0 from the target framework. In that case we do not need IF

src/CommunityToolkit.Maui/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+	public static MauiAppBuilder UseCommunityToolkit(this MauiAppBuilder builder)
+	{
+		builder.ConfigureLifecycleEvents(events =>
+		{
+#if ANDROID
+			events.AddAndroid(android =>
+			{
+				android.OnCreate((a, b) =>
+				{
+					Activity = a;
```
VladislavAntonyuk, line 24: We can use Maui Essentials to get the current activity.
Do we need this code as well?

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
pictos: This is something inevitable because we need to register handlers for views that will be ported from Renderers

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
pictos: I can't see a good reason to not target platform code here, even this one being a Core project we may need to have platform code, .NET MAUI does that as I referenced in the PR's issue.

src/CommunityToolkit.Maui.Core/Platform/iOS/PaddedButton.ios.cs
pictos: why?

src/CommunityToolkit.Maui.Core/Primitives/MathOperator.shared.cs
pictos: Again, following the .NET MAUI structure and I can't see the issue having this in the root namespace

src/CommunityToolkit.Maui/AppBuilderExtensions.cs
pictos: No this can be removed, I forgot to delete it

src/CommunityToolkit.Maui.Core/AssemblyInfo.shared.cs
pictos: Is there any difference?

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
pictos: If we do this we will break our unit test :(

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
brminnick: Yup, it's fine to keep it. We'll be using it for the XMLNS Proposal anyways.

src/CommunityToolkit.Maui.Core/AssemblyInfo.shared.cs
brminnick: No, please remove this and delete this file.

If something in the `CommunityToolkit.Maui` project needs something from `CommunityToolkit.Maui.Core`, that API should be public. There should not be anything internal in `CommunityToolkit.Maui.Core` that is needed by `CommunityToolkit.Maui`.

```suggestion

```

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
brminnick: It's clear that there is confusion around what `CommunityToolkit.Maui.Core` is supposed to be.

Let's please pause this PR and return to the [Proposal](https://github.com/CommunityToolkit/Maui/issues/213)  to define `CommunityToolkit.Maui.Core`.

brminnick (changes requested):
Hey Gang,

Let's pause work on this PR until we've defined `CommunityToolkit.Maui.Core`.

Let's continue the discussion in the Proposal: https://github.com/CommunityToolkit/Maui/issues/213

brminnick:
Adding the `do not merge` tag until we've defined the architecture in the [Proposal](https://github.com/CommunityToolkit/Maui/issues/213#issuecomment-991776228)

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
VladislavAntonyuk: It depends on what we put into Core. If there is no platform specific code, we do not need to add specific target framework.
1. It reduces the package size
2. It simplifies dependencies
3. There is no confusion where to put platform specific code.

src/CommunityToolkit.Maui.Core/Platform/iOS/PaddedButton.ios.cs
VladislavAntonyuk: Because this is a control and it is not used in Core

src/CommunityToolkit.Maui.Core/Primitives/MathOperator.shared.cs
VladislavAntonyuk: 1. Warnings
2. Namespace logically separates the code. If we have the same namespace we do not need it at all

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
VladislavAntonyuk: Hmm, is it required to be net6.0 for test project?
Can it be net6.0-platform?

pictos:
@brminnick I would suggest putting our views, on `CommunityToolkit.Maui.Core` in a folder called `Platform/iOS` as .NET MAUI does [here](https://github.com/dotnet/maui/tree/main/src/Core/src/Platform). The reason is that those views are specific for the platform and maybe cause confusion on users, like `Shouldn't I have this view for all supported platforms?`

brminnick:
I find that structure in the .NET MAUI repo confusing.

For example, there is platform-specific code in the `Handlers` folder, e.g. [`ButtonHandler.android.cs`](https://github.com/dotnet/maui/tree/main/src/Core/src/Handlers/Border) which leads to questions like "Why doesn't not in the `Platform` folder"?.

I think it is more intuitive to nix the `Platform` folder because the suffix (`*.android.cs`) already explicitly explains which platform the code runs on.

pictos:
@brminnick let me know if this makes sense for you. And if makes sense and you still don't want to have the platform code, due to confusion that may cause it's fine to me.

> "Why doesn't not in the Platform folder"?

Because platform code is where you will have code only for that platform, like `MauiLabel` is just for iOS other platforms doesn't need to have a special Label implementation. Now you've `Handlers` for all supported platforms, so you can group them by feature (Button, Label, Entry, etc)

brminnick (dismissed):

brminnick:
Having a `Platform` folder is just going to lead to future confusion because there will be platform-specific code that does not go into `Platform`.

Every PR we review, we will have a conversation around "Does this go into `Platform`, or should it, for example, go into `Extensions`?

Let's continue to create folders based on functionality, not based on platform specificity.

src/CommunityToolkit.Maui.Core/Views/PaddedLabel.ios.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using CoreGraphics;
+using UIKit;
+
+namespace CommunityToolkit.Maui.Core.Views;
+
+class PaddedLabel : UILabel
```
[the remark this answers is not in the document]
brminnick: Done ✅

brminnick (dismissed):

src/CommunityToolkit.Maui/Converters/StateToBooleanConverter.shared.cs
```diff
@@ -1,13 +1,30 @@
 ﻿using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
-using CommunityToolkit.Maui.UI.Views;
-using Microsoft.Maui.Controls;
 
 namespace CommunityToolkit.Maui.Converters;
 
+/// <summary>State used by <see cref="StateToBooleanConverter"/> </summary>
+public enum LayoutState
```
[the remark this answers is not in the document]
brminnick: That's fair - we can create a folder called `Enums`.

We have a lot of single-use enums (e.g. Only `StateToBooleanConverter` uses the `LayoutState`) that were sporadically placed (some were in the `Converters` folder, some were in the `Views` folder etc), which is why I moved the single-use enum into the class that uses it.

src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
```diff
@@ -13,13 +13,13 @@ namespace CommunityToolkit.Maui.Converters;
 public abstract class BaseConverterOneWay<TFrom, TTo> : ValueConverterExtension, ICommunityToolkitValueConverter
```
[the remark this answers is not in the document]
brminnick: No, because it relies on `Microsoft.Maui.Controls`.

`BaseConverterOneWay` implements `ICommunityToolkitValueConverter` which implements `Microsoft.Maui.Controls.IValueConterter`

src/CommunityToolkit.Maui/Behaviors/Validators/TextValidationBehavior.shared.cs
```diff
@@ -7,6 +8,24 @@
 
 namespace CommunityToolkit.Maui.Behaviors;
```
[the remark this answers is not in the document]
brminnick: The Toolkit only uses the first folder name in a file's namespace.

E.g. Something in `*/Behaviors/` will use the namespace `CommunityToolkit.Maui.Behaviors`
Likewise, something in `*/Behaviors/Validators/` will also use `CommunityToolkit.Maui.Behaviors`

This helps us organize the code a bit better. We have many validators so it makes sense to organize them together in one folder. But it's mostly to make it easier for XAML users to import the namespaces. Otherwise, the user would need to import multiple namespaces:

```xml
xmlns:behaviors="clr-namespace:CommunityToolkit.Maui.Behaviors;assembly=CommunityToolkit.Maui"
xmlns:validators="clr-namespace:CommunityToolkit.Maui.Behaviors.Validators;assembly=CommunityToolkit.Maui"
```

The same is true in Xamarin Community Toolkit.

src/CommunityToolkit.Maui/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+/// <summary>
+/// Extensions for MauiAppBuilder
+/// </summary>
+public static class AppBuilderExtensions
+{
+	/// <summary>
+	/// Initializes the .NET Maui Community Toolkit Library
+	/// </summary>
+	/// <param name="builder">MauiAppBuilder</param>
+	/// <returns></returns>
```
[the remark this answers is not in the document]
brminnick: We can keep this.

`UseMauiCommunityToolkit` will be required in an upcoming PR, so we may as well add it now and start getting users in the habit of calling `builder.UseMauiApp<App>().UseMauiCommunityToolkit();`

src/CommunityToolkit.Maui.Core/Properties/AssemblyInfo.shared.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
```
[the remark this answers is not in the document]
brminnick: I'd prefer to keep this. 

Sometimes we'll need to be able to test `internal` structures in our Unit tests.

src/CommunityToolkit.Maui.Core/Primitives/MathOperator.shared.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+
+namespace CommunityToolkit.Maui.Core;
+
+/// <summary>
+/// Precedence for Math Opertator
+/// </summary>
+public enum MathOperatorPrecedence
```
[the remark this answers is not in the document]
brminnick: Yup - that's cool. I'll create an `Enums` folder.

src/CommunityToolkit.Maui.Core/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+/// </summary>
+public static class AppBuilderExtensions
+{
+	/// <summary>
+	/// Initializes the .NET MAUI Community Toolkit Core Library
+	/// </summary>
+	/// <param name="builder"></param>
+	/// <returns></returns>
+	public static MauiAppBuilder UseMauiCommunityToolkitCore(this MauiAppBuilder builder)
+	{
```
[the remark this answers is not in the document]
brminnick: > Do we need it right now?

Right now, no. But we will need it in a future PR, and I think it's helpful to start getting users familiar with calling `builder.UseMaiApp<App>.UseMauiCommunityToolkitCore()` to initialize the toolkit.

src/CommunityToolkit.Maui.Controls/src/CommunityToolkit.Maui.Controls/CommunityToolkit.Maui.Controls.csproj
brminnick: I don't think we need mdoc.

We never actually used it for anything in XCT.

We can leave it out and see if any alarms go off requiring it.

src/CommunityToolkit.Maui.Core/Properties/AssemblyInfo.shared.cs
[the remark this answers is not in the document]
VladislavAntonyuk: In that case let’s move it to csproj

src/CommunityToolkit.Maui/AppBuilderExtensions.cs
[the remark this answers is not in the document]
VladislavAntonyuk: I mean returns block. Please add description or remove

src/CommunityToolkit.Maui/Behaviors/Validators/TextValidationBehavior.shared.cs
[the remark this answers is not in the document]
bijington: If we went ahead with this then I don't think consumers will need more than one declaration:

https://github.com/CommunityToolkit/Maui/discussions/144

This is already set on XCT so if we feel it's cleaner/easier to keep the namespace definitions matching the folder structure then it shouldn't cause XAML pain. I guess it might increase the number of required using statements in C# though

src/CommunityToolkit.Maui/AppBuilderExtensions.cs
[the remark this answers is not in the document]
brminnick: Done ✅

src/CommunityToolkit.Maui.Core/Properties/AssemblyInfo.shared.cs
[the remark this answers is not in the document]
brminnick: Done ✅

src/CommunityToolkit.Maui.Core/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+/// <summary>
+/// <see cref="MauiAppBuilder"/> Extensions
+/// </summary>
+public static class AppBuilderExtensions
+{
+	/// <summary>
+	/// Initializes the .NET MAUI Community Toolkit Core Library
+	/// </summary>
+	/// <param name="builder"></param>
```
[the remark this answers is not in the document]
brminnick: Done ✅

src/CommunityToolkit.Maui/Behaviors/Validators/TextValidationBehavior.shared.cs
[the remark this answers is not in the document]
brminnick: True, #144 definitely solves this.

Since adding new namespaces will make life harder for XAML users until #144 is complete, let's do this:
1. I'll create a Proposal to update the namespaces to match the folder names
2. Once #144 is complete, we'll submit + merge a PR for this new Proposal where all namespaces will match their folder structure

src/CommunityToolkit.Maui.Core/Views/PaddedButton.macios.cs
```diff
@@ -0,0 +1,47 @@
...
+public class PaddedButton : UIButton
+{
+	/// <summary>
+	/// Initialize <see cref="PaddedButton"/>
+	/// </summary>
+	/// <param name="leftPadding"></param>
+	/// <param name="topPadding"></param>
+	/// <param name="rightPadding"></param>
+	/// <param name="bottomPadding"></param>
+	public PaddedButton(double leftPadding, double topPadding, double rightPadding, double bottomPadding)
```
pictos, line 18: shouldn't change those values to be `nfloat` as well?

VladislavAntonyuk: it's easier to use double, because nfloat is specific to iOS

pictos: @VladislavAntonyuk but we changed the `PaddedLabel` to accept `nfloat`
https://github.com/CommunityToolkit/Maui/pull/214/commits/4f5ad7dc9c0b441c2d34308b7fa81a7fbb993b1d

VladislavAntonyuk: Padding and CornerRadius use double. it would be easier to have double as parameter and cast it in one place

brminnick: > shouldn't change those values to be nfloat as well?
The difference is how the values are used in `PaddedLabel` vs `PaddedButton`.

`PaddedLabel` uses `public override CGSize IntrinsicContentSize` and `public override void DrawText` which both require `nfloat`.

`PaddedButton` uses `ContentEdgeInsets` which requires a `double.

pictos: I see now, thanks @brminnick for the clarification.

brminnick: Correct. The cast from `double` to `nfloat` has to be done somewhere.

It's more honest for the constructor of `PaddedLabel` to require `nfloat` rather that letting it accept a `double` and immediately casting it to `nfloat` inside the constructor. 

Having `PaddedLabel`' constructor requiring `nfloat` let's the consumer of `PaddedLabel` decide how to cast their input type.

src/CommunityToolkit.Maui/Behaviors/Validators/TextValidationBehavior.shared.cs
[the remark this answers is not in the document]
brminnick: I've created Proposal #219 for this which'll we'll complete upon the completion of #144

src/CommunityToolkit.Maui.Core/Primitives/MathOperator.shared.cs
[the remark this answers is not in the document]
brminnick: Hey Vlad! I started the process of creating an `Enums` folder and moving all of the enums to it. 

But then I quickly realized that this would require a new `xmlns`, making life harder for XAML devs.

I agree that we should move the files, but I think we should wait until #144 is completed to avoid making life hard for XAML devs before #144 makes it simpler.

I've added this task to a Proposal #219, which we'll complete after #144 👍

src/CommunityToolkit.Maui/Converters/StateToBooleanConverter.shared.cs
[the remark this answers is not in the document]
brminnick: Hey Vlad! I started the process of creating an `Enums` folder and moving all of the enums to it. 

But then I quickly realized that this would require a new `xmlns`, making life harder for XAML devs.

I agree that we should move the files, but I think we should wait until #144 is completed to avoid making life hard for XAML devs before #144 makes it simpler.

I've added this task to a Proposal #219, which we'll complete after #144 👍

src/CommunityToolkit.Maui/Behaviors/Validators/TextValidationBehavior.shared.cs
```diff
@@ -7,6 +8,24 @@
...
 
+/// <summary>Flags to indicate what treatment <see cref="ValidationBehavior.Value"/> should receive prior to validation with <see cref="ValidationBehavior"/> or subclasses. This can be used to trim or ignore whitespace for instance. This value might be ignored by a behavior if <see cref="ValidationBehavior.Value"/> isn't of type <see cref="string"/>.</summary>
+[Flags]
+public enum TextDecorationFlags
```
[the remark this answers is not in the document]
brminnick: Hey Vlad! I started the process of creating an `Enums` folder and moving all of the enums to it. 

But then I quickly realized that this would require a new `xmlns`, making life harder for XAML devs.

I agree that we should move the files, but I think we should wait until #144 is completed to avoid making life hard for XAML devs before #144 makes it simpler.

I've added this task to a Proposal #219, which we'll complete after #144 👍

brminnick (dismissed):
Based on feedback from @VladislavAntonyuk + @bijington, I've created + championed [Proposal 219](https://github.com/CommunityToolkit/Maui/issues/219) in which we'll update the namespaces once #144 is complete.

src/CommunityToolkit.Maui.Core/Views/PaddedButton.macios.cs
brminnick: Oh wait - I lied.

@pictos is right - `PaddedButton` uses `new UIContentEdgeInsets` which _does_ require `float`. I'll update `PaddedButton`'s constructor to use `nfloat` as well 👍

brminnick (approved):
Approving again updating the following `macios.cs`  constructors to accept `nfloat`:
- ToastView.macios.cs
- SnackbarView.macios.cs
- PaddedButton.macios.cs
- PaddedLabel.macios.cs

This update helps consumers of these classes to avoid a hidden cast from `double` to `nfloat` that was occurring inside each constructor.

src/CommunityToolkit.Maui.Core/Views/PaddedLabel.macios.cs
```diff
@@ -0,0 +1,62 @@
...
+
+/// <summary>
+/// <see cref="UILabel"/> with Left, Top, Right and Bottom Padding
+/// </summary>
+public class PaddedLabel : UILabel
+{
+	/// <summary>
+	/// Initialize <see cref="PaddedLabel"/>
+	/// </summary>
+	/// <param name="leftPadding"></param>
```
VladislavAntonyuk, line 19: lets remove these params

src/CommunityToolkit.Maui.Core/Views/PaddedButton.macios.cs
```diff
@@ -0,0 +1,47 @@
...
+
+/// <summary>
+/// <see cref="UIButton"/> with Left, Top, Right and Bottom Padding
+/// </summary>
+public class PaddedButton : UIButton
+{
+	/// <summary>
+	/// Initialize <see cref="PaddedButton"/>
+	/// </summary>
+	/// <param name="leftPadding"></param>
```
VladislavAntonyuk, line 14: lets remove these params

src/CommunityToolkit.Maui.Core/Views/Popup/PopupView.macios.cs
```diff
@@ -139,4 +173,87 @@ public override void Draw(CGRect rect)
 			return path.CGPath;
 		}
 	}
+}
+
+static class UIViewExtensions
```
VladislavAntonyuk, line 178: I think it's definitely should be in a separate file

src/CommunityToolkit.Maui.Core/Views/Popup/PopupView.macios.cs
```diff
@@ -2,40 +2,74 @@
...
 using UIKit;
 
-namespace CommunityToolkit.Maui.Views.Popup;
+namespace CommunityToolkit.Maui.Core.Views;
 
-class PopupViewVisualOptions
+/// <summary>
+/// Visual Options for <see cref="PopupView"/>
+/// </summary>
+public class PopupViewVisualOptions
```
VladislavAntonyuk, line 14: why SnackbarOptions have its own file, but this one is here?

VladislavAntonyuk (approved):

pictos:
I approve this PR. 
✅

src/CommunityToolkit.Maui.Core/Views/Popup/PopupView.macios.cs
brminnick: That's fair. I'll move it 👍

src/CommunityToolkit.Maui.Core/Views/Popup/PopupView.macios.cs
brminnick: I kept getting build errors when putting it in the `Extensions` folder. 

I'll give it another shot 👍

pictos: @brminnick Let me know if you still have issues with this one. I can take a look on my end

src/CommunityToolkit.Maui.Core/Views/Popup/PopupView.macios.cs
brminnick: Done ✅

src/CommunityToolkit.Maui.Core/Views/Popup/PopupView.macios.cs
brminnick: Done ✅ 

(I forgot to add the using declarations in the new file last tie 🤦‍♂️)

src/CommunityToolkit.Maui.Core/Views/PaddedButton.macios.cs
brminnick: Done ✅

src/CommunityToolkit.Maui.Core/Views/PaddedLabel.macios.cs
brminnick: Done ✅

brminnick (approved):
Updated with Vlad's recommendations 👍

brminnick:
Shoot - it looks like `dotnet workload install maui` now installs Preview 11 (must've _just_ been released), which is now breaking the build.

I'm on it 👍

brminnick (dismissed):
Approving again with updates for .NET MAUI Preview 11

(The .NET MAUI team moved `System.nfloat`  to `ObjCRuntime.nfloat` 🤷‍♂️)

brminnick:
Dang. Preview 11 broke the Unit Tests too.

I'll circle back to this tomorrow when (hopefully) all of the Preview 11 workloads have been pushed to the public NuGet feeds 👍

VladislavAntonyuk:
> Approving again with updates for .NET MAUI Preview 11
> 
> (The .NET MAUI team moved `System.nfloat` to `ObjCRuntime.nfloat` 🤷‍♂️)

Exactly what I told yesterday 😄 .

brminnick (dismissed):
Approving again after updating the unit test suite for .NET MAUI Preview 11

brminnick (approved):

Issue #122, opened 2021-09-28
[Proposal] UniformItemsLayout


brminnick:
# UniformGrid

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation https://github.com/MicrosoftDocs/CommunityToolkit/pull/75

## Summary
[summary]: #summary

The UniformItemsLayout is like the Grid, with the possibility of multiple rows and columns, but with one important difference: all rows and columns have the same size. That size is determined by the largest width and height of all the child elements. The child element with the largest width does not necessarily have to be the child element with the largest height.

Use the UniformItemsLayout when you need the Grid behavior without the need to specify different sizes for the rows and columns

## Detailed Design
[design]: #detailed-design

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             x:Class="MyLittleApp.MainPage">

    <toolkit:UniformItemsLayout >
        <BoxView Color="Red" />
        <BoxView Color="Yellow" />
        <BoxView Color="Orange" />
        <BoxView Color="Purple" />
    </toolkit:UniformItemsLayout >

</ContentPage>
```

### C# Usage

```cs
class MyPage : ContentPage
{
  public MyPage()
  {
    Content = new UniformItemsLayout
    {
      Children = 
      {
         new BoxView { BackgroundColor = Colors.Red },
         new BoxView { BackgroundColor = Colors.Yellow },
         new BoxView { BackgroundColor = Colors.Orange },
         new BoxView { BackgroundColor = Colors.Purple },
      }
    };
  }
}
```

VladislavAntonyuk:
@DoubleDBE UniformGrid doesn't have these properties and for now, measure and align all items automatically.
But we can easily add properties like MaxRows and MaxColumns.
@brminnick , @jsuarezruiz what do you think?

DoubleDBE:
Thanks for the feedback!
I created a [discussion](https://github.com/CommunityToolkit/Maui/discussions/156)

jsuarezruiz:
[...]

PR #216, opened 2021-12-10
122 uniform grid

VladislavAntonyuk:
## Fixes

#122 

## Description

Migrate UniformGrid to MCT
image


pictos:
Adding the `do-not-merge` because the #214 has priority

src/CommunityToolkit.Maui.UnitTests/Views/UniformGrid_Tests.cs
```diff
@@ -0,0 +1,98 @@
...
+	const double _childCount = 3;
+	const double _childWidth = 20;
+	const double _childHeight = 10;
+
+	readonly UniformGrid _uniformGrid;
+	readonly IView _uniformChild;
+
+	public UniformGridTests()
+	{
+		_uniformChild = Substitute.For<IView>();
```
brminnick, line 27: Can we accomplish this Unit Test without NSubstitute? 

I'd prefer that we test against the actual public APIs instead of creating substitutes.

src/CommunityToolkit.Maui.UnitTests/Views/UniformGrid_Tests.cs
```diff
@@ -0,0 +1,98 @@
...
+		_uniformGrid.MaxRows = 1;
+
+		var actualSize = _uniformGrid.Measure(double.PositiveInfinity, double.PositiveInfinity);
+
+		Assert.Equal(expectedSize, actualSize);
+	}
+
+	void SetupChildrenSize(Size size)
+	{
+		_uniformChild.Measure(double.PositiveInfinity, double.PositiveInfinity).ReturnsForAnyArgs(size);
```
brminnick, line 95: Can we accomplish this Unit Test without NSubstitute?

I'd prefer that we test against the actual public APIs and not use tools like NSubstitute to mock/inject values.

brminnick (changes requested):
Hey Vlad! 

Let's Update 2 Things:
- Move `UniformGrid` from the `Views` folder/namespace to `Layouts` folder + namespace
  - Technically, `UniformGrid` is a `View`, but since it contains `List<IView> Children`, it is more accurate to define it as a `Layout`
- Rearchitect `Grid`
  - Let's inherit from `Layout` instead `Grid`
  - (More thoughts below)

I think we need to rethink how `UniformGrid` is architected.

A lot of `Grid` properties are not being used (like `RowDefinitions` and `ColumnDefinitions`) because `ILayoutManager` is essentially re-drawing the UI on the screen to imitate a `Grid`. 

Perhaps we don't inherit from `Grid`, and instead `Layout` instead.

👇 Check out this example I added to the sample where I add an `IView` to the `UniformGrid` using `uniformGrid.Children.Add()`. 

The child is added to the Grid, but it is always added to the same column + row.

Once the row becomes wider than the screen, it wraps around creating what looks like a new column.

And if I specify a Row/Column for the child, it is ignored; the child is always added to the same row/column:

```cs
var boxView = new BoxView
{
	HeightRequest = widthRequest,
	WidthRequest = heightRequest,
	Color = randomColor
};
Grid.SetRow(boxView, 1);
Grid.SetColumn(boxView, 1);
uniformGrid.Children.Add(boxView);
```

https://user-images.githubusercontent.com/13558917/146099517-4cc9daba-e603-427d-8d35-834b8b31b53a.mp4

src/CommunityToolkit.Maui.UnitTests/Views/UniformGrid_Tests.cs
VladislavAntonyuk: I tried the actual API and it returns 0 for the real size

VladislavAntonyuk:
Hey Brandon, maybe rename the control to UniformLayout?

src/CommunityToolkit.Maui.UnitTests/Views/UniformGrid_Tests.cs
brminnick: Sounds like you've either found a bug, or we need to "Render" the element. 

Perhaps you could mock something up for the rendering, similar to how we created the [`.EnableAnimations()` extension](https://github.com/CommunityToolkit/Maui/blob/0b992330d8c109863b55ffe193bff1415c6bc905/src/CommunityToolkit.Maui.UnitTests/Mocks/AnimationExtensions.cs#L14) method that allows us to animate an `IView`.

brminnick:
Yea! Naming is tough, but I like where you're going with it!

Along with `Uniform` to let users know everything will be spaced evenly, maybe we also add something that lets users know we'll be putting their controls into rows + columns? 

Here's a couple more ideas:
- `UniformRowsColumnsLayout`
- `EvenlySpacedGridLayout`
- `AutoSpacedLayout`
- `EvenlySpacedRowsColumnsLayout`
- `AutoSpacedRowsColumnsLayout`

I'm not sold on any of these, but sometimes it helps to throw a bunch of names out there 😃

bijington:
> Yea! Naming is tough, but I like where you're going with it!

You are telling me! Child 1 and Child 2 would definitely agree with you 😆 

> 
> Along with `Uniform` to let users know everything will be spaced evenly, maybe we also add something that lets users know we'll be putting their controls into rows + columns? 
> 
> Here's a couple more ideas:
> - `UniformRowsColumnsLayout`
> - `EvenlySpacedGridLayout`
> - `AutoSpacedLayout`
> - `EvenlySpacedRowsColumnsLayout`
> - `AutoSpacedRowsColumnsLayout`
> 
> I'm not sold on any of these, but sometimes it helps to throw a bunch of names out there 😃 

I still quite like `UniformGrid` as the grid suffix implies the rows/columns part. Some other possible names could be:

- `UniformItemsLayout`
- that's all I've got

src/CommunityToolkit.Maui.UnitTests/Views/UniformGrid_Tests.cs
VladislavAntonyuk: I am not sure if it is possible. .NET MAUI team also uses NSubstiture for Measure tests: https://github.com/dotnet/maui/blob/main/src/Controls/tests/Core.UnitTests/Layouts/LayoutCompatTests.cs

VladislavAntonyuk: well, I found a way to remove NSubstitute dependency. I will update PR in a few minutes

brminnick (approved):
Thanks Vlad! I love the new name, love the new `IUniformItemsLayout` interface, and am impressed at the elegance of the implementation of `UniformItemsLayout` 💯 👏

brminnick:
Dang - looks like Preview 13 was just released which scrolled our pipeline.

image

brminnick:
Seems we have to wait for the .NET Team to publish specific NuGet Packages first, like [`Microsoft.Extensions.Configuration v6.0.2-mauipre.1.22102.15`](https://www.nuget.org/packages/Microsoft.Extensions.Configuration/6.0.2-mauipre.1.22054.8)

> CommunityToolkit.Maui/src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj : error NU1102: Unable to find package Microsoft.Extensions.Configuration with version (>= 6.0.2-mauipre.1.22102.15) [/Users/bramin/GitHub/CommunityToolkit.Maui/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj]

image

brminnick (approved):
Approving again now that we've updated the CI Pipeline for .NET MAUI Preview 13

PR #220, opened 2021-12-13
Addition of contributing guide

bijington:
## Description

Adds a contributing guidelines document largely inspired by [https://github.com/xamarin/XamarinCommunityToolkit/blob/main/CONTRIBUTING.md](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/CONTRIBUTING.md)


CONTRIBUTING.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Bug Fixes
+
+If you're looking for something to fix, please browse [open issues](https://github.com/CommunityToolkit/Maui/issues). 
+
+Follow the style used by the [.NET Foundation](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md), with two primary exceptions:
+
+- We do not use the `private` keyword as it is the default accessibility level in C#.
+- We will **not** use `_` or `s_` as a prefix for internal or private field names
+- We will use `camelCaseFieldName` for naming internal or private fields in both instance and static implementations
```
bijington, line 144: This is from XCT but it appears we have a inconsistent approach already in MCT:

* `_` prefix -> https://github.com/CommunityToolkit/Maui/blob/00f8f8de9e81423a43800404c1d6ba932f8a7690/src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs#L11
* `camelCaseFieldName` -> https://github.com/CommunityToolkit/Maui/blob/00f8f8de9e81423a43800404c1d6ba932f8a7690/src/CommunityToolkit.Maui/Behaviors/BaseBehavior.shared.cs#L21

CONTRIBUTING.md
```diff
@@ -0,0 +1,158 @@
+# Contributing
```
bijington, line 1: TODO: add text and flowchart diagram indicating where changes should be located based on #214

bijington:
Also investigate `.editorconfig` changes to make it easier to follow with the guidelines

brminnick:
+1 to `.editorconfig` rules that'll automatically enforce these 💯 

Can you also add not to use the [Null-Forgiving Operator, `!`](https://docs.microsoft.com/dotnet/csharp/language-reference/operators/null-forgiving).

VladislavAntonyuk:
> +1 to `.editorconfig` rules that'll automatically enforce these 💯
> 
> Can you also add not to use the [Null-Forgiving Operator, `!`](https://docs.microsoft.com/dotnet/csharp/language-reference/operators/null-forgiving).

I have one. We can use it as start point:
[editorconfig.zip](https://github.com/CommunityToolkit/Maui/files/7707474/editorconfig.zip)

CONTRIBUTING.md
VladislavAntonyuk: can we fix it in https://github.com/CommunityToolkit/Maui/pull/214 or it will be a new PR with editorConfig rules?

bijington:
> Can you also add not to use the [Null-Forgiving Operator, `!`](https://docs.microsoft.com/dotnet/csharp/language-reference/operators/null-forgiving). 

There is a bullet point in the guidelines stating we should avoid using it. I will see if we can prevent its use entirely with .editorconfig

bijington:
> > +1 to `.editorconfig` rules that'll automatically enforce these 💯
> 
> > 
> 
> > Can you also add not to use the [Null-Forgiving Operator, `!`](https://docs.microsoft.com/dotnet/csharp/language-reference/operators/null-forgiving).
> 
> 
> 
> I have one. We can use it as start point:
> 
> [editorconfig.zip](https://github.com/CommunityToolkit/Maui/files/7707474/editorconfig.zip)
> 
> 
> 
> 

That's really helpful thank you!

CONTRIBUTING.md
bijington: I don't mind. If #214 is near completion I'm happy to include in this or whichever PR includes the .editorconfig addition.

I guess we should decide which approach we prefer first 😃

brminnick: Let's make those changes here in this PR.

#214 is a massive PR and just got 2 Approvals. I'm exhausted just thinking about reviewing it again 😅

brminnick: For what it's worth, I prefer prefixing fields with `_`.

That's just my personal preference (makes it easy to determine at a glance whether a variable's scope is local or global). It's not a hill I'm willing to die on, though, if everyone else prefers to nix the `_` prefix.

bijington: I lean the other way but again I'm not going to fight to the death on it. If it means we have fewer differences to the linked standard from dotnet foundation then that's probably a better approach. To clarify this would be in favour of the `_` prefix

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.android.cs
```diff
@@ -63,19 +63,19 @@ public async Task Show()
 		if (rootView is null)
 			throw new NotSupportedException("Unable to retrieve snackbar parent");
 
-		_nativeSnackbar = AndroidSnackbar.Make(rootView, Text, (int)Duration.TotalMilliseconds);
-		var snackbarView = _nativeSnackbar.View;
+		nativeSnackbar = AndroidSnackbar.Make(rootView, Text, (int)Duration.TotalMilliseconds);
+		var snackbarView = nativeSnackbar.View;
 
 		if (Anchor is not Page)
 		{
```
VladislavAntonyuk, line 70: it would be nice if we define the style for `if` blocks. somewhere we use braces (line 70), somewhere we skip them (line 64).

VladislavAntonyuk:
what do you think about `dotnet format` in git pre-commit hooks? it will automatically validate and fix rules

brminnick:
> what do you think about dotnet format in git pre-commit hooks? it will automatically validate and fix rules

Love it! How do we do that?

VladislavAntonyuk:
> > what do you think about dotnet format in git pre-commit hooks? it will automatically validate and fix rules
> 
> Love it! How do we do that?

https://vladislavantonyuk.azurewebsites.net/articles/How-to-automatically-keep-your-code-clean

bijington:
> > > what do you think about dotnet format in git pre-commit hooks? it will automatically validate and fix rules
> > 
> > Love it! How do we do that?
> 
> https://vladislavantonyuk.azurewebsites.net/articles/How-to-automatically-keep-your-code-clean

I like the idea as it means we shouldn't need the bot that does the same thing and it should result in fewer PRs to review.

Is there any risk in it changing things so much from what the author of the commit originally wrote that it may confuse things?

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.android.cs
bijington: Definitely! Great spot! Time for another vote I think

VladislavAntonyuk:
> > > > what do you think about dotnet format in git pre-commit hooks? it will automatically validate and fix rules
> > > 
> > > 
> > > Love it! How do we do that?
> > 
> > 
> > https://vladislavantonyuk.azurewebsites.net/articles/How-to-automatically-keep-your-code-clean
> 
> I like the idea as it means we shouldn't need the bot that does the same thing and it should result in fewer PRs to review.
> 
> Is there any risk in it changing things so much from what the author of the commit originally wrote that it may confuse things?

We can describe it in the guide (with the link to the issue/discussion) but implement it a bit later.

bijington:
I believe this is ready for review. I have currently changed to encourage/force:

- camelCaseNaming
- `is null` over `== null`
- `{ }` after `if`, `for`, etc.

I have not yet managed to get the `!` null forgiving operator banned. This package https://github.com/tom-englert/Nullable.Extended allows us to but the correcting behaviour is to provide a commit to omit it rather than remove it which feels wrong for this use. Thoughts?

VladislavAntonyuk (approved):

.editorconfig
```diff
@@ -50,4 +50,58 @@ dotnet_diagnostic.CS4014.severity = error
...
+
+## Instance fields are camelCase
+dotnet_naming_rule.instance_fields_should_be_camel_case.severity = error
+dotnet_naming_rule.instance_fields_should_be_camel_case.symbols = instance_fields
+dotnet_naming_rule.instance_fields_should_be_camel_case.style = instance_field_style
+
+dotnet_naming_symbols.instance_fields.applicable_kinds = field
+
+dotnet_naming_style.instance_field_style.capitalization = camel_case
+dotnet_naming_style.instance_field_style.required_prefix =
```
brminnick, line 95: Are we missing something after the equals sign? Or does leaving it blank ensure that we don't use `_` for fields?

bijington: This is left blank to prevent the _ prefix as you suggest

PR #222, opened 2021-12-15
[🔥Hot Fix🔥] Adding WinUI reference

pictos:
- Fixes #
On project reorganization PR we forgot to add the WinUI reference, so I'm doing it right now❣

### DIFF
--- a/src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
+++ b/src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
@@ -75,6 +75,10 @@
     <None Include="**\windows\**\*.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
   </ItemGroup>
 
+  <ItemGroup Condition="$(TargetFramework.Contains('-windows'))">
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.0.0" />
+  </ItemGroup>
+
   <ItemGroup>
     <None Include="..\..\build\nuget.png" PackagePath="icon.png" Pack="true" />
   </ItemGroup>
--- a/src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
+++ b/src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
@@ -75,6 +75,7 @@
 
   <ItemGroup Condition="$(TargetFramework.Contains('-windows'))">
     <PackageReference Include="CommunityToolkit.WinUI.Notifications" Version="7.1.2" />
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.0.0" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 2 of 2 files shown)


src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -75,6 +75,7 @@
 
   <ItemGroup Condition="$(TargetFramework.Contains('-windows'))">
     <PackageReference Include="CommunityToolkit.WinUI.Notifications" Version="7.1.2" />
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.0.0" />
```
brminnick, line 78: Should this go in [`CommunityToolkit.Maui.Core.csproj`](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj)?

pictos: good question... AFAIK we need this reference if we need to access the platform controls. At the end of the day I think that is fine to add it on `Core` just in case

brminnick: Right on! Let's remove it from here and move it to `Core`, then. 

Since `CommunityToolkit.Maui` depends on `CommunityToolkit.Maui.Core`, it'll inherit the `Microsoft.WindowsAppSDK` NuGet package.

pictos: Done

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
```diff
@@ -75,6 +75,10 @@
     <None Include="**\windows\**\*.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
   </ItemGroup>
 
+  <ItemGroup Condition="$(TargetFramework.Contains('-windows'))">
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.0.0" />
```
VladislavAntonyuk, line 79: do we need to include ```<PackageReference Include="Microsoft.Graphics.Win2D" Version="1.0.0.30" />``` as well?

pictos: TBH I don't if we need it on the lib. I'll try to find something

pictos: No, we don't need it, at least not for now. As you can see in the readme is just an API to draw 2D graphics.
- https://github.com/microsoft/Win2D

brminnick (changes requested):
Let's remove it from the Sample app too:
https://github.com/CommunityToolkit/Maui/blob/0b992330d8c109863b55ffe193bff1415c6bc905/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj#L54

VladislavAntonyuk (approved):

brminnick (approved):

PR #223, opened 2021-12-15
Add Additional XML Warnings to `<WarningsAsErrors>` in `Directory.Build.props`

brminnick:
## Description

This PR ensures that future PRs will continue to have properly formatted XML Documentation

## Detailed Solution

Adds the following compiler warnings to `<WarningsAsErrors>` in `DirectoryBuild.props`:
- CS1573: Parameter has no matching param tag in the XML comment 
- CS1734: XML comment has a paramref tag, but there is no parameter by that name

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,7 +6,7 @@
 		<!--CS1574: XML comment has cref attribute that could not be resolved-->
  		<!--CS1573: Parameter has no matching param tag in the XML comment -->
  		<!--CS1734: XML comment has a paramref tag, but there is no parameter by that name-->
- 		<WarningsAsErrors>nullable,CS1574,CS1573,CS137</WarningsAsErrors>
+ 		<WarningsAsErrors>nullable,CS1574,CS1573,CS1734</WarningsAsErrors>
 		<ProduceReferenceAssembly>True</ProduceReferenceAssembly>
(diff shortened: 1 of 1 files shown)


VladislavAntonyuk (approved):

Directory.Build.props
```diff
@@ -3,7 +3,10 @@
 	<PropertyGroup>
 		<LangVersion>latest</LangVersion>
 		<Nullable>enable</Nullable>
-		<WarningsAsErrors>nullable,CS1574</WarningsAsErrors><!--CS1574: XML comment has cref attribute that could not be resolved-->
+		<!--CS1574: XML comment has cref attribute that could not be resolved-->
+ 		<!--CS1573: Parameter has no matching param tag in the XML comment -->
+ 		<!--CS1734: XML comment has a paramref tag, but there is no parameter by that name-->
+ 		<WarningsAsErrors>nullable,CS1574,CS1573,CS1734</WarningsAsErrors>
```
VladislavAntonyuk, line 9: can we enable all warnings as errors? but use Exclude with NoWarn

brminnick: Interesting thought! 

I'm not against it. I'm trying to think of scenarios where this could cause problems, but it may be something that we may need to try it first.

Issue #211, opened 2021-11-29
[Proposal] TextColorTo extension for ITextStyle


PieEatingNinjas:
# TextColorTo extension for ITextStyle

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation

## Link to Discussion

[Extend ColorAnimationExtensions with 'TextColorTo' for ITextStyle](https://github.com/CommunityToolkit/Maui/discussions/203)

## Summary
Extend the `ColorAnimationExtenions` class with a `TextColorTo` extension method on the `ITextStyle` interface, allowing developers to animate the color transition of the `TextColor` property of controls implementing `ITextStyle`. 

## Motivation
Currently the `ColorAnimationExtensions` class has 1 method `BackgroundColorTo` which helps animating the color transition of a `VisualElement`'s `BackgroundColor`. Almost the exact same code can be re-used for this proposed `TextColorTo` extension method. Effort is low, value is pretty descent as `ITextStyle` is implemented by several controls. 
## Detailed Design
```CSharp
public static Task<bool> TextColorTo(this ITextStyle element, Color color, uint rate = 16u, uint length = 250u, Easing? easing = null)
{
	ArgumentNullException.ThrowIfNull(element);
	ArgumentNullException.ThrowIfNull(color);

	element.TextColor ??= Colors.Transparent;

	var animationCompletionSource = new TaskCompletionSource<bool>();

	try
	{
		new Animation
		{
			{ 0, 1, GetRedTransformAnimation(element, color.Red) },
			{ 0, 1, GetGreenTransformAnimation(element, color.Green) },
			{ 0, 1, GetBlueTransformAnimation(element, color.Blue) },
			{ 0, 1, GetAlphaTransformAnimation(element, color.Alpha) },
		}
		.Commit(element, nameof(TextColorTo), rate, length, easing, (d, b) => animationCompletionSource.SetResult(true));
	}
	catch (ArgumentException aex)
	{
		animationCompletionSource.SetResult(false);
	}

	return animationCompletionSource.Task;
}

static Animation GetRedTransformAnimation(ITextStyle element, float targetRed) =>
	new(v => element.TextColor = element.TextColor.WithRed(v), element.TextColor.Red, targetRed);

[...]

PR #224, opened 2021-12-18
Implement `TextColorTo` 

brminnick:
## Fixes
`TextColorTo` Implementation for #211 

## Description

This PR extends the `ColorAnimationExtensions`, adding a `TextColorTo` extension method for each class in `Microsoft.Maui.Controls` that implements the `ITextStyle` interface

`TextColorTo` allows developers to animate the color transition for  `ITextStyle.TextColor`.

## Detailed Solution

Because `ITextStyle` is a read-only interface (its properties only implement `get` and do not implement `set`), to be able to create a `TextColorTo` extension method where we modify `TextColor` (e.g. `label.TextColor = Colors.Red`) for each class that implements `ITextStyle`, we must create a `TextColorTo` extension method for each class that implements `ITextStyleTo`.

Using Source Generators ensures that all current (and future) `Microsoft.Maui.Controls` classes that implement `ITextStyle` do (and will) have support for `TextColorTo`.

### CommunityToolkit.Maui
- ColorAnimationExtensions.shared.cs
  - Move `static Animation GetTransformAnimation` methods to local functions
    - This ensures no method conflicts with `TextColorTo`'s `GetTransformAnimation` methods
  - Define as a `partial` class
    - Allows the `TextColorTo` source generator to create additional `public static partial class ColorAnimationExtensions` classes

### CommunityToolkit.Maui.SourceGenerators
This is a new CSPROJ allowing us to implement source generators for `CommunityToolkit.Maui`

- TextColorToGenerator.cs
  - Generates `public static partial class ColorAnimationExtensions`
  - Generates `public static Task<bool> TextColorTo()` extension method for each `Microsoft.Maui.Controls` class that implements `ITextStyle`
- NamedTypeSymbolExtensions.cs
  - Extension methods for `INamespaceSymbol`
  - `IEnumerable<INamedTypeSymbol> GetNamedTypeSymbols()` returns all nested `INamedTypeSymbol`s in an `INamespaceSymbol`
- `NamedTypeSymbolExtensions.cs`
  - Extension methods for `INamedTypeSymbol`
  - `bool ImplementsInterfaceOrBaseClass` returns `true` if a an `INamedTypeSymbol` implements a specified type

### CommunityToolkit.Maui.UnitTests
- ColorAnimationExtensions_Test.cs
  - Added `TextColorTo` Unit Tests


src/CommunityToolkit.Maui.SourceGenerators/CommunityToolkit.Maui.SourceGenerators.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
```
pictos, line 6: Let's adds this prop to the SG
```suggestion
    <IsRoslynComponent>true</IsRoslynComponent>
  </PropertyGroup>
```

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,113 @@
...
+{
+	public void Execute(GeneratorExecutionContext context)
+	{
+		var textStyleClassNameList = new List<string>();
+
+		var mauiControlsAssemblySymbol = context.Compilation.SourceModule.ReferencedAssemblySymbols.Single(q => q.Name == "Microsoft.Maui.Controls");
+
+		foreach (var namedTypeSymbol in mauiControlsAssemblySymbol.GlobalNamespace.GetNamedTypeSymbols())
+		{
+			if (namedTypeSymbol.ImplementsInterfaceOrBaseClass("ITextStyle"))
```
pictos, line 24: Looks like this just scans the .NET MAUI source, that way this will ignore if the user creates his own control that implements the `ITextStyle` interface.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,113 @@
...
+			{
+				textStyleClassNameList.Add(namedTypeSymbol.Name);
+			}
+		}
+
+		foreach (var textStyleClassName in textStyleClassNameList)
+		{
+			var textColorToBuilder = @"
```
pictos, line 32: we could make this a constant, and replace all `textStyleClassName` with `{0}`, and here we just call  `string.Format(textColorToBuilder, textStyleClassName)`.

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -78,6 +78,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\CommunityToolkit.Maui.SourceGenerators\CommunityToolkit.Maui.SourceGenerators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```
pictos, line 81: can you add this line inside the `PropertyGroup`?
`<CompilerGeneratedFilesOutputPath>$(BaseIntermediateOutputPath)Generated</CompilerGeneratedFilesOutputPath>`

This will create a folder inside `obj` and we can inspect the generated file there... That's useful because sometimes VS stops to generate files and you need to restart it to see it under solution explorer

pictos (commented):
Great insight using SG to solve this issue. I've some suggestions for you

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: Ah - nice! I read that the generated files should appear in `obj`, but I couldn't find them 😆

pictos: Not sure if you know, but you can see it on VS as well... 
image

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
brminnick: Honest question - does that improve performance since we are still injecting the textStyleClassName string?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
brminnick: Dang - I didn't even think about user-generated `ITextStyle` controls.

How do we handle that?

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: I still don't see the files listed in the `Analyzers` drop down. 

Did I implement this correctly? https://github.com/CommunityToolkit/Maui/pull/224/commits/adf99bf15468dc70064aac5b5af14bda62912a1c

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: I believe that will not improve the performance, since we will create a new string every time. My suggestion was to reduce the method lines, moving this string to another place

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
pictos: @brminnick try to re-open the VS

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: we will need to implement a `SyntaxReceiver`, that way we will listen every time that users do some changes in the code... So to improve perf. we can try (I really don't know if that will be possible to do) to move this implementation to the `Initialize` method, this method is executed just once and the Maui types that implement `ITextStyle` will not change.
This will improve perf.

Then, on the `Execute` method we take about the user controls that implements `ITextStyle`.

pictos: I can take a look at that with you next week. Anyway, if you want to play with it on your own [here is a reference for you](https://github.com/pictos/SourceGenerators/blob/main/StructGenerator/StructGenerator/StructGen.cs)

This is a source generator that I did that implements the `IEquatable<T>` interface on structs. So the only thing that the user needs to do to trigger the SG is to create a `partial struct` and implement `IEquatable<T>`. I can see that is the same thing that we want here :D

brminnick: I've also seen `IIncrementalGenerator` which is supposed to improve performance. 

https://andrewlock.net/exploring-dotnet-6-part-9-source-generator-updates-incremental-generators/

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: image
Coooool!

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: I didn't try them yet. But yeah, that is more performant than normal `SourceGenerators`. We can try to implement it here

pictos:
@brminnick I did some experiments with Incremental Source Generators and I could make it work. 100% that will be better than using the old Source Generators. Let me know if you want me to implement it or need any guidance.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: @brminnick Since the Initialize method doesn't support any code generation we can create a collection with all Maui types that implement ITextStyle and generate the method for those types during the initialization. Let me know what do you think about it

brminnick:
FYI - I'm converting this to a Draft PR until I get the source converter working with the NuGet Package to avoid it getting accidentally merged

brminnick:
Screen Shot 2022-01-02 at 3 09 47 PM
Confirmed the Source Generator for user's custom `ITextStyle` controls via the NuGet Package 👍

azure-pipelines.yml
```diff
@@ -67,21 +67,21 @@ jobs:
         condition: and(succeeded(), eq(variables['build.reason'], 'PullRequest'))
       # build, test and pack the packages
       - task: VSBuild@1
-        displayName: 'Build and pack Community Toolkit'
+        displayName: 'Build Community Toolkit Sample'
```
brminnick, line 70: FYI - I had to reorder the Build steps because the packing the NuGet Package requires `CommunityToolkit.Maui.SourceGenerators.dll`.

`dotnet pack` will fail if `CommunityToolkit.Maui.SourceGenerators.dll` doesn't already exist before `dotnet pack` is executed.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,144 @@
...
+			{
+				if (namedTypeSymbol.ImplementsInterfaceOrBaseClass("ITextStyle"))
+				{
+					textStyleClassList.Add((namedTypeSymbol.Name, namedTypeSymbol.ContainingNamespace.ToString()));
+				}
+			}
+			// Collect All Classes in User Library that Implement ITextStyle
+			foreach (var classDeclarationSyntax in collectedValues.CandidateClassesProvider)
+			{
+				if (classDeclarationSyntax.BaseList?.Types.Any(x => x.ToString() == "ITextStyle") is true)
```
pictos, line 45: **Strings are evil**
So, if the user has his own interface `ITextStyle` it will be added to `textStyleClassList`.
In other words, let say that my project or a lib that I consume has an interface called `ITextStyle`, as you can see in the image below

image


Them when we iterate on our SG we will have that class to generate, as you can see in the image below
image

And then our SG will fail.

I would suggest you work with `Symbols` instead of `strings`, the solution will be something like:

```csharp
 var textStyleClassList = new List<ClassDeclarationSyntax>();
 var iTextStyleSymbol = compiliation.GetTypeByMetadataName("Microsoft.Maui.ITextStyle");

 foreach (var classDeclarationSyntax in collectedValues.CandidateClassesProvider)
 {
  var model = compiliation.GetSemanticModel(classDeclarationSyntax.SyntaxTree);
  var classSymbol = model.GetDeclaredSymbol(classDeclarationSyntax);

  if (classSymbol?.AllInterfaces.Any(i => i.OriginalDefinition.Equals(iTextStyleSymbol, SymbolEqualityComparer.Default)) ?? false)
   textStyleClassList.Add(classDeclarationSyntax);
}
```

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,144 @@
...
+			(context, cancellationToken) => (ClassDeclarationSyntax)context.Node);
+
+		// Get Microsoft.Maui.Controls Assymbly Symbol
+		var mauiControlsAssemblySymbolProvider = context.CompilationProvider.Select(
+			(compilation, token) => compilation.SourceModule.ReferencedAssemblySymbols.Single(q => q.Name == "Microsoft.Maui.Controls"));
+
+		var inputs = candidateClassesProvider.Collect()
+						.Combine(mauiControlsAssemblySymbolProvider)
+						.Select((combined, cancellationToken) => (CandidateClassesProvider: combined.Left, MauiControlsAssemblySymbolProvider: combined.Right));
+		context.RegisterSourceOutput(inputs, (context, collectedValues) =>
```
pictos, line 30: we don't have access to the `Compilation` object here, can you double-check that? Maybe I missed something.

pictos (changes requested):
@brminnick I did some tries on my end and I found this scenario

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
brminnick: Argh! I've literally been pulling my hair out  for the last 2 hours trying to figure this out to find inherited interfaces using `IIncrementalGenerator `.

You're right that `AllInterfaces` does the trick, but I've only found samples using `ISourceGenerator` that demonstrate `AllInterfaces`.

There's gotta be a way to do this...

brminnick:
@pictos I got it working with custom interfaces!! 

It is ready for your review (again! 😃).

image

Regarding avoiding strings, I agree. However, I'm unable to reference MAUI types because the Source Generator doesn't have a dependency on MAUI, and I'm not quite sure how to avoid using strings? 

Let me know if you have any ideas! If not, maybe we merge this PR and open a Discussion for how to avoid hard-coded strings in `TextColorToGenerator` 👍

pictos:
@brminnick I have some ideas on how to fix it, I'll play with this on my end and what do you prefer(?):
1. Open a PR against your branch;
2. Commit directly to this PR

I can see that we have a lot of room to use SGs and with that this `csproj` will grow over time, so I would say that we can create a solid basement from this PR

brminnick:
Thanks @pictos!

If you're able to figure it out and get it working, go ahead and commit it to this PR 👍

PieEatingNinjas:
I really love how this is implemented now! 
Awesome work @brminnick and @pictos !

brminnick:
(👆Just making sure the pipeline still works with today's release of .NET MAUI Preview 11 + VS 2022 Preview)

brminnick:
@pictos FYI - I just learned that the `IIncrementalGenerator` docs are available in this PR: https://github.com/dotnet/roslyn/pull/58344

pictos:
@brminnick thanks for pointing that out. I did some work yesterday and could reduce the code to generate the files, I'll try to push the changes later today with the p11 bits. Also I'll see how easy will be to create a UnitTest project just for SG, since I plan to add some `Diagnostics` features

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+	const string iAnimatableInterface = "Microsoft.Maui.Controls.IAnimatable";
+	const string mauiControlsAssembly = "Microsoft.Maui.Controls";
+
+	public void Initialize(IncrementalGeneratorInitializationContext context)
+	{
+		// Get All Classes in User Library
+		var userGeneratedClassesProvider = context.SyntaxProvider.CreateSyntaxProvider(
+			static (syntaxNode, cancellationToken) => syntaxNode is ClassDeclarationSyntax { BaseList: not null },
+			static (context, cancellationToken) => (ClassDeclarationSyntax)context.Node);
+
```
pictos, line 25: I removed the user interface part because we didn't use it anywhere.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+			static (compilation, token) => compilation.SourceModule.ReferencedAssemblySymbols.Single(q => q.Name == mauiControlsAssembly));
+
+		var inputs = userGeneratedClassesProvider.Collect()
+						.Combine(mauiControlsAssemblySymbolProvider)
+						.Select(static (combined, cancellationToken) => (UserGeneratedClassesProvider: combined.Left, MauiControlsAssemblySymbolProvider: combined.Right))
+						.Combine(context.CompilationProvider)
+						.Select(static (combined, cancellationToken) => (combined.Left.UserGeneratedClassesProvider, combined.Left.MauiControlsAssemblySymbolProvider, Compilation: combined.Right));
+
+		context.RegisterSourceOutput(inputs, (context, collectedValues) =>
+		Execute(context, collectedValues.Compilation, collectedValues.UserGeneratedClassesProvider, collectedValues.MauiControlsAssemblySymbolProvider));
```
pictos, line 37: Extracted to a separated method to make it more clear on how to access the parameters

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	static void Execute(SourceProductionContext context, Compilation compilation, ImmutableArray<ClassDeclarationSyntax> userGeneratedClassesProvider, IAssemblySymbol mauiControlsAssemblySymbolProvider)
+	{
+		var textStyleSymbol = compilation.GetTypeByMetadataName(iTextStyleInterface);
+		var iAnimatableSymbol = compilation.GetTypeByMetadataName(iAnimatableInterface);
+
+		if (textStyleSymbol is null || iAnimatableSymbol is null)
+			return;
```
pictos, line 46: TODO: Add a diagnostic telling the user that he doesn't have a reference to the .NET MAUI

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+
+		var textStyleClassList = new List<(string ClassName, string ClassAcessModifier, string Namespace)>();
+
+		// Collect Microsoft.Maui.Controls that Implement ITextStyle
+		var mauiTextStyleImplementors = mauiControlsAssemblySymbolProvider.GlobalNamespace.GetNamedTypeSymbols().Where(x => x.AllInterfaces.Contains(textStyleSymbol, SymbolEqualityComparer.Default)
+				&& x.AllInterfaces.Contains(iAnimatableSymbol, SymbolEqualityComparer.Default));
```
pictos, line 52: For the Animate part, we need to make sure that the control implements the` IAnimatable` interface

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+
+		foreach (var namedTypeSymbol in mauiTextStyleImplementors)
+			textStyleClassList.Add((namedTypeSymbol.Name, "public", namedTypeSymbol.ContainingNamespace.ToDisplayString()));
+
+		// Collect All Classes in User Library that Implement ITextStyle
+		foreach (var classDeclarationSyntax in userGeneratedClassesProvider)
+		{
+			var declarationSymbol = compilation.GetSymbol<INamedTypeSymbol>(classDeclarationSyntax);
+			if (declarationSymbol is null)
+				continue;
```
pictos, line 62: TODO: Add a diagnostic telling that something was wrong with the declarationSyntax

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+			if (declarationSymbol.AllInterfaces.Contains(textStyleSymbol, SymbolEqualityComparer.Default)
+				&& declarationSymbol.AllInterfaces.Contains(iAnimatableSymbol, SymbolEqualityComparer.Default))
+			{
+				textStyleClassList.Add((declarationSymbol.Name, GetClassAccessModifier(declarationSymbol), declarationSymbol.ContainingNamespace.ToDisplayString()));
+			}
+		}
+
+		var options = ((CSharpCompilation)compilation).SyntaxTrees[0].Options as CSharpParseOptions;
```
pictos, line 70: Make sure that the generated file will follow the user options for C#

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+using Microsoft.Maui;
+using Microsoft.Maui.Controls;
+using Microsoft.Maui.Graphics;
+
+namespace " + textStyleClass.Namespace + @";
+
+" + textStyleClass.ClassAcessModifier + @" static partial class ColorAnimationExtensions_" + textStyleClass.ClassName + @"
+{
+	/// <summary>
+	/// Animates the TextColor of an <see cref=""Microsoft.Maui.ITextStyle""/> to the given color
```
pictos, line 89: I added the `FullNamespace` here and also on the generated code. Other SourceGenerators do the same, which is kind of a pattern to avoid naming collision.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+
+		static Animation GetBlueTransformAnimation(" + textStyleClass.Namespace + "." + textStyleClass.ClassName + @"  element, float targetBlue) =>
+			new(v => element.TextColor = element.TextColor.WithBlue(v), element.TextColor.Blue, targetBlue);
+
+		static Animation GetAlphaTransformAnimation(" + textStyleClass.Namespace + "." + textStyleClass.ClassName + @"  element, float targetAlpha) =>
+			new(v => element.TextColor = element.TextColor.WithAlpha(v), element.TextColor.Alpha, targetAlpha);
+	}
+}");
+			var source = textColorToBuilder.ToString();
+			SourceGeneratorHelpers.FormatText(ref source, options);
```
pictos, line 147: Make sure that we will format the generated code following the user code style and compilation options

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+			context.AddSource($"{textStyleClass.ClassName}TextColorTo.g.shared.cs", SourceText.From(source, Encoding.UTF8));
+		}
+	}
+
+	static string GetClassAccessModifier(INamedTypeSymbol namedTypeSymbol) => namedTypeSymbol.DeclaredAccessibility switch
+	{
+		Accessibility.Public => "public",
+		Accessibility.Internal => "internal",
+		_ => throw new NotSupportedException("TextColorTo only supports public and internal classes ineriting from ITextStyle")
```
pictos, line 156: TODO: Change this to a diagnostic

pictos (commented):
@brminnick here's my version, I added some comments in key areas. During this week I'll try to create a unit test for the SG project, but if I can't we can merge this as it's and open an issue to create the Unit Test

src/CommunityToolkit.Maui.SourceGenerators/Helpers/SourceGeneratorHelpers.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Text;
+
+namespace CommunityToolkit.Maui.SourceGenerators.Helpers
+{
+	static class SourceGeneratorHelpers
+	{
+		public static void FormatText(ref string classSource, CSharpParseOptions? options)
+		{
+			var mysource = CSharpSyntaxTree.ParseText(SourceText.From(classSource, Encoding.UTF8), options);
```
brminnick, line 12: Let's change this to something more descriptive like `sourceCodeText`.

We should void using `my` for variable names in production code; variable names like `myObject` are better purposed for samples/demos.

src/CommunityToolkit.Maui.SourceGenerators/Helpers/SourceGeneratorHelpers.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Text;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Text;
+
+namespace CommunityToolkit.Maui.SourceGenerators.Helpers
+{
+	static class SourceGeneratorHelpers
```
brminnick, line 8: Let's avoid creating a file called SourceGeneratorHelpers because it'll just become a dumping ground for future PRs.

brminnick (commented):
Thanks Pedro! I love these updates and am looking forward to your idea of adding diagnostics to help our users 💯 

Honest question - Are we supposed to write Unit Tests for Source Code generators? I took at glance at the Roslyn repo and didn't see any Unit Tests (but I'm not super familiar with the Roslyn repo, so maybe I just missed them!). We currently have Unit Tests for the Source Generator output in `CommunityToolkit.Maui.UnitTests`, which _should_ suffice to verify the `TextColorTo` generator. 

All that being said, I am very new to Source Generators and still learning best practices! This has been a fun learning experience so far!

pictos:
@brminnick 
> Honest question - Are we supposed to write Unit Tests for Source Code generators? I took at glance at the Roslyn repo and didn't see any Unit Tests (but I'm not super familiar with the Roslyn repo, so maybe I just missed them!). We currently have Unit Tests for the Source Generator output in CommunityToolkit.Maui.UnitTests, which should suffice to verify the TextColorTo generator.

Yes, we can write a unit test for SGs, with them we can even test the Diagnostics messages (warning, errors, info) as you can see in [this talk](https://youtu.be/lJCfPhnFLQs?list=PL3GeZ8t0AGwwM72SqgKu_GAM67OkJK2fJ&t=1491). I'm not sure how new it's the UnitTest tooling for SGs.

Yeah, I knew a little bit about SGs but I learned the Incremental Source Generator on this PR. IMHO I didn't like the API it's more verbose to do stuffs, but I can see the perf gain.

src/CommunityToolkit.Maui.SourceGenerators/Helpers/SourceGeneratorHelpers.cs
pictos: Sure, I'll create a ParseHelper (?) or something like that to handle the generated `string`

src/CommunityToolkit.Maui.SourceGenerators/Extensions/CompilationExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace CommunityToolkit.Maui.SourceGenerators.Extensions
+{
+	static class CompilationExtensions
+	{
+		public static TSymbol? GetSymbol<TSymbol>(this Compilation compilation, BaseTypeDeclarationSyntax declarationSyntax)
+			where TSymbol : ISymbol
+		{
+			var model = compilation.GetSemanticModel(declarationSyntax.SyntaxTree);
+			return (TSymbol?)model.GetDeclaredSymbol(declarationSyntax);
```
Youssef1313, line 12: Can this cast fail in error scenarios?

src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamedTypeSymbolExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+		if (typeSymbol.MetadataName == typeToCheck)
+		{
+			return true;
+		}
+
+		if (typeSymbol.BaseType?.MetadataName == typeToCheck)
+		{
+			return true;
+		}
```
Youssef1313, line 23: Is it intended this doesn't go through all base types?

src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+			foreach (var member in namespaceSymbolFromStack.GetMembers())
+			{
+				if (member is INamespaceSymbol memberAsNamespace)
+				{
+					stack.Push(memberAsNamespace);
+				}
+				else if (member is INamedTypeSymbol memberAsNamedTypeSymbol)
+				{
+					yield return memberAsNamedTypeSymbol;
```
Youssef1313, line 26: Is it intended to not return possible nested types?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+	const string iTextStyleInterface = "Microsoft.Maui.ITextStyle";
+	const string iAnimatableInterface = "Microsoft.Maui.Controls.IAnimatable";
+	const string mauiControlsAssembly = "Microsoft.Maui.Controls";
+
+	public void Initialize(IncrementalGeneratorInitializationContext context)
+	{
+		// Get All Classes in User Library
+		var userGeneratedClassesProvider = context.SyntaxProvider.CreateSyntaxProvider(
+			static (syntaxNode, cancellationToken) => syntaxNode is ClassDeclarationSyntax { BaseList: not null },
+			static (context, cancellationToken) => (ClassDeclarationSyntax)context.Node);
```
Youssef1313, line 24: Are records intentionally ignored?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+			return;
+
+		var textStyleClassList = new List<(string ClassName, string ClassAcessModifier, string Namespace)>();
+
+		// Collect Microsoft.Maui.Controls that Implement ITextStyle
+		var mauiTextStyleImplementors = mauiControlsAssemblySymbolProvider.GlobalNamespace.GetNamedTypeSymbols().Where(x => x.AllInterfaces.Contains(textStyleSymbol, SymbolEqualityComparer.Default)
+				&& x.AllInterfaces.Contains(iAnimatableSymbol, SymbolEqualityComparer.Default));
+
+		foreach (var namedTypeSymbol in mauiTextStyleImplementors)
+			textStyleClassList.Add((namedTypeSymbol.Name, "public", namedTypeSymbol.ContainingNamespace.ToDisplayString()));
```
Youssef1313, line 55: I think this line is problematic if the `namedTypeSymbol` is in global namespace.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+
+		// Collect All Classes in User Library that Implement ITextStyle
+		foreach (var classDeclarationSyntax in userGeneratedClassesProvider)
+		{
+			var declarationSymbol = compilation.GetSymbol<INamedTypeSymbol>(classDeclarationSyntax);
```
Youssef1313, line 60: I'm not sure if the semantic model can return an `IErrorTypeSymbol`, for which this call will crash due to invalid cast.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313, line 70: In theory, a compilation can have 0 syntax trees, in this scenario this will crash. Is there any previous check that guarantees unreachability of this code path in this scenario?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+			if (declarationSymbol is null)
+				continue;
+			if (declarationSymbol.AllInterfaces.Contains(textStyleSymbol, SymbolEqualityComparer.Default)
+				&& declarationSymbol.AllInterfaces.Contains(iAnimatableSymbol, SymbolEqualityComparer.Default))
+			{
+				textStyleClassList.Add((declarationSymbol.Name, GetClassAccessModifier(declarationSymbol), declarationSymbol.ContainingNamespace.ToDisplayString()));
```
Youssef1313, line 66: Same concern about global namespace here.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+using System;
+using System.Threading.Tasks;
+using CommunityToolkit.Maui.Core.Extensions;
+using Microsoft.Maui;
+using Microsoft.Maui.Controls;
+using Microsoft.Maui.Graphics;
+
+namespace " + textStyleClass.Namespace + @";
+
+" + textStyleClass.ClassAcessModifier + @" static partial class ColorAnimationExtensions_" + textStyleClass.ClassName + @"
```
Youssef1313, line 86: I don't think the `ClassAccessModifier` is required. The compiler is smart enough.

https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEDMACADhAJwBcoIAbbdAJmwGEB9ABQFdgyoBjAWACgBvXgF9evAPSjsMAPbYIHDgFMAzkuxgpCKADMoCggDo6UsLihk92DlJhKoCC0QAWCtRu26C2a9icupvz3xiUgorAgIFDiIyAE99XiCSckoaBhY2Tl4BHmEeER5YIj0YZMSQlLp6AEkYIoISsiyhfLLk6kqauoamnKA===

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+		static Animation GetBlueTransformAnimation(" + textStyleClass.Namespace + "." + textStyleClass.ClassName + @"  element, float targetBlue) =>
+			new(v => element.TextColor = element.TextColor.WithBlue(v), element.TextColor.Blue, targetBlue);
+
+		static Animation GetAlphaTransformAnimation(" + textStyleClass.Namespace + "." + textStyleClass.ClassName + @"  element, float targetAlpha) =>
+			new(v => element.TextColor = element.TextColor.WithAlpha(v), element.TextColor.Alpha, targetAlpha);
+	}
+}");
+			var source = textColorToBuilder.ToString();
+			SourceGeneratorHelpers.FormatText(ref source, options);
+			context.AddSource($"{textStyleClass.ClassName}TextColorTo.g.shared.cs", SourceText.From(source, Encoding.UTF8));
```
Youssef1313, line 148: Is there any reason for the `.shared` in the name?

I'm concerned about the compiler not considering this as auto generated file in some scenarios. See compiler implementation for reference.

https://github.com/dotnet/roslyn/blob/ffde321ee629f942e346318503766ea367393533/src/Compilers/Core/Portable/InternalUtilities/GeneratedCodeUtilities.cs#L72-L82

src/CommunityToolkit.Maui.SourceGenerators/Extensions/CompilationExtensions.cs
pictos: What will be the error scenario for this context?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: Yes, controls are mutable

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: problematic how?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: Why this will be an `IErrorTypeSymbol`?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: If I'm not wrong, the case when the compilation has no SyntaxTrees is when there's no code, if we don't have code this method will not be executed.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: can you write more about why are you concerned?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: Generated code should be explicit, so we will be verbose here

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: It's fine, the file just enter that loop when his name doesn't end with `.cs`, this file ends with `.cs`, so this will compile (:

src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
pictos: We aren't using this method anymore, I'll remove it in a future commit

src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamedTypeSymbolExtensions.cs
pictos: We aren't using this method anymore, I'll remove it in a future commit

src/CommunityToolkit.Maui.SourceGenerators/Extensions/CompilationExtensions.cs
Youssef1313: @pictos No specific scenario in mind, but I think this can happen. Using `as TSymbol` instead of a direct cast is much safer.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: @pictos The `namedTypeSymbol.ContainingNamespace.ToDisplayString()` will produce `<global-namespace>` or something of that kind. So you'll end up generating a code that contains a compile error.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: Because I think nothing guarantees otherwise. There could be some compile errors that causes roslyn to not have an INamedTypeSymbol for this node. It's safer to keep it an `as TSymbol` instead of `(TSymbol?)`

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: https://github.com/CommunityToolkit/Maui/pull/224#discussion_r780429867

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: > Generated code should be explicit

I don't really see a reason.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: @pictos It's not about "will compile" or not. The `AnalyzerDriver` in roslyn treats generated code differently per the registered [`GeneratedCodeAnalysisFlags`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.diagnostics.generatedcodeanalysisflags?view=roslyn-dotnet-4.0.1). So it's better to drop the `.shared` part IMO if there is no benefit/reason for it.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313, line 55: It looks like this assumes that types in `mauiTextStyleImplementors` are public. Is this assumption really valid?

src/CommunityToolkit.Maui.SourceGenerators/Extensions/CompilationExtensions.cs
pictos: I would say that we want this to fail. If the type isn't the wanted we should see an exception

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: Can you give an example of where that can happen?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: I would say that we want this to fail. If the type isn't the wanted we should see an exception

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: Resolving this one keep the discussion just in one window

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: yes

src/CommunityToolkit.Maui.SourceGenerators/Extensions/CompilationExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+		}
+	}
+}
```
Youssef1313, line 15: The added `﻿using Microsoft.CodeAnalysis.CSharp` will bring in an extension method from `Microsoft.CodeAnalysis.CSharp.CSharpExtensions`, which returns an `INamedTypeSymbol?`

This eliminates the need for a cast.

```suggestion
﻿using Microsoft.CodeAnalysis;
﻿using Microsoft.CodeAnalysis.CSharp;
using Microsoft.CodeAnalysis.CSharp.Syntax;

namespace CommunityToolkit.Maui.SourceGenerators.Extensions
{
	static class CompilationExtensions
	{
		public static INamedTypeSymbol? GetSymbol(this Compilation compilation, BaseTypeDeclarationSyntax declarationSyntax)
		{
			var model = compilation.GetSemanticModel(declarationSyntax.SyntaxTree);
			return model.GetDeclaredSymbol(declarationSyntax);
		}
	}
}
```

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: @pictos Can you elaborate on how the assumption is correct? What prevents me from doing:

```csharp
internal class MyClass : Microsoft.Maui.ITextStyle { }
```

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: @pictos It's not a good experience to see an exception from the generator while typing erroneous code in the IDE.

src/CommunityToolkit.Maui.SourceGenerators/Extensions/CompilationExtensions.cs
pictos: @Youssef1313 I think that you missed something here... This method is an extension method that can be used for other Symbols, like `ITypeSymbol`, `IFieldSymbol`, `IEventSymbol`, etc... So removing the Generic isn't a solution

Youssef1313: @pictos I thought there is a single caller only?

Anyway, there are multiple overloads of the `GetDeclaredSymbol` extension. So there is probably no need for the generic extension and all callers should call `model.GetDeclaredSymbol` directly (they just shouldn't miss the `using Microsoft.CodeAnalysis.CSharp;`)

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: @pictos You can take any test case that currently produce generated code correctly, and just pull the class out of its namespace.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+
+	static void Execute(SourceProductionContext context, Compilation compilation, ImmutableArray<ClassDeclarationSyntax> userGeneratedClassesProvider, IAssemblySymbol mauiControlsAssemblySymbolProvider)
+	{
+		var textStyleSymbol = compilation.GetTypeByMetadataName(iTextStyleInterface);
+		var iAnimatableSymbol = compilation.GetTypeByMetadataName(iAnimatableInterface);
+
+		if (textStyleSymbol is null || iAnimatableSymbol is null)
+			return;
+
+		var textStyleClassList = new List<(string ClassName, string ClassAcessModifier, string Namespace)>();
```
Youssef1313, line 48: Can these classes be generic? Was that tested? I think there could be problems if they're generic.

src/CommunityToolkit.Maui.UnitTests/Extensions/TextColorTo_Tests.cs
```diff
@@ -0,0 +1,129 @@
...
+	public class PublicTextStyleView : View, ICustomTextStyle
+	{
+		public Color TextColor { get; set; } = new();
+
+		public Font Font { get; set; }
+
+		public double CharacterSpacing { get; set; }
+	}
+
+	class InternalTextStyleView : View, ICustomTextStyle
```
Youssef1313, line 115: I think the generated code for this is "public" per the following line:

```csharp
			textStyleClassList.Add((namedTypeSymbol.Name, "public", namedTypeSymbol.ContainingNamespace.ToDisplayString()));
```

and if my assumption is correct, this will be a compile error if the code was "internal class ..." instead of "class ..."

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: Ok I now see these from from maui assembly and not from user code. Is that correct?
If so, what classes in dotnet/maui repository implements `ITextStyle`? I can't find.

src/CommunityToolkit.Maui.UnitTests/Extensions/TextColorTo_Tests.cs
```diff
@@ -0,0 +1,129 @@
...
+			var label = new Label();
+			label.EnableAnimations();
+
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+			await Assert.ThrowsAsync<ArgumentNullException>(() => label.TextColorTo(null));
+#pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
+		}
+	}
+}
+
```
Youssef1313, line 103: This duplicates the same test case but now in a global namespace. I'm expecting this scenario to fail.

```suggestion

	public class PublicTextStyleView : View, ICustomTextStyle
	{
		public Color TextColor { get; set; } = new();

		public Font Font { get; set; }

		public double CharacterSpacing { get; set; }
	}

	class InternalTextStyleView : View, ICustomTextStyle
	{
		public Color TextColor { get; set; } = new();

		public Font Font { get; set; }

		public double CharacterSpacing { get; set; }
	}

	// Ensures custom ITextStyle interfaces are supported
	interface ICustomTextStyle : ITextStyle
	{

	}
	
```

Youssef1313 (changes requested):
I cloned the repository and checkout out the branch locally and here are few observations:

1. I can easily reproduce this error:

    > 4>CSC : warning CS8785: Generator 'TextColorToGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'ArgumentException' with message 'The hintName 'PublicTextStyleViewTextColorTo.g.shared.cs' of the added source file must be unique within a generator.

    Basically I have the same class in different namespaces.

2. I can confirm the issue about `ContainingNamespace.ToString()` Here is a screenshot from the generated code:

    image

    So to me, this looks like a bug.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
Youssef1313: I now see this is not a partial of the `textStyleClass`. So you can ignore this comment.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+			{
+				textStyleClassList.Add((declarationSymbol.Name, GetClassAccessModifier(declarationSymbol), declarationSymbol.ContainingNamespace.ToDisplayString()));
+			}
+		}
+
+		var options = ((CSharpCompilation)compilation).SyntaxTrees[0].Options as CSharpParseOptions;
+		foreach (var textStyleClass in textStyleClassList)
+		{
+			var textColorToBuilder = new StringBuilder(@"
+// AutoGenerated Code
```
Youssef1313, line 74: If we can't change the `.shared.cs` suffix, the compiler also recognizes a comment on this form.

```suggestion
// <auto-generated/>
```

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,158 @@
...
+				&& declarationSymbol.AllInterfaces.Contains(iAnimatableSymbol, SymbolEqualityComparer.Default))
+			{
+				textStyleClassList.Add((declarationSymbol.Name, GetClassAccessModifier(declarationSymbol), declarationSymbol.ContainingNamespace.ToDisplayString()));
+			}
+		}
+
+		var options = ((CSharpCompilation)compilation).SyntaxTrees[0].Options as CSharpParseOptions;
+		foreach (var textStyleClass in textStyleClassList)
+		{
+			var textColorToBuilder = new StringBuilder(@"
```
Youssef1313, line 73: We use a string builder but only produce a single string without using any of the `Append*` methods. Why not a string directly?

pictos:
@Youssef1313 thanks for trying it on your end. I'll take care of this next week

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+			}
+
+			if (declarationSymbol.AllInterfaces.Contains(textStyleSymbol, SymbolEqualityComparer.Default)
+				&& declarationSymbol.AllInterfaces.Contains(iAnimatableSymbol, SymbolEqualityComparer.Default))
+			{
+				var nameSpace = declarationSymbol.ContainingNamespace.ToDisplayString();
+				if (nameSpace == "<global namespace>")
+				{
+					var diag = Diagnostic.Create(TextColorToDiagnostic.GlobalNamespace, Location.None, declarationSymbol.Name);
+					context.ReportDiagnostic(diag);
```
Youssef1313, line 81: I'm not much convinced by reporting no locations for diagnostics.

pictos: I'm ok with not having a location here, since this isn't a code analyzer and I'm not scanning for a region to do a code fix, for example. I'm just telling the dev that a specific class should be inside a namespace. So the location here doesn't matter.

Youssef1313: @pictos Diagnostic location is not for codefixes only. It helps the developer double click the error to be navigated to the class he should fix.

Youssef1313: A location is also useful to suppress the diagnostic via `#pragma` directives.

src/CommunityToolkit.Maui.SourceGenerators/Diagnostics/TextColorToDiagnostic.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Microsoft.CodeAnalysis;
+
+namespace CommunityToolkit.Maui.SourceGenerators
+{
+	static class TextColorToDiagnostic
+	{
+		const string category = "TextColorTo";
+		public static readonly DiagnosticDescriptor GlobalNamespace = new(
+			   "MCT001",
+			   "Global namespace is not support for this Source Generator",
```
Youssef1313, line 10: Users may have no context about what "this" refers to.

```suggestion
			   "Global namespace is not supported for 'TextColorTo' Source Generator",
```

src/CommunityToolkit.Maui.SourceGenerators/Diagnostics/TextColorToDiagnostic.cs
```diff
@@ -0,0 +1,41 @@
...
+			   "Please put '{0}' inside a valid namespace",
+			   category,
+			   DiagnosticSeverity.Warning,
+			   true);
+
+		public static readonly DiagnosticDescriptor MauiReferenceIsMissing = new(
+			   "MCT002",
+			   "Was not possible to find Microsoft.Maui.ITextStyle and or Microsoft.Maui.IAnimatable",
```
Youssef1313, line 18:
```suggestion
			   "Was not possible to find Microsoft.Maui.ITextStyle or Microsoft.Maui.IAnimatable",
```

Youssef1313 (commented):
I think there is still a bug for this:

```csharp
namespace MyNamespace1
{
	public class MyStyleView: View, ITextStyle
	{
		public Color TextColor { get; set; } = new();

		public Font Font { get; set; }

		public double CharacterSpacing { get; set; }
	}
}

namespace MyNamespace2
{
	public class MyStyleView: View, ITextStyle
	{
		public Color TextColor { get; set; } = new();

		public Font Font { get; set; }

		public double CharacterSpacing { get; set; }
	}
}
```

The two classes will attempt to generate source with the same hint name. So the compiler will produce a diagnostic.

src/CommunityToolkit.Maui.SourceGenerators/Helpers/SourceStringExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.CodeAnalysis.Text;
+
+namespace CommunityToolkit.Maui.SourceGenerators.Helpers
+{
+	static class SourceStringExtensions
+	{
+		public static void FormatText(ref string classSource, CSharpParseOptions? options)
+		{
+			var mysource = CSharpSyntaxTree.ParseText(SourceText.From(classSource, Encoding.UTF8), options);
+			var formattedRoot = (CSharpSyntaxNode)mysource.GetRoot().NormalizeWhitespace();
```
Youssef1313, line 13: If you're already producing the generated source as a string and it's well-formatted (or you can make it well-formatted if it's not), then you don't really need to format the code via `NormalizeWhitespace`, nor you need to parse the source.

`NormalizeWhitespace` can be useful only if you're producing the code via `SyntaxFactory` APIs or something similar.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+using System;
+using System.Threading.Tasks;
+using CommunityToolkit.Maui.Core.Extensions;
+using Microsoft.Maui;
+using Microsoft.Maui.Controls;
+using Microsoft.Maui.Graphics;
+
+namespace " + textStyleClass.Namespace + @";
+
+" + textStyleClass.ClassAcessModifier + @" static partial class ColorAnimationExtensions_" + textStyleClass.ClassName + @"
```
Youssef1313, line 114:
```suggestion
" + textStyleClass.ClassAcessModifier + @"static partial class ColorAnimationExtensions_" + textStyleClass.ClassName + @"
```

Youssef1313, line 114: Is using `_` in class names a good name? It doesn't follow the conventions.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+{
+	/// <summary>
+	/// Animates the TextColor of an <see cref=""Microsoft.Maui.ITextStyle""/> to the given color
+	/// </summary>
+	/// <param name=""element""></param>
```
Youssef1313, line 119: Missing documentation for this parameter. Not sure if the suggestion is accurate.

```suggestion
	/// <param name=""element"">The element for which the TextColor will be animated.</param>
```

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+using Microsoft.Maui;
+using Microsoft.Maui.Controls;
+using Microsoft.Maui.Graphics;
+
+namespace " + textStyleClass.Namespace + @";
+
+" + textStyleClass.ClassAcessModifier + @" static partial class ColorAnimationExtensions_" + textStyleClass.ClassName + @"
+{
+	/// <summary>
+	/// Animates the TextColor of an <see cref=""Microsoft.Maui.ITextStyle""/> to the given color
```
Youssef1313, line 117: Probably cref for TextColor?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+	/// </summary>
+	/// <param name=""element""></param>
+	/// <param name=""color"">The target color to animate the <see cref=""Microsoft.Maui.ITextStyle.TextColor""/> to</param>
+	/// <param name=""rate"">The time, in milliseconds, between the frames of the animation</param>
+	/// <param name=""length"">The duration, in milliseconds, of the animation</param>
+	/// <param name=""easing"">The easing function to be used in the animation</param>
+	/// <returns>Value indicating if the animation completed successfully or not</returns>
+	public static Task<bool> TextColorTo(this " + textStyleClass.Namespace + "." + textStyleClass.ClassName + @" element, Color color, uint rate = 16u, uint length = 250u, Easing? easing = null)
```
Youssef1313, line 125: I might be missing something, but why not use `ITextStyle` as the element type. With that, we don't care about the type of the first parameter, and there is a single generated extension that works for all types that implement `ITextStyle`.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -0,0 +1,186 @@
...
+	{
+		ArgumentNullException.ThrowIfNull(element);
+		ArgumentNullException.ThrowIfNull(color);
+
+		if(element is not Microsoft.Maui.ITextStyle)
```
Youssef1313, line 130:
```suggestion
		if (element is not Microsoft.Maui.ITextStyle)
```

src/CommunityToolkit.Maui.SourceGenerators/Helpers/SourceStringExtensions.cs
pictos: Since this will be a method to be used by future SGs I will keep this way

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: Because `ITextStyle` implements the properties that we need as read-only.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: No need, the `FormatText` method fix that

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: cc: @brminnick

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: No need, the `FormatText` method fix that

pictos: Is a generated file, so it's fine

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
pictos: cc: @brminnick

pictos (approved):
I believe this is good enough to be merged, we can work on more enhancements as they show up. I'll create an issue later this week to add the unitTest for SG project.

@brminnick feel free to merge it if this looks good for you!

PR #226, opened 2021-12-19
ImplicitUsings

VladislavAntonyuk:
ImplicitUsings enabled by default in Maui Templates: https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-lib/MauiLib1.csproj
Let's use implicit usings in MCT

### DIFF
--- a/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
+++ b/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
@@ -8,7 +8,6 @@
     <UseMaui>true</UseMaui>
     <SingleProject>true</SingleProject>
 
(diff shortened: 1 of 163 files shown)


samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -7,6 +7,8 @@
     <OutputType>Exe</OutputType>
     <UseMaui>true</UseMaui>
     <SingleProject>true</SingleProject>
+
+    <ImplicitUsings>enable</ImplicitUsings>
```
pictos, line 11: Why not add this to the Build.Props? That way we don't need to duplicate it on all other `csproj`s

brminnick (approved):

Issue #124, opened 2021-09-30
[Proposal] Drop the Effects API and migrate all Effects to be Behaviors 


pictos:
# [Drop the usage of Effects API]

* [x] Proposed
* [x] Prototype: Started
* [ ] Implementation: Not Started
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [ ] Unit Tests: Not Started
* [x] Sample: Started
* [ ] Documentation: Not Started

## Summary
[summary]: #summary

> [Currently Effects don't get wired into Handlers. The need for effects in XF came from the need to be able to register something that was smaller than a renderer. The usefulness of an effect inside MAUI is debatable because inside MAUI users can just subscribe to AttachedHandler/DetachedHandler and achieve basically the same thing. You could even just use a behavior at this point.](https://github.com/dotnet/maui/issues/1226)

With that quote, I would suggest migrating our `Effects` to use the `Behaviors` API.

## Motivation
[motivation]: #motivation

**From my understanding**. 
The .NET MAUI team left some _tips_ in [issues](https://github.com/dotnet/maui/issues/1226) and [PR comment](https://github.com/dotnet/maui/pull/2137#discussion_r694036714) that devs should not use the Effects API and that was added into .NET MAUI just to avoid a hard break change with Forms. That said since we are supporting MAUI Handlers we should move all our Effects to use a _better_ approach.

## Detailed Design
[design]: #detailed-design

The API for the `BasePlatformBehavior` is the following:

```csharp
public abstract partial class BasePlatformBehavior<TView, TNativeView> : BaseBehavior<TView>
	where TView : VisualElement
#if !NET6_0 || IOS || ANDROID || WINDOWS
		where TNativeView : NativeView
#else
		where TNativeView : class
#endif
{
	protected TNativeView? NativeView => View?.Handler?.NativeView as TNativeView;

	protected override void OnAttachedTo(BindableObject bindable)
	{
		base.OnAttachedTo(bindable);
	}

	protected override void OnDetachingFrom(BindableObject bindable)
	{
		base.OnDetachingFrom(bindable);
	}

	protected override void OnAttachedTo(TView bindable)
	{
		base.OnAttachedTo(bindable);
		OnPlatformAttachedBehavior(bindable);
	}

	protected override void OnDetachingFrom(TView bindable)
	{
		base.OnDetachingFrom(bindable);
		OnPlatformDeattachedBehavior(bindable);
	}

[...]

PR #231, opened 2021-12-22
Converting Effects into Behaviors

pictos:
This is just a POC of how the design API will look like. 

I removed the iOS targets because I'm not on .NET MAUI p11 to avoid future conflicts when the VS 2022 P2 releases. 
(Maybe I try to update the MAUI workload on my mac during the holidays).

Please see #124 for more information about this design.


src/CommunityToolkit.Maui/Behaviors/BasePlatformBehavior.cs
```diff
@@ -0,0 +1,65 @@
...
+	protected override void OnDetachingFrom(BindableObject bindable)
+	{
+		base.OnDetachingFrom(bindable);
+	}
+
+	/// <inheritdoc />
+	protected override async void OnAttachedTo(TView bindable)
+	{
+		base.OnAttachedTo(bindable);
+		await System.Threading.Tasks.Task.Delay(5000);
```
VladislavAntonyuk, line 45: it should be removed

src/CommunityToolkit.Maui/Behaviors/SelectAllText/SelectAllTextBehavior.windows.cs
```diff
@@ -0,0 +1,44 @@
...
+
+	protected override void OnPlatformAttachedBehavior(InputView view)
+	{
+		NativeView!.GotFocus += OnGotFocus;
+	}
+	/// <inheritdoc />
+
+	protected override void OnPlatformDeattachedBehavior(InputView view)
+	{
+		NativeView!.GotFocus -= OnGotFocus;
```
VladislavAntonyuk, line 22: let's not use null forgiving operator

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -75,6 +75,7 @@
 
   <ItemGroup Condition="$(TargetFramework.Contains('-windows'))">
     <PackageReference Include="CommunityToolkit.WinUI.Notifications" Version="7.1.2" />
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.0.0" />
```
VladislavAntonyuk, line 78: do we need it since we already have it in Core lib?

src/CommunityToolkit.Maui/Behaviors/BasePlatformBehavior.cs
pictos: It's a workaround to a Maui bug

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
pictos: Don't bother about it, as I mentioned we will not merge this PR, it's just to show the api and how to use it

src/CommunityToolkit.Maui/Behaviors/BasePlatformBehavior.cs
brminnick: @pictos What's the bug? Can you link to it?

pictos: I didn't open an issue yet, because I didn't test the p11. But on p10, at least, when this code runs the `MauiContext`, `NativeView` and `Handler` are null, just the `VirtualView` or `View`  isn't `null`. That's caused by some timing thing. When we add a delay those values aren't `null`.

pictos: @VladislavAntonyuk and @brminnick I tested against p12 and the bug still happens, here's the issue for reference

https://github.com/dotnet/maui/issues/4496

pictos:
Closing this one in favor of 
-  https://github.com/dotnet/maui/issues/4958

PR #239, opened 2022-01-08
#238 Toast

### LINKED ISSUE #238: [Proposal] Toast
### Discussed in https://github.com/CommunityToolkit/Maui/discussions/209

<div type='discussions-op-text'>

<sup>Originally posted by **VladislavAntonyuk** November 27, 2021</sup>

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation: Not Started

# Toast

## Summary
[summary]: #summary

Adds a toast.
It should use Native Toast on Android and Windows.
All other platforms implement a default popup container with message.

## Detailed Design
[design]: #detailed-design

### Toast.shared.cs

```cs
namespace CommunityToolkit.Maui.Alerts.Toast;

public class Toast: IToast
{
	ToastDuration Duration { get; set; }
	string Text { get; set; }

	Task Dismiss();
	Task Show();

    public static IToast Make(string text, ToastDuration duration = ToastDuration.Short);
}

public interface IToast
{
	ToastDuration Duration { get; }
	bool IsShown { get; }
	string Text { get; }

	Task Dismiss();
	Task Show();
}

public enum ToastDuration
{
    Short,
    Long
}
```

VladislavAntonyuk:
## Fixes
#238 

## Description

Implement toast,
Fix Snackbar OnDismissed is not executed
Add optional CancellationToken parameter to all public methods


samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -63,7 +63,7 @@
 
   <PropertyGroup Condition="$(TargetFramework.Contains('-windows'))">
     <OutputType>WinExe</OutputType>
-    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
+    <RuntimeIdentifier>win10-x64</RuntimeIdentifier>
```
VladislavAntonyuk, line 66: missed changes from preview 11

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -116,8 +116,8 @@ public async ValueTask DisposeAsync()
 	/// </summary>
 	protected virtual async ValueTask DisposeAsyncCore()
 	{
-#if NET6_0_ANDROID || NET6_0_IOS || NET6_0_MACCATALYST
-		await Microsoft.Maui.Controls.Device.InvokeOnMainThreadAsync(() => _nativeSnackbar?.Dispose());
+#if ANDROID || IOS || MACCATALYST
```
VladislavAntonyuk, line 119: NET6_0_xxx are not defined.

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.android.cs
```diff
@@ -17,29 +17,27 @@ public partial class Snackbar
 {
 	static readonly SemaphoreSlim _semaphoreSlim = new(1, 1);
 
-	AndroidSnackbar? _nativeSnackbar;
+	static AndroidSnackbar? _nativeSnackbar;
```
VladislavAntonyuk, line 20: otherwise line 28 always returns true

pictos (changes requested):
@VladislavAntonyuk, let's make this PR just for toast. Please implement the Snackbar fixes in another PR

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
VladislavAntonyuk: https://github.com/dotnet/maui/blob/8a4cf99d92f23c1466ae8e0037c26ac1a4b66d23/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj#L54

src/CommunityToolkit.Maui/Alerts/Toast/Toast.macios.cs
```diff
@@ -0,0 +1,57 @@
...
+		{
+			Duration = GetDuration(Duration)
+		};
+
+		nativeToast.Show();
+
+		static T? GetMaximum<T>(params T[] items) => items.Max();
+	}
+
+	static CGRect GetCornerRadius(int defaultRadius = 4)
```
pictos, line 53: I would say that this isn't a good name for this method. I think that `CreateCornerRadius` will describes better what it does. And change the parameter name to `radius`

src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
```diff
@@ -0,0 +1,43 @@
...
+
+	/// <summary>
+	/// Toast message
+	/// </summary>
+	string Text { get; }
+
+	/// <summary>
+	/// Dismiss the toast
+	/// </summary>
+	Task Dismiss(CancellationToken token = default);
```
pictos, line 21: Why this is a Task? The implementation doesn't have any async call

src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
```diff
@@ -0,0 +1,43 @@
...
+
+	/// <summary>
+	/// Show the toast
+	/// </summary>
+	Task Show(CancellationToken token = default);
```
pictos, line 26: Why this is a Task? The implementation doesn't have any async call

src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
```diff
@@ -0,0 +1,43 @@
+﻿namespace CommunityToolkit.Maui.Core;
+
+/// <summary>
+/// Toast
+/// </summary>
+public interface IToast : IAsyncDisposable
```
pictos, line 6: Why this interface implements `IAsyncDisposable`?

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
```diff
@@ -0,0 +1,93 @@
...
+	/// Create new Toast
+	/// </summary>
+	/// <param name="message">Toast message</param>
+	/// <param name="duration">Toast duration</param>
+	/// <returns>New instance of Toast</returns>
+	public static IToast Make(
+		string message,
+		ToastDuration duration = ToastDuration.Short)
+	{
+		return new Toast()
```
pictos, line 24: Sine we have this method to create the Toast. We should implement a private or internal constructor to avoid devs to create a new instance without using this method.

pictos (changes requested):
Thanks for the hard work, I have some code suggestions and didn't run on my end yet.

src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
brminnick: `IToast` is an interface 😜

samples/CommunityToolkit.Maui.Sample/Pages/Alerts/ToastPage.xaml.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Reflection;
+using CommunityToolkit.Maui.Alerts;
+using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.Extensions;
+
+namespace CommunityToolkit.Maui.Sample.Pages.Alerts;
+
+public partial class ToastPage : BasePage
+{
+	IToast? toast;
```
brminnick, line 10: We probably don't need this field since we're re-initializing `IToast` in `ShowToastButtonClicked`

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
brminnick: You're not wrong. 

We're basically matching how Android instantiates `Toast`, via `Toast.Make`, while also csharp-ifying it by allowing devs to initialize it via `new Toast()`.

The hope is that this lowers the friction for devs, allowing them to instantiate `Toast` via the way they're most comfortable/familiar, without needing to look up the docs.  

We did the same implementation for `Snackbar`.

brminnick:
I'd like to have `ISnackbar` inherit from `IToast` (and for `IPopup` to inherit from `ISnackbar`). 

What if we do something like this? Since Snackbars can be displayed indefinitely and Toasts cannot, we'd want to override the `Duration` property for `ISnackbar` using Default Interface Methods.

### `ISnackbar`

```cs
namespace CommunityToolkit.Maui.Core;

/// <summary>
/// Snackbar
/// </summary>
public interface ISnackbar : IToast
{
	/// <summary>
	/// Action to invoke on action button click
	/// </summary>
	Action? Action { get; }

	/// <summary>
	/// Snackbar action button text
	/// </summary>
	string ActionButtonText { get; }

	/// <summary>
	/// Snackbar anchor. Snackbar appears near this view
	/// </summary>
	IView? Anchor { get; }

	/// <summary>
	/// Snackbar duration
	/// </summary>
	new TimeSpan Duration { get; }

	/// <summary>
	/// Snackbar visual options
	/// </summary>
	SnackbarOptions VisualOptions { get; }

	ToastDuration IToast.Duration => Duration.TotalSeconds switch
	{
		>= 3.0 => ToastDuration.Long,
		>= 0 => ToastDuration.Short,
		_ => throw new InvalidOperationException($"{nameof(Duration)} cannot be negative")
	};
}
```

### `IToast`
```cs
namespace CommunityToolkit.Maui.Core;

/// <summary>
/// Toast
/// </summary>
public interface IToast : IAsyncDisposable
{
	/// <summary>
	/// Toast duration
	/// </summary>
	ToastDuration Duration { get; }

	/// <summary>
	/// Toast message
	/// </summary>
	string Text { get; }

	/// <summary>
	/// Dismiss the toast
	/// </summary>
	Task Dismiss(CancellationToken token = default);

	/// <summary>
	/// Show the toast
	/// </summary>
	Task Show(CancellationToken token = default);
}

/// <summary>
/// Toast duration
/// </summary>
public enum ToastDuration
{
	/// <summary>
	/// Displays Toast for 2 seconds
	/// </summary>
	Short,

	/// <summary>
	/// Displays Toast for 3.5 seconds
	/// </summary>
	Long
}
```

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace CommunityToolkit.Maui.Alerts;
+
+/// <inheritdoc/>
+public partial class Toast : IToast
+{
+	/// <inheritdoc/>
+	public string Text { get; set; } = string.Empty;
+
+	/// <inheritdoc/>
+	public ToastDuration Duration { get; set; }
```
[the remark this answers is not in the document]
brminnick: Let's also add a Unit Test to verify the default values of `new Toast()`. 

```cs
[Fact]
public void ToastMake_NewToastCreatedWithDefaultValues()
{
    Assert.Equals(string.Empty, toast.Text);
    Assert.Equals(ToastDuration.Short, toast.Duration);
}

File: src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs (hunk 8 above)
Comment (brminnick) [inline src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs:24]: Let's add a null check here, just in case a user does have Nullable enabled.

```suggestion
		ArgumentNullException.ThrowIfNull(message);
		
		return new Toast()
```

brminnick: Let's then also add a Unit Test to verify this:

```cs
[Fact] 
public void ToastMake_NewToastCreatedWithNullString_ShouldThrowArgumentNullException() 
{ 
    #pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
    Assert.Throws<ArgumentNullException>(() => Toast.Make(null));
    #pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
}
```

src/CommunityToolkit.Maui.UnitTests/Alerts/Toast_Tests.cs
```diff
@@ -0,0 +1,73 @@
+﻿using CommunityToolkit.Maui.Alerts;
+using CommunityToolkit.Maui.Core;
+using FluentAssertions;
+using Xunit;
+
+namespace CommunityToolkit.Maui.UnitTests.Alerts;
+
+public class Toast_Tests : BaseTest
+{
+	IToast toast = new Toast();
```
brminnick, line 10: Should we make this readonly?

```suggestion
	readonly IToast toast = new Toast();
```

src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
```diff
@@ -0,0 +1,43 @@
...
+	Task Show(CancellationToken token = default);
+}
+
+/// <summary>
+/// Toast duration
+/// </summary>
+public enum ToastDuration
+{
+	/// <summary>
+	/// Short
```
brminnick, line 35:
```suggestion
	/// Displays Toast for 2 seconds
```

src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
```diff
@@ -0,0 +1,43 @@
...
+	/// </summary>
+	Short,
+
+	/// <summary>
+	/// Long
```
brminnick, line 40: Displays Toast for 3.5 seconds

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+#if IOS || MACCATALYST || WINDOWS
+	static TimeSpan GetDuration(ToastDuration duration)
+	{
+		return duration switch
+		{
+			ToastDuration.Short => TimeSpan.FromSeconds(2),
+			ToastDuration.Long => TimeSpan.FromSeconds(3.5),
+			_ => throw new ArgumentOutOfRangeException(nameof(duration), duration, null)
```
brminnick, line 89: Let's throw an `InvalidEnumArgumentException`:

```suggestion
			_ => throw new InvalidEnumArgumentException(nameof(Duration), (int)duration, typeof(ToastDuration))
```

src/CommunityToolkit.Maui/Alerts/Toast/Toast.windows.cs
```diff
@@ -0,0 +1,54 @@
+﻿using CommunityToolkit.WinUI.Notifications;
+using Windows.Data.Xml.Dom;
+using Windows.UI.Notifications;
+
+namespace CommunityToolkit.Maui.Alerts;
+
+public partial class Toast
+{
+	static ToastNotification? nativeToast;
+
```
brminnick, line 10: Since `nativeToast` is `static`, let's add and implement a `SemaphoreSlim(1,1)` like we did for `Snackbar.windows.cs` to ensure we don't induce any race conditions.

```suggestion
	static readonly SemaphoreSlim semaphoreSlim = new(1, 1);

```

src/CommunityToolkit.Maui/Alerts/Toast/Toast.android.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Android.Widget;
+using AndroidToast = Android.Widget.Toast;
+
+namespace CommunityToolkit.Maui.Alerts;
+
+public partial class Toast
+{
+	static AndroidToast? nativeToast;
+
```
brminnick, line 9: Since `nativeToast` is `static`, let's add and implement a `SemaphoreSlim(1,1)` like we did for `Snackbar.android.cs` to ensure we don't induce any race conditions.

```suggestion
	static readonly SemaphoreSlim semaphoreSlim = new(1, 1);

```

src/CommunityToolkit.Maui/Alerts/Toast/Toast.macios.cs
```diff
@@ -0,0 +1,57 @@
+﻿using CommunityToolkit.Maui.Core.Views;
+using CoreGraphics;
+using UIKit;
+
+namespace CommunityToolkit.Maui.Alerts;
+
+public partial class Toast
+{
+	static ToastView? nativeToast;
+
```
brminnick, line 10: Since `nativeToast` is `static`, let's add and implement a `SemaphoreSlim(1,1)` like we did for `Snackbar.macios.cs` to ensure we don't induce any race conditions.

```suggestion
	static readonly SemaphoreSlim semaphoreSlim = new(1, 1);

```

brminnick (changes requested):
Thanks Vlad! Just a couple small requests here to update some XML comments, add a couple Unit Tests, and add `SemaphoreSlim` to avoid a race condition with `static nativeToast`.

If possible, I'd also like `ISnackbar` to inherit from `IToast` if we can (e.g. `public interface ISnackbar : IToast`). 

And then, if also possible, I'd like for `Snackbar.shared.cs` to inherit from `Toast.shared.cs` (e.g. `public class Snackbar : Toast`).

This'll help us share some of the common code (assuming it's possible), and hopefully lower our maintenance cost in the future!

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
pictos: I can't see any value-creating an empty Toast. So here I would say to just keep the `Make` method, to create a Toast or change it to be a constructor with two parameters.
IMHO having two ways to create a Toast will be more error-prone than keeping just one way

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
```diff
@@ -0,0 +1,93 @@
+﻿using CommunityToolkit.Maui.Core;
+
+namespace CommunityToolkit.Maui.Alerts;
+
+/// <inheritdoc/>
+public partial class Toast : IToast
```
pictos, line 6: Just like Android, this should be immutable, so we shouldn't implement the `set` for properties.

src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
VladislavAntonyuk: moved to IAlert. I keep Task to not break alerts and snackbar

src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
VladislavAntonyuk: moved to IAlert. I keep Task to not break alerts and snackbar

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
VladislavAntonyuk: Android Toast has a constructor, but it receives Context as a first parameter.
As for me I always use Toast.Make, but maybe we should keep 2 variants to look similar to the snackbar

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
VladislavAntonyuk: it has methods for it, like this 
image


I prefer to keep properties here.

VladislavAntonyuk:
Thank you @brminnick and @pictos 
as for the Snackbar inherits Toast, I think we should try to avoid it. In Android, Snackbar and Toast are absolutely different in terms of System. Snackbar is visual component (java.lang.Object=>com.google.android.material.snackbar.BaseTransientBottomBar (com.google.android.material.snackbar.Snackbar) => com.google.android.material.snackbar.Snackbar), while Toast is just a Widget. I think we should keep this separation in MCT as well, so we can easily extend toast without breaking popup and snackbar. 
We have this inheritance in XCT, most likely because we actually do not have Toast in XCT at all. Snackbar is used in both cases.

brminnick:
Thanks Vlad! I like the idea of using `IAlert` 💯 

I pushed a quick commit to remove the Null Forgiving Operator since we say to avoid it in our [Contribution Guidelines](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#null-checking):
> Avoid using the ! null forgiving operator to avoid the unintended introduction of bugs.

(Although, one could argue that [this unit test](https://github.com/CommunityToolkit/Maui/pull/239/commits/3fb8ca4186f03a049d0d902dea7138f7ac2b410f) is one of the few valid use cases for the Null Forgiving Operator)

I also added in the following [`Assert` for `Snackbar_Tests` and `Toast_Tests`](https://github.com/CommunityToolkit/Maui/pull/239/commits/3fcb23ccfe49ca9f129b3db795b1212f26690b75) to ensure we don't accidentally break the `IAlert` dependency in a future PR:

```cs
public Snackbar_Tests()
{
    Assert.IsAssignableFrom<IAlert>(snackbar);
}
```

```cs
public Toast_Tests()
{
    Assert.IsAssignableFrom<IAlert>(toast);
}
```

brminnick:
### Toast on iOS

For the Toast on iOS, the character spacing looks odd; it's different than the character spacing for Snackbar. Any idea what's going on here?

| Toast | Snackbar |
| ----- | --------- |
|  |  |

VladislavAntonyuk:
Hmm, it maybe related to the method GetEmFromPx which is used to calculate character spacing

brminnick:
Ah - found it: 

We need to set the default CharacterSpacing for `ToastView` to `0.0d` (it is currently set to 1.0).

I'll add a comment in a review 👍  

```cs
nativeToast = new ToastView(Text,
							UIColor.LightGray,
							cornerRadius,
							UIColor.DarkTextColor,
							UIFont.SystemFontOfSize(UIFont.LabelFontSize),
							0.0d,
							padding)
```

image

src/CommunityToolkit.Maui/Alerts/Toast/Toast.macios.cs
```diff
@@ -0,0 +1,66 @@
...
+		token.ThrowIfCancellationRequested();
+
+		var cornerRadius = CreateCornerRadius();
+		var padding = GetMaximum(cornerRadius.X, cornerRadius.Y, cornerRadius.Width, cornerRadius.Height) + ToastView.DefaultPadding;
+		nativeToast = new ToastView(Text,
+											UIColor.LightGray,
+											cornerRadius,
+											UIColor.DarkTextColor,
+											UIFont.SystemFontOfSize(UIFont.LabelFontSize),
+											1,
```
brminnick, line 51: We need to set the character spacing to `0.0d` to use the default font spacing 
```suggestion
											0.0d,
```

brminnick (changes requested):
Thanks again, Vlad! Just a quick edit to the iOS character spacing and I think we're good to go 👍

src/CommunityToolkit.Maui/Alerts/Toast/Toast.macios.cs
VladislavAntonyuk: I will run my mac VM tomorrow and check both snackbar and toast. I hope we'll merge it tomorrow

VladislavAntonyuk: sorry for a delay. I made some changes to be more consistent in style (background, font size, font color). I also found a way to customize toast (at least font size), so I added a new property for it.
I verified Apple version. it looks good. I will verify Android and WinUI.

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
```diff
@@ -81,4 +81,8 @@
     <None Include="..\..\build\nuget.png" PackagePath="icon.png" Pack="true" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="CommunityToolkit.Maui" />
```
pictos, line 85: Why do we have internal on `Maui.Core`? 
@brminnick I remember that you mentioned that we shouldn't have `internals` on `Maui.Core`

src/CommunityToolkit.Maui.Core/Interfaces/IAlert.shared.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+	/// <summary>
+	/// Message text
+	/// </summary>
+	string Text { get; }
+
+	/// <summary>
+	/// Dismiss the alert
+	/// </summary>
+	Task Dismiss(CancellationToken token = default);
```
pictos, line 16: We don't have `async` calls on the implementation. We should change it to be `void` or `ValueTask` (if there's a chance of an async implementation). I know that you mentioned that this will break other APIs, but this is the time to break things and improve the API.

`Task` always allocates an object on the `heap`, even if we don't have an `async` call. With `ValueTask` we don't have that allocation if there's a hot path.

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
pictos: We don't need to keep it just because Google keeps it. 
We can't change the text of a Toast after calling the `Show` method, so this should be `immutable`. Making this immutable will help us to avoid `bugs` in the future. So can't see any good reason to not restrict the `Toast` creation and make the properties `read-only`.

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
```diff
@@ -0,0 +1,112 @@
...
+	/// Dispose Toast
+	/// </summary>
+	protected virtual async ValueTask DisposeAsyncCore()
+	{
+#if ANDROID
+		await Device.InvokeOnMainThreadAsync(() => nativeToast?.Dispose());
+#else
+		await Task.CompletedTask;
+#endif
+	}
```
pictos, line 99: Can't we just return the `Task`? That way we don't create the state machine

pictos (changes requested):
I like the code simplification, but I still have some considerations about the API.

src/CommunityToolkit.Maui/Alerts/Toast/Toast.windows.cs
```diff
@@ -0,0 +1,61 @@
...
+	/// Dismiss Toast
+	/// </summary>
+	public virtual async partial Task Dismiss(CancellationToken token)
+	{
+		if (nativeToast is null)
+		{
+			return;
+		}
+
+		await semaphoreSlim.WaitAsync(token);
```
pictos, line 22: Thinking aloud here... All the Show and Dismiss operations should happen on UIThread, in other words, if another Thread executes this method, the system will throw an exception, so do we need to have this `Semaphore`? Or am I missing the big picture?

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
brminnick: Good catch! Yes, let's remove this. 

If something in `CommunityToolkit.Maui.Core` is needed in `CommunityToolkit.Maui`, let's ensure it's available to everyone. If need be, we can always use `[EditorBrowsable(EditorBrowsableState.Never)]` like the `Microsoft.Maui` repo does.

src/CommunityToolkit.Maui/Alerts/Toast/Toast.windows.cs
brminnick: Dang - you bring up a really good point! If we call `Show` or `Dismiss` from a non-UI thread, the app will crash. 

This means that `nativeView` can remain `static` and not need a `SemaphoreSlim(1,1)` because _only_ one thread (the main thread) can access it. 

Since `IAlert` implements `Task Show()` and `Task Dismiss()`, and is implemented by both `ISnackbar` and `IToast`, and here's what I think the best move forward is:
- Keep `Task` as the return value
  - `Snackbar` needs to return a `Task` to dismiss the Android Snackbar
- Use `MainThread.InvokeOnMainThreadAsync` for `Toast.Show()`, `Toast.Dismiss()`, `Snackback.Show()` and `Snackbar.Dismiss()`
  ```cs
  public virtual partial Task Show(CancellationToken token) => MainThread.InvokeOnMainThreadAsync(async () => 
  {
    // ...
  }
  ```

^This will allows use to get rid of the `SemaphoreSlim`, allow us to keep `nativeView` `static`, and ensure that the app does not crash if `Show` or `Dismiss` is called from a background thread.

brminnick: Removed `SemaphoreSlim` and ensured that `nativeToast` is only called from the Main Thread https://github.com/CommunityToolkit/Maui/pull/239/commits/6d669ee2b84f4072da384e736316a011f5a694f8

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
brminnick: You're totally correct, but I think this is easier to read than the alternative.

The unused state machine has a negligible impact on the DLL size (a couple kilobytes) and a negligible impact on performance (a couple ns).

That being said, I'm fine with either implementation.

### Alternative
```cs
#if ANDROID
	protected virtual async ValueTask DisposeAsyncCore()
	{
		if(NativeToast is not null)
		{
			await Device.InvokeOnMainThreadAsync(() => NativeToast.Dispose());
		}
        }
#else
	protected virtual ValueTask DisposeAsyncCore()
	{
		return new ValueTask();
        }
#endif
```

brminnick (approved):
Thanks Vlad! 

I implemented Pedro's recommendation to remove `SemaphoreSlim` by ensuring that `Snackbar.Show`, `Snackbar.Dismiss`, `Toast.Show` and `Toast.Dismiss` always run on the Main Thread.

I approve of this PR, although it looks like Pedro still has two requests:

- Make `Toast.String` readonly
  - I see a valid use-case for keeping this writeable

```cs
IToast? toast = Toast.Make("5 seconds remaining");

async void HandleToastButtonClicked(sender object, EventArgs e)
{
    toast.Text = "3 Minutes Remaining";
    await toast.Show();
}
```
- Update `DisposeAsyncCore` for `Toast.shared.cs` and `Snackbar.shared.cs`

### Current

```cs
        /// <summary>
	/// Dispose Toast
	/// </summary>
	protected virtual async ValueTask DisposeAsyncCore()
	{
#if ANDROID
		if(NativeToast is not null)
		{
			await Device.InvokeOnMainThreadAsync(() => NativeToast.Dispose());
		}
#else
		await ValueTask.CompletedTask;
#endif
	}
```

### Updated

```cs
#if ANDROID
	protected virtual async ValueTask DisposeAsyncCore()
	{
		if(NativeToast is not null)
		{
			await Device.InvokeOnMainThreadAsync(() => NativeToast.Dispose());
		}
        }
#else
	protected virtual ValueTask DisposeAsyncCore()
	{
		return ValueTask.CompletedTask;
        }
#endif
```

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
pictos: @brminnick I'm wondering if we can remove the `async` modifier for Android implementation as well

protected virtual ValueTask DisposeAsyncCore()
	{
		if(NativeToast is not null)
		{
			return Device.InvokeOnMainThreadAsync(() => NativeToast.Dispose());
		}
        }

src/CommunityToolkit.Maui/Alerts/Toast/Toast.windows.cs
pictos: @brminnick I would say that we don't need to force the code to run on a UI thread, and we should let the dev take care of that since this can cause other issues. For reference, there's this discussion on `CollectionView` from Forms that caused some issues because the framework forces the code to run on the UI thread.
https://github.com/xamarin/Xamarin.Forms/discussions/11823

pictos: I'll mark this as unsolved to make sure that you will see it :) 
You can resolve it after you read

VladislavAntonyuk: We marked this method as Virtual so developers can override it.

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
VladislavAntonyuk: we cannot: InvokeOnMainThreadAsync returns Task
image

src/CommunityToolkit.Maui.UnitTests/CommunityToolkit.Maui.UnitTests.csproj
```diff
@@ -1,30 +1,120 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
```
brminnick, line 1: Let's remove all of these new empty lines

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -1,72 +1,144 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
```
brminnick, line 1: Let's remove all of these new empty lines

samples/CommunityToolkit.Maui.Sample/ViewModels/Alerts/AlertsGalleryViewModel.cs
```diff
@@ -7,13 +7,20 @@ public class AlertsGalleryViewModel : BaseGalleryViewModel
...
 		})
 	{
 	}
 }
 
+
 public class SnackbarViewModel : BaseViewModel
 {
 
+}
```
brminnick, line 21: Let's break this out into its own file

samples/CommunityToolkit.Maui.Sample/ViewModels/Alerts/AlertsGalleryViewModel.cs
```diff
@@ -7,13 +7,20 @@ public class AlertsGalleryViewModel : BaseGalleryViewModel
...
+
+public class ToastViewModel : BaseViewModel
+{
+
```
brminnick, line 25: Let's break this out into its own file

brminnick (commented):
Just FYI - a bunch of new empty lines appeared after merging in the latest commit to `main`

VladislavAntonyuk:
@pictos you are right, that we cannot change the text of Toast after it is shown. But in our implementation, we create a new native toast each time, so we should always see the new text.

src/CommunityToolkit.Maui.UnitTests/CommunityToolkit.Maui.UnitTests.csproj
VladislavAntonyuk: VS got crazy during rebase

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -126,20 +126,26 @@ public async ValueTask DisposeAsync()
 	/// <summary>
 	/// Dispose Snackbar
 	/// </summary>
+#if ANDROID || IOS || MACCATALYST
 	protected virtual async ValueTask DisposeAsyncCore()
 	{
-#if ANDROID || IOS || MACCATALYST
-		if (nativeSnackbar is not null)
+		if (NativeSnackbar is not null)
```
brminnick, line 132: I just realized that if the GC doesn't run on the main thread, calling `NativeSnackbar` will throw an exception. Changing to `nativeSnackbar` bypasses the `MainThread.IsMainThread` check

```suggestion
		if (nativeSnackbar is not null)
```

pictos:
@brminnick @VladislavAntonyuk I'll move the discussion about `mutable`/`immutable` to the issue

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
```diff
@@ -0,0 +1,119 @@
...
+		GC.SuppressFinalize(this);
+	}
+
+	/// <summary>
+	/// Dispose Toast
+	/// </summary>
+#if ANDROID
+	protected virtual async ValueTask DisposeAsyncCore()
+	{
+		if (NativeToast is not null)
```
brminnick, line 93: I just realized that if the GC doesn't run on the main thread, calling `NativeToast` will throw an exception. Changing to `nativeToast` bypasses the `MainThread.IsMainThread` check.


```suggestion
		if (nativeToast is not null)
```

brminnick (commented):
Just fixing a couple possible crashes in the Dispose method

brminnick:
Hey Gang! In the spirit of getting this PR merged before our upcoming release this week, what does everyone think about moving forward with making Toast immutable?

Changing `Toast` later to be Mutable is not a breaking change, whereas if we release it as Mutable now and later change it to Immutable, it _is_ is a breaking change.

This way, we can push out the feature and get feedback/examples from the community if they'd like to add mutability to Toast.

```cs
public string Text { get; init; } = string.Empty;

public ToastDuration Duration { get; init; } = ToastDuration.Short;

public double TextSize { get; init; } = Defaults.FontSize;
```

VladislavAntonyuk:
I like it. The only suggestion is to implement the same for the snackbar.

brminnick:
> The only suggestion is to implement the same for the snackbar (maybe in a new PR) but they should go in 1 release

Sounds good! 

I'm fine with making `Snackbar` immutable in this PR since we've already made a few changes to the `Snackbar.shared.cs` 👍

pictos:
LGTM changing snack bar on this PR

brminnick (approved):
Thanks Vlad!!!

This is a huge PR and I know it wasn't easy, but I LOVE the way it turned out!!

VladislavAntonyuk:
We are getting stronger with such challenges! I love it too!

PR #241, opened 2022-01-08
Snackbar dismissed event is not executed

VladislavAntonyuk:
1. Snackbar.Dismissed is not executed
2. Pass CancellationToken to all Task methods.
3. Add `partial virtual` to ensure all methods have the same signature (feedback from @maxkoshevoi)


src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -152,11 +156,13 @@ public static class SnackbarVisualElementExtension
...
+	/// <param name="token">Cancellation token</param>
 	public static Task DisplaySnackbar(
 		this VisualElement? visualElement,
 		string message,
 		Action? action = null,
 		string actionButtonText = "OK",
 		TimeSpan? duration = null,
-		SnackbarOptions? visualOptions = null) => Snackbar.Make(message, action, actionButtonText, duration, visualOptions, visualElement).Show();
+		SnackbarOptions? visualOptions = null,
+		CancellationToken token = default) => Snackbar.Make(message, action, actionButtonText, duration, visualOptions, visualElement).Show(token);
```
brminnick, line 167: Lazy question - What is the `default` value for `CancellationToken`? Is it `CancellationToken.None`?

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -63,7 +63,7 @@
 
   <PropertyGroup Condition="$(TargetFramework.Contains('-windows'))">
     <OutputType>WinExe</OutputType>
-    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
+    <RuntimeIdentifier>win10-x64</RuntimeIdentifier>
```
brminnick, line 66: Oh cool! I haven't seen `win10` as a `RuntimeIdentifier` before. 

Does this change any behavior? Will this still run on Windows 11?

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.android.cs
```diff
@@ -127,22 +123,31 @@ void SetupActions(AndroidSnackbar nativeSnackbar)
...
+		nativeSnackbar.AddCallback(new SnackbarCallback(this, _dismissedTCS = new()));
 	}
 
 	class SnackbarCallback : BaseTransientBottomBar.BaseCallback
 	{
+		readonly Snackbar snackbar;
 		readonly TaskCompletionSource<bool> _dismissedTCS;
 
-		public SnackbarCallback(in TaskCompletionSource<bool> dismissedTCS)
+		public SnackbarCallback(Snackbar snackbar, in TaskCompletionSource<bool> dismissedTCS)
```
brminnick, line 134: Let's add the `in` modifier to `Snackbar` to match `TaskCompletionSource`, ensuring the reference cannot be modified


```suggestion
		public SnackbarCallback(in Snackbar snackbar, in TaskCompletionSource<bool> dismissedTCS)
```

src/CommunityToolkit.Maui.Core/Interfaces/ISnackbar.shared.cs
```diff
@@ -38,10 +38,10 @@ public interface ISnackbar : IAsyncDisposable
 	/// <summary>
 	/// Dismiss the snackbar
 	/// </summary>
-	Task Dismiss();
+	Task Dismiss(CancellationToken token = default);
```
brminnick, line 41: Let's add a couple Unit Tests to verify `CancellationToken`. 

We'll want to test these scenarios:
- `CancellationToken.None`
- An actual CancellationToken is passed in but not cancelled
- An actual CancellationToken is passed in and is cancelled
- A `null` is passed in (our code should throw a `ArgumentNullException`

brminnick (changes requested):
Thanks Vlad!!

Let's add a couple Unit Tests for the new `CancellationToken` API 👍

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
VladislavAntonyuk: https://github.com/dotnet/maui/blob/8a4cf99d92f23c1466ae8e0037c26ac1a4b66d23/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj#L54

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
VladislavAntonyuk: yes,
image

we cannot use CancellationToken.None because the default parameter must be a compile-time constant

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
VladislavAntonyuk: and yes, it can still run on Windows 11

brminnick: Nice!!

src/CommunityToolkit.Maui.Core/Interfaces/ISnackbar.shared.cs
VladislavAntonyuk: just a question what should happen with Snackbar when the cancellation token is canceled? should it dismiss?

as for the 4th point. Null cannot be passed as CancellationToken is struct.

brminnick: > as for the 4th point. Null cannot be passed as CancellationToken is struct

Ah - good point. Apologies!

brminnick: > what should happen with Snackbar when the cancellation token is canceled

Good question! 

I'm thinking we should throw a `TaskCanceledException` via `CancellationToken.ThrowIfCancellationRequested()` and let the developer decide how to handle it for their app.

brminnick (approved):

brminnick:
Thanks Vlad!!

PR #243, opened 2022-01-10
Implement Dependency Injection in Sample App

brminnick:
## Description

This PR implements Dependency Injection for the .NET MAUI Community Toolkit Sample App.

DI will be the recommended approach going forward for all .NET MAUI apps and it's important that our Sample App reflects this. 

Plus, we will be able to leverage and reference this sample app going forward to new .NET MAUI developers who are looking to better understand MVVM + DI + MAUI.


samples/CommunityToolkit.Maui.Sample/App.xaml.cs
```diff
@@ -9,6 +9,7 @@ public App()
 	{
 		InitializeComponent();
 
-		MainPage = new BaseNavigationPage(new MainGalleryPage());
+		var mainGalleryPage = ServiceProvider.GetRequiredService<MainGalleryPage>();
```
VladislavAntonyuk, line 12: I think it can be injected into constructor

VladislavAntonyuk (approved):

samples/CommunityToolkit.Maui.Sample/App.xaml.cs
brminnick: You're not wrong. However, injecting `MainGalleryPage` into `App.cs` creates an order-of-operations problem in .NET MAUI Apps that also use `Application.Resources.ResourceDictionary` in `App.xaml.cs` + XAML Compilation.

The DI container initializes `App` last, and thus its ResourceDictionary also gets initialized last.

This then causes every `ContentPage` that uses `{StaticResource ...}` (e.g. `{StaticResource ContentPadding}`) to throw an error because the `StaticResource` (e.g. `ContentPadding`) doesn't exist during initialization.

brminnick: If we turn off XAML Compilation or stop using `ResourceDictionary` in App.xaml.cs`, then we could inject `MainGalleryPage` into `App`.

brminnick: In the `CommunityToolkit.Maui.Markup` sample app, we **do** inject `MainPage` into the `App` constructor.

This works fine because the app is not using `App.xaml.cs`.

https://github.com/CommunityToolkit/Maui.Markup/blob/ed7ca904ace3b711ea40c45283111b70ac6c606d/samples/CommunityToolkit.Maui.Markup.Sample/App.cs#L10

samples/CommunityToolkit.Maui.Sample/App.xaml
```diff
@@ -65,9 +66,9 @@
             <Style x:Key="label_section_header" TargetType="Label">
                 <Setter Property="TextColor" Value="{StaticResource HeadingLabelTextColor}" />
             </Style>
-            
+
             <!--  Values  -->
-            <Thickness x:Key="ContentPadding">20, 0</Thickness>
+            <maui:Thickness x:Key="ContentPadding">20, 0</maui:Thickness>
```
pictos, line 71: This should be in the default `xmlns`, no? Is it a bug?

pictos (approved):
Nice!

samples/CommunityToolkit.Maui.Sample/App.xaml
brminnick: Honestly - I'm not a XAML expert and have no clue.

When I opened the file in VS for Mac, it had a red squiggle under `<Thickness x:Key="ContentPadding">20, 0</Thickness>`. 

When I clicked the lightbulb in the IDE, it recommended this....so I did it 😊

brminnick: I'll try removing it and see if the Azure DevOps pipelines still work 👍

brminnick: Done ✅ 

Looks like it was just a false negative in VS for Mac

PR #250, opened 2022-01-18
Implement Flyout Shell in Sample App

VladislavAntonyuk:
### Description of Change ###

Remove ViewModel dependency from ContentPage. (No need to register Content page. Only ViewModels).

Use Shell.

So how does navigation work now?

1. Page knows about ViewModel, ViewModel doesn't know about pages. It only knows about other ViewModels.
2. we map ViewModel to a specific Shell route.
3. we map the Shell route to a specific page.
2. we navigate to ViewModel. extract shell route by ViewModel from mapping. Shell navigates by route to a specific page.

 ### PR Checklist ###
 

samples/CommunityToolkit.Maui.Sample/Pages/Base/BasePage.cs
```diff
@@ -8,9 +8,12 @@ namespace CommunityToolkit.Maui.Sample.Pages;
 
 public abstract class BasePage<TViewModel> : BasePage where TViewModel : BaseViewModel
 {
+	public BasePage() : this(ServiceProvider.GetRequiredService<TViewModel>())
```
VladislavAntonyuk, line 11: empty ctor is required for MAUI (it will be fixed in preview 12)

samples/CommunityToolkit.Maui.Sample/Pages/Base/ViewModelLocator.cs
```diff
@@ -0,0 +1,119 @@
...
+
+namespace CommunityToolkit.Maui.Sample.Pages;
+
+static class ViewModelLocator
+{
+	internal static readonly Dictionary<Type, string> mappings;
+
+	static ViewModelLocator()
+	{
+		mappings = new Dictionary<Type, string>();
```
pictos, line 17:
```suggestion
	internal static readonly Dictionary<Type, string> mappings = new();

	static ViewModelLocator()
	{
```

pictos (changes requested):
That looks cleaner to me, I have some comments and I found an issue on Windows (so far) I can't navigate to the sample page if I navigate back, I need to go to another sample and navigate again.

On video:
1. I navigate to `DoubleToIntConverter` 
2. Navigate back
3. clicked again on `DoubleToIntConverter`
4. Nothing happened


https://user-images.githubusercontent.com/20712372/150696067-95088523-69ac-46df-9cca-e956e73d465a.mp4

brminnick:
Thanks Vlad! I'm concerned with how this increases the complexity of the sample app. 

We have a tough job of balancing the need to ensure that the sample app both follows best practices, and ensuring the sample app is simple enough for a community member to easily add content too.

I agree that it is a good idea to remove the `ContentPage` references from the `BaseGalleryViewModel` classes. This currently causes our app to consume a lot of memory because every of `ContentPage` is initialized in the `BaseGalleryViewModel` (and each `ContentPage`'s View Model is also instantiated).

## Alternative Implementation

👇 Here's an idea I came up with tonight that might accomplish everything we need without making the sample app too complicated.

I'll open a PR for this tonight (assuming it works) and let's chat about both options.

- **SectionModel.cs**
  - Replace `ContentPage` Parameter /Property with `Type` Parameter / Property

- **ServiceProvider.cs**
  - Add an Overload for `GetRequiredService`
  ```cs
  public static object GetRequiredService(Type type) => Current.GetRequiredService(type);
  ```
- **BaseGalleryPage.cs**
  - Update `HandleSelectionChanged`

  ```cs
  async void HandleSelectionChanged(object? sender, SelectionChangedEventArgs e)
  {
	  ArgumentNullException.ThrowIfNull(sender);
  
	  var collectionView = (CollectionView)sender;
	  collectionView.SelectedItem = null;
  
	  if (e.CurrentSelection.FirstOrDefault() is SectionModel sectionModel)
	  {
		  var page = (BasePage)ServiceProvider.GetRequiredService(sectionModel.PageType)
		  await Navigation.PushAsync(page);
	  }
  }
  ```

brminnick:
It worked! Here's the link to the PR:
https://github.com/CommunityToolkit/Maui/pull/258

This week (and maybe also in the upcoming standup) let's chat about the pros + cons of each, and we can likely combine the pros from both into the Sample App 👍

brminnick:
Hey Vlad! I tweaked a couple things in the sample app to hopefully make it easier for devs to add their new sample pages and view models :

1. Merged `LocatorService` into `AppShell.xaml.cs`
    - Since `App.xaml.cs` sets `MainPage = new AppShell()`, new devs should easily find `AppShell` and see where to add their View Model mappings and register routes
2. Refactored `SectionModel` to use Factory Pattern
    - Made `SectionModel` constructor private
    - Created `public SectionModel Create<TViewModel> where T : BaseViewModel`  
      -  Ensures that every `public Type ViewModelType` is of type `BaseViewModel`
3. Create `WelcomePage`
    - Makes a more user-friendly landing page when the sample app launches

brminnick:
And just FYI - it looks like the Flyout Shell hasn't ben fully implemented on iOS/MacCatalyst yet.

(The top navigation bar is missing)

image

samples/CommunityToolkit.Maui.Sample/Pages/Base/BaseGalleryPage.cs
```diff
@@ -31,10 +31,26 @@ async void HandleSelectionChanged(object? sender, SelectionChangedEventArgs e)
...
+		var uri = new UriBuilder("", GetPagePathForViewModel(viewModelType));
+		return uri.Uri.OriginalString[..^1];
+	}
+
+	static string GetPagePathForViewModel(Type viewModelType)
+	{
+		if (!ViewModelLocator.mappings.ContainsKey(viewModelType))
+		{
+			throw new KeyNotFoundException($"No map for ${viewModelType} was found on navigation mappings");
+		}
```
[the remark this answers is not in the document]
VladislavAntonyuk: it throws rather abstract error message:
image
I will change message a bit to return more informative message

brminnick (approved):
Hey Vlad! I just added a couple commits to simplify the Routing Registration, and I also added a work-around for the iOS Shell appearance bugs. 

If you're cool with the updates, go ahead and merge this PR! 🙌 

@pictos I confirmed that the Windows App bug you pointed out is currently happening on the `main` branch. I haven't dug into it, but I think you're right that it is a `CollectionView` bug. Since it's not related to this PR/Shell, I think we're good to move forward with this PR 👍

VladislavAntonyuk:
Thank you @brminnick!

PR #251, opened 2022-01-19
Remove PNG from Source Code

brminnick:
### Description of Change ###

Small change that removes the `png` file from the source code, replacing it with a link to GitHub's blob storage.

Every image that is uploaded to GitHub in a Pull Request, Issue, Discussion, etc., is stored by GitHub in their blob storage. Since GitHub has already stored this PNG in [Proposal #213](https://github.com/CommunityToolkit/Maui/issues/213), we can leverage that image in lieu of storing it in our repository.

This slightly reduces the size of our repo by 257 KB and allows for quicker diffing + cloning of the repository.


pictos (approved):
LGTM

PR #278, opened 2022-02-05
isNullOrEmptyConverter should check for empty instead of WhiteSpace

VladislavAntonyuk:
### Description of Change ###

Check if string is empty instead of WhiteSpace in IsNullOrEmptyConverter
Allow to check for WhiteSpace by passing parameter

 ### Linked Issues ###
 

 It was created in XCT. Let's fix it here: https://github.com/xamarin/XamarinCommunityToolkit/issues/1742

 ### PR Checklist ###
 

bijington:
I like this change. Can I suggest we add an `IsNullOrWhitespaceConverter` also so that check is still possible?

src/CommunityToolkit.Maui/Converters/IsNullOrEmptyConverter.shared.cs
```diff
@@ -32,5 +32,5 @@ public class IsNullOrEmptyConverter : ValueConverterExtension, ICommunityToolkit
 		=> throw new NotImplementedException();
 
 	internal static bool ConvertInternal(object? value) =>
-		value == null || (value is string str && string.IsNullOrWhiteSpace(str));
-}
\ No newline at end of file
+		value == null || (value is string str && string.IsNullOrEmpty(str));
```
bijington, line 35: It looks like the `is` vs `==` checking in `.editorconfig` isn't being forceful here. I can check on the severity in a new PR

src/CommunityToolkit.Maui.UnitTests/CommunityToolkit.Maui.UnitTests.csproj
```diff
@@ -5,7 +5,7 @@
     <UseMaui>true</UseMaui>
     <IsPackable>false</IsPackable>
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
-    <CompilerGeneratedFilesOutputPath>$(BaseIntermediateOutputPath)\GeneratedFiles</CompilerGeneratedFilesOutputPath>
+    <CompilerGeneratedFilesOutputPath>$(BaseIntermediateOutputPath)\GF</CompilerGeneratedFilesOutputPath>
```
pictos, line 8: I know the reason behind this change, but can you do it in another PR? That way will be easy to us to track it in the future, if needed

brminnick:
Good catch, @VladislavAntonyuk!

I agree with @bijington - let's scope this PR to fix the `IsNullOrEmptyConverter` and `IsNullOrEmptyConverter` so that they only check `string.IsNullOrEmpty(str)`.

Then let's open a new Proposal for `IsNullOrWhiteSpaceConverter` and `IsNotNullOrWhiteSpaceConverter` 👍 

Breaking these into two separate converters should help devs more easily discover them since devs are already familiar with the `string.IsNullOrWhiteSpace` vs `string.IsNullOrEmpty` syntax.

VladislavAntonyuk:
What do you think is better? New converter or add parameter here?
New converter
Pros:
It is specific for the string and does exactly how it’s named: check string is null or white space 
Cons: new code to support, we change the current behavior and require to use new converter

Parameter
Pros: we implement the parameter which was unused. We reuse our code for multiple scenarios
Cons: we change the current behavior anyway. 

Also what I found. IsNullOrEmpty doesn’t check if collection is empty (we have ListIsEmptyConverter). It’s only check for null and check for string. 

What do you think if we remove string validation from this converter at all and create new converter StringIsNullOrEmpty (maybe with parameter AllowWhiteSpaces)?

brminnick:
Interesting! I never thought about how folks may be using this `Converter` as a generic null check.

I think we have a combination of a naming problem and an implementation problem. 

Below are what I think is the best path forward. 

React with a "👍" emoji if you agree, or respond with a comment if you want to tweak it a bit! 

If we get a few "👍" emojis and no comments, I'll open the proposals to do the work and we can vote on it in our next standup.

### Recommendations

1. Rename `IsNullOrEmptyConverter` to `IsStringNullOrEmptyConverter`
    - Throw an `InvalidArgumentException` if `object? value` is not `string`
    - Use `string.IsNullOrEmpty` to evaluate `value`
2. Rename `IsNotNullOrEmptyConverter` to `IsStringNotNullOrEmptyConverter`
    - Throw an `InvalidArgumentException` if `object? value` is not `string`
    - Use `string.IsNullOrEmpty` to evaluate `value`
3. Create `IsStringNullOrWhiteSpaceConverter`
    - Throw an `InvalidArgumentException` if `object? value` is not `string`
    - Use `string.IsNullOrWhiteSpace` to evaluate `value`
4. Create `IsStringNotNullOrWhiteSpaceConverter`
    - Throw an `InvalidArgumentException` if `object? value` is not `string`
    - Use `string.IsNullOrWhiteSpace` to evaluate `value`
5. Rename `ListIsNotNullOrEmptyConverter` to `IsListNotNullOrEmptyConverter`
    - Moving the word `Is` to the front of the converter should improve discoverability by aligning with our other Converters that return a `bool`, e.g. `IsInRangeConverter`, `IsStringNullOrEmptyConverter` etc
6. Rename `ListIsNullOrEmptyConverter` to `IsListNullOrEmptyConverter`
    - Moving the word `Is` to the front of the converter should improve discoverability by aligning with our other Converters that return a `bool`, e.g. `IsInRangeConverter`, `IsStringNullOrEmptyConverter` etc

bijington:
> Interesting! I never thought about how folks may be using this `Converter` as a generic null check.
> 
> 
> 
> I think we have a combination of a naming problem and an implementation problem. 
> 
> 
> 
> Below are what I think is the best path forward. 
> 
> 
> 
> React with a "👍" emoji if you agree, or respond with a comment if you want to tweak it a bit! 
> 
> 
> 
> If we get a few "👍" emojis and no comments, I'll open the proposals to do the work and we can vote on it in our next standup.
> 
> 
> 
> ### Recommendations
> 
> 
> 
> 1. Rename `IsNullOrEmptyConverter` to `IsStringNullOrEmptyConverter`
> 
>     - Throw an `InvalidArgumentException` if `object? value` is not `string`
> 
>     - Use `string.IsNullOrEmpty` to evaluate `value`
> 
> 2. Rename `IsNotNullOrEmptyConverter` to `IsStringNotNullOrEmptyConverter`
> 
>     - Throw an `InvalidArgumentException` if `object? value` is not `string`
> 
>     - Use `string.IsNullOrEmpty` to evaluate `value`
> 
> 3. Create `IsStringNullOrWhiteSpaceConverter`
> 
>     - Throw an `InvalidArgumentException` if `object? value` is not `string`
> 
>     - Use `string.IsNullOrWhiteSpace` to evaluate `value`
> 
> 4. Create `IsStringNotNullOrWhiteSpaceConverter`
> 
>     - Throw an `InvalidArgumentException` if `object? value` is not `string`
> 
>     - Use `string.IsNullOrWhiteSpace` to evaluate `value`
> 
> 5. Rename `ListIsNotNullOrEmptyConverter` to `IsListNotNullOrEmptyConverter`
> 
>     - Moving the word `Is` to the front of the converter should improve discoverability by aligning with our other Converters that return a `bool`, e.g. `IsInRangeConverter`, `IsStringNullOrEmptyConverter` etc
> 
> 6. Rename `ListIsNullOrEmptyConverter` to `IsListNullOrEmptyConverter`
> 
>     - Moving the word `Is` to the front of the converter should improve discoverability by aligning with our other Converters that return a `bool`, e.g. `IsInRangeConverter`, `IsStringNullOrEmptyConverter` etc

No tweak but I like the naming convention with the Is at the start as it makes it clearer what the result is going to be

brminnick:
Closing in favor of https://github.com/CommunityToolkit/Maui/pull/289, [as discussed above](https://github.com/CommunityToolkit/Maui/pull/278#issuecomment-1031761255).

PR #289, opened 2022-02-10
Rename `IsNullOrEmptyConverter` to `IsStringNullOrEmptyConverter`

brminnick:
### Description of Change ###

This PR renames the following converters:
1. Rename `IsNullOrEmptyConverter` to `IsStringNullOrEmptyConverter`
    - Throw an `InvalidArgumentException` if `object? value` is not `string`
    - Use `string.IsNullOrEmpty` to evaluate `value`
2. Rename `IsNotNullOrEmptyConverter` to `IsStringNotNullOrEmptyConverter`
    - Throw an `InvalidArgumentException` if `object? value` is not `string`
    - Use `string.IsNullOrEmpty` to evaluate `value`

 ### Linked Issues ###
 

 - Fixes https://github.com/CommunityToolkit/Maui/pull/278#issuecomment-1031761255

 ### PR Checklist ###
 

src/CommunityToolkit.Maui/Converters/IsStringNullOrEmptyConverter.cs
```diff
@@ -30,7 +35,4 @@ public class IsNullOrEmptyConverter : ValueConverterExtension, ICommunityToolkit
 	/// <returns>N/A</returns>
 	public object? ConvertBack(object? value, Type? targetType, object? parameter, CultureInfo? culture)
 		=> throw new NotImplementedException();
```
bijington, line 32: This probably isn't within the scope of this PR but I wanted to ask your thoughts on `NotImplementedException` vs `NotSupportedException`?

Clearly I have a view 😃  `NotSupportedException` makes it clear it isn't supported and probably won't be `NotImplementedException` can give the impression it might be supported/implemented at some point.

src/CommunityToolkit.Maui/Converters/IsStringNullOrEmptyConverter.cs
```diff
@@ -7,18 +7,23 @@ namespace CommunityToolkit.Maui.Converters;
 /// <summary>
 /// Converts the incoming value to a <see cref="bool"/> indicating whether or not the value is null or empty.
 /// </summary>
-public class IsNullOrEmptyConverter : ValueConverterExtension, ICommunityToolkitValueConverter
+public class IsStringNullOrEmptyConverter : ValueConverterExtension, ICommunityToolkitValueConverter
 {
 	/// <summary>
-	/// Converts the incoming value to a <see cref="bool"/> indicating whether or not the value is null or empty.
+	/// Converts the incoming string to a <see cref="bool"/> indicating whether or not the string is null and not empty using string.IsNullOrEmpty.
```
bijington, line 13: I think this should read:

`the string is null or empty`

bijington:
```suggestion
	/// Converts the incoming string to a <see cref="bool"/> indicating whether or not the string is null or empty using string.IsNullOrEmpty.
```

src/CommunityToolkit.Maui.UnitTests/Converters/IsStringNotNullOrEmptyConverter_Tests.cs
```diff
@@ -0,0 +1,33 @@
...
+	[InlineData(null, false)]
+	[InlineData("", false)]
+	[InlineData(" ", true)]
+	public void IsNotNullOrEmptyConverter_ValidStringValue(string? value, bool expectedResult)
+	{
+		var isNotNullOrEmptyConverter = new IsStringNotNullOrEmptyConverter();
+
+		var result = (bool)isNotNullOrEmptyConverter.Convert(value, null, null, null);
+
+		Assert.Equal(expectedResult, result);
```
bijington, line 20: This wasn't your original test but I am beginning to love FluentAssertions

```suggestion
        result.Should().Be(expectedResult);
```

Actually in this scenario I am not sure it is more readable. What are your thoughts?

bijington (changes requested):

src/CommunityToolkit.Maui/Converters/IsStringNullOrEmptyConverter.cs
brminnick: Good point - I agree! 

Could you open a Proposal to change all of the Converters from using 
`NotImplementedException` to using `NotSupportedException`?

Maybe we also update the Contribution Guide and (if it's possible) add an `.editorconfig` rule to remove `NotImplementedException` since it implies that we forgot to import a feature?

src/CommunityToolkit.Maui.UnitTests/Converters/IsStringNotNullOrEmptyConverter_Tests.cs
brminnick: Yea, it's probably a toss-up for readability. 

As a tie breaker, I'll leave it to use the classic `Assert.Equals` since the other unit test also uses `Assert.*`.

src/CommunityToolkit.Maui/Converters/IsStringNullOrEmptyConverter.cs
```diff
@@ -7,18 +7,23 @@ namespace CommunityToolkit.Maui.Converters;
...
+	/// Converts the incoming string to a <see cref="bool"/> indicating whether or not the string is null or empty using string.IsNullOrEmpty.
 	/// </summary>
 	/// <param name="value">The value to convert.</param>
 	/// <param name="targetType">The type of the binding target property. This is not implemented.</param>
 	/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
 	/// <param name="culture">The culture to use in the converter. This is not implemented.</param>
 	/// <returns>A <see cref="bool"/> indicating if the incoming value is null or empty.</returns>
 	[return: NotNull]
-	public object? Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture) => ConvertInternal(value);
+	public object Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
```
brminnick, line 21: Should return a null object to match the interface 

```suggestion
	public object? Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
```

src/CommunityToolkit.Maui/Converters/IsStringNullOrEmptyConverter.cs
bijington: It is open at: https://github.com/CommunityToolkit/Maui/issues/292

bijington (approved):

bijington (approved):
Awesome work @brminnick

IeuanWalker:
@brminnick Hi, think this change has broken my use case.

Was using this converter to check if `double? ` was null or not, but now I'm getting a `InvalidCastException`

image

Any chance of a converter to check if an `object?` is null or not. I'm happy to do a PR for it.

VladislavAntonyuk:
@IeuanWalker I suggest to create a discussion. The new control by its name can be used for String only.
As an alternative you can use EqualConverter and check if value is null.
But in discussion we can discuss if we need a new IsNullConverter or we replace casting with System.Convert or we leave it as it is.

PR #260, opened 2022-01-27
Remove Null-Coalescing Assignment Operator Workaround from Sample App

brminnick:
### Description of Change ###

This PR Removes the Null-Coalescing Assignment Operator (`??=`) Workaround that was required for instantiating XAML fields when Nullable is enabled.
 
This workaround is now handled natively via Source Generators in .NET MAUI https://github.com/dotnet/maui/issues/1796#event-5661151291

PR #290, opened 2022-02-10
Add Popup Control

### LINKED ISSUE #114: [Proposal] Popup
# Popup

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation

## Summary
[summary]: #summary

A customizable popup

## Detailed Design
[design]: #detailed-design

### Popup.shared.cs

```cs
public class Popup<T> : BasePopupView
{  
  public static readonly BindableProperty ContentProperty;
  public static readonly BindableProperty ColorProperty;
  public static readonly BindableProperty SizeProperty;
  public static readonly BindableProperty VerticalOptionsProperty;
  public static readonly BindableProperty HorizontalOptionsProperty;
  
  public Task<T?> Result { get; }
  public virtual View? Content { get; set; }
  public Color Color { get; set; }
  public LayoutOptions VerticalOptions { get; set; }
  public LayoutOptions HorizontalOptions { get; set; }
  public View? Anchor { get; set; }
  public Size Size { get; set; }
  public bool IsLightDismissEnabled { get; set; }
  public event EventHandler<PopupDismissedEventArgs> Dismissed;
  public event EventHandler<PopupOpenedEventArgs> Opened;
  
  public void Reset();
  public void Dismiss(T? result);
}

// Both `Popup` and `Snackbar` should inherit from `BasePopupView` 
public abstract class BasePopupView
{
  public static readonly BindableProperty IsShown; // One-way binding
  public static readonly BindableProperty AnchorProperty; // Two-way binding
  public static readonly BindableProperty AppearingAnimationProperty; // Two-way binding
  public static readonly BindableProperty DisappearingAnimationProperty; // Two-way binding

  public bool IsShown { get; private set; }

  public IView Anchor { get; set; } = GetDefaultAnchor();
  public Animation AppearingAnimation { get; set; } = GetDefaultAnimation();
  public Animation DisappearingAnimation { get; set; } = GetDefaultAnimation();

  public event EventHandler Shown;
  public event EventHandler Dismissed;

  public abstract Task Show();
  public abstract Task Dismiss();

  private Animation GetDefaultAnimation();

  private IView GetDefaultAnchor();
}
```

pictos:
### Description of Change ###

This PR is the initial work of the Popup handler and the basement for other Handlers in MCT.

Somethings to be aware of:

- WinUI is not so great, we can open issues to fix it and get community help on that;
- Platform-specific APIs are not contemplated in this PR;
- Not have automated tests, I need to work in the Unit device test suite;

 ### Linked Issues ###
 

 - Fixes #114 

 ### PR Checklist ###
 

src/CommunityToolkit.Maui.Core/Platform/Windows/MCTPopup.windows.cs
```diff
@@ -0,0 +1,255 @@
...
+using XamlStyle = Microsoft.UI.Xaml.Style;
+using WindowsThickness = Microsoft.UI.Xaml.Thickness;
+using System.Diagnostics;
+
+namespace CommunityToolkit.Core.Platform;
+
+/// <summary>
+/// The navite implementation of Popup control.
+/// </summary>
+public class MCTPopup : Flyout
```
brminnick, line 17: Let's avoid using `MCT` in `CommunityToolkit.Maui` and `CommunityToolkit.Maui.Core`. 

Long story short, marketing has strict rules around branding, and "MAUI Community Toolkit" always needs to be referred to as ".NET MAUI Community Toolkit". I agree it's silly, but renaming this is easier than causing a ruckus internally. 

This also now introduces two `Popup*` classes in `CommunityToolkit.Maui.Core`. 

Here's my recommendation:
- Rename this to `Popup`
- Rename `CommunityToolkit.Maui.Core.Views.Popup` to `CommunityToolkit.Maui.Core.View.Alert` since it's exclusively used for `IAlert`

src/CommunityToolkit.Maui.Core/Platform/iOS/MCTPopup.macios.cs
```diff
@@ -0,0 +1,176 @@
...
+using Microsoft.Maui.Handlers;
+using Microsoft.Maui.Platform;
+using UIKit;
+
+namespace CommunityToolkit.Core.Platform;
+
+/// <summary>
+/// The navite implementation of Popup control.
+/// </summary>
+public class MCTPopup : UIViewController
```
brminnick, line 11: Let's avoid using `MCT` in `CommunityToolkit.Maui` and `CommunityToolkit.Maui.Core`. 

Long story short, marketing has strict rules around branding, and "MAUI Community Toolkit" always needs to be referred to as ".NET MAUI Community Toolkit". I agree it's silly, but renaming this is easier than causing a ruckus internally. 

This also now introduces two `Popup*` classes in `CommunityToolkit.Maui.Core`. 

Here's my recommendation:
- Rename this to `Popup`
- Rename `CommunityToolkit.Maui.Core.Views.Popup` to `CommunityToolkit.Maui.Core.View.Alert` since it's exclusively used for `IAlert`

brminnick (commented):
Thanks Pedro! I'm still working my way through this review, but wanted to share with you some thoughts before I got too deep into it!

## Folder Structure Recommendation

I know the .NET MAUI team uses a `Platform` folder, but I find that it makes files harder to locate.

Let's keep the toolkit folders based on functionality, e.g. Views, Primitives, Handlers, etc. and leverage the platform-suffix, `*.android.cs`, to denote the platform.

For this PR, I'd like to move `MCTPopup.*.cs` and `PopupExtensions.*.cs` files to the /Views/Popup/ folder.

(FYI - I'm opening a PR now to re-name the existing `Popup` folder to `Alert` since its controls are used exclusively by `IAlert`).

```
CommunityToolkit.Maui.Core 
└───Views
│   └───Alert
│   └───Popup
|   |    MCTPopup.*.cs
|   |    PopupExtensions.*.cs
|   |
│   └───Snackbar
│   └───Toast
```

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -14,7 +14,7 @@
     <SupportedOSPlatformVersion Condition="'$(TargetFramework)' == 'net6.0-maccatalyst'">14.0</SupportedOSPlatformVersion>
     <SupportedOSPlatformVersion Condition="'$(TargetFramework)' == 'net6.0-android'">21.0</SupportedOSPlatformVersion>
     <SupportedOSPlatformVersion Condition="$(TargetFramework.Contains('-windows'))">10.0.17763.0</SupportedOSPlatformVersion>
```
VladislavAntonyuk, line 16: should we update this as well?

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -75,10 +75,17 @@
...
+  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard')) != true AND '$(TargetFramework)' != 'net6.0'">
+    <Compile Remove="**\*.netstandard.cs" />
+    <None Include="**\*.netstandard.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <Compile Remove="**\netstandard\**\*.cs" />
+    <None Include="**\netstandard\**\*.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+  </ItemGroup>
+
   <ItemGroup Condition="$(TargetFramework.Contains('-windows'))">
     <PackageReference Include="CommunityToolkit.WinUI.Notifications" Version="7.1.2" />
+    <!--<PackageReference Include="Microsoft.WindowsAppSDK" Version="1.0.0" />-->
```
VladislavAntonyuk, line 87:
```suggestion
```

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -75,10 +75,17 @@
     <None Include="**\windows\**\*.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
   </ItemGroup>
 
+  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard')) != true AND '$(TargetFramework)' != 'net6.0'">
```
VladislavAntonyuk, line 78: I think we can remove it because we do not have netstandard target framework

src/CommunityToolkit.Maui/Extensions/Navigation/NavigationExtensions.netstandard.cs
```diff
@@ -0,0 +1,16 @@
...
+using CommunityToolkit.Maui.Views;
+using Microsoft.Maui;
+using Microsoft.Maui.Controls;
+
+namespace CommunityToolkit.Maui.Extensions;
+
+public static partial class NavigationExtensions
+{
+	static void PlatformShowPopup(BasePopup popup, IMauiContext mauiContext) =>
+		throw new NotSupportedException($"The current platform '{Device.RuntimePlatform}' does not support Xamarin.CommunityToolkit.BasePopup");
```
VladislavAntonyuk, line 12:
```suggestion
		throw new NotSupportedException($"The current platform '{Device.RuntimePlatform}' does not support CommunityToolkit.Maui.Core.BasePopup");
```

src/CommunityToolkit.Maui/Extensions/Navigation/NavigationExtensions.netstandard.cs
```diff
@@ -0,0 +1,16 @@
...
+
+	static Task<object?> PlatformShowPopupAsync(Popup popup, IMauiContext mauiContext) =>
+		throw new NotSupportedException($"The current platform '{Device.RuntimePlatform}' does not support Xamarin.CommunityToolkit.Popup.");
```
VladislavAntonyuk, line 15:
```suggestion
		throw new NotSupportedException($"The current platform '{Device.RuntimePlatform}' does not support CommunityToolkit.Maui.Core.Popup.");
```

src/CommunityToolkit.Maui/Extensions/Navigation/NavigationExtensions.netstandard.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Threading.Tasks;
+using CommunityToolkit.Maui.Views;
+using Microsoft.Maui;
+using Microsoft.Maui.Controls;
+
+namespace CommunityToolkit.Maui.Extensions;
+
+public static partial class NavigationExtensions
```
VladislavAntonyuk, line 9: actually, we can remove this file at all. it is not used

src/CommunityToolkit.Maui/Extensions/Navigation/NavigationExtensions.shared.cs
```diff
@@ -0,0 +1,70 @@
...
+	{
+		var mauiContext = GetMauiContext(navigation);
+		var popupNative = popup.ToHandler(mauiContext);
+		popupNative.Invoke(nameof(IPopup.OnOpened));
+	}
+
+	static IMauiContext GetMauiContext(INavigation navigation)
+	{
+		return (Shell.Current is null ? 
+			navigation.NavigationStack[0].Handler?.MauiContext
```
VladislavAntonyuk, line 67: possible index out of range exception:
```
var page = new MainPage()
page.ShowPopup();
```

src/CommunityToolkit.Maui/Extensions/Navigation/NavigationExtensions.windows.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.Maui.Platform;
+
+namespace CommunityToolkit.Maui.Extensions;
+
+public static partial class NavigationExtensions
+{
+	static void PlatformShowPopup(BasePopup popup, IMauiContext mauiContext)
+	{
+		var window = mauiContext.GetNativeWindow().GetWindow();
+		popup.Parent = GetCurrentPage((Page)window!.Content);
```
VladislavAntonyuk, line 12: please avoid using null forgiving operator

src/CommunityToolkit.Maui.Core/Interfaces/IPopup.shared.cs
```diff
@@ -0,0 +1,61 @@
...
+
+	/// <summary>
+	/// Occurs when the Popup is Opened.
+	/// </summary>
+	void OnOpened();
+
+	/// <summary>
+	/// Occurs when the Popup is closed by a LightDismiss.
+	/// </summary>
+	void LightDismiss();
```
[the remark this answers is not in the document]
bijington: Could we consider some alternatives such as:

`Close()` - called by close button
`Dismiss()` - popup is dismissed rather than closed

Or with a parameter

`Dismiss(bool wasClosed)`

brminnick:
Auto-closed incidentally by #291

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net6.0;net6.0-android;net6.0-ios;net6.0-maccatalyst</TargetFrameworks>
-    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) and '$(MSBuildRuntimeType)' == 'Full'">$(TargetFrameworks);net6.0-windows10.0.17763.0</TargetFrameworks>
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) and '$(MSBuildRuntimeType)' == 'Full'">$(TargetFrameworks);net6.0-windows10.0.18362.0</TargetFrameworks>
```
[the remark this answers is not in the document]
VladislavAntonyuk: maui-lib uses target framework 19041. was there a reason to change it to 17763?  https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-lib/MauiLib1.csproj

samples/CommunityToolkit.Maui.Sample/Pages/Views/Popups/OpenedEventSimplePopup.xaml.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace CommunityToolkit.Maui.Sample;
+
+public partial class OpenedEventSimplePopup
+{
+	public OpenedEventSimplePopup()
+	{
+		InitializeComponent();
+		Opened += OnOpened;
+		Title ??= new();
```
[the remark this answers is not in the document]
brminnick: I tested it a few weeks ago with .NET MAUI Preview 12, and it was still required on Windows 

https://github.com/CommunityToolkit/Maui/pull/260

samples/CommunityToolkit.Maui.Sample/ViewModels/Views/Popups/CsharpBindingPopupViewModel.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace CommunityToolkit.Maui.Sample.ViewModels.Views.Popups;
+
+public sealed class CsharpBindingPopupViewModel
+{
+	public string Title { get; } = "Xaml Binding Popup";
+
+	public string Message { get; } = "This is a native popup with a Xamarin.Forms View being rendered. The behaviors of the popup will confirm to 100% native look and feel, but still allows you to use your Xamarin.Forms controls.";
```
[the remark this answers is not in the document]
brminnick: We should adhere  avoid the word `native` to abide by the inclusive code comments: https://github.com/jonathanpeppers/inclusive-code-comments

samples/CommunityToolkit.Maui.Sample/ViewModels/Views/Popups/XamlBindingPopupViewModel.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace CommunityToolkit.Maui.Sample.ViewModels.Views.Popups;
+public sealed class XamlBindingPopupViewModel
+{
+	public string Title { get; } = "Xaml Binding Popup";
+
+	public string Message { get; } = "This is a native popup with a Xamarin.Forms View being rendered. The behaviors of the popup will confirm to 100% native look and feel, but still allows you to use your Xamarin.Forms controls.";
```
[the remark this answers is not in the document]
brminnick: We should also avoid the word `native` to abide by the inclusive code comments: https://github.com/jonathanpeppers/inclusive-code-comments

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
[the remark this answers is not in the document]
brminnick: If the Library template now uses `windows10.0.19041`, we should also use `windows10.0.19041` 👍

src/CommunityToolkit.Maui.Core/Handlers/Popup/PopUpViewHandler.android.cs
```diff
@@ -0,0 +1,123 @@
...
+	public static void MapSize(PopupViewHandler handler, IPopup view)
+	{
+		handler.NativeView?.SetSize(view, handler.Container);
+		handler.NativeView?.SetAnchor(view);
+	}
+
+	/// <inheritdoc/>
+	protected override MCTPopup CreateNativeElement()
+	{
+		_ = MauiContext ?? throw new NullReferenceException("MauiContext is null, please check your MauiApplication.");
```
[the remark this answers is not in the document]
brminnick: Maybe an `InvalidOperationException`?

src/CommunityToolkit.Maui.Core/Interfaces/IPopup.shared.cs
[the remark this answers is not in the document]
brminnick: Just adding some research here:

Light Dismiss seems to be a Windows term: https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.primitives.popup.islightdismissenabled?view=winrt-22000

It does not seem to be an Android term: https://www.google.com/search?q=light+dismiss+Android

It does not seem to be an iOS term: https://www.google.com/search?q=light+dismiss+ios

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: For sure. 

Let's use `windows10.0.19041` to match the .NET MAUI Library templates

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
brminnick: Agreed. We already target `net6.0` which has become the new .NET Standard now that we have `net6.0-ios` and `net6.0-android`,

src/CommunityToolkit.Maui/Extensions/Navigation/NavigationExtensions.shared.cs
pictos: This will not happen, because the user needs to have a `NavigationPage`, as you can see the `NavigationExtensions` receives an `INavigation` as a parameter. If we think that using a Page instead is better (and will work), we can change this in a future PR

src/CommunityToolkit.Maui.Core/Platform/Android/MCTPopup.android.cs
```diff
@@ -0,0 +1,92 @@
...
+using CommunityToolkit.Maui.Core;
+using Microsoft.Maui.Platform;
+using AView = Android.Views.View;
+
+namespace CommunityToolkit.Core.Platform;
+
+/// <summary>
+/// The navite implementation of Popup control.
+/// </summary>
+public class MCTPopup : Dialog, IDialogInterfaceOnCancelListener
```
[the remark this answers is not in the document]
pictos: This will be solved when we define how to name those kinds of controls. I'll resolve others since they will be solved at once and to reduce noise on this PR

src/CommunityToolkit.Maui.Core/Handlers/Popup/PopUpViewHandler.android.cs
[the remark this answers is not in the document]
pictos: Well, the exception is throwing because MauiContext is null, so NRE makes sense to me

src/CommunityToolkit.Maui.Core/Handlers/Popup/PopupViewHandler.netstandard.cs
```diff
@@ -0,0 +1,76 @@
+﻿using CommunityToolkit.Maui.Core;
+using Microsoft.Maui.Handlers;
+
+namespace CommunityToolkit.Core.Handlers;
+
+public partial class PopupViewHandler : ElementHandler<IPopup, object>
```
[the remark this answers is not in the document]
pictos: We need this, trust me not having this makes our lives harder since the compiler gets crazy

src/CommunityToolkit.Maui.Core/Platform/Windows/MCTPopup.windows.cs
```diff
@@ -0,0 +1,255 @@
...
+		FlyoutStyle.Setters.Add(new Microsoft.UI.Xaml.Setter(FlyoutPresenter.BackgroundProperty, color.ToWindowsColor()));
+
+		if (VirtualView.Color == Colors.Transparent)
+		{
+			FlyoutStyle.Setters.Add(new Microsoft.UI.Xaml.Setter(FlyoutPresenter.IsDefaultShadowEnabledProperty, false));
+		}
+
+		//Configure border
+
+		FlyoutStyle.Setters.Add(new Microsoft.UI.Xaml.Setter(FlyoutPresenter.PaddingProperty, 0));
```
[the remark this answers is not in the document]
pictos: In resume I can't do it.

Long answer:
On Windows there's a special case, that I can't change the `Style` after applying it to the control, if I do it I'll break the app. So the `SetColor` extension is to set the Content's `BackgroundColor` and this is to set the Flyout `BackgroundColor`, and when I change it or change other Flyout property I need to recreate all the style and apply it to the Flyout Control.

src/CommunityToolkit.Maui.Core/Platform/iOS/MCTPopup.macios.cs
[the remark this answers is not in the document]
pictos: which Popup control?

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
pictos: See this [comment](https://github.com/CommunityToolkit/Maui/pull/290#discussion_r807389661)

When we implement the `PlatformConfiguration` features we will need to implement handlers on this project and this rule will be needed.

src/CommunityToolkit.Maui/Extensions/Navigation/NavigationExtensions.netstandard.cs
pictos: No we can't

samples/CommunityToolkit.Maui.Sample/Pages/Views/Popups/PopupPositionPage.xaml
```diff
@@ -0,0 +1,115 @@
...
+            </Style>
+        </ResourceDictionary>
+    </ContentPage.Resources>
+
+    <VerticalStackLayout>
+        <ScrollView>
+            <VerticalStackLayout>
+                <Label Style="{StaticResource Header}"
+                           Text="Popup's can be positioned anywhere on the screen using VerticalOptions and HorizontalOptions. Tap the arrows below to see how this works." />
+                <Grid HorizontalOptions="CenterAndExpand"
```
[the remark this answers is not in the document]
pictos: Let's create another ticket for it, since we have this all the place on our samples

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
```diff
@@ -72,6 +72,12 @@
     <Compile Remove="**\windows\**\*.cs" />
     <None Include="**\windows\**\*.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
   </ItemGroup>
+  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard')) != true AND '$(TargetFramework)' != 'net6.0'">
```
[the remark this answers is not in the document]
pictos: See this [comment](https://github.com/CommunityToolkit/Maui/pull/290#discussion_r807389661)

src/CommunityToolkit.Maui.Core/Extensions/Workaround/WrapperControl.windows.cs
```diff
@@ -0,0 +1,97 @@
+﻿using Microsoft.Maui.Platform;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using WRect = Windows.Foundation.Rect;
+using Microsoft.Maui.Controls.Platform;
+
+namespace CommunityToolkit.Core.Extensions.Workarounds;
+
+// We still have to keep this class, due to some random exceptions that occours when the 
+// .NET MAUI implementation is used.
```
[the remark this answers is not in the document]
pictos: Didn't have time to create a repo yet. Since this is a `internal` feature I'll need to make a repo on this branch, so will do it after finish this

pictos:
@VladislavAntonyuk do you have an idea why the CI didn't find the method? I can build it on my end using the .NET MAUI p13

pictos:
If the CI is targeting p14 is because we are using the main branch, no? If so when the team releases p14 our pipeline will use p15.

brminnick:
We're not targeting anything special with the CI Pipeline; it just runs [`dotnet workload install maui`](https://github.com/CommunityToolkit/Maui/blob/0c935358e9d1d2850df381c90a5a61179e4bbb53/azure-pipelines.yml#L51). 

This has never happened before. This is the first time that the dotnet workload installer has installed a different version of .NET MAUI than what has been "officially released". Someone may have goofed and accidentally pushed Preview 14 too early. 

As you pointed out, Preview 14 introduces some breaking changes that we'll need to correct in the future. 

Here's our options:
- Update the CI pipeline to only install Preview 13 
  - [ ] Before Preview 14 releases, finish + merge this PR
  - [ ] After Preview 14 releases, revert the changes to the CI Pipeline 
  - [ ] After Preview 14 releases, update Popup to address breaking changes 
- Target Preview 14 for this PR
  - [ ] Update local dev machines to Preview 14
  - [ ] Review + merge this PR
- Pause this PR until Preview 14 Releases
  - [ ] After Preview 14 releases, update this PR to address breaking changes
  - [ ] After Preview 14 releases, finish + merge this PR

src/CommunityToolkit.Maui.Core/Interfaces/IPopup.shared.cs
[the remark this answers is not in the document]
pictos: On android is the [Cancel](https://developer.android.com/reference/android/app/Dialog#cancel()) method
On iOS is the [ModalPresentation](https://developer.apple.com/documentation/uikit/uiviewcontroller/3229894-modalinpresentation)

So I believe that `Light Dismiss` was used to be the _best_ name (at that time), but I'm open to suggestions

src/CommunityToolkit.Maui/Extensions/Navigation/NavigationExtensions.windows.cs
```diff
@@ -0,0 +1,46 @@
...
+		// https://github.com/xamarin/Xamarin.Forms/blob/0c95d0976cc089fe72476fb037851a64987de83c/Xamarin.Forms.Platform.iOS/PageExtensions.cs#L44
+		var native = popup.ToHandler(mauiContext);
+		native?.Invoke(nameof(IPopup.OnOpened));
+		Page GetCurrentPage(Page currentPage)
```
VladislavAntonyuk, line 16: Does MAUI have this implementation?  can we reuse it?

src/CommunityToolkit.Maui.Core/Views/Popup/MCTPopup.macios.cs
```diff
@@ -0,0 +1,176 @@
...
+using Microsoft.Maui.Handlers;
+using Microsoft.Maui.Platform;
+using UIKit;
+
+namespace CommunityToolkit.Core.Platform;
+
+/// <summary>
+/// The navite implementation of Popup control.
+/// </summary>
+public class MCTPopup : UIViewController
```
VladislavAntonyuk, line 11: sorry, I missed the previous comment. Can we extend https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.Core/Views/Alert/AlertView.macios.cs? I think it already allows to add content. we only need to add PopoverDelegate Dismissed.

pictos: I don't think so. AlertView extends from `UIView`, here we extend from `UIViewController`

src/CommunityToolkit.Maui.Core/Views/Popup/MCTPopup.macios.cs
```diff
@@ -0,0 +1,176 @@
...
+		{
+			var index = stackCount - 1;
+			currentPageRenderer = mainPage.Navigation.NavigationStack[index];
+		}
+		else
+		{
+			currentPageRenderer = mainPage;
+		}
+
+		var viewController = (currentPageRenderer.Handler as PageHandler)?.ViewController;
```
VladislavAntonyuk, line 101: I think we should set controller only when it is null: https://github.com/xamarin/XamarinCommunityToolkit/blob/main/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Popup/iOS/PopupRenderer.ios.cs#L156

We had issue in XCT: https://github.com/xamarin/XamarinCommunityToolkit/pull/1762

pictos: Good catch! Probably that was on `develop` branch (before v2) and I took the `main` as a reference. I'll update it ASAP

src/CommunityToolkit.Maui.Core/Views/Popup/MCTPopup.macios.cs
VladislavAntonyuk: let's skip it for now. we may back to it later

samples/CommunityToolkit.Maui.Sample/Pages/Views/Popups/ButtonPopup.xaml.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace CommunityToolkit.Maui.Sample;
+
+public partial class ButtonPopup : Popup
+{
+	public ButtonPopup()
+	{
+		InitializeComponent();
+	}
+
+	void Button_Clicked(object? sender, System.EventArgs e) => Dismiss(null);
```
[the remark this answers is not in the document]
pictos: Still thinking about it

brminnick:
Hey @pictos! Just FYI - To maintain consistency with `Snackbar`'s implementation, I removed the `netstandard` items, replacing them with the same ifdef we used for `Snackbar`, ` #if !(IOS || ANDROID || MACCATALYST || WINDOWS)`:
https://github.com/CommunityToolkit/Maui/blob/2fba2f4fb8297a507f065ab2aae58e1fe52aa675/src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs#L103-L123

👆 This implementation in `Snackbar.shared.cs` was a one-off to handle code that ran only on `net6.0` (ie not a platform like `net6.0-ios` or `net6.0-android`). To be honest, we only added it so that that we could write Unit Tests for `Snackbar` (because our Unit Tests run on `net6.0`).

I'm not sure I love this implementation and now that we're using it in multiple places with `Snackbar` + `Popup`, I'll open a Discussion about how to handle `net6.0`-only code going forward, and we can add that in a future PR once we hammer out the details in the Discussion.

My first thought is to copy your idea, but using `net` in lieu of `netstandard` since we don't have .NET Standard as a Target Framework for the .NET MAUI Community Toolkit.

I'll open that Discussion now and defer the conversation over to there 🙌

brminnick:
Hey Pedro! 

I added a couple additional extension methods that extend `Page`:

```cs
public static void ShowPopup<TPopup>(this Page page, TPopup popup) where TPopup : BasePopup
		=> ShowPopup(page.Navigation, popup);

public static Task<object?> ShowPopupAsync<TPopup>(this Page page, TPopup popup) where TPopup : Popup
		=> ShowPopupAsync(page.Navigation, popup);
```

This allows us to use `Popup` similar to `DisplayAlert`:

```cs
async void HandleSimplePopupButtonClicked(object sender, EventArgs e)
{
	var simplePopup = new SimplePopup();
	await this.ShowPopupAsync(simplePopup);
}
```

I also created `CommunityToolkit.Maui.UnitTests.PopupTests` and added the following Unit Test that purposely throws an `NotImplementedException` to remind us to implement Unit Tests for Popup before we merge this PR 👍 

```cs
[Fact]
public void PlaceholderTest()
{
  // This is a place holder to remind us to implement Unit Tests for Popup
  // This test should be deleted once Popup Unit Tests are added
  throw new NotImplementedException("Popup Unit Tests Not completed");
}
```

pictos:
@brminnick but we should run the test on a device, since Popup is a control, no?

brminnick:
> we should run the test on a device, since Popup is a control, no?

Ideally, we should do both 💯 

But right now we don't have device tests, so we'll focus on only implementing Unit Tests in this PR like we did for [`Snackbar`](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.UnitTests/Alerts/SnackBar_Tests.cs) and [`Toast`](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.UnitTests/Alerts/Toast_Tests.cs).

src/CommunityToolkit.Maui.Core/Platform/Android/MCTPopup.android.cs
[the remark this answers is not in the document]
brminnick: Hey Pedro! I updated the name from `MCTPopup` to `MauiPopup`.

After taking a look at the .NET MAUI repo, this aligns with their naming convention, e.g. `public class MauiLabel : UILabel`

src/CommunityToolkit.Maui/Extensions/Popup/PopupExtensions.shared.cs
```diff
@@ -0,0 +1,108 @@
...
+	/// <summary>
+	/// Displays a popup.
+	/// </summary>
+	/// <param name="page">
+	/// The current <see cref="Page"/>.
+	/// </param>
+	/// <param name="popup">
+	/// The <see cref="BasePopup"/> to display.
+	/// </param>
+	public static void ShowPopup<TPopup>(this Page page, TPopup popup) where TPopup : BasePopup
```
pictos, line 21: @brminnick why not use the `IPopup` interface instead?

brminnick: In `PopupExtensions.windows.cs`, `PlatformShowPopup` sets the value of `popup.Parent`:

https://github.com/CommunityToolkit/Maui/blob/7afed071cd46498267fd1a9b4d6d8c8737f1e623/src/CommunityToolkit.Maui/Extensions/Popup/PopupExtensions.windows.cs#L13

pictos: Oh yeah, we can't use the `IPopup`. I just looked at the old code, and we can use the `BasePopup` since all Popups will be inherited from it

src/CommunityToolkit.Maui/Extensions/Popup/PopupExtensions.shared.cs
```diff
@@ -0,0 +1,108 @@
...
+	/// <param name="page">
+	/// The current <see cref="Page"/>.
+	/// </param>
+	/// <param name="popup">
+	/// The <see cref="Popup"/> to display.
+	/// </param>
+	/// <returns>
+	/// A task that will complete once the <see cref="Popup"/> is dismissed.
+	/// </returns>
+	public static Task<object?> ShowPopupAsync<TPopup>(this Page page, TPopup popup) where TPopup : Popup
```
pictos, line 55: @brminnick 

We have a situation here, where we need to decide if we will use `Page` or `INavigation` (to make things easier to us). I just tested the code and if we do this:
```csharp
var page = new ContentPage
{
	Content = new Label
	{
		Text = "Hello there"
	}
};

page.ShowPopup((MockPopup)popup);
```

this will fail, because the `page.Navigation.NavigationStack.Count` will be zero and this will throw an exception on the `GetMauiContext` method (because we will not have anything to grab the `Handler.MauiContext`). Now if we rewrite this API to work with `Page` we can grab the `MauiContext`. So my vote here is to make it work with `Page`.

The code that works (using INavigation) is
```csharp
		var page = new ContentPage
		{
			Content = new Label
			{
				Text = "Hello there"
			}
		};
		var np = new NavigationPage(page);
		page.ShowPopup((MockPopup)popup);
```
Sadly this isn't obvious for devs that not creating a `NavigationPage` will throw an exception. thoughts??

src/CommunityToolkit.Maui.UnitTests/Views/PopupTests.cs
```diff
@@ -0,0 +1,47 @@
...
+		var page = new ContentPage
+		{
+			Content = new Label
+			{
+				Text = "Hello there"
+			}
+		};
+		var np = new NavigationPage(page);
+		_ = page.Navigation;
+		page.ShowPopup((MockPopup)popup);
```
pictos, line 30: @brminnick This will throw an exception because the `Handler` and `MauiContext` will be `null`. If I'm not wrong, the `Handler` concept doesn't exist on `net6`. That said we can't test the `Popup` because it uses the `Handler`, `Snackbar` and `Toast` don't have a `Handler` implementation then we can test it using `UnitTest`.
I remember that Peppers have a package that mocks `Xamarin.Forms` so you can unit test it, but I can't see something for `Maui`. Do you have any idea on how to workaround that?

src/CommunityToolkit.Maui/Extensions/Popup/PopupExtensions.shared.cs
brminnick: Agreed that we should use `Page`.

Seeing that you're referencing `MockPopup`, I assume that you're getting that exception while unit testing. If that's the case, you'll need to create a `ViewHandler` that implements its own `MauiContext` like we did for MockAnimationHandler:

https://github.com/CommunityToolkit/Maui/blob/c4d7d72b1512a187f9ca0d1560e00dcab3fe4543/src/CommunityToolkit.Maui.UnitTests/Mocks/AnimationExtensions.cs#L11-L141

pictos: Thanks for the reference

pictos: First I'll update the API to use `Page` and then I move to tests

src/CommunityToolkit.Maui.UnitTests/Views/PopupTests.cs
brminnick: You'll need to create a mock `ViewHandler` that implements its own `MauiContext` like we did for MockAnimationHandler:

https://github.com/CommunityToolkit/Maui/blob/c4d7d72b1512a187f9ca0d1560e00dcab3fe4543/src/CommunityToolkit.Maui.UnitTests/Mocks/AnimationExtensions.cs#L11-L141

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
```diff
@@ -0,0 +1,175 @@
...
+		SetElementSize(new Size(View.Bounds.Width, View.Bounds.Height));
+	}
+
+	/// <summary>
+	/// Method to initialize the native implementation.
+	/// </summary>
+	/// <param name="element">An instance of <see cref="IPopup"/>.</param>
+	public void SetElement(IPopup element)
+	{
+		var mainPage = Application.Current?.MainPage ?? throw new InvalidOperationException($"{nameof(Application.Current.MainPage)} cannot be null");
```
brminnick, line 59: We can't reference `Application` in `CommunityToolkit.Maui.Core` because `Application` is in the `Microsoft.Maui.Controls` namespace

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
```diff
@@ -0,0 +1,175 @@
...
+			presentationController.Delegate = null;
+		}
+	}
+
+	PageHandler CreateControl(in IPopup virtualView)
+	{
+		var view = (View?)virtualView.Content;
+		var contentPage = new ContentPage { Content = view, BackgroundColor = Colors.Orange };
+
+		contentPage.Parent = Application.Current?.MainPage;
```
brminnick, line 99: We can't reference `Application` in `CommunityToolkit.Maui.Core` because `Application` is in the `Microsoft.Maui.Controls` namespace

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
```diff
@@ -0,0 +1,175 @@
...
+		if (presentationController is not null)
+		{
+			presentationController.Delegate = null;
+		}
+	}
+
+	PageHandler CreateControl(in IPopup virtualView)
+	{
+		var view = (View?)virtualView.Content;
+		var contentPage = new ContentPage { Content = view, BackgroundColor = Colors.Orange };
```
brminnick, line 97: We can't reference `ContentPage` in `CommunityToolkit.Maui.Core` because `ContentPage` is in the `Microsoft.Maui.Controls` namespace

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
```diff
@@ -0,0 +1,175 @@
...
+		var presentationController = (UIPopoverPresentationController)PresentationController;
+		if (presentationController is not null)
+		{
+			presentationController.Delegate = null;
+		}
+	}
+
+	PageHandler CreateControl(in IPopup virtualView)
+	{
+		var view = (View?)virtualView.Content;
```
brminnick, line 96: We can't reference `View` in `CommunityToolkit.Maui.Core` because `View` is in the `Microsoft.Maui.Controls` namespace

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
```diff
@@ -0,0 +1,175 @@
...
+
+		contentPage.Parent = Application.Current?.MainPage;
+		contentPage.SetBinding(VisualElement.BindingContextProperty, new Binding { Source = virtualView, Path = VisualElement.BindingContextProperty.PropertyName });
+
+		return (PageHandler)contentPage.ToHandler(mauiContext);
+	}
+
+	UIViewController CreateViewController(Page mainPage)
+	{
+		Page currentPageRenderer;
```
brminnick, line 107: We can't reference `Page` in `CommunityToolkit.Maui.Core` because `Page` is in the `Microsoft.Maui.Controls` namespace

brminnick (changes requested):
Hey Pedro! I just noticed today that we have a couple references to `Microsoft.Maui.Controls` in the `CommunityToolkit.Maui.Core` project.

We'll want to pull those out to ensure `CommunityToolkit.Maui.Core` will work with other frameworks that don't use `Microsoft.Maui.Controls`, like Comet.

pictos:
@brminnick I believe that I removed all Maui.Controls from the Core... And did some improvements in the API, merging `BasePopup` with `Popup`

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
```diff
@@ -0,0 +1,153 @@
...
+	public void CleanUp()
+	{
+		if (VirtualView is null)
+		{
+			return;
+		}
+
+		VirtualView = null;
+
+		var presentationController = (UIPopoverPresentationController)PresentationController;
```
VladislavAntonyuk, line 85:
```suggestion
		if (PresentationController is UIPopoverPresentationController presentationController)
```

src/CommunityToolkit.Maui.Core/Views/Popup/PopupExtensions.macios.cs
```diff
@@ -0,0 +1,88 @@
...
+	/// </summary>
+	/// <param name="popup">An instance of <see cref="MauiPopup"/>.</param>
+	/// <param name="basePopup">An istance of <see cref="IPopup"/>.</param>
+	public static void SetBackgroundColor(this MauiPopup popup, in IPopup basePopup)
+	{
+		if (popup.Control is null)
+		{
+			return;
+		}
+		var color = basePopup.Color?.ToNative() ?? null;
```
VladislavAntonyuk, line 36: it is null anyway, isn't it

src/CommunityToolkit.Maui/AppBuilderExtensions.cs
```diff
@@ -14,6 +16,12 @@ public static class AppBuilderExtensions
 	/// <returns><see cref="MauiAppBuilder"/> initialized for <see cref="CommunityToolkit.Maui"/></returns>
 	public static MauiAppBuilder UseMauiCommunityToolkit(this MauiAppBuilder builder)
 	{
+		builder.ConfigureMauiHandlers(h =>
+		{
+		   h.AddHandler(typeof(Popup), typeof(PopupViewHandler));
+		   //h.AddHandler(typeof(Popup), typeof(PopupViewHandler));
```
VladislavAntonyuk, line 22: can be removed

src/CommunityToolkit.Maui/Extensions/Popup/PopupExtensions.windows.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.Maui.Platform;
+
+namespace CommunityToolkit.Maui.Extensions;
+
+/// <summary>
+/// Extension methods for <see cref="Popup"/>.
+/// </summary>
+public static partial class PopupExtensions
+{
+	static void PlatformShowPopup(Popup popup, IMauiContext mauiContext)
```
VladislavAntonyuk, line 13: can we name it CreatePopup and remove this if? https://github.com/CommunityToolkit/Maui/pull/290/files#diff-dee80352ab8b3f5f51f2a55578e4b93a9fb981cafa99db2bd3991edb660368cbR25

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui/HandlerImpl/Popup/BasePopup.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+	public static CommandMapper<IPopup, PopupViewHandler> ControlPopUpCommandMapper = new(PopupViewHandler.PopUpCommandMapper)
+	{
+#if IOS
+		[nameof(IPopup.OnOpened)] = MapOnOpened
+#endif
+	};
+
+	internal static void RemapForControls()
+	{
+		PopupViewHandler.PopUpCommandMapper = ControlPopUpCommandMapper;
```
brminnick: This overwrites the original `ControlPopupCommandMapper` assignment. 

Is this intended? No worries if it is, but just want to make sure!

```cs
	public static CommandMapper<IPopup, PopupViewHandler> PopUpCommandMapper = new(ElementCommandMapper)
	{
		[nameof(IPopup.OnDismissed)] = MapOnDismissed,
#if !(IOS || MACCATALYST)
		[nameof(IPopup.OnOpened)] = MapOnOpened,
#endif
		[nameof(IPopup.LightDismiss)] = MapOnLightDismiss
	};
```

brminnick (changes requested):
Thanks Pedro! 

I love that we got rid of `BasePopup` (to be honest, I never really understood the need for it).

👇 There's a strange problem on iOS where the Popup is always much bigger than its content. Here's an example from the sample app. I'm not sure if this is a bug in the sample or a bug in the iOS implementation.
Screen Shot 2022-03-02 at 4 25 09 PM

I just pushed a couple commits, most just remove unused code. But there was also a couple larger updates:
- Rename `BasePopup.shared.cs` to `Popup.shared.cs`
- Add Nullable Static Analyzers 
- Rename `HandlerImpl` to `HandlerImplementation`
  - I know the .NET MAUI team uses `HandlerImpl`, but I'm strongly against abbreviating folder, file, + variable names (that was something devs did 10-20 years ago when storage was expensive)
- Move `PopupExtensions` from `Extensions/PopupExtensions/` to `Views/Popup/`
  - Since the `Extensions/` folder contains extension methods for `Microsoft.Maui` controls (i.e. not `CommunityToolkit.Maui` controls), these extensions methods fit better in the same folder as `Popup.shared.cs`
    - For example, the Sample App's Extensions Page shows the `ColorAnimationExtensions` that extend `VisualElement`
- Change default value of `Popup.ColorProperty` from `null` to `Colors.Transparant`

src/CommunityToolkit.Maui/HandlerImpl/Popup/BasePopup.shared.cs
pictos: Yes, it's the same approach they use [here](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/HandlerImpl/Label/Label.cs). We can use this _movement_ when we need to implement something on Handlers that needs the `Maui. Controls` namespace

pictos:
> I love that we got rid of BasePopup (to be honest, I never really understood the need for it).

This makes sense when we had the generic implementation (XCT old times), so if your Popup doesn't return a result you just inherit from `BasePopup`, now if your Popup returns a result you inherit from `Popup<T>`. 

```csharp
// On XCT
public abstract class Popup<T> : BasePopup 
{
 // All the jazz
}

//User code

public class PopupNewUser : Popup<User> // Or any object that you want
{
}
```
We can't do that on MAUI because it doesn't work well with generics. In the future, we can spike and see if we can bring that back (if the community want it)

>  There's a strange problem on iOS where the Popup is always much bigger than its content. Here's an example from the sample app. I'm not sure if this is a bug in the sample or a bug in the iOS implementation.

Yeah... Measurement is hard and iOS debug isn't helping me on this version of VS. As you can see in the image below it says that 414/2 is zero.

image

brminnick (dismissed):
Hey Pedro! I'm still concerned about the sizing problem on iOS vs Android.

Today I tried running the Popup samples on MacCatalyst and discovered that the Popup doesn't appear.

👇 Here's a screen recording of what I'm seeing

ScreenFlow

VladislavAntonyuk:
Probably on MacCatalyst it has the same background?

pictos:
I'll give some love to the iOS. For maccatalyst I don't anything about it, hope that is easy to solve.

samples/CommunityToolkit.Maui.Sample/Views/Popups/MultipleButtonPopup.xaml
```diff
@@ -0,0 +1,59 @@
...
+            </ResourceDictionary>
+        </VerticalStackLayout.Resources>
+
+
+        <Label Style="{StaticResource Title}" 
+                Text="Button Popup" />
+        <BoxView Style="{StaticResource Divider}" />
+        <Label Style="{StaticResource Content}" 
+                Text="This is a native popup with a .NET MAUI View being rendered. The behaviors of the popup will confirm to 100% native look and feel, but still allows you to use your .NET MAUI Controls." />
+        <StackLayout Style="{StaticResource ButtonGroup}">
```
VladislavAntonyuk, line 50: Let's use VerticalStackLayout

samples/CommunityToolkit.Maui.Sample/Views/Popups/ToggleSizePopup.xaml.cs
```diff
@@ -0,0 +1,26 @@
...
+
+	public ToggleSizePopup()
+	{
+		InitializeComponent();
+		originalSize = Size;
+	}
+
+	void Button_Clicked(object? sender, System.EventArgs e)
+	{
+		if (originalSize == Size)
```
VladislavAntonyuk, line 17: Can be simplified and written in 1 line. But it is not critical. I would prefer readability.

samples/CommunityToolkit.Maui.Sample/Views/Popups/TransparentPopup.xaml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<mct:Popup
+    x:Class="CommunityToolkit.Maui.Sample.TransparentPopup"
+    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    xmlns:models="clr-namespace:CommunityToolkit.Maui.Sample.Models"
+    xmlns:mct="clr-namespace:CommunityToolkit.Maui.Views;assembly=CommunityToolkit.Maui"
+    Size="{x:Static models:PopupSizeConstants.Small}">
+
+    <Grid>
```
VladislavAntonyuk, line 10: Do we need grid here?

src/CommunityToolkit.Maui.Core/Handlers/Popup/PopUpViewHandler.android.cs
```diff
@@ -0,0 +1,126 @@
...
+		handler.NativeView?.SetAnchor(view);
+	}
+
+	/// <inheritdoc/>
+	protected override MauiPopup CreateNativeElement()
+	{
+		_ = MauiContext ?? throw new InvalidOperationException("MauiContext is null, please check your MauiApplication.");
+		_ = MauiContext.Context ?? throw new InvalidOperationException("Android Context is null, please check your MauiApplication.");
+
+		return new MauiPopup(MauiContext.Context, MauiContext);
```
VladislavAntonyuk, line 105: You can use Context directly. Instead if MauiContext.Context

src/CommunityToolkit.Maui.Core/Handlers/Popup/PopupViewHandler.shared.cs
```diff
@@ -0,0 +1,127 @@
...
+	/// <summary>
+	/// Default Constructor for <see cref="PopupViewHandler"/>.
+	/// </summary>
+	public PopupViewHandler()
+		: base(PopUpMapper, PopUpCommandMapper)
+	{
+	}
+}
+
+#if !(ANDROID || IOS || MACCATALYST || WINDOWS)
```
VladislavAntonyuk, line 54: Can you move it to netstandard.cs?

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.android.cs
```diff
@@ -0,0 +1,103 @@
...
+	public MauiPopup(Context context, IMauiContext mauiContext)
+		: base(context)
+	{
+		this.mauiContext = mauiContext ?? throw new ArgumentNullException(nameof(mauiContext));
+	}
+
+	/// <summary>
+	/// An instace of the <see cref="IPopup"/>.
+	/// </summary>
+	public IPopup? VirtualView { get; private set; }
```
VladislavAntonyuk, line 32: Is it correct that we use VirtualView in NativeView?

src/CommunityToolkit.Maui/HandlerImplementation/Popup/Popup.shared.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace CommunityToolkit.Maui.Views;
+
+public partial class Popup
+{
+	/// <summary>
+	/// 
+	/// </summary>
+	public static CommandMapper<IPopup, PopupViewHandler> ControlPopUpCommandMapper = new(PopupViewHandler.PopUpCommandMapper)
+	{
+#if IOS
```
VladislavAntonyuk, line 13: Probably MacCatalyst as well

VladislavAntonyuk (commented):
Just some minor comments and suggestions. I have no objections to merge it

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.android.cs
pictos: Yes. We need it here because we want to have access to values from the shared code

src/CommunityToolkit.Maui.Core/Handlers/Popup/PopUpViewHandler.android.cs
pictos: We don't have the `Context` here, because the Popup extends from `ElementHandler` and not from `ViewHandler`

src/CommunityToolkit.Maui.Core/Handlers/Popup/PopupViewHandler.shared.cs
pictos: Sure. Actually it was in a separated file, don't remember to change it

src/CommunityToolkit.Maui/HandlerImplementation/Popup/Popup.shared.cs
pictos: Just figured out that you discovered why the popup didn't show for MacCatalyst. Great catch!

VladislavAntonyuk:
I checked it on Windows:
I found some issues:
1. MultiplePopupPage->ToggleSizePopup->ToggleSize crashes the app:
image

2. Background color on popup looks transparent:
image

3. Popup disappear on window resize

brminnick:
Dang! I didn't realize that `WeakEventManager` lives in the `Microsoft.Maui.Controls` namespace.

I wonder if we can petition the .NET MAUI team to move it to the core namespace before GA?

image

brminnick:
> Dang! I didn't realize that `WeakEventManager` lives in the `Microsoft.Maui.Controls` namespace.
> 
> I wonder if we can petition the .NET MAUI team to move it to the core namespace before GA?

**Update:** I spoke with Matthew Leibowitz and opened a PR for it: https://github.com/dotnet/maui/pull/5139

brminnick:
Nah, we can still merge this 💯 

In this instance we don't _need_ `WeakEventManager` since it was for an Event being used inside of a sealed private class. 

If the .NET MAUI team approves my PR, then we can always circle back and update this code for `WeakEventManager`.

pictos:
@VladislavAntonyuk 
1. Fixed, this issue is caused because we can't change the values from Style, so we need to instantiate a new one
2. It's expected since I don't set a background color on ou popups, and the default is transparent.
3. It's the normal behavior, if you run the XCT sample will be the same


Now I believe that we're good to go

src/CommunityToolkit.Maui.Core/Handlers/Popup/PopupViewHandler.shared.cs
```diff
@@ -49,4 +49,78 @@ public PopupViewHandler()
 		: base(PopUpMapper, PopUpCommandMapper)
 	{
 	}
-}
\ No newline at end of file
+}
+
+#if !(IOS || ANDROID || MACCATALYST || WINDOWS)
```
pictos, line 54: @brminnick why not use a separated file?

brminnick: We will! 

I opened Discussion #301 to chat about the best/recommended way to do it.

For this PR, we'll keep with the current convention of `#if !(IOS || ANDROID || MACCATALYST || WINDOWS)` like we used in `Snackbar` and `Toast`. 

Then we'll move this into a separate file in the future once we've hammered out the best way to go about it in #301.

Long story short, I don't want that Discussion + future Proposal holding up this PR.

brminnick (commented):
Thanks Pedro! I just had a thought about the Popup border on Windows. Let me know what you think!

src/CommunityToolkit.Maui.Core/Interfaces/IPopup.shared.cs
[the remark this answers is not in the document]
brminnick: Hey Gang! Now is probably a good time to update/resolve this conversation now that we're getting close to merging Popup!

My vote is to move forward with Shaun's naming recommendation:
-  `OnClosed()` 
    - Triggers when the user taps a close button to dismiss a Popup 
- `OnDismissed()` 
    -  Triggers when the user taps outside of the Popup to dismiss it

As someone who primarily works on iOS + Android (and rarely works on Windows), I was unfamiliar with Windows' `Light Dismiss` nomenclature and I personally had to google `LightDismiss` in order to  understand this feature.

My concern is that other devs may have a similar experience and using `Light Dismiss` will also slow down their development velocity and decrease their discoverability of this feature.

That being said, I'm happy to be outvoted if most folks **are** familiar with the term `Light Dismiss`!

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.windows.cs
```diff
@@ -0,0 +1,293 @@
...
+using XamlStyle = Microsoft.UI.Xaml.Style;
+
+namespace CommunityToolkit.Core.Views;
+
+/// <summary>
+/// The navite implementation of Popup control.
+/// </summary>
+public class MauiPopup : Flyout
+{
+	const double defaultBorderThickness = 2;
```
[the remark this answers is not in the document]
pictos: Let's remove the border here. And add in the future if needed

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui/AssemblyInfo.shared.cs
```diff
@@ -1,6 +1,7 @@
 ﻿[assembly: XmlnsDefinition(Constants.XamlNamespace, Constants.CommunityToolkitNamespacePrefix + nameof(CommunityToolkit.Maui.Alerts))]
 [assembly: XmlnsDefinition(Constants.XamlNamespace, Constants.CommunityToolkitNamespacePrefix + nameof(CommunityToolkit.Maui.Behaviors))]
 [assembly: XmlnsDefinition(Constants.XamlNamespace, Constants.CommunityToolkitNamespacePrefix + nameof(CommunityToolkit.Maui.Converters))]
+[assembly: XmlnsDefinition(Constants.XamlNamespace, Constants.CommunityToolkitNamespacePrefix + nameof(CommunityToolkit.Maui.Views))]
```
VladislavAntonyuk, line 4: could you please add Layouts here as well

VladislavAntonyuk (dismissed):

VladislavAntonyuk (dismissed):

pictos:
@brminnick there's something else that you want before merge this?

brminnick:
Yes! I wanted to circle back to our discussion about the naming for LightDismiss now that we're getting close to merging Popup!

### Concern

As someone who primarily works on iOS + Android (and rarely works on Windows), I was unfamiliar with Windows' `Light Dismiss` nomenclature and I personally had to google `LightDismiss` in order to  understand this feature.

My concern is that other devs may have a similar experience and using `Light Dismiss` will also slow down their development velocity and decrease their discoverability of this feature.

That being said, I'm happy to be outvoted if most folks **are** familiar with the term `Light Dismiss`!

### Research

(Copy/Pasting the info from the [Code Review Conversation](https://github.com/CommunityToolkit/Maui/pull/290/files/679322913e3686043e49055601ccdebbf5f886bc#r805105055)):
> Light Dismiss seems to be a Windows term: https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.primitives.popup.islightdismissenabled?view=winrt-22000
> 
> It does not seem to be an Android term: https://www.google.com/search?q=light+dismiss+Android
> 
> It does not seem to be an iOS term: https://www.google.com/search?q=light+dismiss+ios

### Shaun's Recommendation

My vote is to move forward with [@bijington naming recommendation](https://github.com/CommunityToolkit/Maui/pull/290/files/679322913e3686043e49055601ccdebbf5f886bc#r804490776):
-  `OnClosed()` 
    - Triggers when the user taps a close button to dismiss a Popup 
- `OnDismissed()` 
    -  Triggers when the user taps outside of the Popup to dismiss it

## Let's Vote! 

Leave the following emoji reaction for your preference:
🚀 Keep `LightDismiss`
👀 Use `Dismiss` and `Close`
❤️ Use other naming convention (leave a comment with recommendation)

VladislavAntonyuk:
We can call it Hide as the simplest variant. 
But another proposal:
We can add boolean Backdrop property.
Backdrop=true prevents the popup closing on background click.
Add new event "hidePrevented" which is raised when user clicks on background.

pictos:
@VladislavAntonyuk maybe we could call cancel instead of hide. Like hide gives an idea that the pop-up is not visible, but actually it was destroyed

jfversluis:
I'm a bit biased here as I already knew what light dismiss does, but I still voted for close/dismissed.

I feel cancelled wouldn't really cover it because there is still a way to get some kind of result out of the popup even when dismissed so it's not really cancelled? I don't know... Naming is hard :P

VladislavAntonyuk:
I agree with Gerald. Cancel for me is like an action. reverting some changes.
Does it matter for us how it was dismissed?
To keep it simple we can use only 1 method Dismiss with the reason parameter/result.

bijington:
That is a good point @VladislavAntonyuk. Sorry one bit I am missing, is there a reason we need to know the difference between a LightDismiss and a Dismiss? Other Popup controls seem to only offer a Closed command/method/event

VladislavAntonyuk (dismissed):

brminnick (approved):
image

PR #296, opened 2022-02-14
Replaces `NotImplementedException` with `NotSupportedException`

bijington:
### Description of Change ###

Replaces all usages of `NotImplementedException` with `NotSupportedException`.
 

 ### Linked Issues ###
 

 - Fixes #292

 ### PR Checklist ###
 
 - [ ] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [ ] Has tests (if omitted, state reason in description)
 - [ ] Has samples (if omitted, state reason in description)
 - [x] Rebased on top of `main` at time of PR
 - [x] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#contributing-code---best-practices)
 - [ ] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pulls/XYZ 

 ### Additional information ###

PR #310, opened 2022-03-02
Add Null Support for BaseOneWayConverter<TFrom, TTo> 

brminnick:
> Note: This PR targets the `feature/not-supported-exception` branch, aka #296. It does not target the `main` branch

 ### Description of Change ###

This PR adds `null` support for `BaseOneWayConverter<TFrom, TTo>`.

Adding support for `null` enables us to use nullable generics, eg `<TFrom?, TTo?>` which then introduces the possibility to update the following converters from `BaseOneWayConverter` to `BaseOneWayConverter<TFrom, TTo>`:
- `public class ImageResourceConverter : BaseOneWayConverter<TFrom, TTo>`
- `public class IsStringNotNullEmptyConverter : BaseOneWayConverter<TFrom, TTo>`
- `public class IsListNotNullOrEmptyConverter : BaseOneWayConverter<TFrom, TTo>`
- `public class IsListNullOrEmptyConverter : BaseOneWayConverter<TFrom, TTo>`
- `public class ItemSelectedEventArgsConverter : BaseOneWayConverter<TFrom, TTo>`
- `public class ItemTappedEventArgsConverter : BaseOneWayConverter<TFrom, TTo>`

### DIFF
--- a/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
+++ b/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
@@ -27,7 +27,7 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
 #pragma warning restore CS8604 // Possible null reference argument.
 		}
 
-		if (value is not TFrom)
+		if (value is not TFrom convertedValue)
 		{
 			throw new ArgumentException($"value needs to be of type {typeof(TFrom)}");
 		}
@@ -37,7 +37,7 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
 			throw new ArgumentException($"targetType needs to be typeof {typeof(TTo)}");
 		}
 
-		return ConvertFrom((TFrom)value);
+		return ConvertFrom(convertedValue);
 	}
 
 	/// <summary>
--- a/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
+++ b/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
@@ -56,7 +56,7 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
 			return true; // ref-type
 		}
 
-		if (Nullable.GetUnderlyingType(type) != null)
+		if (Nullable.GetUnderlyingType(type) is not null)
 		{
 			return true; // Nullable<T>
 		}
--- a/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
+++ b/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
@@ -27,7 +27,7 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
 #pragma warning restore CS8604 // Possible null reference argument.
 		}
 
-		if (value?.GetType() != typeof(TFrom))
+		if (value is not TFrom)
 		{
 			throw new ArgumentException($"value needs to be of type {typeof(TFrom)}");
 		}
--- a/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
+++ b/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
@@ -51,17 +51,17 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
 	{
 		var type = typeof(T);
 
-		if (!type.IsValueType)
+		if (type.IsValueType)
 		{
-			return true; // ref-type
+			return true; // value type
 		}
 
-		if (Nullable.GetUnderlyingType(type) != null)
+		if (Nullable.GetUnderlyingType(type) is not null)
 		{
 			return true; // Nullable<T>
 		}
 
-		return false; // value-type
+		return true; // ref type
 	}
 }
 
--- a/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
+++ b/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
@@ -51,17 +51,17 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
 	{
 		var type = typeof(T);
 
-		if (type.IsValueType)
+		if (!type.IsValueType)
 		{
-			return true; // value type
+			return true; // ref-type
 		}
 
-		if (Nullable.GetUnderlyingType(type) is not null)
+		if (Nullable.GetUnderlyingType(type) != null)
 		{
 			return true; // Nullable<T>
 		}
 
-		return true; // ref type
+		return false; // value-type
 	}
 }
 
--- a/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
+++ b/src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
@@ -28,7 +28,7 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
 		}
 		else if (value is null && !IsNullable<TFrom>())
 		{
-			throw new ArgumentException($"value cannot be null because {nameof(TFrom)} is not Nullable");
+			throw new ArgumentNullException(nameof(value), $"value cannot be null because {nameof(TFrom)} is not Nullable");
 		}
 
 		if (value is not TFrom convertedValue)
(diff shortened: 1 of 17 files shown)


bijington (commented):
@brminnick Great work! I really like this approach.

Just one point on the `IsNullable` implementation as it always returns true. 

I plan to have a play with using Nullable and non nullable types to help my understanding

src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
```diff
@@ -39,6 +46,23 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
...
 	/// <returns>An object of type <typeparamref name="TTo"/>.</returns>
 	public abstract TTo? ConvertFrom(TFrom value);
+
+	static bool IsNullable<T>()
+	{
+		var type = typeof(T);
+
+		if (type.IsValueType)
+		{
+			return true; // value type
```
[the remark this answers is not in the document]
brminnick: Sun of a gun! I made a last-minute change and forgot to negate that. Fantastic catch 🙌

bijington:
I love the direction of this PR. It opens the door to so many of our converters for using this base implementation.

brminnick:
Yup, I'm happy to merge it now, assuming you approve! 

> do you believe this will allow us to remove the BaseOneWayConverter that I added in https://github.com/CommunityToolkit/Maui/pull/296?

This PR targets #296, and these updates will merge into your existing PR. So there's no need to remove anything from #296.

Screen Shot 2022-03-02 at 9 32 26 AM

bijington:
> Yup, I'm happy to merge it now, assuming you approve! 
> 
> 
> 
> > do you believe this will allow us to remove the BaseOneWayConverter that I added in https://github.com/CommunityToolkit/Maui/pull/296?
> 
> 
> 
> This PR targets #296, and these updates will merge into your existing PR. So there's no need to remove anything from #296.
> 
> 
> 
> Screen Shot 2022-03-02 at 9 32 26 AM
> 
> 

I couldn't see a way to approve it as it doesn't mention needing a review. I'll take another look over and hit merge if all is good.

As for the no work part. Your rework of the base one way converter can remove the need for a class I added in that PR. I'll take a look once it merges and see 😃

brminnick:
Ooooh - I understand now! You're refereeing to the non-generic `BaseConverterOneWay` class. 

The non-generic `BaseConverterOneWay` is still being used by converters that need the additional parameters from `IValueConverter.Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)`.

`BaseConverterOneWay<TFrom, TTo>.ConvertFrom` only provides the `value` parameter:
```cs
public abstract TTo? ConvertFrom(TFrom value);
```

We could probably update `ConvertFrom` to something like this, and then we could delete the non-generic `BaseConverterOneWay` class.

```cs
public abstract TTo? ConvertFrom(TFrom value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null);
```

brminnick:
Right on! Yup, we just need to remember to update Unit Tests to test both `.Convert` and `.ConvertFrom` since both are valid paths once we port a converter from `BaseConverterOneWay` to `BaseConverterOneWay<TFrom, TTo>`.

src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
```diff
@@ -18,9 +18,16 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
 	/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
 	/// <param name="culture">The culture to use in the converter. This is not implemented.</param>
 	/// <returns>An object of type <typeparamref name="TTo"/>.</returns>
-	public override object? Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
+	public override sealed object? Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
 	{
-		if (value is not TFrom valueFrom)
+		if (value is null && IsNullable<TFrom>())
+		{
+#pragma warning disable CS8604 // Possible null reference argument.
```
bijington, line 25: I don't understand the need for this. I expect that if `TFrom` is not nullable then default would not return `null`. Take a `bool` for example, `default(bool)` is `false` not `null` whereas `default(bool?)` would be `null`.

Do you know if this is needed and why?

brminnick: Yea, it's a bit of a weird case and it took me a minute to understand it as well. 

In `if (value is not TFrom valueFrom)`, `valueFrom` will never be `null`.

So if a user passes in `null` to the converter, `if (value is not TFrom valueFrom)` will resolve to `true` and we'll throw an `ArgumentException`.

You can see this by removing the `if (value is null && IsNullable<TFrom>())` block and running the unit tests.

bijington: Hmm ok, I think I might be getting it, `TFrom` is null and `ConvertFrom` doesn't specify that `TFrom` can be null (we require the inheritor declares that which we check inside `IsNullable<>`.

But wait a minute why is `ConvertFrom(default(TFrom)` not valid as it's claiming `default(TFrom)` could be null but if it could be null surely that means the signature of `ConvertFrom(TFrom ...)` would be allowing for `TFrom` to be null? Now I am even more confused...

Although on this topic given the above does that mean we could change the signature of `ConvertFrom` from:

```csharp
public abstract TTo? ConvertFrom(TFrom value);
```

to:

```csharp
public abstract TTo ConvertFrom(TFrom value);
```

This would then let the inheriting converter implementation determine whether a null can be returned?

brminnick: The reason I use `return ConvertFrom(default)` instead of `return ConvertFrom(null)` or `return ConvertFrom((TFrom)null)` is mostly thanks to the Roslyn analyzers.

In this case, since we're confident that `value` is `null` and that `TFrom` is nullable, return `default` is the same as returning `null`.

Roslyn gets mad if we try to do something like `return `ConvertFrom((TFrom)null)`, because it doesn't know whether or not `TFrom` is nullable.

bijington: OK that makes some sense, so I guess in an ideal world we could inform Roslyn that `TFrom` is in fact nullable through our `IsNullable` method call? e.g. using something like `[NullWhen]` or similar. I don't know if there is something that fits the bill here though.

brminnick: Right! But there's no support for generics in the [Nullable Static Analyzers](https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/nullable-analysis), so there's no way for us to tell Roslyn that TFrom is nullable and/or that `value` is null. 

And you're correct that the class that inherits/implements `BaseConverterOneWay<TFrom, TTo>` chooses whether or not `TFrom` and `TTo` are nullable.

So we're kinda caught in this weird in-between place where this is the best solution we can do until a new version of C#/Roslyn helps us out.

bijington: OK. Thanks for helping my understanding here 😄

src/CommunityToolkit.Maui/Converters/BaseConverterOneWay.shared.cs
```diff
@@ -18,9 +18,20 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : BaseConverterOneWay
...
+			return ConvertFrom(default);
+#pragma warning restore CS8604 // Possible null reference argument.
+		}
+		else if (value is null && !IsNullable<TFrom>())
+		{
+			throw new ArgumentException($"value cannot be null because {nameof(TFrom)} is not Nullable");
```
bijington, line 31: Sorry one last thing (I hope) could/should this be `ArgumentNullException` instead here?

brminnick: Good question! I think that's valid. 

[ArgumentNullException](https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception?view=net-6.0):

>An [ArgumentNullException](https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception?view=net-6.0) exception is thrown when a method is invoked and at least one of the passed arguments is null but should never be null.
An [ArgumentNullException](https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception?view=net-6.0) exception is thrown at run time in the following two major circumstances, both of which reflect developer error:
An uninstantiated object is passed to a method. To prevent the error, instantiate the object.
>
> An object returned from a method call is then passed as an argument to a second method, but the value of the original returned object is null. To prevent the error, check for a return value that is null and call the second method only if the return value is not null.
> 
> [ArgumentNullException](https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception?view=net-6.0) behaves identically to [ArgumentException](https://docs.microsoft.com/en-us/dotnet/api/system.argumentexception?view=net-6.0). It is provided so that application code can differentiate between exceptions caused by null arguments and exceptions caused by arguments that are not null.

brminnick: Updated ✅

bijington:
Thanks for this work @brminnick!

PR #318, opened 2022-03-07
Add `MaskedBehavor` Unit Tests

### LINKED ISSUE #46: [Proposal] Masked Behavior
# MaskedBehavior

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Documentation
* [x] Sample

## Summary
[summary]: #summary

The `MaskedBehavior` is a `Behavior` that allows the user to define an input mask for data entry. Adding this behavior to an `Entry` control will force the user to only input values matching a given mask. Examples of its usage include input of a credit card number or a phone number

## Detailed Design
[design]: #detailed-design

### MaskedBehavior.shared.cs

```cs
public class MaskedBehavior : BaseBehavior<InputView>
{
    public static readonly BindableProperty MaskProperty;
    public static readonly BindableProperty UnMaskedCharacterProperty;

    public string? Mask { get; set; }
    public char UnMaskedCharacter { get; set; }
}
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<Entry>
    <Entry.Behaviors>
        <xct:MaskedBehavior 
            Mask="AA-AA-AA" 
            UnMaskedCharacter="A"
        />
    </Entry.Behaviors>
</Entry>
```

### C# Usage

```cs
var entry = new Entry();
entry.Behaviors.Add(new MaskedBehavior
{
    Mask="AA-AA-AA" 
    UnMaskedCharacter="A"
});
```

brminnick:
### Description of Change ###

This PR implements the Unit Tests for #46 

 ### Linked Issues ###
 

 - Fixes #46 

 ### PR Checklist ###

I also updated `BaseBehavior.OnViewPropertyChanged` to use `TView`:

**Updated**

```cs
protected virtual void OnViewPropertyChanged(TView sender, PropertyChangedEventArgs e)
{

}
 ```

**Previous**

```cs
protected virtual void OnViewPropertyChanged(object? sender, PropertyChangedEventArgs e)
{

}
```

### DIFF
--- a/src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
@@ -43,7 +43,7 @@ public class MaskedBehavior : BaseBehavior<InputView>
 	}
 
 	/// <inheritdoc />
-	protected override async void OnViewPropertyChanged(object? sender, PropertyChangedEventArgs e)
+	protected override async void OnViewPropertyChanged(InputView sender, PropertyChangedEventArgs e)
 	{
 		base.OnViewPropertyChanged(sender, e);
 
@@ -134,7 +134,7 @@ public class MaskedBehavior : BaseBehavior<InputView>
 		{
 			if (!string.IsNullOrWhiteSpace(text) && maskPositions is not null)
 			{
-				if (text.Length > (Mask?.Length ?? 0))
+				if (Mask is not null && text.Length > Mask.Length)
 				{
 					text = text.Remove(text.Length - 1);
 				}
--- a/src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
@@ -1,4 +1,5 @@
 ﻿using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 
 namespace CommunityToolkit.Maui.Behaviors;
 
@@ -8,20 +9,20 @@ namespace CommunityToolkit.Maui.Behaviors;
 public class MaskedBehavior : BaseBehavior<InputView>
 {
 	/// <summary>
-	/// Backing BindableProperty for the <see cref="Mask"/> property.
+	/// BindableProperty for the <see cref="Mask"/> property.
 	/// </summary>
 	public static readonly BindableProperty MaskProperty =
 		BindableProperty.Create(nameof(Mask), typeof(string), typeof(MaskedBehavior), propertyChanged: OnMaskPropertyChanged);
 
 	/// <summary>
-	/// Backing BindableProperty for the <see cref="UnMaskedCharacter"/> property.
+	/// BindableProperty for the <see cref="UnmaskedCharacter"/> property.
 	/// </summary>
-	public static readonly BindableProperty UnMaskedCharacterProperty =
-		BindableProperty.Create(nameof(UnMaskedCharacter), typeof(char), typeof(MaskedBehavior), 'X', propertyChanged: OnUnMaskedCharacterPropertyChanged);
+	public static readonly BindableProperty UnmaskedCharacterProperty =
+		BindableProperty.Create(nameof(UnmaskedCharacter), typeof(char), typeof(MaskedBehavior), 'X', propertyChanged: OnUnmaskedCharacterPropertyChanged);
 
-	IDictionary<int, char>? positions;
+	readonly SemaphoreSlim applyMaskSemaphoreSlim = new(1, 1);
 
-	bool applyingMask;
+	IReadOnlyDictionary<int, char>? maskPositions;
 
 	/// <summary>
 	/// The mask that the input value needs to match. This is a bindable property.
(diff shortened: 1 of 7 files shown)


src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
```diff
@@ -35,127 +36,135 @@ public class MaskedBehavior : BaseBehavior<InputView>
...
 	/// <inheritdoc />
-	protected override void OnViewPropertyChanged(object? sender, PropertyChangedEventArgs e)
+	protected override async void OnViewPropertyChanged(InputView sender, PropertyChangedEventArgs e)
 	{
 		base.OnViewPropertyChanged(sender, e);
 
 		if (e.PropertyName == InputView.TextProperty.PropertyName)
 		{
-			OnTextPropertyChanged();
+			await OnTextPropertyChanged();
```
VladislavAntonyuk, line 52: Can we and should we use Cancellation token here?

src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
```diff
@@ -35,127 +36,135 @@ public class MaskedBehavior : BaseBehavior<InputView>
...
+
+		await ApplyMask(originalText);
 	}
 
-	string? RemoveMaskNullableString(string? text)
+	[return: NotNullIfNotNull("text")]
+	string? RemoveMask(string? text)
 	{
-		if (text == null || string.IsNullOrEmpty(text))
+		if (text is null)
```
VladislavAntonyuk, line 111: Instead of 2 ifs you can write 1: if (string.IsNullOrEmpty(text)) return text;

src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
```diff
@@ -35,127 +36,135 @@ public class MaskedBehavior : BaseBehavior<InputView>
...
-	void ApplyMask(string? text)
+	async Task ApplyMask(string? text)
 	{
-		if (text != null && !string.IsNullOrWhiteSpace(text) && positions != null)
-		{
-			if (text.Length > (Mask?.Length ?? 0))
-			{
-				text = text.Remove(text.Length - 1);
-			}
+		await applyMaskSemaphoreSlim.WaitAsync();
```
VladislavAntonyuk, line 131: Should we pass cancellation token here?

src/CommunityToolkit.Maui/Behaviors/MaxLengthReachedBehavior.shared.cs
```diff
@@ -50,9 +50,10 @@ public bool ShouldDismissKeyboardAutomatically
 	}
 
 	/// <inheritdoc/>
-	protected override void OnViewPropertyChanged(object? sender, PropertyChangedEventArgs e)
+	protected override void OnViewPropertyChanged(InputView sender, PropertyChangedEventArgs e)
```
VladislavAntonyuk, line 53: Maybe we should use interface here

src/CommunityToolkit.Maui/Behaviors/Validators/ValidationBehavior.shared.cs
```diff
@@ -245,15 +245,13 @@ protected override void OnDetachingFrom(VisualElement bindable)
 	}
 
 	/// <inheritdoc/>
-	protected override async void OnViewPropertyChanged(object? sender, PropertyChangedEventArgs e)
+	protected override async void OnViewPropertyChanged(VisualElement sender, PropertyChangedEventArgs e)
```
VladislavAntonyuk, line 248: Can we use interface here?

src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
brminnick: Absolutely! Good catch 💯

src/CommunityToolkit.Maui/Behaviors/MaxLengthReachedBehavior.shared.cs
pictos: Since we are on `Community.Maui` projects it's safe to use the `Maui.Controls` objects. Actually could be safer than using the interface... Because here we force the dev to really use an InputView and we designed this behavior to work on this control. An interface will be more "generic".

brminnick: For Behaviors, we can't use interfaces because `public IList<Behavior> Behaviors` isn't defined in any interface; it is only defined in `VisualElement`:

https://github.com/dotnet/maui/blob/ec33802c4ab453e13a24b1fc63bb019cd28152f9/src/Controls/src/Core/VisualElement.cs#L421-L424

This means that for `BaseBehavior<TView>`, the lowest that `TView` can be constrained is to `VisualElement`:
```cs
public abstract class BaseBehavior<TView> : Behavior<TView> where TView : VisualElement
```

src/CommunityToolkit.Maui/Behaviors/Validators/ValidationBehavior.shared.cs
brminnick: For Behaviors, we can't constrain to interfaces because `public IList<Behavior> Behaviors` isn't defined in any interface; it is only defined in `VisualElement`:

https://github.com/dotnet/maui/blob/ec33802c4ab453e13a24b1fc63bb019cd28152f9/src/Controls/src/Core/VisualElement.cs#L421-L424

This means that for `BaseBehavior<TView>`, the lowest that `TView` can be constrained is to `VisualElement`:
```cs
public abstract class BaseBehavior<TView> : Behavior<TView> where TView : VisualElement
```

And since `ValidationBehavior ` implements `BaseBehavior<TView>`, the lowest we can constrain `TView` is `InputView`:

```cs
public class ValidationBehavior : BaseBehavior<VisualElement>
```

src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
brminnick: Nice! 

Yea, I was trying to be too cute by adding `[return: NotNullIfNotNull("text")]`, thinking that the analyzers would need me to check for `null` and return `null` explicitly.

But you're right - this totally works 💯 

```cs
[return: NotNullIfNotNull("text")]
string? RemoveMask(string? text)
{
  if (string.IsNullOrEmpty(text))
  {
	  return text;
  }
// ...
```

brminnick: Updated ✅

src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
brminnick: I added the `CancellationToken token = default` as a parameter, but then realized that there's no way to pass in a token from .NET MAUI. I.e. These methods are (currently) only called from `PropertyChanged` which doesn't provide a token. 

But, I still think it's a good idea to have a `CancellationToken` as a parameter so that these methods can be leveraged by future code that maybe can supply a token 👍

src/CommunityToolkit.Maui/Behaviors/MaskedBehavior.shared.cs
brminnick: I added the `CancellationToken token = default` as a parameter, but then realized that there's no way to pass in a token from .NET MAUI. I.e. These methods are (currently) only called from `PropertyChanged` which doesn't provide a token. 

But, I still think it's a good idea to have a `CancellationToken` as a parameter so that these methods can be leveraged by future code that maybe can supply a token 👍

VladislavAntonyuk (approved):

PR #331, opened 2022-03-21
Remove Non-Generic `BaseConverter`

brminnick:
### Description of Change ###

This PR removes the non-Generic implementation of `BaseConverter`. This accomplished by expanding the `abstract` methods of `BaseConverter<T, U>` to support additional converter parameters: `Type? targetType`, `object? parameter`, `CultureInfo? culture`.

```cs
public abstract class BaseConverter<TFrom, TTo> : ValueConverterExtension, ICommunityToolkitValueConverter
{
	public abstract TTo ConvertFrom(TFrom value, Type targetType, object? parameter, CultureInfo? culture);

	public abstract TFrom ConvertBackTo(TTo value, Type targetType, object? parameter, CultureInfo? culture);
}
```

- Updated The Following Converters to use `BaseOneWayConverter<T>`:
  - `BoolToObjectConverter`
  - `EnumToBoolConverter`
  - `IsEqualConverter`
  - `IsNotEqualConverter`
  - `ListToStringConverter`
  - `MathExpressionConverter`
  - `StateToBooleanConverter`
  - `StringToListConverter`
  - `TextCaseConverter`
- Updated The Following Converters to use `BaseConverter<T, U>`:
  - `ByteArrayToImageSourceConverter`
  - `CompareConverter`
  - `DateTimeOffsetConverter`
  - `DoubleToIntConverter`
  - `EnumToIntConverter`
  - `IndexToArrayItemConverter`
  - `InvertedBoolConverter`
  - `TimeSpanToSecondsConverter`
- Fix `IsEqualConverter` Naming
  - `EqualConverter` -> `IsEqualConverter`
  - `NotEqualConverter` -> `IsNotEqualConverter`
  - This aligns with every other Converters that return `bool` such as `IsListNullOrEmptyConverter`, `IsStringNullOrEmptyConverter`, etc and should improve discoverability
- Fix `SelectedItemEventArgsConverter` Naming
  - `ItemSelectedEventArgsConverterPage` -> `SelectedItemEventArgsConverterPage`
  - This naming aligns with its return value of `SelectedItemEventArgs` and should improve discoverability
- Fix `TimeSpanToSecondsConverter` Naming
  - `TimeSpanToDoubleConverter` -> `TimeSpanToSecondsConverter`
  - Because this converter returns `TimeSpan.TotalSeconds`, including `Seconds` in the name ensures devs understand that we are _not_ returning `TimeSpan.TotalMinutes`, `TimeSpan.TotalHours`, etc which each also return `double`
- Updated associated Unit Tests


src/CommunityToolkit.Maui.UnitTests/Converters/ColorToColorConverter_Tests.cs
```diff
@@ -5,75 +5,75 @@ namespace CommunityToolkit.Maui.UnitTests.Converters;
 
 public class ColorToColorConverter_Tests : BaseTest
 {
-	public static IReadOnlyList<object?[]> ColorToBlackOrWhiteData { get; } = new[]
```
VladislavAntonyuk, line 8: Please check if it works. According to xunit documentation it should be object

brminnick:
Todo:
- [ ] Fix Sample App crash when `IsEqualConverterPage` is loaded
- [ ] Fix Sample App crash when `IsNotEqualConverterPage` is loaded
- [ ] Fix Sample App crash when `StringToListConverterPage` is loaded

src/CommunityToolkit.Maui/Converters/IndexToArrayItemConverter.shared.cs
```diff
@@ -17,24 +16,19 @@ public class IndexToArrayItemConverter : ValueConverterExtension, ICommunityTool
...
 		}
 
-		if (index < 0 || index >= array.Length)
+		if (value < 0 || value >= array.Length)
 		{
 			throw new ArgumentOutOfRangeException(nameof(value), "Index was out of range");
 		}
 
-		return array.GetValue(index);
+		return array.GetValue(value);
```
VladislavAntonyuk, line 31: Let's name it index

brminnick:
Thanks @bijington! I've now fixed the errors and moved the PR from `Draft` to `Ready for Review`!

Yea, this PR ended up being a bigger effort than I had anticipated 😅

src/CommunityToolkit.Maui.UnitTests/Converters/ColorToColorConverter_Tests.cs
brminnick: Screen Shot 2022-03-21 at 4 35 07 PM
Screen Shot 2022-03-21 at 4 36 04 PM
It does! I was surprised too

src/CommunityToolkit.Maui/Converters/IndexToArrayItemConverter.shared.cs
brminnick: This is just the name of the parameter, `int value`, that is required by `abstract class  BaseConverter<int, object?>`:

```cs
public abstract TTo ConvertFrom(TFrom value, Type targetType, object? parameter, CultureInfo? culture);
```

bijington (commented):
@brminnick thanks for the mammoth effort! Just 2 comments and they might just help me understand bits rather than require change

src/CommunityToolkit.Maui/Converters/ListToStringConverter.shared.cs
```diff
@@ -22,25 +21,16 @@ public class ListToStringConverter : BaseConverterOneWay
...
-		if (value == null)
-		{
-			return string.Empty;
-		}
-
-		if (value is not IEnumerable enumerable)
-		{
-			throw new ArgumentException("Value cannot be casted to IEnumerable", nameof(value));
-		}
+		ArgumentNullException.ThrowIfNull(value);
```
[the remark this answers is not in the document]
brminnick: Correct! 

I did intentionally for a couple reasons:
1. I think it's best to nudge devs away from using `null`
2. We can add  support for `null` later (after GA) without introducing a breaking change
  - If we leave in support for `null` now, we cannot remove it later without a breaking change
  - If we get feedback that many devs are using `null` lists, we can easily add support for it later

src/CommunityToolkit.Maui.UnitTests/Converters/StringToListConverter_Tests.cs
```diff
@@ -19,30 +19,40 @@ public class StringToListConverter_Tests : BaseTest
 
 	[Theory]
 	[MemberData(nameof(ListData))]
-	public void StringToListConverter(object? value, object? parameter, object? expectedResult)
+	public void StringToListConverter(string? value, object? parameter, object? expectedResult)
 	{
 		var stringToListConverter = new StringToListConverter();
 
-		var result = (IEnumerable<string>)stringToListConverter.Convert(value, typeof(IEnumerable<string>), parameter, null);
+		var convertFromResult = stringToListConverter.ConvertFrom(value, typeof(IList<string>), parameter, null);
```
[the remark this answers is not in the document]
brminnick: Good question! 

I initially made `StringToListConverter` require an `IList` (mostly because its name contains the word `List`. But, after some testing, I lowered it from `IList` to `IEnumerable`.

That being said, I think it's best to keep this Unit Test as `IList` because it demonstrates/tests the updated validation method of  `BaseConverter`:
```cs
if (!typeof(TTo).IsAssignableFrom(targetType) && !IsValidTargetType<TTo>(targetType))
	throw new ArgumentException($"targetType needs to be assignable from {typeof(TTo)}", nameof(targetType));
```

In this unit test, `typeof(List<string>).IsAssignableFrom(IEnumerable)` evaluates to `true`.

VladislavAntonyuk (approved):

VladislavAntonyuk:
I don't remember if we use parameter in any Converter, but what do you think about this proposal: https://github.com/CommunityToolkit/Maui/discussions/205

bijington:
> I don't remember if we use parameter in any Converter, but what do you think about this proposal: #205

We do in the `DoubleToIntConverter`(https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui/Converters/DoubleToIntConverter.shared.cs) sadly this would probably still want object given how open we are for the accepted types.

It could work well for `EnumToBoolConverter`(https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui/Converters/EnumToBoolConverter.shared.cs)

bijington (approved):
Looks great thanks @brminnick

PR #338, opened 2022-03-24
Ensure `Snackbar` + `Toast` work in Modal Page

### LINKED ISSUE #334: [Bug] Snackbar isn't shown in a modal page
### Description

The snackbar does not show when invoked from within a modal page.

### Steps to Reproduce

1. Create a new Maui project
2. Add a new page `Page2`
3. From within the main page add a button and in the click handler: `Navigation.PushModalAsync(new Page2());`
4. In `Page2` add a button and in the click handler: `await Snackbar.Make("test").Show();`

### Expected Behavior

The snackbar should show.

### Actual Behavior

The snackbar does not show.

### Basic Information

- Version with issue: 1.0.0-pre8
  - Android: 12.1

### LINKED ISSUE #344: [Bug] iOS SnackBar Message Doesn't Have Action Button
### Description
I've attached a sample project that demonstrates the issue.
The snackbar message doesn't show the action button in iOS. The action button for the snackbar does
show in Android.
As an additional note changing font size doesn't seem to affect this.

### Steps to Reproduce
In iOS
1. Run sample project.
2. Click "Open Snack Bar Message"
3. The snackbar message appears without an action OK button.

### Expected Behavior
The snackbar message should have an action button in iOS.

### Actual Behavior
The snackbar message does not have an action button in iOS.

### Basic Information
- Version with issue: CommunityToolkit.Maui 1.0.0-pre8, Maui 14
[MauiFlyout.zip](https://github.com/CommunityToolkit/Maui/files/8383946/MauiFlyout.zip)

VladislavAntonyuk:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #334 
 - Fixes #344 

 ### PR Checklist ###
 
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pulls/XYZ 


src/CommunityToolkit.Maui/Alerts/Toast/Toast.android.cs
```diff
@@ -25,7 +25,7 @@ public partial class Toast
 		var styledText = new SpannableStringBuilder(Text);
 		styledText.SetSpan(new AbsoluteSizeSpan((int)TextSize, true), 0, Text.Length, 0);
 
-		NativeToast = Android.Widget.Toast.MakeText(Platform.AppContext, styledText, GetToastLength(Duration))
+		NativeToast = Android.Widget.Toast.MakeText(Platform.CurrentActivity.Window?.DecorView.FindViewById(Android.Resource.Id.Content)?.RootView?.Context, styledText, GetToastLength(Duration))
```
pictos, line 28: Why can't we use the `AppContext`?

Can we avoid the `Native` word? Maybe we can use `PlatformToast`

brminnick (approved):
Thanks Vlad!!

✅ Confirmed Snackbar appears on modal pages on iOS, Android + macOS
✅ Confirmed Toast appears on modal pages on iOS, Android + macOS

brminnick:
Dang - looks like the Unit Tests are complaining about `MainThread` not being implemented on `net6.0`.

You'll probably just have to create a mock for `IMainThread`. I had to do something similar for my app, GitTrends: https://github.com/brminnick/GitTrends/blob/main/GitTrends.UnitTests/MockServices/MockMainThread.cs

src/CommunityToolkit.Maui/Alerts/Toast/Toast.android.cs
VladislavAntonyuk: Because we need a context of the RootView, while the AppContext is Application context

VladislavAntonyuk:
> 

IMainThread doesn't exist. MainThread is a static class

pictos (commented):
@VladislavAntonyuk thanks for renaming everything <3.
For me, this is ready to merge, but I'll wait for @brminnick approval since there are issues with the unit test.

I know that the MAUI team is refactoring the essentials to use interfaces, we can wait on that... So, should we merge this as it's and create a ticket to improve the unittest?

VladislavAntonyuk:
@pictos I try to replace MainThread with Dispatcher.
Do we need MainThread.IsMainThread if we have Dispatcher.GetForCurrentThread()? If I am not mistaken Dispatcher ensures that we run on UI thread

VladislavAntonyuk:
We have some natives in popup, but it is for a separate PR

pictos:
@VladislavAntonyuk 

> @pictos I try to replace MainThread with Dispatcher. Do we need MainThread.IsMainThread if we have Dispatcher.GetForCurrentThread()? If I am not mistaken Dispatcher ensures that we run on UI thread

So, the `Dispacher` should work, since it's in a `shared` level and will take care of platform specifics. But I would suggest to keep the Essentials APIs, just because MAUI will rely on this API.



> We have some natives in popup, but it is for a separate PR

Yeap, I'll take a look at those when I've time. Rush week for me

VladislavAntonyuk:
MainThread will be moved to Compatibility with the next release, so we can replace it with Dispatcher: https://github.com/dotnet/maui/pull/5407#issuecomment-1076965497

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
```diff
@@ -150,27 +150,7 @@ protected virtual ValueTask DisposeAsyncCore()
...
-		set
-		{
-			if (!MainThread.IsMainThread)
-			{
-				throw new InvalidOperationException($"{nameof(nativeSnackbar)} can only be called from the Main Thread");
-			}
-
-			nativeSnackbar = value;
-		}
-	}
```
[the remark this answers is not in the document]
brminnick: This code doesn't run on `net6.0`, only on `net6.0-ios`, `net6.0-android`, `net6.0-maccatalyst`, and `net6.0-windows`, so it shouldn't impact our Unit Tests.

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
```diff
@@ -102,46 +102,27 @@ static TimeSpan GetDuration(ToastDuration duration)
...
-		set
-		{
-			if (!MainThread.IsMainThread)
-			{
-				throw new InvalidOperationException($"{nameof(nativeToast)} can only be called from the Main Thread");
-			}
-
-			nativeToast = value;
-		}
-	}
```
brminnick, line 124: Can we keep this, replacing `MainThread` with `Dispatcher`? 

We added this after getting random crashes in the sample app that contained cryptic error messages. 

The crashes occurred because `Snackbar` was being modified from a non-UI thread, which makes sense. However, the StackTrace / Error Message provided from OS was very vague and unclear about _why_ the crash happened. 

Long story short, we added this `MainThread.IsMainThread` logic so that when future devs try to consume the `Snackbar` from a background thread will get a clear + concise error message that tells them exactly why it crashed, rather than some confusing OS exception like `SIGSEV blahblahblah`.

brminnick: This code doesn't run on `net6.0`, only on `net6.0-ios`, `net6.0-android`, `net6.0-maccatalyst`, and `net6.0-windows`, so it shouldn't impact our Unit Tests.

brminnick (changes requested):
Thanks Vlad!!

I'd like to keep the `IsMainThread` check (moving it to use `Dispatcher`) if we can! I added more details on my comment 👍

samples/CommunityToolkit.Maui.Sample/Platforms/Windows/App.xaml.cs
```diff
@@ -20,4 +18,22 @@ public App()
...
+
+		if (!createdNew)
+		{
+			Process.GetCurrentProcess().Kill();
+		}
+		else
+		{
+			base.OnLaunched(args);
+		}
+	}
```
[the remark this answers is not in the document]
VladislavAntonyuk: It's for the Snackbar action to make sure we have only 1 application instance. (Before it opened a new window).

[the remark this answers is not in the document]
VladislavAntonyuk: We need static field mutex, otherwise it's disposed and cleaned by GC

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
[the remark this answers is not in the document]
VladislavAntonyuk: MainThread as I left a comment is going to be removed. Also according to implementation Dispatcher checks if it needs to be dispatched and run the code on platform MainThread.
Because our methods Show, Dismiss, Dispose use Dispatcher (and it's the only one way to show Snackbar/Toast) from my point of view this check is redundant.

samples/CommunityToolkit.Maui.Sample/Platforms/Windows/App.xaml.cs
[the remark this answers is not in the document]
pictos: But we create a new mutex when this method is called and after that we just use the out parameter, so do you know why gc collecting the mutex object is bad?

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs
```diff
@@ -51,27 +52,27 @@ public partial class Snackbar
...
 		OnShown();
 	}
 
 	void OnActivated(ToastNotification sender, object args)
 	{
-		if (NativeSnackbar is not null && Action is not null)
+		if (PlatformSnackbar is not null && Action is not null)
 		{
-			MainThread.BeginInvokeOnMainThread(Action);
+			Dispatcher.GetForCurrentThread().DispatchIfRequired(Action);
```
pictos, line 69: While ago we discussed forcing the code to execute on `MainThread` inside the lib, and how that isn't a good practice. One reason is that this can cause some bugs and then could be hard to identify. For example, here, the `Action` will not be executed right after you step from this method, it will be queued to be executed when the `MainThread` is free, so at the time that this `Action` execute the state of the application can be invalid and this can lead to some crash or odd behavior. 

From the usage of this, this `delegate handler` will be subscribed in the thread that calls `ShowPlatform`, and the **user** must call it from the `MainThread`. What we can do here to protect the user is adding a check to make sure that we're on the `MainThread` and throw a good exception message. More than that I would say that we're shooting ourselves in the foot.

VladislavAntonyuk: It doesn't work when you use MainThread class: image

samples/CommunityToolkit.Maui.Sample/Platforms/Windows/App.xaml.cs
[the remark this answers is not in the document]
VladislavAntonyuk: updated

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs
pictos: What I mentioned is that we shouldn't use either `Dispatcher` or `MainThread`.
we should just do `Action();` or `Action.Invoke()`

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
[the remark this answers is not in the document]
brminnick: Yup, you're right that it's totally redundant because `Show()` and `Dismiss()` always push to the UI thread. 

I added the redundancy on purpose for two reasons:
- If `Show()` and `Dismiss()` are ever modified to remove the Dispatcher, we will still validate that we're using the correct thread
- Since `Snackbar` is a partial class, if anything in `Snackbar.[platorm].cs` references `Snackbar`, this ensures `NativeSnackbar` is always referenced from the UI thread

All that said, this isn't really something I care enough about to hold up this PR! If you'd like to keep it out, that's cool. Just let me know and I'll approve this PR 👍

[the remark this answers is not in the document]
brminnick: Fair enough. We can refactor `Show()` and `Dismiss()` to remove the marshaling to the UI thread.

I strongly disagree that a library should provide code that throws a preventable exception, but I'm happy to be outvoted by Vlad, Pedro and Shaun 🙌

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs
VladislavAntonyuk, line 69: it is important to keep Dispatcher here:
image

brminnick (approved):
Thanks Vlad! Just FYI - now that we don't need to `await` anything for the `Dispose` workflow, I changed `IAlert : IAsyncDisposable` to `IAlert : IDisposable`

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
```diff
@@ -78,36 +79,47 @@ public double TextSize
...
+		{
+			return;
+		}
+
+		if (isDisposing)
+		{
+#if ANDROID
+			PlatformToast?.Dispose();
 #endif
+		}
```
VladislavAntonyuk, line 121: isDisposed=true

brminnick (approved):

src/CommunityToolkit.Maui.Core/Views/Alert/AlertView.macios.cs
```diff
@@ -8,7 +8,7 @@ namespace CommunityToolkit.Maui.Core.Views;
 /// <summary>
 /// <see cref="UIView"/> for <see cref="Alert"/>
 /// </summary>
-public class AlertView : UIView
+public sealed class AlertView : UIView
```
pictos, line 11: In `Maui/Forms` isn't common to have don't allow inheritance on controls, do you have a good motivation for that?

src/CommunityToolkit.Maui.Core/Views/Alert/AlertView.macios.cs
```diff
@@ -8,7 +8,7 @@ namespace CommunityToolkit.Maui.Core.Views;
...
 {
 	readonly List<UIView> children = Array.Empty<UIView>().ToList();
```
pictos, line 13:
```suggestion
	readonly List<UIView> children = Enumerable.Empty<UIView>().ToList();
```

This will be faster

```
|               Method |     Mean |    Error |   StdDev | Ratio |  Gen 0 | Allocated |
|--------------------- |---------:|---------:|---------:|------:|-------:|----------:|
|       UsingArrayEmpty | 17.36 ns | 0.164 ns | 0.145 ns |  1.00 | 0.0051 |      32 B |
| UsingEnumerableEmpty | 10.22 ns | 0.060 ns | 0.047 ns |  0.59 | 0.0051 |      32 B |
```

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.macios.cs
```diff
@@ -26,15 +26,15 @@ public partial class Snackbar
...
 	{
-		await DismissNative(token);
+		await DismissPlatform(token);
 		token.ThrowIfCancellationRequested();
 
 		var cornerRadius = GetCornerRadius(VisualOptions.CornerRadius);
 
-		var padding = GetMaximum(cornerRadius.X, cornerRadius.Y, cornerRadius.Width, cornerRadius.Height) + ToastView.DefaultPadding;
-		NativeSnackbar = new SnackbarView(Text,
+		var padding = GetMaximum(cornerRadius.X, cornerRadius.Y, cornerRadius.Width, cornerRadius.Height) + PlatformToast.DefaultPadding;
```
pictos, line 36: I see here that we use the `DefaultPadding`, but what happens if the user changes it?

pictos (commented):
@VladislavAntonyuk I have some questions and suggestions before merging this one.

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.macios.cs
VladislavAntonyuk: The user will get different padding. 
It is a public setting.

pictos: Default padding is a read-only and we don't set it inside the `PlatformToast`[ ctor](https://github.com/CommunityToolkit/Maui/blob/eb2606b34d56b96e8cb71eb0dd9f923ffa2d4e3a/src/CommunityToolkit.Maui.Core/Views/Toast/PlatformToast.macios.cs#L35-L50). So even if the user sets a padding on ctor, the DefaultPadding will be the same

VladislavAntonyuk: Ah, you are right. I need a cup of coffee. I thought it is in Defaults.shared.cs.
I don't remember why it is a public read only. It should be private constant. We need this property to look similar to Android Snackbar.

src/CommunityToolkit.Maui.Core/Views/Alert/AlertView.macios.cs
VladislavAntonyuk: Inspired by https://www.meziantou.net/performance-benefits-of-sealed-class.htm

pictos: Yeah, I'm very cautious with inheritance as well. But in this case, we should allow it and we're prepared for it.

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.macios.cs
VladislavAntonyuk: so to summarize, it is used as a default padding. so if the user sets the corner radius equal to 0, we still have some padding equal to 10. but it was not correct to set it here. Now set default padding internally

VladislavAntonyuk:
Added additional documentation PR: https://github.com/MicrosoftDocs/CommunityToolkit/pull/55

brminnick (approved):

pictos (approved):

PR #347, opened 2022-04-02
`ListToStringConverter`: Remove Unused Null Coalescing Operator + Update XML Documentation

brminnick:
### Description of Change ###

This PR updates `ListToStringConverter`, to ensure ``public string Separator { get; set; }` cannot be `null`. 

In the `ConvertFrom` logic, we can then remove the unused `??` and adding a comment explaining that the `object? parameter` takes precedence over `public string Separator { get; set; }`.

This PR also updates the XML Documentation for `public string Separator { get; set; }` to better help the user understand the order of precedence between `object? parameter` and `public string Separator { get; set; }`.

These updates are inspired by the documentation for `ListToStringConverter`: https://github.com/MicrosoftDocs/CommunityToolkit/blob/73bd78afad79aa07c9c8329119ca45b5688d54d4/docs/maui/converters/list-to-string-converter.md

### DIFF
--- a/src/CommunityToolkit.Maui/Converters/ListToStringConverter.shared.cs
+++ b/src/CommunityToolkit.Maui/Converters/ListToStringConverter.shared.cs
@@ -9,7 +9,7 @@ namespace CommunityToolkit.Maui.Converters;
 public class ListToStringConverter : BaseConverterOneWay<IEnumerable, string>
 {
 	/// <summary>
-	/// The separator value that separates each item in the collection
+	/// The value that separates each item in the collection
 	/// This value is superceded by the ConverterParameter, if provided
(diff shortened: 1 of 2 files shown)


src/CommunityToolkit.Maui/Converters/ListToStringConverter.shared.cs
```diff
@@ -25,7 +27,9 @@ public override string ConvertFrom(IEnumerable value, Type targetType, object? p
 	{
 		ArgumentNullException.ThrowIfNull(value);
 
-		if ((parameter ?? Separator ?? string.Empty) is not string separator)
```
brminnick, line 28: Since `public string Separator { get; set; }` is not nullable, we can remove the unused `??`

VladislavAntonyuk (approved):

PR #349, opened 2022-04-02
Add `Null` Validation to `Toast` and `Snackbar`

brminnick:
### Description of Change ###

This PR adds `null` validation checks for the following:
- `Snackbar.Text `
- `Snackbar.ActionButtonText `
- `Toast.Text `

This PR also added `ArgumentNullException` Unit Tests for these properties in `Snackbar_Tests` and `Toast_Tests`.

I added Unit Tests to verify both `Snackbar` and `Toast` default values.

I also added `[EditorBrowsable(EditorBrowsableState.Never)]` to `CommunityToolkit.Maui.Core.Defaults`

### DIFF
--- a/src/CommunityToolkit.Maui.Core/Primitives/Defaults.shared.cs
+++ b/src/CommunityToolkit.Maui.Core/Primitives/Defaults.shared.cs
@@ -5,6 +5,7 @@ namespace CommunityToolkit.Maui.Core;
 /// <summary>
 /// Default Values for CommunityToolkit.Maui
 /// </summary>
+[EditorBrowsable(EditorBrowsableState.Never)]
 public static class Defaults
 {
 	/// <summary>
--- a/src/CommunityToolkit.Maui.Core/Primitives/Defaults.shared.cs
+++ b/src/CommunityToolkit.Maui.Core/Primitives/Defaults.shared.cs
@@ -21,6 +21,12 @@ public static class Defaults
 	public const double CharacterSpacing = 0.0d;
 
 	/// <summary>
+	/// Default Character Spacing
+	/// </summary>
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public const string ActionButtonText = "OK";
+
+	/// <summary>
 	/// Default Text Color
 	/// </summary>
 	[EditorBrowsable(EditorBrowsableState.Never)]
--- a/src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
+++ b/src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
@@ -14,11 +14,10 @@ namespace CommunityToolkit.Maui.Alerts;
 /// <inheritdoc/>
 public partial class Snackbar : ISnackbar
 {
-	internal const string defaultActionButtonText = "OK";
 	static readonly WeakEventManager weakEventManager = new();
 
 	string text = string.Empty;
-	string actionButtonText = defaultActionButtonText;
+	string actionButtonText = Defaults.ActionButtonText;
 
 	/// <summary>
 	/// Initializes a new instance of <see cref="Snackbar"/>
@@ -85,7 +84,7 @@ public partial class Snackbar : ISnackbar
 	public static ISnackbar Make(
 		string message,
 		Action? action = null,
-		string actionButtonText = defaultActionButtonText,
+		string actionButtonText = Defaults.ActionButtonText,
 		TimeSpan? duration = null,
 		SnackbarOptions? visualOptions = null,
 		IView? anchor = null)
@@ -206,7 +205,7 @@ public partial class Snackbar : ISnackbar
 public static class SnackbarVisualElementExtension
 {
 	/// <summary>
-	/// Display snackbar with the anchor
+	/// Display snackbar anchored to <see cref="VisualElement"/>
 	/// </summary>
 	/// <param name="visualElement">Anchor element</param>
 	/// <param name="message">Text of the snackbar</param>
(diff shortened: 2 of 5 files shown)


src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -14,36 +14,46 @@ namespace CommunityToolkit.Maui.Alerts;
 /// <inheritdoc/>
 public partial class Snackbar : ISnackbar
 {
+	internal const string defaultActionButtonText = "OK";
```
VladislavAntonyuk, line 17: Maybe it should be in Defaults?

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
```diff
@@ -209,8 +222,14 @@ public static class SnackbarVisualElementExtension
...
 		string message,
 		Action? action = null,
-		string actionButtonText = "OK",
+		string actionButtonText = Snackbar.defaultActionButtonText,
 		TimeSpan? duration = null,
 		SnackbarOptions? visualOptions = null,
-		CancellationToken token = default) => Snackbar.Make(message, action, actionButtonText, duration, visualOptions, visualElement).Show(token);
+		CancellationToken token = default)
+	{
+		ArgumentNullException.ThrowIfNull(message);
```
VladislavAntonyuk, line 230: Do we need it as we have a validation in property?

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
brminnick: Do we use `"OK"` as the default text in any other class?

I didn't add it to `Defaults` because I think it is only used in `Snackbar`.

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
brminnick: > Do we need it as we have a validation in property

No, but I added it here for another layer of protection. You know how I like to bake in redundancy in our codebase 😀 

If anything changes in the future, to `DisplaySnackbar`, `Snackbar.Make`, `Snackbar.Text` or `Snackbar.ActionButtonText` in the future, I want to ensure that have multiple points of null validation. 

This is also why I explicitly added null validation for `DisplaySnackbar` in the Unit Tests.
```cs
	[Fact]
	public void SnackbarNullValuesThrowArgumentNullException()
	{
#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
		Assert.Throws<ArgumentNullException>(() => new Snackbar { Text = null });
		Assert.Throws<ArgumentNullException>(() => new Snackbar { ActionButtonText = null });
		Assert.Throws<ArgumentNullException>(() => Snackbar.Make(null));
		Assert.Throws<ArgumentNullException>(() => Snackbar.Make(string.Empty, actionButtonText: null));
		Assert.ThrowsAsync<ArgumentNullException>(() => new Button().DisplaySnackbar(null));
		Assert.ThrowsAsync<ArgumentNullException>(() => new Button().DisplaySnackbar(string.Empty, actionButtonText: null));
#pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
	}
```

brminnick: Long story short, we may get a PR from a community member in the future who isn't familiar with the complete architecture of `Snackbar`, and this'll make sure that they don't accidentally break anything.

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
VladislavAntonyuk: I think about Defaults as the way to have a global one time configuration for the app, so optional parameters can be skipped

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
VladislavAntonyuk: As the property always used by the class (more often than the static method), I believe this is redundant. Also you will still get an exception if you call this method.

And something change in Text property (e.g. it became nullable) then we do not need this check in the Make method as well.

from my point of view identical logic on different layers leads to complexity and possible bugs

brminnick: Fair enough! I'll keep the null-check on the property and remove it from the method 👍 

The unit tests I added will protect us from any future PR breaking things

src/CommunityToolkit.Maui/Alerts/Snackbar/SnackBar.shared.cs
brminnick: Added `Defaults.ActionButtonText` ✅

src/CommunityToolkit.Maui.Core/Primitives/Defaults.shared.cs
```diff
@@ -5,6 +5,7 @@ namespace CommunityToolkit.Maui.Core;
 /// <summary>
 /// Default Values for CommunityToolkit.Maui
 /// </summary>
+[EditorBrowsable(EditorBrowsableState.Never)]
```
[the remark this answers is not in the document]
brminnick: I assumed that we preferred to keep it `internal` since it's only used internally for default values in the toolkit. But since it lives in `CommunityToolkit.Maui.Core` and we needed to leverage it in `CommunityToolkit.Maui`, we had to make it `pubic`.

That's just a guess though.

VladislavAntonyuk (approved):

PR #350, opened 2022-04-03
Add `Popup.Color` Null Validation

brminnick:
### Description of Change ###

This PR ensures that `Popup.Color` cannot be set to `null`.

Because `Popup.Color` can be used as a bindable property, I implemented the `null` validation on `public static readonly BindableProperty ColorProperty` to ensure validation also happens if a `null` value is passed in via a binding.

I also added `ArgumentNullException.ThrowIfNull` to `WrapperControl` which is used by `Popup` on Windows.

### DIFF
--- a/src/CommunityToolkit.Maui/Views/WrapperControl.windows.cs
+++ b/src/CommunityToolkit.Maui/Views/WrapperControl.windows.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Maui.Platform;
+﻿using System;
+using Microsoft.Maui.Platform;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using WRect = Windows.Foundation.Rect;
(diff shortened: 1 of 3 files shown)


pictos:
With this we can remove some defensive to handle `null` values, like [this one](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.Core/Views/Popup/PopupExtensions.windows.cs#L19). Do you mind if I add these changes on this PR?

brminnick:
Good catch! Please do! 🙌

src/CommunityToolkit.Maui/Views/WrapperControl.windows.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Microsoft.Maui.Platform;
+﻿using System;
```
VladislavAntonyuk, line 1:
```suggestion
```

VladislavAntonyuk (approved):

VladislavAntonyuk (approved):

pictos (approved):
We can't remove the null checks, because on `IPopup` we set the Color as `Color?`, so other implementations can allow null if they want. The only way to remove the defensive code is by changing `Color?` to `Color`.

PR #358, opened 2022-04-13
Update Sample App to `CommunityToolkit.Maui.Markup` v1.0.0-pre9

brminnick:
### Description of Change ###

This PR updates the Sample App to use `CommunityToolkit.Maui.Markup` v1.0.0-pre9.
 
It also adds `WeakEventManager` to `MauiPopup.PopoverDelegate.PopoverDismissedEvent` now that `WeakEventManager` lives in the `Microsoft.Maui` namespace.

(`WeakEventManager ` was previously in the `Microsoft.Maui.Controls` namespace).

### DIFF
--- a/samples/CommunityToolkit.Maui.Sample/Pages/Base/BaseGalleryPage.cs
+++ b/samples/CommunityToolkit.Maui.Sample/Pages/Base/BaseGalleryPage.cs
@@ -40,12 +40,10 @@ public abstract class BaseGalleryPage<TViewModel> : BasePage<TViewModel> where T
 		var collectionView = (CollectionView)sender;
 		collectionView.SelectedItem = null;
 
-		if (e.CurrentSelection.FirstOrDefault() is not SectionModel sectionModel)
+		if (e.CurrentSelection.FirstOrDefault() is SectionModel sectionModel)
 		{
-			return;
+			await Shell.Current.GoToAsync(AppShell.GetPageRoute(sectionModel.ViewModelType));
 		}
-
-		await Shell.Current.GoToAsync(AppShell.GetPageRoute(sectionModel.ViewModelType));
 	}
 
 	class GalleryDataTemplate : DataTemplate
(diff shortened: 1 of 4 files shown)


samples/CommunityToolkit.Maui.Sample/Pages/Base/BaseGalleryPage.cs
```diff
@@ -40,12 +40,10 @@ async void HandleSelectionChanged(object? sender, SelectionChangedEventArgs e)
 		var collectionView = (CollectionView)sender;
 		collectionView.SelectedItem = null;
 
-		if (e.CurrentSelection.FirstOrDefault() is not SectionModel sectionModel)
+		if (e.CurrentSelection.FirstOrDefault() is SectionModel sectionModel)
 		{
-			return;
+			await Shell.Current.GoToAsync(AppShell.GetPageRoute(sectionModel.ViewModelType));
```
pictos, line 45: If I'm not wrong it's safe to add the ConfigureAwait here
```suggestion
			await Shell.Current.GoToAsync(AppShell.GetPageRoute(sectionModel.ViewModelType)).ConfigureAwait(false);
```

pictos (approved):
I've one suggestion but it doesn't invalidate this PR.

I disabled the Auto-merge to give a chance to think about it

samples/CommunityToolkit.Maui.Sample/Pages/Base/BaseGalleryPage.cs
brminnick: You're not wrong; adding `.ConfigureAwait(false)` here would add a performance improvement. 

As a personal rule-of-thumb when implementing an MVVM architecture, I try to avoid using `.ConfigureAwait(false)` in the View layer since the View code typically _does_ require the UI Thread. 

Since `.ConfigureAwait(false)` does not return to the UI thread, any code added afterwards in a future PR would not be on the main thread. 

In the Services + ViewModel layers, however, I _always_ use `.ConfigureAwait(false)` 💯 

Long story short, I try to avoid `.ConfigureAwait(false)` in the View layer to avoid future bugs where future code may need to be running on the UI Thread.

pictos: Got it, yeah I suggested it because this is the last async call in the method, so doesn't matter which thread will run. But I understand your point, in terms that this might lead developers - who use this example as a reference - to dragons

PR #363, opened 2022-04-15
fix xml docs for color extensions

bijington:
### Description of Change ###

 

Fixes mistakes in the xml docs discovered when writing up docs.


brminnick (approved):
Thanks Shaun!

PR #364, opened 2022-04-15
Rename ColorToBlackKeyConverter to ColorToPercentKeyConverter

bijington:
### Description of Change ###

 

Rename `ColorToBlackKeyConverter` to `ColorToPercentKeyConverter` to be more consistent with the other CMYK related converters


brminnick:
Should it be `ColorToPercentBlackKey` since the method runs the extension method with a similar name, `return value.GetPercentBlackKey();`?

This would match our other `ColorToPercent*Key` APIs. 

This is more of a question than a critique. I'm not much of an expert in color theory.

bijington:
> Should it be `ColorToPercentBlackKey` since the method runs the extension method with a similar name, `return value.GetPercentBlackKey();`?
> 
> This would match our other `ColorToPercent*Key` APIs. 
> 
> This is more of a question than a critique. I'm not much of an expert in color theory. 

Yeah I think that is better as it provides a better level of consistency 👍

brminnick (approved):
Thanks Shaun!

PR #404, opened 2022-05-14
Fixed measurement on Android

### LINKED ISSUE #388: [Bug] Popup height doesn't resize based on content
### Description

It appears that the popup has a fixed height regardless of the content that is placed inside it.
The yellow part indicates the space occupied by the popup; regardless of the content it will always have that fixed height.

image

### Steps to Reproduce

1. Create a popup;
2. Place something inside, in this case a VerticalStackLayout with some stacklayouts inside;
3. Run app on android;
4. Open popup.

### Expected Behavior
The expected behavior would be to occupy only the orange space that is a basic verticalStackLayout (and the other space for the stackLayout for buttons).

image

### Actual Behavior
The current behavior on the other hand does not dynamically resize the popup based on its content.

image

### Basic Information

- Version with issue: 1.0.0-rc2 (and 1.0.0-rc1 too)
- Last known good version: none
- IDE: VS2022 Preview
- Platform Target Frameworks: 
  - Android: 11.0

### Workaround
None.

pictos:
### Description of Change ###

Added more "tries" in order to get the correct size.

 ### Linked Issues ###
 

 - Fixes #388 

 ### PR Checklist ###
 

bijington:
@pictos Do you think this measurement behaviour is something we should include in the docs? Does it vary per platform?

pictos:
@bijington the idea is to have the same behavior at all platforms, but the way to calculate it could be different

bijington:
> @bijington the idea is to have the same behavior at all platforms, but the way to calculate it could be different 

Ok great if the behaviour is consistent.

bijington:
Are there any rules around when it will resize and it won't?

pictos:
> Are there any rules around when it will resize and it won't?

AFAIK just when the platform can't give us the correct size when we call the `Measure` method on the `Popup.Content`. TBH, I'm not an expert on that subject, I didn't implement it this way in the first place because I didn't know how that works. Now it feels right.

bijington:
> > Are there any rules around when it will resize and it won't?
> 
> AFAIK just when the platform can't give us the correct size when we call the `Measure` method on the `Popup.Content`. TBH, I'm not an expert on that subject, I didn't implement it this way in the first place because I didn't know how that works. Now it feels right.

OK great, I don't think we need any extra docs updates then

src/CommunityToolkit.Maui.Core/Views/Popup/PopupExtensions.android.cs
```diff
@@ -186,6 +180,38 @@ public static void SetSize(this Dialog dialog, in IPopup popup, in AView contain
...
+		static void CalculateSizes(IPopup popup, Context context, ref int realWidth, ref int realHeight, ref int realContentWidth, ref int realContentHeight)
+		{
+			ArgumentNullException.ThrowIfNull(popup.Content);
+
+			if (!popup.Size.IsZero)
+			{
+				realWidth = (int)context.ToPixels(popup.Size.Width);
+				realHeight = (int)context.ToPixels(popup.Size.Height);
+			}
+			if (double.IsNaN(popup.Content.Width) && double.IsNaN(popup.Content.Height))
```
bijington, line 194: Is there ever a chance that one of these could be `NaN` and not the other? I am not sure I would expect this scenario to happen but it just struck me as a possibility when reading this statement

bijington (approved):
Thanks @pictos . I have just added one comment that probably just helps improved my understanding rather than require any change 😄

src/CommunityToolkit.Maui.Core/Views/Popup/PopupExtensions.android.cs
pictos: Good catch, probably an `||` will be safer here. I'll change it asap

VladislavAntonyuk (approved):

PR #412, opened 2022-05-18
Fix #410 - [Bug] Unable to use MultiConverter with non matching targetTypes

### LINKED ISSUE #410: [Bug] Unable to use MultiConverter with non matching targetTypes
### Description

It appears impossible to bind a `MultiConverter` containing at least one converter that does not match the target type of the property it is bound to. It may be easier to explain through an example using the sample page:

```csharp
<pages:BasePage 
    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
    xmlns:mct="http://schemas.microsoft.com/dotnet/2022/maui/toolkit"
    xmlns:pages="clr-namespace:CommunityToolkit.Maui.Sample.Pages"
    x:Class="CommunityToolkit.Maui.Sample.Pages.Converters.MultiConverterPage"
    xmlns:vm="clr-namespace:CommunityToolkit.Maui.Sample.ViewModels.Converters"
    x:TypeArguments="vm:MultiConverterViewModel"
    x:DataType="vm:MultiConverterViewModel">

    <pages:BasePage.Resources>
        <ResourceDictionary>
            <mct:MultiConverter x:Key="MyMultiConverter">
                <mct:TextCaseConverter />
                <mct:IsNotEqualConverter />
            </mct:MultiConverter>
            <x:Array x:Key="MultiParams"
                     Type="{x:Type mct:MultiConverterParameter}">
                <mct:MultiConverterParameter
                    ConverterType="{x:Type mct:TextCaseConverter}"
                    Value="{x:Static mct:TextCaseType.Upper}" />
                <mct:MultiConverterParameter
                    ConverterType="{x:Type mct:IsNotEqualConverter}"
                    Value="MAUI" />
            </x:Array>
        </ResourceDictionary>
    </pages:BasePage.Resources>
    <pages:BasePage.Content>
        <VerticalStackLayout Padding="20"
                             Spacing="20">
            <Label Text="This sample demonstrates how to use the MultiConverter with the IsNotEqualConverter and the TextCaseConverter. It converts the entered text to Upper Case and checks that it is Not Equal to the string 'MAUI'."/>
            <Entry Text="{Binding EnteredName}"/>
            <Label IsVisible="{Binding EnteredName, Converter={StaticResource MyMultiConverter}, ConverterParameter={StaticResource MultiParams}, Mode=OneWay}"
                   HorizontalOptions="CenterAndExpand"/>
        </VerticalStackLayout>
    </pages:BasePage.Content>
</pages:BasePage>
```

The second Label here has its `IsVisible` property bound to the result of the converters. This will throw an exception due to the check on this line: https://github.com/CommunityToolkit/Maui/blob/2af75d3e6eb522d3b17ffc7cd2d5fef005ca885a/src/CommunityToolkit.Maui/Extensions/ValueConverterExtension.shared.cs#L62

This is down to the fact that the first converter expects a `targetType` of `string`.

I don't know what the right answer is for a fix. I like the validation of the `targetType`s when dealing with single converters but when we are chaining we only care that the last converter has a valid `targetType`.

### Link to Reproduction Sample

This unit test reproduces the issue:

```sharp
[Fact]
public void ChainingConvertersOfDifferentTargetTypes()
{
    var multiConverter = new MultiConverter();
    
    multiConverter.Add(new TextCaseConverter());
    multiConverter.Add(new IsEqualConverter());

    var multiParams = new List<MultiConverterParameter>();
    
    multiParams.Add(new MultiConverterParameter { ConverterType = typeof(TextCaseConverter), Value = TextCaseType.Upper });
    multiParams.Add(new MultiConverterParameter { ConverterType = typeof(IsEqualConverter), Value = "MAUI" });
    
    var result = multiConverter.Convert("JOHN", typeof(bool), multiParams, CultureInfo.CurrentCulture);
    
    Assert.Equal(false, result);
}
```

### Steps to Reproduce

Use the example of unit test from above.

### Expected Behavior

No exception should be thrown

### Actual Behavior

An exception is thrown

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

### Workaround

Don't use it

brminnick:
### Description of Change ###

This PR adds two properties to `ICommunityToolkitValueConverter `:
- `Type FromType { get; }`
- `Type ToType { get; }`

These properties allow us to retrieve `TFrom` and `TTo` used by `BaseConverter<TFrom, To>` and `BaseConverter<TFrom, TTo, TParam>`. 

 ### Linked Issues ###
 

 - Fixes #410

 ### PR Checklist ###
 
#### `MultiConverter`
In `MultiConverter` I implemented both `ToType` and `FromType` explicitly to "hide" these properties since `MultiConverter` does not leverage these two properties.

#### Unit Tests
I've also added the previously failing Unit Test provided in #410.

### Additional Questions
@bijington - do we need to update the docs for `ICommunityToolkitValueConverter`? I don't see it currently documented.


src/CommunityToolkit.Maui/Converters/MultiConverter.shared.cs
```diff
@@ -5,6 +5,10 @@
 /// </summary>
 public class MultiConverter : List<ICommunityToolkitValueConverter>, ICommunityToolkitValueConverter
 {
+	Type ICommunityToolkitValueConverter.FromType => throw new NotSupportedException($"{nameof(ICommunityToolkitMultiValueConverter)} does not implement {nameof(ICommunityToolkitValueConverter.FromType)}");
+
+	Type ICommunityToolkitValueConverter.ToType => throw new NotSupportedException($"{nameof(ICommunityToolkitMultiValueConverter)} does not implement {nameof(ICommunityToolkitValueConverter.ToType)}");
+
```
bijington, line 11: What you mean we won't be able to nest `MultiConverters`???

Just kidding 😄

src/CommunityToolkit.Maui/Converters/ICommunityToolkitValueConverter.shared.cs
```diff
@@ -5,6 +5,16 @@ namespace CommunityToolkit.Maui.Converters;
 /// <inheritdoc />
 public interface ICommunityToolkitValueConverter : IValueConverter
 {
+	/// <summary>
+	/// Type Converting From
```
bijington, line 9: Perhaps something like

```suggestion
	/// Gets the <see cref="Type" /> this converter expects to <see cref="Convert" /> from or <see cref="ConvertBack" /> to.
```

src/CommunityToolkit.Maui/Converters/ICommunityToolkitValueConverter.shared.cs
```diff
@@ -5,6 +5,16 @@ namespace CommunityToolkit.Maui.Converters;
...
+	/// </summary>
+	Type FromType { get; }
```
bijington, line 11: I suspect this is here for symmetry but do we need it?

src/CommunityToolkit.Maui/Converters/ICommunityToolkitValueConverter.shared.cs
```diff
@@ -5,6 +5,16 @@ namespace CommunityToolkit.Maui.Converters;
...
+
+	/// <summary>
+	/// Type Converting To
```
bijington, line 14: Perhaps something like

```suggestion
	/// Gets the <see cref="Type" /> this converter expects to <see cref="Convert" /> to or <see cref="ConvertBack" /> from.
```

bijington (approved):
@brminnick I like the idea! I had a similar thought while struggling to get to sleep last night but I was worried it would be more breaking (I forgot we had the base classes to limit the damage).

Just added some minor XML doc improvements

bijington:
@brminnick re: docs... I am not sure to be honest, we could add a little detail to the `converters/index.md` file but I don't think it needs to be more than that

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui/Converters/ICommunityToolkitValueConverter.shared.cs
brminnick: Yup, exactly 💯.

It doesn't affect the bug in #410, but I figured it'd be good to add it for symmetry.

PR #426, opened 2022-05-25
SelectAllTextBehavior

### LINKED ISSUE #82: [Proposal] SelectAllTextEffect
# SelectAllTextEffect

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [ ] Unit Tests: Not Started
* [x] Sample

## Summary
[summary]: #summary

Highlights all text when an InputView is selected

## Detailed Design
[design]: #detailed-design

### SelectAllTextEffect.shared.cs

```cs
public class SelectAllTextEffect : RoutingEffect
{
}
```

### SelectAllTextEffect.android.cs

```cs
public class SelectAllTextEffect : PlatformEffect
{
  EditText EditText => (EditText)Control;
  
  protected override void OnAttached() => EditText?.SetSelectAllOnFocus(true);
  
  protected override void OnDetached() => EditText?.SetSelectAllOnFocus(false);
}
```

### SelectAllTextEffect.ios.cs

```cs
public class SelectAllTextEffect : PlatformEffect
{
  protected override void OnAttached() => ApplyEffect(true);
  
  protected override void OnDetached() => ApplyEffect(false);
  
  void ApplyEffect(bool apply) => ApplyToControl(Control, apply);

  bool ApplyToControl<T>(T controlType, bool apply) => controlType switch
  {
	UITextField textField => ApplyToUITextField(textField, apply),
	UITextView => ApplyToUITextView(apply),
	_ => throw new NotSupportedException($"Control of type: {controlType?.GetType()?.Name} is not supported by this effect.")
  };
}
```

## Usage Syntax
[usage]: #usage-syntax

<!-- Here is an example from CharactersValidationBehavior:

### XAML Usage

```xml
<Entry
    <Entry.Effects>
        <local: SelectAllTextEffect/>
    </Entry.Effects>
</Entry>

```

### C# Usage

```cs
var entry = new Entry();
entry.Effects.Add(new SelectAllTextEffect());
```

pictos:
### Description of Change ###

Since MAUI is GA 🎊 I presume that our code freeze is ending. With that here's the first `Effect` converted to `Behavior` using the new `Behavior API`.
 ### Linked Issues ###
 

 - Fixes #82

 ### PR Checklist ###
 
I'm looking at how to implement the unit test since I need some platform code to validate if the text is selected. Any idea is welcome.

### DIFF
--- a/samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/SelectAllTextBehaviorPage.xaml
+++ b/samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/SelectAllTextBehaviorPage.xaml
@@ -13,7 +13,7 @@
 	<ScrollView>
 		<VerticalStackLayout
 			Padding="10,10"
-			BackgroundColor="{StaticResource AppBackgroundColor}"
+			BackgroundColor="{StaticResource PrimaryColor}"
 			HorizontalOptions="Fill"
 			Spacing="20"
 			VerticalOptions="Fill">
--- a/samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/SelectAllTextBehaviorPage.xaml
+++ b/samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/SelectAllTextBehaviorPage.xaml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8" ?>
+﻿<?xml version="1.0" encoding="utf-8" ?>
 <pages:BasePage
 	x:Class="CommunityToolkit.Maui.Sample.Pages.Behaviors.SelectAllTextBehaviorPage"
 	xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
@@ -10,10 +10,9 @@
 	x:DataType="vm:SelectAllTextBehaviorViewModel"
 	x:TypeArguments="vm:SelectAllTextBehaviorViewModel">
 
-	<ScrollView>
+	<ScrollView BackgroundColor="{StaticResource PrimaryColor}">
 		<VerticalStackLayout
 			Padding="10,10"
-			BackgroundColor="{StaticResource PrimaryColor}"
 			HorizontalOptions="Fill"
 			Spacing="20"
 			VerticalOptions="Fill">
@@ -29,6 +28,7 @@
 			<Label FontAttributes="Italic" Text="Entry without the effect, when focussed no text will be selected" />
 			<Frame BackgroundColor="{StaticResource SoftFrameBackgroundColor}" HasShadow="False">
 				<Entry
+					BackgroundColor="{StaticResource PrimaryColor}"
 					PlaceholderColor="{StaticResource DarkLabelPlaceholderColor}"
 					Text="https://github.com/xamarin/XamarinCommunityToolkit"
 					TextColor="{StaticResource DarkLabelTextColor}" />
@@ -39,6 +39,7 @@
 			<Label FontAttributes="Italic" Text="Entry with the effect, when focussed all text will be selected" />
 			<Frame BackgroundColor="{StaticResource SoftFrameBackgroundColor}" HasShadow="False">
 				<Entry
+					BackgroundColor="{StaticResource PrimaryColor}"
 					PlaceholderColor="{StaticResource DarkLabelPlaceholderColor}"
 					Text="https://github.com/xamarin/XamarinCommunityToolkit"
 					TextColor="{StaticResource DarkLabelTextColor}">
@@ -56,7 +57,9 @@
 
 			<Label FontAttributes="Italic" Text="Editor without the effect, when focussed no text will be selected" />
 			<Frame BackgroundColor="{StaticResource SoftFrameBackgroundColor}" HasShadow="False">
-				<Editor PlaceholderColor="{StaticResource DarkLabelPlaceholderColor}" TextColor="{StaticResource DarkLabelTextColor}">
+				<Editor PlaceholderColor="{StaticResource DarkLabelPlaceholderColor}"
+						TextColor="{StaticResource DarkLabelTextColor}"
+						BackgroundColor="{StaticResource PrimaryColor}">
 					<Editor.Text>
 						Yeah, but your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should. Must go faster... go, go, go, go, go! You know what? It is beets. I've crashed into a beet truck. Is this my espresso machine? Wh-what is-h-how did you get my espresso machine?
 						Just my luck, no ice. I gave it a cold? I gave it a virus. A computer virus. Must go faster... go, go, go, go, go! God creates dinosaurs. God destroys dinosaurs. God creates Man. Man destroys God. Man creates Dinosaurs. Yeah, but your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should.
@@ -68,7 +71,9 @@
 
 			<Label FontAttributes="Italic" Text="Editor with the effect, when focussed all text will be selected" />
 			<Frame BackgroundColor="{StaticResource SoftFrameBackgroundColor}" HasShadow="False">
-				<Editor PlaceholderColor="{StaticResource DarkLabelPlaceholderColor}" TextColor="{StaticResource DarkLabelTextColor}">
+				<Editor PlaceholderColor="{StaticResource DarkLabelPlaceholderColor}"
+						TextColor="{StaticResource DarkLabelTextColor}"
+						BackgroundColor="{StaticResource PrimaryColor}">
 					<Editor.Behaviors>
 						<mct:SelectAllTextBehavior />
 					</Editor.Behaviors>
--- a/samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/SelectAllTextBehaviorPage.xaml
+++ b/samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/SelectAllTextBehaviorPage.xaml
@@ -1,10 +1,15 @@
 <?xml version="1.0" encoding="utf-8" ?>
-<ContentPage
-	x:Class="CommunityToolkit.Maui.Sample.SelectAllTextPage"
+<pages:BasePage
+	x:Class="CommunityToolkit.Maui.Sample.Pages.Behaviors.SelectAllTextBehaviorPage"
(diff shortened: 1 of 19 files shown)


src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.ios.cs
```diff
@@ -0,0 +1,97 @@
...
+
+		return true;
+	}
+
+	void OnEditingDidBegin(object? sender, EventArgs e)
+	{
+		if (sender is not UITextField textfield)
+		{
+			return;
+		}
```
brminnick, line 61: Should this throw an `ArgumentException` or an `InvalidOperationException`?


Is any other type subscribing to this event handler that isn't a `UITextField`? If not, let's throw an Exception.


Or, better, just cast it:

```suggestion
		var texfield = (UITextField)sender;
```

samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/SelectAllTextBehaviorPage.xaml
```diff
@@ -0,0 +1,85 @@
...
+							<mct:SelectAllTextBehavior />
+						</Editor.Behaviors>
+						<Editor.Text>
+							Yeah, but your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should. Must go faster... go, go, go, go, go! You know what? It is beets. I've crashed into a beet truck. Is this my espresso machine? Wh-what is-h-how did you get my espresso machine?
+							Just my luck, no ice. I gave it a cold? I gave it a virus. A computer virus. Must go faster... go, go, go, go, go! God creates dinosaurs. God destroys dinosaurs. God creates Man. Man destroys God. Man creates Dinosaurs. Yeah, but your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should.
+						</Editor.Text>
+					</Editor>
+				</Frame>
+			</VerticalStackLayout>
+		</VerticalStackLayout>
```
brminnick, line 83: Let's avoid nesting multiple VerticalStackLayouts inside of a VerticalStackLayout. 

Let's either use a `Grid`, or a `FlexLayout`, or edit each control's `Padding` + `Margin` + `Spacing` to fine-tune the groupings. 

https://docs.microsoft.com/en-us/xamarin/xamarin-forms/deploy-test/performance#choose-the-correct-layout
> ... don't attempt to reproduce the appearance of a specific layout by using combinations of other layouts, as this results in unnecessary layout calculations being performed. For example, don't attempt to reproduce a Grid layout by using a combination of StackLayout instances.

brminnick (commented):
Thanks Pedro! Exciting to start working on v1.1.0 now!!

Here's a couple comments after a quick glance at the code. I'll spin it up on my machine and play around with it tomorrow 🙌

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.ios.cs
```diff
@@ -0,0 +1,97 @@
...
+			_ => throw new NotSupportedException($"Control of type: {platformView?.GetType()?.Name} is not supported by this effect.")
+		};
+	}
+
+	bool ApplyToUITextField(UITextField textField, bool apply)
+	{
+		if (textField is null)
+		{
+			return false;
+		}
```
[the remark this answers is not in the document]
pictos: From XCT we didn't throw the exception, so not sure if we should change the behavior or not

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.ios.cs
pictos: same as before, I just maintained the previous behavior, as you can see [here](https://github.com/xamarin/XamarinCommunityToolkit/blob/22b1d4ccf45555c2ed89ecb94e9ff93a5cf61ab3/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.ios.cs#L73-L74). Do you think that we should throw an exception?

If I'm not wrong, throwing an exception at this level could crash the app, not sure if the behavior is too important to the point of being able to break the application

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.ios.cs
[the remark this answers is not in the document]
brminnick: We don't need to adhere to the `Xamarin.CommunityToolkit` implementation. In fact, I highly recommend we don't for every feature we add to `CommunityToolkit.Maui`!

I always advocate for throwing Exceptions over printing an error message to the console for two reasons:
1. They immediately help the developer locate the problem in their code
2. If the method fails without throwing an exception, the developer will experience an unexpected behavior from our library 
    * This leads to them wasting time debugging it, looking at our source code, etc. trying to figure out the problem
    * The developer may open an Issue on our repo, leading to wasted time for us explaining the expected implementation and behavior to them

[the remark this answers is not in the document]
pictos: @brminnick  I'm fine with that, I'll update the code to throw exceptions

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.ios.cs
pictos: Resolved in another discussion

brminnick (changes requested):
Hey Pedro! 

I found a bug when selecting the **Editor** text on macOS; when I tap on the Editor, it does not select all of the text. 

It works fine on iOS though, which is odd.

Here's a screen recording demoing iOS (right) and macOS (left):

ScreenFlow

pictos:
Thanks, @brminnick, maybe it's some API detail. I'll set up my mac environment and try it here on my end.

pictos:
@brminnick I ran the sample using the VSCode, and I would like to confirm that the environment is not deceiving me. Do you mind running our sample and putting a breakpoint [here](https://github.com/CommunityToolkit/Maui/blob/8a6b34725d7e2bf5b6feda1030c45f7d98344794/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.ios.cs#L83) and [here](https://github.com/CommunityToolkit/Maui/blob/8a6b34725d7e2bf5b6feda1030c45f7d98344794/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.ios.cs#L95).

On my test the `Focused` event isn't trigger on Editor.

pictos:
@brminnick FYI I opened an issue report on `Xamarin-macios` repo.
- https://github.com/xamarin/xamarin-macios/issues/15156

brminnick:
Thanks @pictos! 

I just saw Rolf's response to https://github.com/xamarin/xamarin-macios/issues/15156.

What do you think? Should we add a work-around for macOS that modifies this default behavior?

pictos:
@brminnick I would say that we can just mark it as not supported on mac-catalyst and I'll do some investigation in the future. I didn't find anything on the internet about how to workaround this.

This one was supposed to be easy peasy 😄

pictos:
@brminnick the `NotSupportedOSAttribute` should be used on a `public` method, but we don't have it. I thought to add it inside the `SelectAllTextBehavior` class, but it works well for `Entry` control. Do you have an idea where we can add it?

brminnick:
Dang - good point. 

You're right - if we put the attribute on the class (like this), then users will assume that it doesn't work at all on macOS, which isn't accurate because it _does_ work for `Entry`:

```cs
[UnsupportedOSPlatform("maccatalyst")]
public class SelectAllTextBehavior : PlatformBehavior<InputView>
```

Let's do this:
1. Open an Issue in this repo documenting the behavior for Editor + macOS
2. Update the comment and the `Console.Writeline` to point to this known issue
3. Ship it 🚀

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.macios.cs
```diff
@@ -0,0 +1,109 @@
...
+
+		textfield.PerformSelector(new Selector("selectAll"), null, 0.0f);
+	}
+
+	// MacCatalyst support blocked: https://github.com/xamarin/xamarin-macios/issues/15156
+	bool ApplyToUITextView(bool shouldApply)
+	{
+		if (OperatingSystem.IsMacCatalyst())
+		{
+			Console.WriteLine("MacCatalyst doesn't support this feature on the editor. for more info see https://github.com/CommunityToolkit/Maui/issues/432");
```
brminnick, line 77:
```suggestion
			Console.WriteLine("WARNING: `SelectAllTextBehavior` does not support `Microsoft.Maui.Controls.Editor` on MacCatalyst. For more information, see https://github.com/CommunityToolkit/Maui/issues/432");
```

brminnick: Just a quick suggestion to include `SelectAllTextBehavior` in the message and add WARNING to help it stand out in the logs

brminnick (approved):

brminnick (changes requested):
Apologies for approving than requesting changes. 

I just noticed we're missing unit tests. Let's add those and submit a Docs PR before merging this PR 👍

pictos:
@brminnick I tried to create the unit test and it couldn't be done... Because the platformBehavior uses the native APIs, if you see the code there's no `.shared` file, so it doesn't target .NET6. I'll move the device_test forward to make sure that we can test everything.



image

PR #434, opened 2022-06-02
Handle Generic Implementations of `ITextStyle` by TextColorToGenerator 

### LINKED ISSUE #409: [Bug] TextColorToGenerator produces error for generic views
### Description

After adding `CommunityToolkit.Maui` to a project with generic view I'm no longer able to build it with following error

### Stack Trace

/Users/pa/GenericViewBug/CommunityToolkit.Maui.SourceGenerators/CommunityToolkit.Maui.SourceGenerators.Generators.TextColorToGenerator/EnumPickerTextColorTo_e8fe61f7-6659-4d2d-8617-dff3e09df8e7.g.shared.cs(23,62): error CS0305: Using the generic type 'EnumPicker<T>' requires 1 type arguments [/Users/pa/GenericViewBug/GenericViewBug.csproj]

### Link to Reproduction Sample

[repro](https://github.com/anpin/GenericViewBug)
You can find generated files inside the  `Generated` folder

### Steps to Reproduce

1.  git clone https://github.com/anpin/GenericViewBug
2. cd GenericViewBug
3. dotnet build

### Expected Behavior
project builds
### Actual Behavior
error
### Basic Information

- Version with issue: 1.0.0-rc3
- Last known good version: - 
- SDK: 
```
dotnet --info
.NET SDK (reflecting any global.json):
 Version:   6.0.300
 Commit:    8473146e7d

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  12.3
 OS Platform: Darwin
 RID:         osx.12-x64
 Base Path:   /usr/local/share/dotnet/sdk/6.0.300/

Host (useful for support):
  Version: 6.0.5
  Commit:  70ae3df4a6

.NET SDKs installed:
  3.1.418 [/usr/local/share/dotnet/sdk]
  3.1.419 [/usr/local/share/dotnet/sdk]
  5.0.407 [/usr/local/share/dotnet/sdk]
  5.0.408 [/usr/local/share/dotnet/sdk]
  6.0.202 [/usr/local/share/dotnet/sdk]
  6.0.300 [/usr/local/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.24 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.25 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.16 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.17 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.4 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.5 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.24 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.25 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.16 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.17 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.4 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.5 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
```
### Workaround
Remove package or disable the generator with 
```
<Target Name="DisableAnalyzers"
     BeforeTargets="CoreCompile">
        <ItemGroup>
            <Analyzer Remove="@(Analyzer)" Condition="'%(Filename)' == 'CommunityToolkit.Maui.SourceGenerators'" />
        </ItemGroup>
</Target>
```

anpin:
### Description of Change ###
- Added extensions to serialize generic type arguments 
- Using new extensions in TextColorToGenerator 
 ### Linked Issues ###
#409 

 - Fixes #409

 ### PR Checklist ###
 - [X] Has tests (if omitted, state reason in description)

### DIFF
--- a/src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
+++ b/src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
@@ -1,5 +1,7 @@
 ﻿using Microsoft.CodeAnalysis;
-
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Text;
 namespace CommunityToolkit.Maui.SourceGenerators.Extensions;
 
 static class NamespaceSymbolExtensions
@@ -25,5 +27,125 @@ static class NamespaceSymbolExtensions
 				}
 			}
 		}
+	}
+	
+	public static string GetFullTypeString(this INamedTypeSymbol type) => 
+		type.Name + 
+		(type.NullableAnnotation == NullableAnnotation.Annotated ? "?" : string.Empty) +
+		type.TypeArguments.GetGenericTypeArgumentsString();
+
+	public static string GetGenericTypeArgumentsString(this IEnumerable<ITypeSymbol> typeArguments)
+	{
+		string result = string.Empty;
+
+		if (!typeArguments.Any())
+		{
+			return result;
+		}
+
+		result += "<";
+
+		bool isFirstArgument = true;
+		foreach (ITypeSymbol typeArg in typeArguments)
+		{
+			if (isFirstArgument)
+			{
+				isFirstArgument = false;
+			}
+			else
+			{
+				result += ", ";
+			}
+
+			switch (typeArg)
+			{
+				case ITypeParameterSymbol typeParameterSymbol:
+					result += typeParameterSymbol.Name;
+					break;
+				case INamedTypeSymbol namedTypeSymbol:
+					result += namedTypeSymbol.GetFullTypeString();
+					break;
+				default:
+					//Maybe should handle this somehow, maybe inject SourceProductionContext and call ReportDiagnostic on it?
+					break;
+			}
+
+		}
+
+		result += ">";
+
+		return result;
+	}
+
+	public static string GetWhereStatement(this INamedTypeSymbol type)
+	{
+		string result = string.Empty;
+		if (!type.TypeParameters.Any())
+		{
+			return result;
+		}
+
+		foreach (var typeParameterSymbol in type.TypeParameters)
+		{
+			string constraints = "";
+
+			bool isFirstConstraint = true;
+
+			if (typeParameterSymbol.HasNotNullConstraint)
+			{
+				constraints += "notnull";
+
+				isFirstConstraint = false;
+			}
+			else if (typeParameterSymbol.HasUnmanagedTypeConstraint)
+			{
+				constraints += "unmanaged";
+
+				isFirstConstraint = false;
+			}
+			else if (typeParameterSymbol.HasReferenceTypeConstraint)
+			{
+				constraints += "class";
+
+				isFirstConstraint = false;
+			}
+			else if (typeParameterSymbol.HasValueTypeConstraint)
+			{
+				constraints += "struct";
+
+				isFirstConstraint = false;
+			}
+
+			foreach (INamedTypeSymbol contstraintType in typeParameterSymbol.ConstraintTypes)
+			{
+				if (!isFirstConstraint)
+				{
+					constraints += ", ";
+				}
+				else
+				{
+					isFirstConstraint = false;
+				}
+
+				constraints += contstraintType.GetFullTypeString();
+			}
+
+			if (typeParameterSymbol.HasConstructorConstraint)
+			{
+				if (!isFirstConstraint)
+				{
+					constraints += ", ";
+				}
+				constraints += "new()";
+			}
+
+			if (!string.IsNullOrEmpty(constraints))
+			{
+				result += "where " + typeParameterSymbol.Name + " : " + constraints + @"
+";
+			}
+		}
+
+		return result;
 	}
 }
--- a/src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
+++ b/src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.CodeAnalysis;
+
 namespace CommunityToolkit.Maui.SourceGenerators.Extensions;
 
 static class NamespaceSymbolExtensions
@@ -103,6 +104,11 @@ static class NamespaceSymbolExtensions
 			else if (typeParameterSymbol.HasReferenceTypeConstraint)
 			{
 				constraints += "class";
+
+				if (typeParameterSymbol.ReferenceTypeConstraintNullableAnnotation == NullableAnnotation.Annotated)
+				{
+					constraints += "?";
+				}
 
 				isFirstConstraint = false;
 			}
--- a/src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
+++ b/src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
@@ -1,7 +1,4 @@
 ﻿using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.CSharp;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
-using Microsoft.CodeAnalysis.Text;
 namespace CommunityToolkit.Maui.SourceGenerators.Extensions;
 
 static class NamespaceSymbolExtensions
--- a/src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
+++ b/src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.CodeAnalysis;
+﻿using System.Text;
+using Microsoft.CodeAnalysis;
 
 namespace CommunityToolkit.Maui.SourceGenerators.Extensions;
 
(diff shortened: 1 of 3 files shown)


src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
```diff
@@ -26,4 +26,129 @@ public static IEnumerable<INamedTypeSymbol> GetNamedTypeSymbols(this INamespaceS
...
+	{
+		string result = string.Empty;
+		if (!type.TypeParameters.Any())
+		{
+			return result;
+		}
+
+		foreach (var typeParameterSymbol in type.TypeParameters)
+		{
+			string constraints = "";
```
pictos, line 88: Can you please use a StringBuilder instead?

src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
```diff
@@ -26,4 +26,129 @@ public static IEnumerable<INamedTypeSymbol> GetNamedTypeSymbols(this INamespaceS
...
\ No newline at end of file
+	
+	public static string GetFullTypeString(this INamedTypeSymbol type) => 
+		type.Name + 
+		(type.NullableAnnotation == NullableAnnotation.Annotated ? "?" : string.Empty) +
+		type.TypeArguments.GetGenericTypeArgumentsString();
+
+	public static string GetGenericTypeArgumentsString(this IEnumerable<ITypeSymbol> typeArguments)
+	{
+		string result = string.Empty;
```
pictos, line 37: Can you please use a StringBuilder instead?

src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
```diff
@@ -26,4 +26,129 @@ public static IEnumerable<INamedTypeSymbol> GetNamedTypeSymbols(this INamespaceS
...
+			switch (typeArg)
+			{
+				case ITypeParameterSymbol typeParameterSymbol:
+					result += typeParameterSymbol.Name;
+					break;
+				case INamedTypeSymbol namedTypeSymbol:
+					result += namedTypeSymbol.GetFullTypeString();
+					break;
+				default:
+					//Maybe should handle this somehow, maybe inject SourceProductionContext and call ReportDiagnostic on it?
```
pictos, line 67: Going to `default` will cause what?

src/CommunityToolkit.Maui.UnitTests/Extensions/TextColorTo_Tests.cs
```diff
@@ -115,4 +143,37 @@ interface ICustomTextStyle : ITextStyle
...
+	public class ClassConstraint
+	{
+
+	}
+
+	public struct StructConstraint
+	{
+
+	}
+	public class GenericPicker<TA, TB, TC, TD, TE, TF, TG, TH, TI, TJ, TK> : Picker
```
pictos, line 164: Will this work with records? Can you add them to the test case?

pictos (changes requested):
@anpin thanks for this PR, I've some suggestions and also a question.

AFAIK, .NET MAUI Handlers don't work well with generic, that's the main reason why PopupView isn't generic like it's in XCT. I saw your sample on the GitHub (that you linked on your issue) and there's no real application there.

Also, if you do something like `MyPicker<T> : Picker`, the extension will work and will use the generated methods for the `Picker` control. I did a [test here](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYZQAs6AsgBQCU+hxTjTRAbgIYBO6AHugF50AOwCmAd3QAhADYcAPFDgAGAHx0A3C1a8AdABVRAZwAum7QF9kVpNu1pMMaXPn7V6EOgAKASwDGANaiPOIAFsGi6PoejtoESEw2zAnEDli+gcFxlsjISgBsjugAorzI8UxpcIXkUcZmJqE+Rt7+QTwm9bTZKUQ2FkA===) to make sure that it will compile

With that, do you have a real case scenario?

src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
anpin: yes, this makes total sense

src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
anpin: I suppose other derivatives of `ITypeSymbol` must be handled here, but that is my first experience with roslyn and I'm not sure if they can appear in this loop, maybe if one will define `TArrayOfTypes[]` as a generic type parameter or in some other case

src/CommunityToolkit.Maui.UnitTests/Extensions/TextColorTo_Tests.cs
anpin: @pictos yes I do have a case scenario. I'm porting existing solution from `Xamarin.Forms` to `MAUI` and it has `public class EnumPicker<T> : Picker where T : Enum {}` where body of the class populates `ItemsSource` via all possible `Enum` values and takes displayed names from `DisplayAttribute`. I wasn't really interested in this particular color extension, but rather wanted to drop support of custom `IAsyncCommand` in the project in favor of `IAsyncRelayCommand` from this library. After installing the nuget my project stopped building with error mentioned above until I implemented the changes.

pictos:
@anpin not sure if you can share the real code, if don't are you able to create a small POC showing that? Because if are inheriting from `Picker` the roslyn should be happy using the extension method for the `Picker`.
If someone needs to create a generic control that doesn't have a handler, the person will need to implement everything and will need a not generic type in order to make it work as expected with Handlers. If nothing changed since the last time that I did the PopupView

brminnick:
@pictos FYI - @anpin included a reproduction in the bug report: https://github.com/CommunityToolkit/Maui/issues/409

pictos:
@brminnick I saw that repo but isn't using the control... It's just showing that in C# like in the test case. So before merging this I would like to confirm that this scenario is valid on .NET MAUI, if not we don't need to care about it (?)

anpin:
@pictos yes it is totally viable use case, at least for me, I just omitted usage as it wasn't relevant, without the toolkit nuget MAUI project was working as expected, but with toolkit it started producing error on compile time

pictos:
@anpin I just had an insight about this! So let's say that you've `MyPicker<T> : Picker` the SG shouldn't generate the extension method for `MyPicker<T>` because it can be used the extension method generated by the base class `Picker`.
That said, instead of caring about generating more code, we can just check if the base type is one of the Controls that .NET MAUI implements and don't generate the extension method.

pictos:
> > we can just check if the base type is one of the Controls that .NET MAUI implements and don't generate the extension method.
> 
> Love it! That'll also help remove some compiler warnings about duplicate code that I've also seen some community members share.
> 
> I still agree that should merge this PR to add support for the following scenario too:
> 
> ```cs
> public class MyPicker<T> : ITextStyle
> ```

The problem is, as I mentioned on my review, the Handler doesn't play nicely with generics, so if have a class that inherit from the interface you have to create the Handler, and the `PlatformView` used to wire up everything can't be generic. I ran with issues on PopupView because of that (on XCT it's generic).

brminnick:
Ah! I understand now!!

anpin:
I've included use case to the [repro repo](https://github.com/anpin/GenericViewBug) and it works with this PR
image

but fails to compile with nuget `1.0.0`
image

src/CommunityToolkit.Maui.SourceGenerators/Extensions/NamespaceSymbolExtensions.cs
anpin: I couldn't identify any case where default would happen, if you can point me at possible condition let me know

pictos:
Hey @anpin, I truly believe that your PR fixes the issue. But that brings something else to light.
There's, do we need to generate code for all types? Like in your case, we don't need to generate an extension method, because we can use the extension method generated for the `Picker` control. That said, we shouldn't generate the extension method for any control that is inherent in the .NET MAUI Controls that implement` ITextStyle`.  Does it make sense for you?

So, if it's possible the fix will be for the SG to look at the `MyPicker<T> : Picker`  and see that it can use the method generated for `Picker` Control, instead of creating a new one. Does it make sense for you? If so, are you able to implement this?

anpin:
@pictos I agree that generating code for each type seems unnecessary. Shouldn't we implement extension for interface `ITextStyle` instead of a concrete type implementing it? Like I said I haven't considered the actual purpose of the extension generated, just wanted the compiler to pass :)

anpin:
@pictos oh I see now the `ITextStyle` doesn't have setter for color, but the element such as `Picker` does which is why we have extensions. It seems that it would be better if every class supporting setting of color would have implement something like `IUpdatableTextStyle`, so we can do a single extension covering all those classes and no need for SG at all. 
```
interface IUpdatableTextStyle : ITextStyle
{
    new Color TextColor { get; set; }
}
```

pictos:
@anpin So, some interfaces don't have the setter by design. There's something intentional, and the class that implements the interface can implement the setter. From .NET MAUI, just properties that can be set by platform-specific area have a setter, for example, the [Text property](https://github.com/dotnet/maui/blob/67311a359358452a50407aab32f193f026a29349/src/Core/src/Core/ITextInput.cs#L11) in the `ITextInput` (Entry, Editor), you can set the text in the shared layer (code-behind, XAML, ViewModel), but you also can set code thought the platform, using the keyboard as input.

In this case, we need to rely on the class implementation and we can't generate any new interface, because it doesn't live here.

It's time to sleep here, so I'll take some time to answer any further messages, if you want to play with SGs, you need to see if the control has a Maui Control as a base type, if it has we shouldn't generate the extension method. Otherwise, I can play with it and ping you to review the code and see if that work for your scenario. I can't promise that will fix it fast because I've a blocker bug to investigate 😅.

@anpin have a good weekend and thanks for all this effort I really appreciate it!

anpin:
realize that I was wrong stating that I needed this lib for `IAsyncRelayCommand` as it is in another `CommunityToolkit`, this one I needed for the popup implementation  🤦

pictos:
Closing this in favor of #453 

@anpin please test the generated nuget (after we fix the build)

brminnick:
Hey Pedro! I'm going to reopen this now that we've merged #453.

I still see the value in supporting generic types for instances such as this:

```cs
class GenericCustomPicker<T> : View, ITextStyle
{
}
```

brminnick (approved):
This looks good to me, but I'll defer to @pictos for the final review + merge.

I did confirm that the unit tests added in this PR fail when tested against the main branch.

pictos (approved):

brminnick:
Thanks @anpin!!!

anpin:
@brminnick thanks for considering the PR!

PR #444, opened 2022-06-08
Force string comparison usage

bijington:
### Description of Change ###

Taking inspiration from this blog post: https://devblogs.microsoft.com/dotnet/performance-improvements-in-dotnet-maui/#dont-use-culture-aware-string-comparisons and @brminnick here we include rules to make developers explicitly define their string comparison.

 

Some key points:

#### 1. `string.Equals` vs `==`

It appears that the use of `string.Equals` using these new rules suggests that we should always specify the `StringComparison` override however upon closer inspection of the runtime code both `string.Equals` (without `StringComparison`) and `==` appear to use the same underlying check which doesn't appear to rely on `CultureInfo`.

`string.Equals` (without `StringComparison`) and `==`

Screenshot 2022-06-08 at 13 01 00

`EqualsHelper`

Screenshot 2022-06-08 at 13 00 43

With this in mind are we happy to stick with either using `==` for string equality checks or move to `string.Equals(string, string, StringComparison`?

#### 2. `TextValidationBehavior`s

It appears we only violated these new rules in 3 places in main code and once in unit tests. I do suspect that we still want to utilise the `CurrentCulture` for string comparisons in the UI based validation. This leads me on to question whether we should allow the developer to define the `StringComparison` when using the behaviors and put the responsibility/power on them. Of course we can default to the more performant option and document why.

 ### Linked Issues ###
 

n/a

 ### PR Checklist ###
 

bijington:
> Thanks Shaun! Let's move this rule from the solution-wide `.editorconfig` to the nested `.editorconfig`s in the [`CommunityToolkit.Maui` folder](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui/.editorconfig) and in the [`CommunityToolkit.Maui.Core` folder](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.Core/.editorconfig) so that it doesn't impact Unit Testing.
> 
> For example, I can see a future Unit Test maybe needing to test the `CultureInfo.CurrentCulture`.

Good call! I completely missed that we had 2 `.editorconfig` files

brminnick (approved):
Thanks Shaun!

bijington:
@jfversluis @brminnick @pictos @VladislavAntonyuk just to confirm did any of you have any thoughts on whether we should be using `StringComparison.CurrentCulture` in the `RequiredStringValidationBehavior` or whether we should expand on this and allow the consumer to provide which `StringComparison` they wish to use?

brminnick:
I think it makes sense to use `CurrentCulture` to ensure comparisons across other languages + alphabets. 

For example, how would `graçias` compare to `gracias` with/without `CurrentCulture`?

But I'll defer to the multi-lingual maintainers in this one 👍

bijington:
> I think it makes sense to use `CurrentCulture` to ensure comparisons across other languages + alphabets. 
> 
> For example, how would `graçias` compare to `gracias` with/without `CurrentCulture`?
> 
> But I'll defer to the multi-lingual maintainers in this one 👍 

Good point! I should consider adding some unit tests in around this area!

pictos:
@bijington, can you add a unit test with currency and big numbers? I can remember some headache about converting those in localized applications

bijington:
> @bijington, can you add a unit test with currency and big numbers? I can remember some headache about converting those in localized applications

Those can be painful when converting between their data types and strings but this only compares strings so I think we should be safe on that part.

VladislavAntonyuk (approved):

PR #459, opened 2022-06-15
Pin .NET MAUI Version to v6.0.312

### LINKED ISSUE #424: [CI/CD] Pin .NET MAUI Version
# [CI/CD] Pin .NET MAUI Version

In the CI pipeline, let's see if it's possible to pin the version of .NET MAUI to today's GA version, [.NET MAUI v6.0.302](https://www.nuget.org/packages/Microsoft.Maui.Dependencies/6.0.312).

Similar to how we keep the [Xamarin.CommunityToolkit NuGet Package](https://www.nuget.org/packages/Xamarin.CommunityToolkit) pinned to a lower version of Xamarin.Forms than the current stable version of Xamarin.Forms, we should pin CommunityToolkit.Maui's .NET MAUI Dependency to .NET MAUI v6.0.302 for backwards compatibility which enables our library to work for as many users as possible.

VladislavAntonyuk:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #424 

 ### PR Checklist ###
 
### DIFF
--- a/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
+++ b/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <MauiVersion>6.0.312</MauiVersion>
     <TargetFrameworks>net6.0-ios;net6.0-android;net6.0-maccatalyst</TargetFrameworks>
     <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);net6.0-windows10.0.19041.0</TargetFrameworks>
     <OutputType>Exe</OutputType>
--- a/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
+++ b/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <MauiVersion>6.0.312</MauiVersion>
     <TargetFrameworks>net6.0-ios;net6.0-android;net6.0-maccatalyst</TargetFrameworks>
     <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);net6.0-windows10.0.19041.0</TargetFrameworks>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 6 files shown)


samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <MauiVersion>6.0.312</MauiVersion>
```
brminnick, line 4: What do you think about __not__ adding this to the Sample App?

I'd have to test this theory, but I  think it would allow us to ensure that the Sample App continues to work with newer versions of .NET MAUI.

I.e. If I have .NET MAUI v6.0.400 installed on my machine, it would build + run the sample app using v6.0.400, right?

Maybe we could also add this scenario to the CI workflow to ensure that newer versions of .NET MAUI successfully compile and are still compatible with `CommunityToolkit.Maui`?

brminnick (commented):
Thanks Vlad!! All of this looks great 💯 

I just had a thought about leveraging the Sample App to ensure that `CommunityToolkit.Maui` remains compatible with newer versions of .NET MAUI.

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
VladislavAntonyuk: Well, this is used in case you have multiple workloads installed on your device. 
On the one hand we can remove it at all because we set specific MAUI version on ci. 
On the other hand if we use different versions for development and testing we may receive unexpected behavior.

brminnick: Yup! My thought is that by removing it we can test newer versions of MAUI with the Sample App and then we'll know if the .NET MAUI Community Toolkit is working with the newest version of .NET MAUI instead of relying on a user to open an Issue.

VladislavAntonyuk: I've just updated MAUI to 6.0.400 and was not able to build the project. Seems like this tag pins .NET MAUI version and looks for exactly this version installed.
But because we pin it on yaml, we should not break development on a local devices. If build fails, we'll need to upgrade .NET MAUI

brminnick (approved):
Hey Vlad! 

I created two new jobs in the CI Pipeline, `build_windows_sample` + `build_macos_sample`, that install the latest .NET 6 SDK and the latest .NET MAUI Workload and then builds the sample app.

This accomplishes two things:
- Ensures the Sample App works on the latest version of .NET MAUI
- Shortens our build pipeline times
  - Instead of building the Sample App serially in the `build_windows_library` step, it is now built in parallel 

This'll help give us a heads-up if the Toolkit breaks on an upcoming .NET MAUI update.

I'll defer to you to merge this PR as long as you're cool with keeping these updates to the build pipeline!

Issue #110, opened 2021-09-28
[Proposal] GravatarImageSource


brminnick:
# GravatarImageSource

* [x] Proposed
* [x] Prototype
* [ ] Implementation: Not Started
  * [ ] iOS Support
  * [ ] Android Support
  * [ ] macOS Support
  * [ ] Windows Support
* [ ] Unit Tests: Not Started
* [ ] Sample: Not Started
* [ ] Documentation: Not Started

## Summary
[summary]: #summary

Allows the use of Gravatar Images

## Detailed Design
[design]: #detailed-design

### GravatarImageSource.shared.cs
```cs
public class GravatarImageSource : ImageSource
{
  public static readonly BindableProperty EmailProperty;
  public static readonly BindableProperty SizeProperty;
  public static readonly BindableProperty DefaultProperty;
  public static readonly BindableProperty CachingEnabledProperty;
  public static readonly BindableProperty CacheValidityProperty;
  
  public string? Email { get; set; }
  public int Size { get; set; }
  public DefaultGravatar Default { get; set; }
  public bool CachingEnabled { get; set; }
  public TimeSpan CacheValidity { get; set; }
}
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<Image>
  <Image.Source>
     <local:GravatarImageSource Email="{Binding Email}"
                                                 Size="{Binding Size}"
                                                 CachingEnabled="{Binding EnableCache}" />
  </Image.Source>
</Image>
```

### C# Usage

```cs
new Image
{
  Source = new GravatarImageSource
                    .Bind(GravatarImageSource.EmailProperty, nameof(ViewModel.Email))
                    .Bind(GravatarImageSource.SizeProperty, nameof(ViewModel.Size))
                    .Bind(GravatarImageSource.CachingEnabled, nameof(ViewModel.EnableCache))
};
```

GeorgeLeithead:
I looked at this proposal and having worked on AvatarView, I thought I would have a look as it appeared to be in the same "ball park".  Taking advantage of the fact that when an `Image` control has its _source_ property set, it's `FromUri` returns an `UriImageSource` that downloads and reads an image from a specified URI, we can then take advantage of the this as it supports Image Caching.

See [Load a remote image](https://docs.microsoft.com/en-us/dotnet/maui/user-interface/controls/image#load-a-remote-image).

[...]

PR #526, opened 2022-07-29
110 Enhance AvatarView to support Gravatar

GeorgeLeithead:
### Description of Change ###

 
Added capability to allow for Gravatar avatar's to be used.  Also provided properties for specifying `CacheValidity` and `CachingEnabled` for any `UriImageSource` used (for both existing AvatarView and new Gravatar capability).

 ### Linked Issues ###
 

#110 

 ### PR Checklist ###
 
 #### Platforms Tested ####
- [x] Android
- [ ] iOS
- [ ] MacCatalyst
- [x] Windows


src/CommunityToolkit.Maui.Core/Interfaces/IAvatarView.shared.cs
```diff
@@ -9,6 +9,43 @@ public interface IAvatarView : IBorderView, ILabel, Microsoft.Maui.IImage, IImag
 	/// <summary>Gets a value indicating the avatar border width.</summary>
 	double BorderWidth { get; }
 
+	/// <summary>Gets or sets a value of the control image cache validity.</summary>
+	TimeSpan CacheValidity { get; }
```
pictos, line 13: Do we need it? The ImageSource already has this out of the box as you can see [here](https://docs.microsoft.com/en-us/dotnet/maui/user-interface/controls/image#image-caching)

src/CommunityToolkit.Maui.Core/Interfaces/IAvatarView.shared.cs
```diff
@@ -9,6 +9,43 @@ public interface IAvatarView : IBorderView, ILabel, Microsoft.Maui.IImage, IImag
...
+
+	/// <summary>Gets or sets a value indicating whether the control image cache is enabled.</summary>
+	bool CachingEnabled { get; }
+
 	/// <summary>Gets a value indicating the avatar corner radius <see cref="Microsoft.Maui.CornerRadius"/>.</summary>
 	CornerRadius CornerRadius { get; }
+
+	/// <summary>Gets or sets a value of the control default gravatar property.</summary>
+	DefaultGravatarImage DefaultGravatar { get; }
```
pictos, line 22: Can we change this to be `Image`, then devs can use this not only for GravatarImage. Of course we can create a behavior or attached properties to abstract the Gravatar implementation, but I would like to have this more _global_

src/CommunityToolkit.Maui/Views/AvatarView.shared.cs
```diff
@@ -321,9 +355,47 @@ void HandleFontChanged()
...
+			{
+				DefaultGravatarImage.FileNotFound => "404",
+				DefaultGravatarImage.MysteryPerson => "mp",
+				_ => $"{defaultGravatar}".ToLower(),
+			};
+
+	static string GetMd5Hash(string str)
+	{
+		using MD5 md5 = MD5.Create();
+		byte[] hash = md5.ComputeHash(Encoding.UTF8.GetBytes(str));
```
pictos, line 369: Can we use `Span<byte>`? Not sure about the size of this array

src/CommunityToolkit.Maui/Views/AvatarView.shared.cs
```diff
@@ -332,6 +404,38 @@ void HandleImageChanged(ImageSource? newValue)
...
+						gravatarSize = 1;  // Images minimum is 1px
+						break;
+					case > 2048:
+						gravatarSize = 2048; // Images maximum is 2048px, however note that many users have lower resolution images, so requesting larger sizes mar result in pixelation/low-quality images.
+						break;
+				}
+
+				newValue = new UriImageSource()
+				{
+					Uri = new Uri(string.Format(requestUriFormat, GetMd5Hash(fileImageSource.File), gravatarSize, DefaultGravatarName(DefaultGravatar))),
```
pictos, line 424: String interpolation has been improved by the .NET team, so let's use it instead of `string.Format`

pictos (commented):
Hey, @GeorgeLeithead I did a quick review here, thanks for the work... A friendly reminder, we do prefer to use `var` instead of the explicit type, please refactor where it makes sense, thanks in advance.

src/CommunityToolkit.Maui.Core/Interfaces/IAvatarView.shared.cs
GeorgeLeithead: @pictos  I'm personally not sure we should do this.  It could cause a lot of network traffic, which may not be appropriate for all consumers.

Instead, if we want to provide default Images, we *could* provide a selection from embeded resources, but this would increase the size of the package for all users, if they use AvatarView or not.

The consumer COULD if they want, provide a bad email address and set the default gravatar image, thus forcing it to display an Image.  This would then is the consumers choice to implement.  We are not stopping them from doing so, should they wish.

src/CommunityToolkit.Maui/Views/AvatarView.shared.cs
GeorgeLeithead: Changed whole method to use `Span<byte>` and `BitConverter`.

src/CommunityToolkit.Maui/Views/AvatarView.shared.cs
GeorgeLeithead: Changed to use string interpolation.

src/CommunityToolkit.Maui.Core/Interfaces/IAvatarView.shared.cs
GeorgeLeithead: Removed unnecessary properties, added examples, added GravatarEmail property instead.

src/CommunityToolkit.Maui.Core/Interfaces/IAvatarView.shared.cs
pictos: Well, how this can be different from using the Image from the web in terms of `network traffic` ? I can see a lot of users that would like to use the avatar view with other web resources, like apps that have a profile, etc.

brminnick:
Hey George! We don't need to add anything from `GravatarImageSource` into `AvatarView`. 

`AvatarView` already has the property `public ImageSource ImageSource { get; set; }` and since `GravatarImageSource : ImageSource`, users are already able to use `GravatarImageSource` for their `AvatarView.ImageSource`.

brminnick:
Closing this PR because adding GravatarImageSource to AvatarView does not have an approved Proposal.

This does not implement #110 as indicated in the PR comments.

GeorgeLeithead:
> Closing this PR because adding GravatarImageSource to AvatarView does not have an approved Proposal.
> 
> This does not implement #110 as indicated in the PR comments.

@brminnick You are correct. I'll work in this ASAP.

GeorgeLeithead:
> Hey George! We don't need to add anything from `GravatarImageSource` into `AvatarView`.
> 
> `AvatarView` already has the property `public ImageSource ImageSource { get; set; }` and since `GravatarImageSource : ImageSource`, users are already able to use `GravatarImageSource` for their `AvatarView.ImageSource`.

@brminnick I wasn't focused in the right direction.  Always happy to take onboard positive advice, learn and improve.

PR #535, opened 2022-08-04
Add `Verify Code Formatting` to CI Pipeline

brminnick:
This PR adds a step to our CI/CD pipeline that will verify the submitted code has been formatted properly using `dotnet format`

- [x] Adds 2 New CI Pipeline Steps:
  - Verify Code Formatting on Windows
    - Required to verify `*.windows.cs` files
  - Verify Code Formatting on macOS
    - Required to verify `*.ios.cs` files

#### Additional Information

After adding the CI Pipeline steps, I also had to tun `dotnet format` which is why this PR includes 41 files


brminnick:
✅ Confirmed `Verify Code Formatting` step fails when `dotnet format` has not been run

brminnick:
✅ Confirmed error message appears in CI/CD logs

azure-pipelines.yml
```diff
@@ -296,4 +297,35 @@ jobs:
...
+          script: echo '##vso[task.setvariable variable=MD_APPLE_SDK_ROOT;]'/Applications/Xcode_$(XcodeVersion).app;sudo xcode-select --switch /Applications/Xcode_$(XcodeVersion).app/Contents/Developer
+      
+      - task: UseDotNet@2
+        displayName: Install Latest .NET SDK
+        inputs:
+          packageType: 'sdk'
+          version: '$(LATEST_NET6_VERSION)'
+
+      - task: CmdLine@2
+        displayName: 'Install Latest .NET MAUI Workload'
```
VladislavAntonyuk, line 319: You don’t need MAUI to run dotnet format

azure-pipelines.yml
```diff
@@ -296,4 +297,35 @@ jobs:
...
-          script: 'dotnet build $(PathToCommunityToolkitSampleCsproj) -c Release'
\ No newline at end of file
+          script: 'dotnet build $(PathToCommunityToolkitSampleCsproj) -c Release'
+
+  - job: verify_dotnet_format
+    displayName: Verify Code Formatting
+    pool:
+      vmImage: macos-12
+    steps:
+      - task: CmdLine@2
```
VladislavAntonyuk, line 307: You don’t need XCode to run dotnet format

azure-pipelines.yml
```diff
@@ -296,4 +297,35 @@ jobs:
...
+          script: dotnet workload install maui
+
+      - task: CmdLine@2
+        displayName: 'Verify Formatting'
+        inputs:
+          script: |
+            dotnet format $(PathToSamplesSolution) --verify-no-changes
+            status=$?
+            [ $status -eq 0 ] && echo "No errors found"
+            [ $status -ne 0 ] && echo "Formatting errors found. In the command line, please run \`dotnet format CommunityToolkit.Maui.Sample.sln\`, then commit + push the newly formatted code"
```
VladislavAntonyuk, line 330: Let's use variable instead of path to solution

azure-pipelines.yml
```diff
@@ -296,4 +297,35 @@ jobs:
...
+            exit $status
```
VladislavAntonyuk, line 331: I can already see the list of issues: "CI fails but I can build locally"

azure-pipelines.yml
brminnick: I thought that that `dotnet format` does use `dotnet build` under the hood?

If so, then wouldn't we need the .NET MAUI Workload?

azure-pipelines.yml
brminnick: Yup, and that's ok. 

Either the dev reads the error message and runs `dotnet format` themselves, or we can do it as the PR reviewers.

azure-pipelines.yml
brminnick: Done

azure-pipelines.yml
VladislavAntonyuk: I don't see any reason to run build on formatting. Just need to check.

If it doesn't run build then we can run this job only on one os

VladislavAntonyuk: seems like you are right. Approved

VladislavAntonyuk (approved):

VladislavAntonyuk:
I found a way to remove dotnet and maui installation by providing `--no-restore` key. If you think it is better to run restore packages before formatting please close this PR: https://github.com/CommunityToolkit/Maui/pull/537

brminnick:
Thanks for confirming Vlad!

Let's keep this one that does the full restore/build. 

It's probably overkill, but even with the additional steps in these tasks, these new CI tasks still complete faster than the other steps like "Build Windows Sample".

PR #563, opened 2022-08-21
Add Additional Rules to `.editorconfig`

brminnick:
### Description of Change ###

This PR adds the following rules:
- `.editorconfig`
  - CA1001: Type owns disposable field(s) but is not disposable
    - Ensures that we continue to implement `IDisposable` correctly
  - CA1822: Member does not access instance data and can be marked as static
    - Improve performance to avoid unnecessary stack allocations
  - CA1050: Declare types in namespaces
    - Ensures each file has a namespace declared
  - CA2016: Forward the 'CancellationToken' parameter methods that take one
    - Ensures we properly propagate `CancellationToken`
  - CA2208: Method passes parameter as the paramName argument to a ArgumentNullException constructor. Replace this argument with one of the method's parameter names. Note that the provided parameter name should have the exact casing as declared on the method.
    -  Ensures we properly pass in the arguments to `ArgumentNullException`; the constructor accepts `(string paramName, string message)` and sometimes the parameters can be accidentally swapped
  - CA1834: Use 'StringBuilder.Append(char)' instead of 'StringBuilder.Append(string)' when the input is a constant unit string
    - Small performance improvement for `StringBuilder`
  - IDE0220: Add explicit cast
    - Prevents implicit casting
    - Example:
      ```cs
      foreach (Match? match in matches) // `MatchCollection matches` is an `IEnumerable` whose content can be implicitly cast to `Match`
      // changes to
      foreach (var match in matches.Cast<Match?>()) // Ensures we continue to cast to `Match` if/when the implicit cast in `MatchCollection` changes
      ```
- `Directory.build.props`
  - <NoWarn>NETSDK1023</NoWarn>
  - Resolves #558

### DIFF
--- a/src/CommunityToolkit.Maui.Core/Views/Alert/AlertView.macios.cs
+++ b/src/CommunityToolkit.Maui.Core/Views/Alert/AlertView.macios.cs
@@ -14,10 +14,10 @@ public class AlertView : UIView
 	/// <summary>
 	/// Parent UIView
 	/// </summary>
-	public UIView ParentView => UIApplication.SharedApplication.ConnectedScenes.ToArray()
-		.Select(x => x as UIWindowScene)
-		.FirstOrDefault()?
-		.Windows.FirstOrDefault(x => x.IsKeyWindow) ?? throw new InvalidOperationException("KeyWindow is not found");
+	public static UIView ParentView => UIApplication.SharedApplication.ConnectedScenes.ToArray()
+										.Select(x => x as UIWindowScene)
+										.FirstOrDefault()?
+										.Windows.FirstOrDefault(x => x.IsKeyWindow) ?? throw new InvalidOperationException("KeyWindow is not found");
 
 	/// <summary>
 	/// PopupView Children
--- a/src/CommunityToolkit.Maui.Core/Views/DrawingView/PlatformView/MauiDrawingView.shared.cs
+++ b/src/CommunityToolkit.Maui.Core/Views/DrawingView/PlatformView/MauiDrawingView.shared.cs
@@ -6,15 +6,24 @@ namespace CommunityToolkit.Maui.Core.Views;
 /// <summary>
 /// DrawingView Platform Control
 /// </summary>
+#if !(IOS || MACCATALYST || ANDROID)
+public partial class MauiDrawingView : IDisposable
+#else
 public partial class MauiDrawingView
+#endif
 {
 	readonly WeakEventManager weakEventManager = new();
 
-	bool isDrawing;
+	bool isDrawing, isDisposed;
 	PointF previousPoint;
 	PathF currentPath = new();
 	MauiDrawingLine? currentLine;
 	Paint paint = new SolidPaint(DrawingViewDefaults.BackgroundColor);
+
+#if !(IOS || MACCATALYST || ANDROID)
+	/// <inheritdoc />
+	~MauiDrawingView() => Dispose(false);
+#endif
 
 	/// <summary>
 	/// Event raised when drawing line completed 
@@ -96,6 +105,30 @@ public partial class MauiDrawingView
 		currentPath.Dispose();
 		Lines.CollectionChanged -= OnLinesCollectionChanged;
 	}
+
+#if !(IOS || MACCATALYST || ANDROID)
+	/// <inheritdoc />
+	public void Dispose()
+	{
+		// Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+		Dispose(disposing: true);
+		GC.SuppressFinalize(this);
+	}
+
+	/// <inheritdoc />
+	protected virtual void Dispose(bool disposing)
+	{
+		if (!isDisposed)
+		{
+			if (disposing)
+			{
+				currentPath.Dispose();
+			}
+
+			isDisposed = true;
+		}
+	}
+#endif
 
 	void OnStart(PointF point)
 	{
@@ -185,12 +218,17 @@ public partial class MauiDrawingView
 		currentPath = new PathF();
 	}
 
+#if IOS || ANDROID || MACCATALYST || WINDOWS
 	void Redraw()
 	{
-#if IOS || ANDROID || MACCATALYST || WINDOWS
 		Invalidate();
-#endif
 	}
+#else
+	static void Redraw()
+	{
+
+	}
+#endif
 
 	class DrawingViewDrawable : IDrawable
 	{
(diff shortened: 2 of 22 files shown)


src/CommunityToolkit.Maui.Core/Views/Snackbar/PlatformSnackbar.macios.cs
```diff
@@ -5,116 +5,98 @@ namespace CommunityToolkit.Maui.Core.Views;
 /// <summary>
 /// UIView for Snackbar on iOS
 /// </summary>
-public class PlatformSnackbar : PlatformToast, IDisposable
+public class PlatformSnackbar : PlatformToast
```
brminnick, line 8: Moved `IDisposable` to `PlatformToast`

src/CommunityToolkit.Maui.Core/Views/Snackbar/PlatformSnackbar.macios.cs
```diff
@@ -5,116 +5,98 @@ namespace CommunityToolkit.Maui.Core.Views;
...
+    {
+        get => actionButton.TitleLabel.Font;
+        private init => actionButton.TitleLabel.Font = value;
+    }
+
+    /// <summary>
+    /// Dispose <see cref="PlatformSnackbar"/>
+    /// </summary>
+    /// <param name="isDisposing"></param>
+    protected override void Dispose(bool isDisposing)
```
brminnick: Override existing `IDisposable` implementation in `PlatformToast`

src/CommunityToolkit.Maui.UnitTests/Behaviors/IconTintColorBehavior_Tests.cs
```diff
@@ -3,6 +3,7 @@
 
 namespace CommunityToolkit.Maui.UnitTests.Behaviors;
 
+#pragma warning disable CA1416 // Validate platform compatibility
```
brminnick, line 6: Suppresses the following warning which is not relevant to Unit Tests:
> Warning | CA1416 | This call site is reachable on all platforms. 'IconTintColorBehavior' is unsupported on: 'windows'.

src/CommunityToolkit.Maui.Core/Views/Snackbar/PlatformSnackbar.macios.cs
```diff
@@ -5,12 +5,10 @@ namespace CommunityToolkit.Maui.Core.Views;
 /// <summary>
 /// UIView for Snackbar on iOS
 /// </summary>
-public class PlatformSnackbar : PlatformToast, IDisposable
+public class PlatformSnackbar : PlatformToast
```
brminnick, line 8: Move `IDisposable` to the base class, `PlatformToast`

src/CommunityToolkit.Maui.Core/Views/Snackbar/PlatformSnackbar.macios.cs
```diff
@@ -87,29 +80,18 @@ public UIFont ActionButtonFont
...
-		Dispose(true);
-		GC.SuppressFinalize(this);
-	}
-
 	/// <summary>
 	/// Dispose <see cref="PlatformSnackbar"/>
 	/// </summary>
 	/// <param name="isDisposing"></param>
-	protected virtual void Dispose(bool isDisposing)
+	protected override void Dispose(bool isDisposing)
```
brminnick, line 87: Override existing `Dispose` method from base class

src/CommunityToolkit.Maui/Alerts/Toast/Toast.shared.cs
```diff
@@ -139,24 +139,27 @@ static TimeSpan GetDuration(ToastDuration duration)
 	static PlatformToast? PlatformToast { get; set; }
 #endif
 
-
+#if !(IOS || ANDROID || MACCATALYST || WINDOWS)
+	private static partial void ShowPlatform(CancellationToken token);
```
brminnick, line 143: `ShowPlatform` can be `static` on net6.0

pictos: but, will the behavior be the same? With out the `static` the method is part of the `Toast` object, now that's static isn't part of it. In this scenario I would say that we want this to not be `static`. Thoughs?

src/CommunityToolkit.Maui.Core/Views/Alert/AlertView.macios.cs
```diff
@@ -14,10 +14,10 @@ public class AlertView : UIView
 	/// <summary>
 	/// Parent UIView
 	/// </summary>
-	public UIView ParentView => UIApplication.SharedApplication.ConnectedScenes.ToArray()
-		.Select(x => x as UIWindowScene)
-		.FirstOrDefault()?
-		.Windows.FirstOrDefault(x => x.IsKeyWindow) ?? throw new InvalidOperationException("KeyWindow is not found");
+	public static UIView ParentView => UIApplication.SharedApplication.ConnectedScenes.ToArray()
```
pictos, line 17: Just a warning, this will be a breaking change

pictos (approved):
This looks good to me in general, just want to point out one thing

src/CommunityToolkit.Maui/Converters/MathExpressionConverter/MathExpression.shared.cs
```diff
@@ -140,7 +140,7 @@ IEnumerable<string> GetReversePolishNotation(string expression)
 		var output = new List<string>();
 		var stack = new Stack<(string Name, MathOperatorPrecedence Precedence)>();
 
-		foreach (Match? match in matches)
+		foreach (var match in matches.Cast<Match>())
```
VladislavAntonyuk, line 143: Does Cast throw exception or return null?

src/CommunityToolkit.Maui.Core/Views/DrawingView/PlatformView/MauiDrawingView.shared.cs
```diff
@@ -6,16 +6,29 @@ namespace CommunityToolkit.Maui.Core.Views;
 /// <summary>
 /// DrawingView Platform Control
 /// </summary>
+#if !(IOS || MACCATALYST || ANDROID)
```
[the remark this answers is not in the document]
brminnick: Because the Windows implementation uses a class that already implements `IDisposable`

src/CommunityToolkit.Maui/Converters/MathExpressionConverter/MathExpression.shared.cs
brminnick: It throws an `InvalidCastException` if the cast is invalid (and `ArgumentNullException` if `matches` was `null`, which it isn't).

`.Cast<T>()` basically just does an `as` cast on the entire `IEnumerable`:
```cs
IEnumerable<T> typedSource = source as IEnumerable<T>;
```

https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/System.Core/System/Linq/Enumerable.cs#L1031-L1036

PR #576, opened 2022-09-01
Add Support for `*.net.cs` File Type

brminnick:
### Description of Change ###

This PR adds support for the `*.net.cs` file type, e.g. `DrawingView.net.cs` by adding the following to `CommunityToolkit.Maui.csproj` and `CommunityToolkit.Maui.Core.csproj`

```xml
<ItemGroup Condition="!($(TargetFramework.StartsWith('net')) == true AND $(TargetFramework.EndsWith('.0')) == true AND $(TargetFramework.Contains('-')) != true)"> <!-- e.g net6.0 or net7.0 -->
  <Compile Remove="**\*.net.cs" />
  <None Include="**\*.net.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
  <Compile Remove="**\net\**\*.cs" />
  <None Include="**\net\**\*.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
</ItemGroup>
```

 ### Linked Discussion ###

 - Fixes https://github.com/CommunityToolkit/Maui/discussions/301

 ### PR Checklist ###
 
 - [x] Has a linked Discussion

Adding support for `*.net.cs` allows us to move large chunks of code that used `#if !(IOS || ANDROID || MACCATALYST || WINDOWS)` into a `*.net.cs` file, like this section from `PopupHandler.shared.cs`: https://github.com/CommunityToolkit/Maui/blob/6b0f4e6a687f80436b4d6abc78bcd9617d84f699/src/CommunityToolkit.Maui.Core/Handlers/Popup/PopupHandler.shared.cs#L52-L124


src/CommunityToolkit.Maui.Core/Handlers/DrawingView/DrawingViewHandler.shared.cs
```diff
@@ -180,79 +180,4 @@ void OnPlatformViewLinesCollectionChanged(object? sender, NotifyCollectionChange
 		VirtualView.Lines.CollectionChanged += OnVirtualViewLinesCollectionChanged;
 	}
 }
-
```
VladislavAntonyuk, line 183: Can we remove #if block or replace it with #if !net?

brminnick: We could remove the `#if` block by moving the code to `DrawingViewHandler.macios.cs`, `DrawingViewHandler.android.cs` and `DrawingViewHandler.windows.cs`, but then we'd have duplicated ~150 lines of code in each new file. 

And we can't use `#if !NET` because that is also used by `net-ios`, `net-maccatalyst`, `net-android` and `net-windows`
image

brminnick: I think this is the best way to do it to reduce code duplication. 

Plus, if we ever add support for a future TFM, e.g. `net6.0-tizen`, this current `#if IOS || ...` implementation will ensure a build error happens here which then ensures we remember to verify `DrawingViewHander` specifically for the new TFM.

brminnick: Let me know what you think! If you're cool with it we can Approve/Merge this PR 🙌

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui.Core/Handlers/DrawingView/DrawingViewHandler.shared.cs
VladislavAntonyuk: Agree, the less code we duplicate, the more benefit we get from cross-platform development.

PR #577, opened 2022-09-01
Bug Report Form

VladislavAntonyuk:
### Description of Change ###

Bug Report Form to ask user  attach files:
See in action: https://github.com/VladislavAntonyuk/TemplateSample

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report_template.md
+++ b/.github/ISSUE_TEMPLATE/bug_report_template.md
@@ -1,6 +1,6 @@
 name: Bug Report
 description: Report a bug in the .NET MAUI Toolkit
-title: "[BUG] <title>"
+title: "[BUG] "
 labels: [bug, unverified]
 body:
 - type: checkboxes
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report_template.md
```diff
@@ -1,57 +1,75 @@
...
+      1. Click the button and observe the bug 🐞
+  validations:
+    required: true
+- type: input
+  attributes:
+    label: Link to public reproduction project repository
+    description: Please add a link to a public [reproduction project](https://github.com/dotnet/maui/blob/main/.github/repro.md) repository. Otherwise the issue will be marked with "Needs reproduction" label and will have a lower priority. Attached zip files cannot be opened by us.
+  validations:
+    required: true
+- type: dropdown
```
kphillpotts, line 40: I'm wondering if having the Community Toolkit version in a drop down will mean we have to keep remembering to update this bug template with each release.  Also, if we are requiring a repro project, wouldn't we be able to see the version of the toolkit from that?

VladislavAntonyuk: even if we ask a user to provide the link, there is no guarantee he will do it (Or he can use the path to the source instead of using nuget package). Having a version may help us to figure out what version is affected.

I can rename it to affected versions and allow multiple selections.

kphillpotts: Sorry what I meant was if the Community Toolkit version number is a drop down with values - won't we have to update the bug template every release to add the new version number to the drop down values.

Should it just be an entry field?

VladislavAntonyuk: yes, we'll need to update the template. I found it is used in .NET MAUI repo. Probably it would be better to have a similar form. But I have no objections if we choose input field

pictos: I'm OK to try with a field. If we see the community members are having difficulties with that we can change to be a drop down

jfversluis: > I found it is used in .NET MAUI repo

To be honest I'm thinking of removing it there, exactly because we constantly need to remember to upgrade it. Additional problem with .NET MAUI is t hat people have no clue which version they are actually using most of the time. I'm not sure what a good alternative is though, but that's a whole different discussion :D

VladislavAntonyuk: I removed that field and combined it with Environment. At least they should know the OS version

jfversluis (approved):
Looks good to me for now, let's just try it out and see where we end up!

PR #616, opened 2022-09-08
CommunityToolkit.Maui.Options

### LINKED ISSUE #540: [Bug] NumericValidationBehavior throws ArgumentNullException when Text is not set or bound to null value
### Description

When attaching the `NumericValidationBehavior` to an Entry element without setting the Entry's Text property, or when binding it to a Property that is (still) null, the `NumericValidationBehavior` throws an `ArgumentNullException`. Where it should properly validate the Entry instead. For example, the `EmailValidationBehavior` checks if value is null or empty string and then returns false instead of throwing an exception. This behavior should be the same through all validation behaviors.

### Steps to Reproduce

1. Add an Entry element to your Page. Don't set the Text property (so it remains null), or bind it to a string property that is set to null.
2. Add the NumericValidationBehavior to that Entry element.
3. Run the code and an ArgumentNullException will be thrown.

### Expected Behavior

No ArgumentNullException should be thrown, instead false should be returned (as the value is not a proper numeric value).

### Actual Behavior

An ArgumentNullException is thrown.

### Basic Information

- Version with issue: 1.2.0
- Last known good version: N/A
- IDE: Visual Studio 2022 17.4 preview 1
- Platform Target Frameworks:
  - iOS:  N/A
  - Android: N/A
  - UWP:  N/A
- Nuget Packages: None
- Affected Devices: N/A

### Workaround

Setting the Text property to an empty string in Xaml or through a binding property.

VladislavAntonyuk:
### Description of Change ###

Allows to return a default value instead of throwing an exception in case Converter or Animation receives invalid data.

 ### Linked Issues ###
 

 - Fixes #540 
 - Discussion: https://github.com/CommunityToolkit/Maui/discussions/448

 ### PR Checklist ###
 
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/139


src/CommunityToolkit.Maui/MauiCommunityToolkitOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+using CommunityToolkit.Maui.Converters;
+
+namespace CommunityToolkit.Maui;
+
+/// <summary>
+/// .NET MAUI Community Toolkit Options.
+/// </summary>
+public class MauiCommunityToolkitOptions
+{
+#pragma warning disable CA1822 // Mark members as static
```
brminnick, line 11: I'm starting to think that setting `CA1822` to `error` was too aggressive.

Let's update the `.editorconfig` in this PR , changing it to `suggestion`:
```cs
# CA1822: Member does not access instance data and can be marked as static
dotnet_diagnostic.CA1822.severity = suggestion
```

src/CommunityToolkit.Maui/MauiCommunityToolkitOptions.cs
```diff
@@ -0,0 +1,28 @@
+using CommunityToolkit.Maui.Behaviors;
+using CommunityToolkit.Maui.Converters;
+
+namespace CommunityToolkit.Maui;
+
+/// <summary>
+/// .NET MAUI Community Toolkit Options.
+/// </summary>
+public class MauiCommunityToolkitOptions
```
brminnick, line 9: What do you think about moving this file to `CommunityToolkit.Maui.Core` and bringing the same functionality to `CommunityToolkit.Maui.Core.AppBuilderExtensions`? 

Then here in `CommunityToolkit.Maui`, we can add `CommunityToolkit.Maui.MauiCommunityToolkit.Options : CommunityToolkit.Maui.Core.MauiCommunityToolkitOptions`. 

I imagine that `CommunityToolkit.Maui.Core.AppBuilderExtensions` will be empty for now, so let's make it `public abstract class`, and then we have it implemented for whenever we need it in the future!

brminnick: In other words, let's keep the methods + properties you've already implemented here, but let's lay the framework for this same functionality in `CommunityToolkit.Maui.Core`

VladislavAntonyuk: I would prefer to keep it only in CommunityToolkit.Maui.
1. It is used only in this lib (and will be used only in this lib)
2. Properties that we have here (ThrowException) are specific to our implementation and may be unnecessary and not used in other implementations like Comet.

Or did you mean in the last sentence `CommunityToolkit.Maui.Core.MauiCommunityToolkitOptions`?

brminnick: Yup - let's keep everything here, but also add the same options functionality to `CommunityToolkit.Maui.Core`:

1. Create `CommunityToolkit.Maui.Core.MauiCommunityToolkit.Options`
  - It'll be an empty class for now until we add more properties
2. In `CommunityToolkit.Maui.Core.AppBuilderExtensions`, update `UseMauiCommunityToolkit`
   ```cs
   public static MauiAppBuilder UseMauiCommunityToolkit(this MauiAppBuilder builder, Action<MauiCommunityToolkitOptions>? options = default)
   ```
3. In `CommunityToolkit.Maui.MauiCommunityToolkitOptions`, inherit from `CommunityToolkit.Maui.MauiCommunityToolkitOptions`:
   ```cs
   public class MauiCommunityToolkitOptions : Core.MauiCommunityToolkitOptions
   ```
5. In `CommunityToolkit.Maui.AppBuilderExtensions`, pass `options`  into `UseMauiCommunityToolkitCore`:
   ```cs
   return builder.UseMauiCommunityToolkitCore(options);
   ```

brminnick, line 9: After typing the class name out a few times here, it's kind of repetitive. 

What do you think about renaming it to `public class Options`? 

Then we'll have the following two classes: 
- `CommunityToolkit.Maui.Options` 
- `CommunityToolkit.Maui.Core.Options`

VladislavAntonyuk: the problem #4 is that it is a different type `Argument type 'System.Action<CommunityToolkit.Maui.MauiCommunityToolkitOptions>?' is not assignable to parameter type 'System.Action<CommunityToolkit.Maui.Core.MauiCommunityToolkitOptions>?'`

brminnick: Good point! We'd need to cast it to `CommunityToolkit.Maui.Core.MauiCommunityToolkitOptions`.

Since the first inherits the other, it should be a valid cast, right? 

```cs
return builder.UseMauiCommunityToolkitCore((Core.MauiCommunityToolkitOptions)options);
```

VladislavAntonyuk: almost. we pass action. not sure if we can cast it to `Action<Core.Options>`

VladislavAntonyuk: but we pass action. and in this case, we need to pass argument Core.Options. while we already made some changes to the Maui.Options. let's pass Options instead of `Action<Options>` to Core

VladislavAntonyuk: But to be honest, as for me our order is not correct. we should call Core and then configure the lib

src/CommunityToolkit.Maui/AppBuilderExtensions.shared.cs
```diff
@@ -17,9 +17,12 @@ public static class AppBuilderExtensions
 	/// <returns><see cref="MauiAppBuilder"/> initialized for <see cref="CommunityToolkit.Maui"/></returns>
 	public static MauiAppBuilder UseMauiCommunityToolkit(this MauiAppBuilder builder, Action<Options>? options = default)
 	{
+		builder.UseMauiCommunityToolkitCore(options is not null
+											? new Action<Core.Options>(coreOptions => options((Options)coreOptions))
+											: null);
+
```
brminnick, line 23: Hey Vlad! I found a way to cast `Action<Options>` to `Action<Core.Options>`: https://stackoverflow.com/a/3444404/5953643

src/CommunityToolkit.Maui/Behaviors/BaseBehavior.shared.cs
```diff
@@ -37,15 +38,16 @@ internal bool TrySetBindingContext(Binding binding)
 
 	internal bool TryRemoveBindingContext()
 	{
-		if (defaultBindingContextBinding != null)
+		if (defaultBindingContextBinding == null)
```
brminnick, line 41: Let's use `is null` to stick with our [Coding Standards](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#null-checking)

```suggestion
		if (defaultBindingContextBinding is null)
```

src/CommunityToolkit.Maui/AppBuilderExtensions.shared.cs
VladislavAntonyuk: what should we do with lines 24-25? it will be different options? do we want both?

brminnick: Whoops! I just realized that since `Options : Core.Options`, we only need to run `options?.Invoke(options)` once.

In other words, we can safely call `builder.UseMauiCommunityToolkitCore(null);` in `UseMauiCommunityToolkit()` because we know for a fact that the options in `Core.Options` have been set.

brminnick (changes requested):
Thanks Vlad!! 

The one big change I'd like to make is to invert the naming:
- Change `ShouldThrowExceptionInAnimations` -> `ShouldSupressExceptionsInAnimations`
- Change `ShouldThrowExceptionInConverters` -> `ShouldSpressExceptionsInConverters `
- Change `ShouldThrowExceptionInBehaviors` -> `ShouldSupressExceptionsInBehaviors`

The current naming kinda implies that we'll purposely throw exceptions to crash your app; in reality, we just throw an exception to alert the dev to an invalid use case.

Changing the name to `ShouldSupressExceptions*` let's the dev purposefully specify that they want to suppress Exceptions.

src/CommunityToolkit.Maui/AppBuilderExtensions.shared.cs
brminnick: I fixed it in this commit: https://github.com/CommunityToolkit/Maui/pull/616/commits/2b45efc8ca1c3c50663bb4bb6dd0a083bc10afe4

brminnick (commented):
Hey Vlad! What do you think about making `BaseConverter.DefaultReturnType` an `abstract` property?

I think it'd help force us to make a thoughtful decision on current + future converters about what their default return value should be, instead of making them all `default`.

For example, some of the Color converters use `Color` instead of `Color?`; would we want to make their default return value `Colors.Transparent` or `new Color()` instead of `null`?

src/CommunityToolkit.Maui/Converters/BaseConverter.shared.cs
```diff
@@ -11,6 +11,11 @@ namespace CommunityToolkit.Maui.Converters;
 /// <typeparam name="TParam">Type of parameter</typeparam>
 public abstract class BaseConverter<TFrom, TTo, TParam> : ValueConverterExtension, ICommunityToolkitValueConverter
 {
+	/// <summary>
+	/// Default value to return when the value is null.
+	/// </summary>
+	public TTo? DefaultReturnType { get; set; } = default;
```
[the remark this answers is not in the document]
VladislavAntonyuk: I think the name is not correct. It should be DefaultReturnValue. Should we make it abstract or virtual?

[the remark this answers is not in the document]
brminnick: Good catch! Fixed the name to `DefaultReturnValue` ✅

I vote for `abstract` because it'll force contributors who add new converters to thoughtfully consider the `DefaultReturnValue`. This way, a new contributor can't accidentally overlook (and forget about) configuring the `DefaultReturnValue`

brminnick:
Hey Vlad! I just realized that we should have two different properties in `BaseConverter` and `ICommunityToolkitValueConverter` to allow the user to set different values for `TFrom` and `TTo`.

I'll open a quick PR that merges into this PR this morning to demonstrate what I'm thinking 👍

### ICommunityToolkitValueConverter

```cs
{
  object? DefaultConvertReturnValue { get; }
  object? DefaultConvertBackReturnValue { get; }
  Type FromType { get; }
  Type ToType { get; }

   // ...
}
```

### BaseConverter

```cs
BaseConverter<TFrom, TTo, TParam> : ValueConverterExtension, ICommunityToolkitValueConverter
{
  public abstract TTo DefaultConvertReturnValue { get; set; }
  public abstract TFrom DefaultConvertBackReturnValue { get; set; }
  public Type FromType { get; } = typeof(TFrom);
  public Type ToType { get; } = typeof(TTo);

  object? ICommunityToolkitValueConverter.DefaultConvertReturnValue => DefaultConvertReturnValue;
  object? ICommunityToolkitValueConverter.DefaultConvertBackReturnValue => DefaultConvertBackReturnValue;
  // ...
}
```

brminnick (approved):
I think we're good to go! 

I'm happy to merge this if you're good with it, Vlad! 🙌

VladislavAntonyuk:
Looks good to me! Merge!

PR #623, opened 2022-09-12
[CommunityToolkit.Maui.Options] Change `DefaultReturnValue` to `abstract`

brminnick:
> Note: This PR targets the `maui-options` branch, not the `main` branch. In other words, when this PR is merged, it will not be merged to `main`; it will be merged into PR #616.

Hey @VladislavAntonyuk!

I took a stab at changing `DefaultReturnValue` to `abstract` and wanted to share it with you before we add it to your PR, #616.

### Changes

```cs
public TTo DefaultReturnValue { get;  set; } = default;
```
becomes
```cs
public abstract TTo DefaultReturnValue { get; }
```

### Recommendation 
I prefer to move forward with this approach for a couple reasons:
1. It will force future contributors to make a thoughtful decision for the `DefaultReturnValue` of their new Converter (instead of just letting it default it to `DefaultReturnValue = default`)
2. It allows us to use a readonly-property
    - This matches the other readonly properties in `ICommunityToolkitValueConverter`
2. It allows us to use `TTo` instead of `TTo?` for its Type
    - This allows each Converter to decide whether or not `TTo` is nullable
      - E.g., `CompareConverter` uses `object` (not `object?`) for `TTo` type whereas `ByteArrayToImageSourceConverter` uses `ImageSource?` for its `TTo` type
4. It ensures that Converters that don't return a nullable value won't return `null`
    - E.g. `ColorToGrayScaleColorConverter` uses `Color` (not `Color?`) for its `TTo` type, so I set its `DefaultReturnValue` to `Colors.Transparant` instead of `null` to avoid returning `null`

### DIFF
--- a/src/CommunityToolkit.Maui/Converters/StringToListConverter.shared.cs
+++ b/src/CommunityToolkit.Maui/Converters/StringToListConverter.shared.cs
@@ -12,7 +12,7 @@ public class StringToListConverter : BaseConverterOneWay<string?, IEnumerable, o
 	IList<string> separators = Array.Empty<string>();
 
 	/// <inheritdoc/>
-	public override IEnumerable DefaultReturnValue { get; } = Enumerable.Empty<string>();
+	public override IEnumerable DefaultReturnValue { get; } = Array.Empty<string>();
 
 	/// <summary>
 	/// The string that delimits the substrings in this string
--- a/src/CommunityToolkit.Maui/Converters/StringToListConverter.shared.cs
+++ b/src/CommunityToolkit.Maui/Converters/StringToListConverter.shared.cs
@@ -11,6 +11,9 @@ public class StringToListConverter : BaseConverterOneWay<string?, IEnumerable, o
 	string separator = " ";
 	IList<string> separators = Array.Empty<string>();
 
+	/// <inheritdoc/>
+	public override IEnumerable DefaultReturnValue { get; } = Enumerable.Empty<string>();
+
 	/// <summary>
 	/// The string that delimits the substrings in this string
 	/// This value is superseded by the ConverterParameter, if provided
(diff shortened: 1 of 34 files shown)


src/CommunityToolkit.Maui/Converters/StringToListConverter.shared.cs
```diff
@@ -11,6 +11,9 @@ public class StringToListConverter : BaseConverterOneWay<string?, IEnumerable, o
 	string separator = " ";
 	IList<string> separators = Array.Empty<string>();
 
+	/// <inheritdoc/>
+	public override IEnumerable DefaultReturnValue { get; } = Enumerable.Empty<string>();
```
brminnick, line 15: `Enumerable.Empty<string>()` matches the default return value in the `ConvertFrom()` method.

(snippet from `ConvertFrom()`):
```cs
if (value is null)
{
  return Array.Empty<string>();
}
```

src/CommunityToolkit.Maui/Converters/StringToListConverter.shared.cs
```diff
@@ -11,6 +11,9 @@ public class StringToListConverter : BaseConverterOneWay<string?, IEnumerable, o
 	string separator = " ";
 	IList<string> separators = Array.Empty<string>();
 
+	/// <inheritdoc/>
+	public override IEnumerable DefaultReturnValue { get; } = Array.Empty<string>();
```
brminnick, line 15: `Array.Empty<string>()` matches the default return value in the `ConvertFrom()` method.

(snippet from `ConvertFrom()`):
```cs
if (value is null)
{
  return Array.Empty<string>();
}
```

VladislavAntonyuk:
Thanks, Brandon, I started working on it, but you are faster!
The only question is, why do you make it getter only? I would like to make the property with getter and setter to allow user specify the default value.

brminnick:
Oh! I assumed the setter was for the creator of the converter to override the default value.

That's a good point that the user may want to customize the `DefaultReturnValue` to suit their needs 

Fixed ✅

VladislavAntonyuk (approved):

PR #625, opened 2022-09-14
[CommunityToolkit.Maui.Options] Add `BaseConverterTest`

brminnick:
> Note: This PR targets the `maui-options` branch, not the `main` branch. In other words, when this PR is merged, it will not be merged to `main`; it will be merged into PR #616.

Hey @VladislavAntonyuk!

I took a stab at adding unit tests to cover every Converter's  `DefaultConvertReturnValue` and `DefaultConvertBackReturnValue` and wanted to share it with you before we add it to your PR, #616.

## Explanation

I created `BaseConverterTest` that every* Converter Unit Test now inherits from. 

```cs
public abstract class BaseConverterTests<TConverter> : BaseTest where TConverter : ICommunityToolkitValueConverter
{
  public void InvalidConvertBackValue_ShouldThrowException(); // Skipped for One Way converters
  public void InvalidConvertBackValue_ShouldSuppressExceptionsInConverters_ShouldReturnDefaultConvertValue(); // Skipped for One Way converters
  public void InvalidConvertValue_ShouldThrowException();
  public void InvalidConverterValue_ShouldSuppressExceptionsInConverters_ShouldReturnDefaultConvertValue();
}
```

*Certain Converters will never throw an Exception, like `EqualsConverter`, `IsNullConverter` and `IsNotNullConverter` because they accept any `Type` and will always be valid based on the input provided to the converter. The Unit Test for these converters  inherit from `BaseTest` instead of `BaseConverterTests`

### Additional Information

I also took the opportunity to replace `new Random().Next` with `Random.Shared.Next` to ensure Thread Safety (thanks to this blog post: https://andrewlock.net/building-a-thread-safe-random-implementation-for-dotnet-framework/)


VladislavAntonyuk:
Thank you @brminnick.
I am not a fan of skipped tests.
I will split it on BaseOneWayConverterTests and BaseTwoWayConverterTests. In that case we can avoid methods like IsOneWayConverter.

Probably that tests which I added to BoolConverter can be moved to these base classes

VladislavAntonyuk:
@brminnick I split the BaseConverter into 2 classes and fix the naming for our Test classes. please let me know what do you think

brminnick:
> I split the BaseConverter into 2 classes
Ha! I actually did the same when first working on it. 

Looks good to me 👍

PR #633, opened 2022-09-16
Try fix Source Link

jfversluis:


VladislavAntonyuk (approved):

jfversluis:
Hm I see no change in the build artifacts. I guess they should contain the pdb files as well for this to work?

pictos:
The NuGet looks healthy
image

jfversluis:
Yeah I'm worried that this browser might not tell the full story here. If I download the artifact I still don't see any PDB files in there. Also not sure if that's supposed to be the case. I'd hate for this to be dragging on and it's only something we can test when actually publishing NuGet packages to a feed...

marcmognol:
Hi @jfversluis 
Is there any reason not using Dotnet.ReproducibleBuilds package ?

[https://github.com/dotnet/reproducible-builds](https://github.com/dotnet/reproducible-builds)

jfversluis:
@marcmognol you mean this one? :)

https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj#L100

marcmognol:
😁 yes !

VladislavAntonyuk:
It is fixed in .NET 7

jfversluis:
OK, so I think we have established that Source Link works according to @kphillpotts and the error then goes away with .NET 7. So let's just wait for that then.

Issue #107, opened 2021-09-28
[Proposal] DockLayout


brminnick:
# DockLayout

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation: Not Started

## Summary
[summary]: #summary

The DockLayout makes it easy to dock content in all four directions (top, bottom, left and right).
This makes it a great choice in many situations, where you want to divide the screen into specific areas. By default, the last element inside the DockLayout will automatically fill the rest of the space (center), unless this feature is specifically disabled.

The dock position on the child elements are set through an attached property.

Inspired by WPF DockPanel.

## Detailed Design
[design]: #detailed-design

### DockLayout.shared.cs

```cs
public class DockLayout : Layout<View>
{
  public static readonly BindableProperty DockProperty;
  public static readonly BindableProperty LastChildFillProperty;
  
  public static Dock GetDock(BindableObject bindable);
  public static void SetDock(BindableObject bindable, Dock value);
  
  public Dock Dock { get; set; }
  public bool LastChildFill { get; set; }
}
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             x:Class="MyLittleApp.MainPage">

     <StackLayout>

        <xct:DockLayout
          LastChildFill="False">
          <Button xct:DockLayout.Dock="Top" Text="Top" HeightRequest="50"/>
          <Button xct:DockLayout.Dock="Bottom" Text="Bottom" HeightRequest="50"/>
          <Button xct:DockLayout.Dock="Left" Text="Left" WidthRequest="60"/>
          <Button xct:DockLayout.Dock="Left" Text="Left" WidthRequest="60"/>
          <Button xct:DockLayout.Dock="Right" Text="Right" WidthRequest="80"/>
          <Button xct:DockLayout.Dock="Right" Text="Right" WidthRequest="80"/>
      </xct:DockLayout>

    </StackLayout>

</ContentPage>
```

### C# Usage

```cs
class MyPage : ContentPage
{
  public MyPage()
  {
    Content = new DockLayout
    {
      Children =
      {
[...]

PR #635, opened 2022-09-16
Add DockLayout (docks views to the edges of the layout container)

profix898:
### Description of Change ###

This PR implements a DockLayout which lets you dock views to the edges (top, left, right, bottom) of the DockLayout container.

 ### Linked Issues ###

[Proposal] DockLayout #107

 ### PR Checklist ###
 - [x] Documentation created or updated: MicrosoftDocs/CommunityToolkit#160

I tried to write tests, but there are several issues atm:
1. Two issues/errors with the solution
```
Error	CA1001: Type 'MaskedBehavior' owns disposable field(s) 'applyMaskSemaphoreSlim' but is not disposable
.\src\CommunityToolkit.Maui\Behaviors\MaskedBehavior.shared.cs
Error	CA1001: Type 'ValidationBehavior' owns disposable field(s) 'isAttachingSemaphoreSlim' but is not disposable
.\src\CommunityToolkit.Maui\Behaviors\Validators\ValidationBehavior.shared.cs
```
I disabled dotnet_diagnostic.CA1001.severity = error in .editorconfig temporarily.
2. CommunityToolkit.Maui.UnitTests fails to build in VS 17.4 Preview 2 with
```
Error	CS0016: Could not write to output file '.\src\CommunityToolkit.Maui.UnitTests\obj\GF\CommunityToolkit.Maui.SourceGenerators\CommunityToolkit.Maui.SourceGenerators.Generators.TextColorToGenerator\PublicTextStyleViewTextColorTo.g.shared.cs' -- 'Could not find a part of the path '.\src\CommunityToolkit.Maui.UnitTests\obj\GF\CommunityToolkit.Maui.SourceGenerators\CommunityToolkit.Maui.SourceGenerators.Generators.TextColorToGenerator\PublicTextStyleViewTextColorTo.g.shared.cs'.'
```
Unfortunately, I have no idea why this happens and what to do about it :( SourceGenerators work just fine for the 'Sample' project.


VladislavAntonyuk:
As for the issue 2, your path is too long. Max length 256 symbols. 
as for the issue 1, we’ll fix it

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -60,6 +60,12 @@
     <ProjectReference Include="..\..\src\CommunityToolkit.Maui.Analyzers.CodeFixes\CommunityToolkit.Maui.Analyzers.CodeFixes.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
   </ItemGroup>
 
+  <ItemGroup>
```
VladislavAntonyuk, line 63: This can be removed

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
```diff
@@ -0,0 +1,63 @@
+﻿using Microsoft.Maui.Layouts;
+
+namespace CommunityToolkit.Maui.Layouts;
+
+/// <summary>DockLayout positions its child elements along the edges of the layout container.</summary>
+public class DockLayout : Layout
```
VladislavAntonyuk, line 6: Please add IDockLayout interface

VladislavAntonyuk (changes requested):

VladislavAntonyuk:
Thank you for your PR. Overall it looks good. 
Could you please add tests and IDockLayout interface in Core project. See example IUniformItemsLayout.

profix898:
@VladislavAntonyuk: Thanks for the review (and the pointer to the MaxPath issue). I hope that I have addressed all your points.

**Note:** I found that UniformItemsLayout combines Layout and LayoutManager into one. Therefore, violating your own guidelines ;)
From CONTRIBUTING.md:
- Layout Managers, were introduced on .NET MAUI and they live on Microsoft.Maui.Core so makes sense to have our managers on Core as well.
- Layout, will be the implementation of ours custom layouts and will use the Layout Managers on Core

brminnick:
Thanks @profix898! When you get a chance, could you leave a comment on the DockLayout Proposal: https://github.com/CommunityToolkit/Maui/issues/107

GitHub won't allow me to assign it to you until you've left a comment.

VladislavAntonyuk:
Thank you @profix898. I will check how it works on different platforms. If everything looks good, I will approve it.

src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs
```diff
@@ -0,0 +1,156 @@
...
+		var spacing = dockLayout.Spacing;
+
+		var x = bounds.Left + padding.Left;
+		var y = bounds.Top + padding.Top;
+
+		var width = bounds.Width - padding.HorizontalThickness;
+		var height = bounds.Height - padding.VerticalThickness;
+
+		var i = 0;
+		foreach (var child in dockLayout)
```
pictos, line 85: Why not use the `for` loop instead? Since we need the index

src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs
```diff
@@ -0,0 +1,156 @@
...
+				{
+					childX = x;
+					childY = y + height - childHeight;
+					childWidth = width;
+					height -= (childHeight + spacing.Height);
+					break;
+				}
+				default:
+				{
+					goto case DockEnum.Top;
```
pictos, line 147: Instead using the `goto`, can we stack the default with the `case DockEnum.Top`? Something like

```csharp
case DockEnum.Top:
case DockEnum.None:
case default:
//code
```

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
```diff
@@ -0,0 +1,61 @@
...
+using CommunityToolkit.Maui.Core.Layouts;
+using Microsoft.Maui.Layouts;
+
+namespace CommunityToolkit.Maui.Layouts;
+
+/// <summary>DockLayout positions its child elements along the edges of the layout container.</summary>
+public class DockLayout : Layout, IDockLayout
+{
+	/// <summary>Docking position for a view.</summary>
+	public static readonly BindableProperty DockProperty = BindableProperty.Create(nameof(DockEnum), typeof(DockEnum), typeof(DockLayout), DockEnum.None, BindingMode.TwoWay);
```
pictos, line 11: Why does this need to be `BindingMode.TwoWay` by default?

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
```diff
@@ -0,0 +1,61 @@
...
+
+	/// <summary>If true, the last child fills the remaining space.</summary>
+	public static readonly BindableProperty LastChildFillProperty = BindableProperty.Create(nameof(LastChildFill), typeof(bool), typeof(DockLayout), true, BindingMode.TwoWay);
```
pictos, line 14: Why does this need to be `BindingMode.TwoWay` by default?

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
```diff
@@ -0,0 +1,61 @@
...
+
+	/// <summary>Horizontal (width) and vertical (height) spacing between docked views.</summary>
+	public static readonly BindableProperty SpacingProperty = BindableProperty.Create(nameof(Spacing), typeof(SizeF), typeof(DockLayout), SizeF.Zero, BindingMode.TwoWay);
```
pictos, line 17: Why does this need to be `BindingMode.TwoWay` by default?

pictos (changes requested):
Hey, @profix898 thanks for taking the time to implement this. I've some questions and requests. Let me know if you have any question (:

src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs
profix898: Absolutely. Will do.

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
profix898: I guess, we dont ...

src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs
profix898: Was not easily possible before my last refactoring to use explicit fallback value (DockEnum.None). But now  ... of course ...

profix898:
@pictos: Thanks for your feedback. I made the changes as requested/suggested.

> Let me know if you have any question (:

I guess the biggest question would be the overall structure: What goes into Core? Are you happy with the interfaces? Layout vs. LayoutManager (see [my note above](https://github.com/CommunityToolkit/Maui/pull/635#issuecomment-1249536529)).

Also, since updating to VS 17.4 preview 2, I have serious issues testing on Android. So I hope it still works Ok.

brminnick:
> What goes into Core? 

If a class uses the namespace `Microsoft.Maui.Controls.*`, it belongs in `CommunityToolkit.Maui`. 

Otherwise, it can go in `CommunityToolkit.Maui.Core`

Here's more info in our Contribution Guidelines:

https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#where-to-make-your-changes

profix898:
Thanks! Should be Ok then ... Let me know if you'd like any additional changes.

src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs
```diff
@@ -81,21 +81,19 @@ public override Size ArrangeChildren(Rect bounds)
...
 			}
 
-			i++;
-
 			var childWidth = Math.Min(width, child.DesiredSize.Width);
 			var childHeight = Math.Min(height, child.DesiredSize.Height);
 
-			var lastItem = (i == dockLayout.Count);
-			if (lastItem && dockLayout.LastChildFill)
+			var isLastChild = (i == dockLayout.Count - 1);
```
pictos, line 95: Does this syntax work? This is just for readability

```suggestion
			var isLastChild = (i == dockLayout[^1]);
```

src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs
```diff
@@ -107,17 +105,9 @@ public override Size ArrangeChildren(Rect bounds)
...
-					childX = x;
-					childY = y;
-					childHeight = height;
-					x += (childWidth + spacing.Width);
-					width -= (childWidth + spacing.Width);
-					break;
-				}
 				case DockEnum.Top:
 				case DockEnum.None:
+				default:
```
pictos, line 110: Not sure if this should be at the end of the switch statement. But if the tests worked I'm ok with this here

pictos (commented):
Hey @profix898 thanks for the quick updates, I've now a few comments, that doesn't invalidate the PR. After these I think we're good to go

src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs
profix898: According to https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/selection-statements

> The default case can appear in any place within a switch statement. Regardless of its position, the default case is always evaluated last and only if all other case patterns aren't matched.

I also tested this, and it seems to work Ok (I wasn't sure myself ;))

src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs
profix898: I guess you mean `var isLastChild = (child == dockLayout[^1]);`?

profix898: Then, I can do `foreach` again (as the index is no longer needed).

pictos: Nevermind, I just read this wrong.

pictos (dismissed):
LGTM! Thanks @profix898

src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace CommunityToolkit.Maui.Core.Interfaces;
+
+/// <summary>Docking position for a view in <see cref="IDockLayout"/>.</summary>
+public enum DockEnum
```
brminnick, line 4: Let's rename this enum to `DockPosition`
```suggestion
public enum DockPosition
```

src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs
```diff
@@ -0,0 +1,31 @@
...
+	/// <summary>If true, the last child fills the remaining space.</summary>
+	public bool LastChildFill { get; }
+
+	/// <summary>Horizontal (width) and vertical (height) spacing between docked views.</summary>
+	public SizeF Spacing { get; }
+
+	/// <summary>Gets the docking position for a view (or Dock.Top as default value).</summary>
+	/// <param name="view">A view that belongs to the DockLayout.</param>
+	/// <returns>DockEnum that signifies where the item will dock.</returns>
+	DockEnum GetDock(IView view);
```
brminnick, line 30: Let's rename this method to `GetDockPosition`

```suggestion
	DockPosition GetDockPosition(IView view);
```

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
```diff
@@ -0,0 +1,61 @@
...
+	/// <inheritdoc/>
+	public DockEnum GetDock(IView view)
+	{
+		if (view is BindableObject bindable)
+		{
+			return (DockEnum)bindable.GetValue(DockProperty);
+		}
+
+		return DockEnum.None;
+	}
```
brminnick, line 54: 1. Let's rename this method to `GetDockPosition`
2. Every `Microsoft.Maui.Controls.View` control implements `Microsoft.Maui.IView`, so we can directly cast `IView` to `View`

```suggestion
	public DockPoistion GetDockPostion(IView view)
	{
		var mauiView = (Microsoft.Maui.Controls.View)view;
		return (DockEnum)mauiView.GetValue(DockProperty);
	}
```

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
```diff
@@ -0,0 +1,61 @@
...
+
+	/// <inheritdoc/>
+	protected override ILayoutManager CreateLayoutManager()
+	{
+		return new DockLayoutManager(this);
+	}
```
brminnick, line 60: Let's use an expression bodied member here

```suggestion
	protected override ILayoutManager CreateLayoutManager() => new DockLayoutManager(this);
```

src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+/// <summary>DockLayout positions its child elements along the edges of the layout container.</summary>
+public interface IDockLayout : ILayout
+{
+	/// <summary>If true, the last child fills the remaining space.</summary>
+	public bool LastChildFill { get; }
+
+	/// <summary>Horizontal (width) and vertical (height) spacing between docked views.</summary>
+	public SizeF Spacing { get; }
```
brminnick, line 25: Let's use `double` instead of `SizeF` for `Spacing`.

Using `double` aligns better with the spacing properties inside of other Layouts:
- `public double Grid.ColumnSpacing { get; set; }`,
- `public double Grid.RowSpacing { get; set; }`
- `public double VerticalStackLayout.Spacing { get; set; }`
- `public double HorizontalStackLayout.Spacing { get; set; }`

This'll be used as vertical spacing when `DockPosition.Left` and `DockPosition.Right`, and it'll be used for horizontal spacing when `DockPosition.Top` and `DockPosition.Bottom`, similar to how `StackLayout` auto-implements the Spacing based upon `StackLayout.Orientation`.

```suggestion
	public double Spacing { get; }
```

brminnick (changes requested):
Thanks @profix898 + @pictos!

I have a couple recommendations for the naming and types before we start writing the docs

src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs
```diff
@@ -0,0 +1,31 @@
...
+	None = 0,
+	/// <summary>Dock at the top.</summary>
+	Top = 1,
+	/// <summary>Dock to the left.</summary>
+	Left = 2,
+	/// <summary>Dock to the right.</summary>
+	Right = 3,
+	/// <summary>Dock at the bottom.</summary>
+	Bottom = 4
+}
```
brminnick, line 16: Crazy thought - what if we make this enum a flag by adding the `[Flags]` attribute?

Then, for example, users could place a Dock on both the top + bottom:

var dockLayout = new DockLayout();
dockLayout.SetPosition(DockPosition.Top | DockPosition.Bottom);

I imagine most devs won't use it as a flag, but implementing the enum as a Flag now is easy whereas if we ever wanted to change it later would be a breaking change.


```suggestion
[Flags]
public enum DockEnum
{
	/// <summary>Do not dock.</summary>
	None = 0,
	/// <summary>Dock at the top.</summary>
	Top = 1,
	/// <summary>Dock to the left.</summary>
	Left = 2,
	/// <summary>Dock to the right.</summary>
	Right = 4,
	/// <summary>Dock at the bottom.</summary>
	Bottom = 8
}
```

src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace CommunityToolkit.Maui.Core.Interfaces;
+
+/// <summary>Docking position for a view in <see cref="IDockLayout"/>.</summary>
+public enum DockEnum
+{
+	/// <summary>Do not dock.</summary>
+	None = 0,
```
brminnick, line 7: What's the purpose of `None`? Can we remove it?

Or could we rename it to `Unassigned` and then throw an `Exception` in `DockLayoutManager.Measure` if the dev hasn't specified a position?   

```cs
public override Size Measure(double widthConstraint, double heightConstraint)
{
  // ...

  switch (position)
  {
    case DockPosition.Unassigned:
throw new InvalidOperationException($"{position} invalid. {nameof(DockPosition)} must be assigned.");
    // ...
  }

  // ...
}

File: src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs (hunk 15 above)
Comment (pictos) [reply src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs:16]: But how this works in the switch? I believe we will need to add a case for the `Top | Bottom` case, no?

Comment (brminnick) [reply src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs:16]: Good point! We'd convert the `switch` to a series of `if` statements:

```cs
if(position is DockPosition.Top)
{
  // ...
}
if(position is DockPosition.Bottom)
{
  // ...
}
if(position is DockPosition.Left)
{
  // ...
}
if(position is DockPosition.Right)
{
  // ...
}

File: src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs (hunk 14 above)
Comment (profix898) [reply src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs:25]: > This'll be used as vertical spacing when DockPosition.Left and DockPosition.Right, and it'll be used for horizontal spacing when DockPosition.Top and DockPosition.Bottom.

The way it's implemented now, you can have different spacing for horizontal/vertical and for my purposes I use that a lot. I would therefore suggest having both `double VerticalSpacing` and `double HorizontalSpacing`.

File: src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs (hunk 15 above)
Comment (profix898) [reply src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs:16]: > Then, for example, users could place a Dock on both the top + bottom:

Not sure, I get your point. DockPosition (or DockEnum) refers to the position of the child view inside the DockLayout container. It's not `dockLayout.SetPosition(DockPosition.Top | DockPosition.Bottom);`, but `DockLayout.SetPosition(childView, DockPosition.Top | DockPosition.Bottom);`. How can you dock a view both top and bottom at the same time? What does that mean? Do you think in terms of 'anchoring' the view? If so, that would also drastically change the layout calculations.

File: src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs (hunk 16 above)
Comment (profix898) [reply src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs:7]: _None_ is the default value. It essentially means, do not dock the view anywhere to the sides (top, left, right, bottom), but place it as 'content' (= center, if all other positions are occupied). I thought to make the default case explicit (also simplifies the logic a little bit).

Comment (brminnick) [reply src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs:7]: In what scenarios would a developer use a DockLayout with `DockLayout.None`?

File: src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs (hunk 12 above)
Comment (profix898) [reply src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs]: Not a fan, but Ok.

**Comment by profix898** [comment]
I renamed DockEnum to DockPosition (and all related methods as well). Didn't use Github suggestions directly to make sure usages (incl. sample page) are updated accordingly.

File: src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs (hunk 16 above)
Comment (profix898) [reply src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs:7]: If you look at the sample page, the last child (which typically fills the remaining space) is specified with `DockPosition.None`. `DockPosition.None` is the same as omitted it (default value).
`None` works similar to `Top`, but here both width+height are considered. For `Top` and `Bottom` only the height is applied explicitely (width is given by the surrounding view proportions). For `Left`and `Right` its only the width.

File: src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs (hunk 15 above)
Comment (VladislavAntonyuk) [reply src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs]: I agree with @profix898 . We can add additional values like TopLeft or BottomRight, but the flag is not correct, because it allows Top | Bottom. Anyway it won't be a breaking change, so we can enhance it by request.

**Comment by VladislavAntonyuk** [comment]
Thank you @profix898, I updated the sample to show different layouts. I also moved DockPosition to the primitives folder. I also updated Spacing type from SizeF to Thickness.

**Comment by profix898** [comment]
Thanks @VladislavAntonyuk for your review and changes. However, I didn't use `Thickness` for the `Spacing`property for a reason ;) It allows you to specify separate values for Left/Top/Right/Bottom, but these values do not make any sense in the context of DockLayout IMO. We only have vertical and horizontal spacing. I think we should keep SizeF (with Width and Height) or move to two doubles (for vertical/horizontal) as suggested by @brminnick.

**Comment by VladislavAntonyuk** [comment]
Maybe I incorrectly understand spacing, but what I understand it as margin of the center element. So we may have different spaces of Top,Bottom, Left,Right

**Comment by profix898** [comment]
You can have multiple views docked to each side (top,left,right,bottom). Currently, the spacing is applied vertically/horizontally between each of them.
For example, if you dock 2 views to the left, there is 1x horizontal spacing between the 2 docked views AND 1x horizontal spacing between the right-most docked view and the content/center view. So, spacing works as with the Grid layout, which has column/row spacing, respectively. If you want individual 'spacing' around a specific view (e.g the 'center'), you can still apply a `Padding` to that view.

**Comment by profix898** [comment]
Pushed a version with HorizontalSpacing/VerticalSpacing. Feel free to revert in case you disagree.

File: src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs [hunk 17]
```diff
@@ -0,0 +1,152 @@
...
+					break;
+
+				case DockPosition.None:
+					width += childSize.Width;
+					widthLimit -= childSize.Width;
+					height += childSize.Height;
+					heightLimit -= childSize.Height;
+					break;
+
+				default:
```
VladislavAntonyuk, line 65: Let's combine None and default, what do you think?

src/CommunityToolkit.Maui.Core/Primitives/DockPosition.shared.cs
```diff
@@ -0,0 +1,18 @@
+﻿using CommunityToolkit.Maui.Core.Interfaces;
+
+namespace CommunityToolkit.Maui.Core.Primitives;
```
VladislavAntonyuk, line 3: Please remove primitives from the namespace

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
```diff
@@ -0,0 +1,65 @@
...
+	public static readonly BindableProperty LastChildFillProperty = BindableProperty.Create(nameof(LastChildFill), typeof(bool), typeof(DockLayout), true);
+
+	/// <summary>Horizontal spacing between docked views.</summary>
+	public static readonly BindableProperty HorizontalSpacingProperty = BindableProperty.Create(nameof(HorizontalSpacing), typeof(double), typeof(DockLayout), 0.0);
+
+	/// <summary>Vertical spacing between docked views.</summary>
+	public static readonly BindableProperty VerticalSpacingProperty = BindableProperty.Create(nameof(VerticalSpacing), typeof(double), typeof(DockLayout), 0.0);
+
+	/// <inheritdoc/>
+	public bool LastChildFill
```
VladislavAntonyuk, line 24: Maybe name it: ShouldFillLastChild?

src/CommunityToolkit.Maui.Core/Layouts/DockLayoutManager.shared.cs
brminnick: I purposely kept them separate. 

As a best practice, when performing a `switch` on an `enum`, it's best to use the `default` case to throw a `NotSupportedException`. 

This ensures that when a new value is added to the `enum` that we don't forget to implement the new value in every `switch`.

VladislavAntonyuk: In that case let’s add it to arrange method as well

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
profix898: How about `ExpandLastChild`? `ShouldFillLastChild` sounds very strange. We are not filling the child; we are filling the container space with the child (view).

profix898:
I think I addressed all your latest comments. Property is called `ExpandLastChild` now. Anything else?

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
pictos: To be more clear for devs to find/know this is a boolean property, we would like to follow a pattern, that's adding a prefix (Is/Has/Should/Will/Do/Must/etc) to the property

e.g.: `IsLastChildFill`

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
```diff
@@ -21,10 +21,10 @@ public class DockLayout : Layout, IDockLayout
 	public static readonly BindableProperty VerticalSpacingProperty = BindableProperty.Create(nameof(VerticalSpacing), typeof(double), typeof(DockLayout), 0.0);
 
 	/// <inheritdoc/>
-	public bool ExpandLastChild
+	public bool IsLastChildFill
```
pictos, line 24: I'm not a fluent english speaker, so I would like @brminnick to confirm if the new name is correct/makes sense

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
brminnick: Yup! Variable names for Booleans should a question that the value of the Boolean answers, eg IsBusy, IsVisible, etc

profix898: @brminnick: With this criterium, I could also use `ExpandLastChild`, not? Anyway, what name would you suggest/approve? So that we can resolve this discussion?

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
profix898: Thanks. Renamed to `ShouldExpandLastChild` as per brminnick.

VladislavAntonyuk (approved):

brminnick (approved):
Thanks @profix898! Excellent work 💯 

The next step is to write the docs by opening a Pull Request on our Docs Repo: https://github.com/MicrosoftDocs/CommunityToolkit/

We've updated the Proposal, #107, moving it to  [Pull Request Approved, Pending Documentation](https://github.com/CommunityToolkit/Maui/projects/1#column-16085871) on our [Project Board](https://github.com/CommunityToolkit/Maui/projects/1).

Once the docs are complete, we will merge this PR 🙌

profix898:
Is anyone working on the documentation? Or should I give it a go?

bijington:
> Is anyone working on the documentation? Or should I give it a go?

Sorry I don't think it was clear in our initial comments. We usually do prefer the author of the PR to write the documentation as they have the most amount of context. That being said if you need any assistance in writing it please reach out to me and I can assist you 👍

VladislavAntonyuk:
Hey @profix898 , are you working on adding docs? Do you need any help?

profix898:
Yes, I started working on documentation. Unfortunately, I'm currently quite busy with my day job. So, it might take 1-2 weeks to complete an initial PR for docs.

profix898:
Added PR for docs and a new `DockLayout.Add(IView, DockPosition)` method to make it easier to populate the layout from C# (see docs for details).

src/CommunityToolkit.Maui.Core/Interfaces/IDockLayoutManager.shared.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Microsoft.Maui.Layouts;
+
+namespace CommunityToolkit.Maui.Core.Interfaces;
+
+/// <summary>LayoutManager for <see cref="IDockLayout"/>.</summary>
+public interface IDockLayoutManager : ILayoutManager
```
pictos, line 6: What's the value to have an empty interface?

src/CommunityToolkit.Maui.Core/Primitives/DockPosition.shared.cs
```diff
@@ -0,0 +1,18 @@
+﻿using CommunityToolkit.Maui.Core.Interfaces;
```
pictos, line 1:
```suggestion
namespace CommunityToolkit.Maui.Core;
```

src/CommunityToolkit.Maui/Layouts/DockLayout.shared.cs
```diff
@@ -0,0 +1,75 @@
+﻿using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.Core.Interfaces;
+using CommunityToolkit.Maui.Core.Layouts;
+using Microsoft.Maui.Layouts;
+
+namespace CommunityToolkit.Maui.Layouts;
+
+/// <summary>DockLayout positions its child elements along the edges of the layout container.</summary>
```
pictos, line 8: Just curious, why your IDE didn't follow the pattern for adding a `summary`? Something like this

```
/// <summary>
/// Visual Options for <see cref="CommunityToolkit.Maui.Core.Views.AlertView"/>
/// </summary>
```

Can you change that easily?

src/CommunityToolkit.Maui.Core/Interfaces/IDockLayout.shared.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace CommunityToolkit.Maui.Core.Interfaces;
```
pictos, line 1:
```suggestion
﻿namespace CommunityToolkit.Maui.Core;
```

src/CommunityToolkit.Maui.Core/Interfaces/IDockLayoutManager.shared.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Microsoft.Maui.Layouts;
+
+namespace CommunityToolkit.Maui.Core.Interfaces;
```
pictos, line 3:
```suggestion
namespace CommunityToolkit.Maui.Core;
```

pictos (changes requested):
Hey @profix898 thanks for moving this forward, I've some questions and suggestions.

src/CommunityToolkit.Maui.Core/Interfaces/IDockLayoutManager.shared.cs
profix898: Originally, I introduced that interface to ensure that you can't inject any (potentially incompatible) ILayoutManager. For that, a more specific (but empty) interface IDockLayoutManager works well. However, I just checked the latest MAUI sources and they do not take any precautions in that direction either. Removing the interface makes sense then, I guess.

VladislavAntonyuk (dismissed):

profix898:
Sidenote to MCT maintainers: It's kind of difficult to recognize a pattern of best practices from the source code. It's simply not consistent enough (yet).
Two quick examples:
1. UniformItemsLayout combines ILayout and ILayoutManager into a single class. From MAUI docs they should be separate. dotNET 7 MAUI even introduces a new concept (similar to Handlers) where you (as a user) can replace the LayoutManager of existing layouts (see dotnet/maui#9980), which obviously is impossible if both are combined.
2. You asked me to put enums into Primitives. But for IToast the ToastDuration enum is in the same file as the interface.
3. .editorconfig is still too incomplete (thus leading to inconsistent style in VS)
4. `dotnet format` (which I used to overcome problem 3) does not currently work with .NET 7 RC SDKs

profix898:
@VladislavAntonyuk: You were probably 5 mins ahead of me ;) Thanks for helping out.

VladislavAntonyuk (approved):

pictos:
@profix898 thanks for your feedback, I'm working on more docs on how to contribute... Right now the best you can do is to look into our code and find some patterns. During .NET7 release we will work to make the pattern more solid

bijington:
Yes thanks for the feedback. I do have an item on my list to improve the .editorconfig file so that it gives a better experience for devs, if you have any concrete examples of issues with it I would really appreciate them

profix898:
@pictos: Absolutely. Keep at it. Improving consistency throughout the repo is the best (and probably only) way to help with this.

@bijington: I poked the documentation a little to better understand the issues with editorconfig. Not sure it's really correct, but my understanding is that starting with a (user-specific) default IDE configuration, the editorconfig rules are applied hierarchically to override the defaults. In that case, every options that is not specifically listed in editorconfig would use the local IDE setting instead. That would certainly, explain a lot of issues, I was seeing with this repo. For one, there are no editorconfig options for XML comments (at least I could not find any). Second, my typical code style (my defaults) appear to creep through at various places, so I suggest that there are no explicit values for these settings in your editorconfig. I'm usually using ReSharper on top (but temporarily disabled it for this), where you have very detailed control over code style (there are approx 5x more settings in R# than in editorconfig), but it additionally messes with the supposed IDE defaults. Here, I attach an editorconfig file with my IDE defaults for reference (although the export appears to be quite incomplete as well). I'm not sure which options are a problem, but maybe it helps you figuring it out!? I did find 1.5 though: `csharp_space_after_cast`, also blank line before returns (not sure there is an editorconfig option for that). When running `dotnet format` on the console, there are no IDE defaults (applies mostly for VS Code, etc. as well). For that reason, it appears that at least most styles are what you expect there. So much for my 2 cents ...

src/CommunityToolkit.Maui.UnitTests/Layouts/UniformItemsLayoutTests.cs
```diff
@@ -80,20 +80,7 @@ public void MaxRowsArrangeChildrenUniformItemsLayout()
...
-
-	[Fact]
-	public void ArrangeChildrenUniformItemsLayout()
-	{
-		var expectedSize = new Size(childWidth, childHeight);
-		uniformChild = new TestView(expectedSize);
-		uniformItemsLayout.Measure(double.PositiveInfinity, double.PositiveInfinity);
-		var rect = new Rect(0, 0, childWidth * childCount, childHeight * childCount);
-		uniformItemsLayout.Layout(rect);
-		var actualSize = uniformItemsLayout.ArrangeChildren(rect);
```
brminnick, line 96: Any concerns about removing this entire Unit Test?

brminnick (dismissed):
Thanks @profix898!! Looks good to me! 

I noticed one of `UniformItemsLayout` tests got removed and just want to double-check that it was done on purpose and that wasn't temporarily removed when debugging/testing.

brminnick:
@VladislavAntonyuk As the Proposal Champion for `DockLayout`, I'll defer to you to merge this. If everything looks good to you, let's merge this tonight or tomorrow morning in preparation for the [v1.4.0 release](https://github.com/CommunityToolkit/Maui/milestone/3)!

src/CommunityToolkit.Maui.UnitTests/Layouts/UniformItemsLayoutTests.cs
profix898: I didn't touch anything related to UniformItemsLayout! @VladislavAntonyuk refactored UniformItemsLayout (split into Layout and LayoutManager) in one of his recent commits and removed the test. Since merging is apparently his call anyway, I would let it for him to check whether the removal was on purpose or accidental.

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui.UnitTests/Layouts/UniformItemsLayoutTests.cs
VladislavAntonyuk: fixed

pictos (approved):

PR #667, opened 2022-10-01
Implement MediaElement

### LINKED ISSUE #113: [Proposal] MediaElement
# MediaElement

* [x] Proposed
* [x] Prototype: Not Started
* [x] Implementation: Not Started
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests: Not Started
* [x] Sample: Not Started
* [x] Documentation: Not Started

## Summary
[summary]: #summary

MediaElement is a view for playing video and audio. Media that's supported by the underlying platform can be played from the following sources:
- The web, using a URI (HTTP or HTTPS).
- A resource embedded in the platform application, using the ms-appx:/// URI scheme.
- Files that come from the app's local and temporary data folders, using the ms-appdata:/// URI scheme.
The device's library.

## Detailed Design
[design]: #detailed-design

### MediaElement.shared.cs

```cs
public class MediaElement : View, IMediaElementController
{
  public static readonly BindableProperty AspectProperty;
  public static readonly BindableProperty AutoPlayProperty;
  public static readonly BindableProperty BufferingProgressProperty;
  public static readonly BindableProperty CurrentStateProperty;
  public static readonly BindableProperty DurationProperty;
  public static readonly BindableProperty IsLoopingProperty;
  public static readonly BindableProperty KeepScreenOnProperty;
  public static readonly BindableProperty PositionProperty;
  public static readonly BindableProperty ShowsPlaybackControlsProperty;
  public static readonly BindableProperty SourceProperty;
  public static readonly BindableProperty VideoHeightProperty;
  public static readonly BindableProperty VideoWidthProperty;
  public static readonly BindableProperty VolumeProperty;
  public static readonly BindableProperty SpeedProperty;
  
  public double BufferingProgress { get; }
  public bool CanSeek { get; }
  public MediaElementState CurrentState { get; }
  public TimeSpan? Duration { get; }
  public int VideoHeight { get; }
  public int VideoWidth  { get;}
  
  public Aspect Aspect { get; set; }
  public bool AutoPlay { get; set; }
  public bool IsLooping { get; set; }
  public bool KeepScreenOn { get; set; }
  public bool ShowsPlaybackControls { get; set; }
  public TimeSpan Position { get; set; }
  [Forms.TypeConverter(typeof(MediaSourceConverter))]
  public MediaSource? Source { get; set; }
  public double Volume { get; set; }
  public double Speed { get; set; }
  
  public event EventHandler? MediaEnded;
  public event EventHandler? MediaFailed;
  public event EventHandler? MediaOpened;
  public event EventHandler? SeekCompleted;
  
  public void Play() ;
  public void Pause();
  public void Stop();
}
```

## Usage Syntax
[usage]: #usage-syntax

<!-- Here is an example from CharactersValidationBehavior:

### XAML Usage

```xml
<MediaElement Source="https://sec.ch9.ms/ch9/5d93/a1eab4bf-3288-4faf-81c4-294402a85d93/XamarinShow_mid.mp4"
              ShowsPlaybackControls="True" />
```

### C# Usage

```cs
new MediaElement
{
  Source = "https://sec.ch9.ms/ch9/5d93/a1eab4bf-3288-4faf-81c4-294402a85d93/XamarinShow_mid.mp4"
}
```

### LINKED ISSUE #717: MediaElement TODO list
Just putting this issue out here with some things we might want to look into regarding MediaElement without littering the repo with issues for a control that doesn't exist yet.

If you are reading this as a "outside" user, feel free to add any feedback as comments under this issue while MediaElement is still being implemented.

* [x] Got [this repo](https://github.com/Varmod/LocalVideoWontPlay) through someone on my blogpost, just pointing to a filename doesn't work. TODO: Check if we support this scenario and why it doesn't work or how we can easily accommodate this scenario.

Implemented the ResourceMediaSource for loading local media. We might need to make the detection smarter if the users adds a URL or filename in there though, but at least it should be possible to play this easily now

* [x] Fix iOS 16 platform playback control stuff; somewhat done, it shows, but there is a weird top offset thing going on, possibly a bug in .NET MAUI
* [x] Implement MediaManager (underway by @pictos)
* [x] See why [this link](https://mintbackend.azurewebsites.net/api/getuploads/RDpcaG9tZVxzaXRlXHd3d3Jvb3RcdXBsb2Fkc1xjY2dsbHMyZS5qNWEubXA0/dmlkZW8vbXA0) isn’t working on iOS **
* [x] See why [this link](https://samcloud.spacial.com/api/listen?sid=73841&rid=259088&f=mp3,any&br=128000,any&m=sc&t=ssl) isn't working on Android. Has to do with no duration being reported and that overflowing the TimeSpan we try to set for the duration. We need to account for that. Fixed with #787 
* [x] Why is this happening on Android: https://github.com/CommunityToolkit/Maui/issues/113#issuecomment-1303161996
* [x] Make the status (playing, paused, etc.) consistent

** This link doesn't work and fails with this error on iOS
> LocalizedFailureReason: "The server is not correctly configured."
	LocalizedRecoveryOptions: (null)
	LocalizedRecoverySuggestion: (null)
	UnderlyingErrors: {Foundation.NSError[1]}
	UserInfo: {{
    NSLocalizedDescription = "Operation Stopped";
    NSLocalizedFailureReason = "The server is not correctly configured.";
    NSUnderlyingError = "Error Domain=CoreMediaErrorDomain Code=-12939 \"byte range length mismatch - should be length 2 is length 189217134\" UserInfo={NSDescription=byte range length mismatch - should be length 2 is length 189217134, NSURL=https://mintbackend.azurewebsites.net/api/getuploads/RDpcaG9tZVxzaXRlXHd3d3Jvb3RcdXBsb2Fkc1xjY2dsbHMyZS5qNWEubXA0/dmlkZW8vbXA0}";

It seems that this is simply something that is incompatible with iOS and required extra configuration on the server-side.

* [x] Add ILogger messages where useful
* [x] Make SeekComplete work on iOS for scrubbing with platform control (already works with seeking through custom control)
* [x] Make SeekComplete not fire multiple times on Android when using a custom control (already works when using platform control)
* [x] Implement MediaEnded on Windows
* [x] Add KeepScreenOn
* [x] Investigate if we can implement KeepScreenOn for Windows
* [x] Double-check/fix the TODOs in code
* [x] Check if volume/speed/etc. is in sync between platform player controls and our own when setting it through MediaElement or vice versa

jfversluis:
### Description of Change ###

Adds `MediaElement` to the .NET MAUI Community Toolkit! This is the core functionality that we can expand on. I've tried my best to fix all the issues that we've seen for the Xamarin Community Toolkit version of this.

Current public API:

```csharp
/// <summary>
/// With MediaElement you can play multimedia inside of your app.
/// </summary>
public interface IMediaElement : IView
{
	/// <summary>
	/// The current state of the <see cref="MediaElement"/>.
	/// </summary>
	MediaElementState CurrentState { get; }

	/// <summary>
	/// Gets total duration of the loaded media.
	/// </summary>
	/// <remarks>Might not be available for some types, like live streams.</remarks>
	TimeSpan Duration { get; set; }

	/// <summary>
	/// Occurs when the media has ended playing successfully.
	/// </summary>
	/// <remarks>This does not trigger when the media has failed during playback.</remarks>
	void MediaEnded();

	/// <summary>
	/// Occurs when the media has failed loading.
	/// </summary>
	/// <param name="args">Event arguments containing extra information about this event.</param>
	void MediaFailed(MediaFailedEventArgs args);

	/// <summary>
	/// Occurs when the media has been loaded and is ready to play.
	/// </summary>
	void MediaOpened();

	/// <summary>
	/// Pauses the currently playing media.
	/// </summary>
	void Pause();

	/// <summary>
	/// Starts playing the loaded media.
	/// </summary>
	void Play();

	/// <summary>
	/// The current position of the playing media.
	/// </summary>
	TimeSpan Position { get; internal set; }

	/// <summary>
	/// Occurs when the <see cref="Position"/> changes;
	/// </summary>
	event EventHandler<MediaPositionChangedEventArgs> PositionChanged;

	/// <summary>
	/// Occurs when a seek action has been completed.
	/// </summary>
	event EventHandler SeekCompleted;

	/// <summary>
	/// Seek to a specific position in the currently playing media.
	/// </summary>
	/// <param name="position">The requested position to seek to.</param>
	/// <remarks>If <paramref name="position"/> is outside of the range of the current media item, nothing will happen.</remarks>
	void SeekTo(TimeSpan position);

	/// <summary>
	/// Gets or sets whether the media should start playing as soon as it's loaded.
	/// </summary>
	bool ShouldAutoPlay { get; set; }

	/// <summary>
	/// Gets or sets if the media playback will restart from the beginning when it reaches the end.
	/// </summary>
	bool ShouldLoopPlayback { get; set; }

	/// <summary>
	/// Gets or sets if media playback will prevent the device display from going to sleep.
	/// </summary>
	/// <remarks>If media is paused, stopped or has completed playing, the display will turn off.</remarks>
	bool ShouldKeepScreenOn { get; set; }

	/// <summary>
	/// Gets or sets whether the player should show the platform playback controls.
	/// </summary>
	bool ShouldShowPlaybackControls { get; set; }

	/// <summary>
	/// Gets or sets the source of the media to play.
	/// </summary>
	MediaSource? Source { get; set; }

	/// <summary>
	/// Gets or sets the speed with which the media should be played.
	/// </summary>
	/// <remarks>A value of 1 means normal speed.
	/// Anything more than 1 is faster speed, anything less than 1 is slower speed.</remarks>
	double Speed { get; set; }

	/// <summary>
	/// Occurs when <see cref="CurrentState"/> changed.
	/// </summary>
	event EventHandler<MediaStateChangedEventArgs> StateChanged;

	/// <summary>
	/// Stops playing the currently playing media and resets the <see cref="Position"/>.
	/// </summary>
	void Stop();

	/// <summary>
	/// Gets the height (in pixels) of the loaded media in pixels.
	/// </summary>
	/// <remarks>Not reported for non-visual media.</remarks>
	int MediaHeight { get; }

	/// <summary>
	/// Gets the width (in pixels) of the loaded media in pixels.
	/// </summary>
	/// <remarks>Not reported for non-visual media.</remarks>
	int MediaWidth { get; }

	/// <summary>
	/// Gets or sets the volume of the audio for the media.
	/// </summary>
	/// <remarks>A value of 1 means full volume, 0 is silence.</remarks>
	double Volume { get; set; }
}
```

Android implementation is based on ExoPlayer (and the wonderful [ExoPlayer bindings](https://github.com/Baseflow/ExoPlayerXamarin) package), iOS uses `AVPlayer` and Windows uses WinUI `MediaPlayer`. Tizen support should follow soon.

I should make a list of the exact formats that are supported, but from what I tested a good amount works including HLS and other livestream formats.

All this goodness is implemented as a separate NuGet package!

 ### Linked Issues ###

 - Fixes #113 
 - Fixes #717 

 ### PR Checklist ###
 - [ ] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/195


samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -58,10 +58,19 @@
     <ProjectReference Include="..\..\src\CommunityToolkit.Maui.SourceGenerators\CommunityToolkit.Maui.SourceGenerators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
     <ProjectReference Include="..\..\src\CommunityToolkit.Maui.Analyzers\CommunityToolkit.Maui.Analyzers.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
     <ProjectReference Include="..\..\src\CommunityToolkit.Maui.Analyzers.CodeFixes\CommunityToolkit.Maui.Analyzers.CodeFixes.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="..\..\src\CommunityToolkit.Maui.MediaElement\CommunityToolkit.Maui.MediaElement.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
VladislavAntonyuk, line 64: Please remove

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
```diff
@@ -0,0 +1,22 @@
...
+	/// <param name="builder"><see cref="MauiAppBuilder"/> generated by <see cref="MauiApp"/> </param>
+	/// <returns><see cref="MauiAppBuilder"/> initialized for <see cref="CommunityToolkit.Maui.MediaElement"/>.</returns>
+	public static MauiAppBuilder UseMauiCommunityToolkitMediaElement(this MauiAppBuilder builder)
+	{
+		builder.ConfigureMauiHandlers(h =>
+		{
+			h.AddHandler<MediaElement, MediaElementHandler>();
+		});
+
+		return builder;
```
VladislavAntonyuk, line 20: Do we need to call Core library initializer here?

src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs
```diff
@@ -0,0 +1,41 @@
...
+	/// <summary>
+	/// The current position of the playing media.
+	/// </summary>
+	TimeSpan Position { get; set; }
+
+	/// <summary>
+	/// The total duration of the loaded media.
+	/// </summary>
+	/// <remarks>Might not be available for some types, like live streams.</remarks>
+	TimeSpan Duration { get; }
```
VladislavAntonyuk, line 17: What do you think if we remove it from this interface?
To follow interface segregation principle. 
We may have ILiveStreamMediaElement and IFileSourceMediaElement.

src/CommunityToolkit.Maui.MediaElement/MediaSource/StreamMediaSource.shared.cs
```diff
@@ -0,0 +1,135 @@
...
+	protected void OnLoadingCompleted(bool cancelled)
+	{
+		if (!IsLoading || completionSource == null)
+		{
+			return;
+		}
+
+		var tcs = Interlocked.Exchange<TaskCompletionSource<bool>?>(ref completionSource, null);
+
+		if (tcs != null)
```
VladislavAntonyuk, line 95: Let's use is/is not null

src/CommunityToolkit.Maui.MediaElement/PlatformView/MauiMediaElement.macios.cs
```diff
@@ -0,0 +1,254 @@
...
+				asset = AVUrlAsset.Create(nsUrl);
+			}
+			else
+			{
+				throw new InvalidOperationException($"{nameof(uriSource.Uri)} is not initialized");
+			}
+		}
+		else
+		{
+			if (mediaElement?.Source is FileMediaSource fileSource)
```
VladislavAntonyuk, line 73: From the latest PR in XCT by Pedro: we also need to check if file is null

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
jfversluis: I don't think so...? It doesn't even have a dependency on Core at this point.

anvuks:
Can one of the maintainers take a look at this PR for Android MediaElement: https://github.com/xamarin/XamarinCommunityToolkit/pull/1913 before you port it to Maui toolkit? It's a fix for a critical bug that both xamarin and maui users can benefit from.

I'm also willing to fix this issue https://github.com/xamarin/XamarinCommunityToolkit/issues/408 (only for Android), but I am waiting for my previous PR to get merged to avoid annoying merge conflicts. If someone can fix it for iOS too, we can close that issue.

pictos:
@anvuks don't worry about that, the implementation will be a new code here (if I'm not wrong), so there's no chance of conflicts.
I'll take a look to your PR soon-sh, don't have the time to checkout and test on my end yet.

src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
```diff
@@ -0,0 +1,24 @@
...
+
+public partial class MediaElement : View, IMediaElement
+{
+	public MediaElement()
+	{
+	}
+
+	public MediaSource? Source { get; set; }
+
+	public TimeSpan Position { get; set; }
```
Mephisztoe: I am not convinced that the base class "MediaElement" should provide properties bond to files such as Position. When playing a stream for instance, this property is kinda useless. Same might go for duration when there is a live stream which does not have a duration. I'd vote for seperating these into different abstractions or at least interface implementations.

jfversluis: Thanks for the feedback! The API is still in flux, it's marked as a draft still :)

jfversluis: Also, I'm not really sure if that's true. If you look at a live stream on YouTube, I can set the position back in time to still watch from that point on, so it might be useful for that just as well. Duration might not be, this was also already pointed out above.

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.android.cs
```diff
@@ -0,0 +1,255 @@
...
+	public void OnPlayerStateChanged(bool playWhenReady, int playbackState)
+	{
+		if (player is null)
+		{
+			return;
+		}
+
+		if (playbackState == IPlayer.StateReady)
+		{
+			mediaElement.Duration = TimeSpan.FromMilliseconds(player.Duration);
```
Mephisztoe, line 213: Might be causing exceptions on specific types of streams. I am still investigating, but this is an error I received when trying to play an URL on Android: **System.OverflowException:** 'TimeSpan overflowed because the duration is too long.'. It could be possibly related to what people are discussing here: https://stackoverflow.com/questions/35298125/exoplayer-getduration

Mephisztoe:
I am not 100% sure yet - still testing - but could anyone of you please have a look at the MediaElementState implementation for Android? 

Here's the thing:
I am using an ImageButton that shows a Pause icon when the stream is currently playing and a Play icon when it is paused. I implemented that behaviour with DataTriggers like this:

```xaml
<ImageButton Grid.Row="2"
  BindingContext="{x:Reference mediaElement}"
  x:Name="playButton"     
  Clicked="OnPlayPauseButtonClicked"
  HeightRequest="64"
  HorizontalOptions="CenterAndExpand"
  Source="player_pause.png"
  VerticalOptions="Center"
  WidthRequest="64">
  <ImageButton.Triggers>
    <DataTrigger
      Binding="{Binding CurrentState}"
      TargetType="ImageButton"
      Value="{x:Static media:MediaElementState.Playing}">
      <Setter Property="Source" Value="player_pause.png" />
    </DataTrigger>                  
    <DataTrigger
      Binding="{Binding CurrentState}"
      TargetType="ImageButton"
      Value="{x:Static media:MediaElementState.Paused}">
      <Setter Property="Source" Value="player_play.png" />
    </DataTrigger>                   
  </ImageButton.Triggers>
</ImageButton>
```

This works amazingly well on iOS.
On Android however, the State seems to be always "Playing".

I looked at the code, and the Method PlatformUpdateStatus (MediaManager.android.cs) checks the players' PlaybackState which is always PlaybackStateCode.Playing (value 3) even when its paused. The funny thing is, I tried to get a deeper understanding of how ExoPlayer works internally and this state _seems_ to actually refer to its inner Player.State which is like "STATE_READY" for play all the time... which is - coincidantally the value 3.

Maybe I am mixing something up here. All I can say that there's something wrong and I didn't figure it out on my own (yet).

jfversluis:
Please remember that this is work in progress. One of the things that is very much in progress is the reporting of the states, so please don't focus on that as it will probably drastically change.

src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs
jfversluis: @VladislavAntonyuk how would you like to see this implemented?

src/CommunityToolkit.Maui.MediaElement/EventArgs/MediaFailedEventArgs.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace CommunityToolkit.Maui.MediaElement;
+
+public class MediaFailedEventArgs : EventArgs
+{
+	/// <summary>
+	/// Gets a description of why the media failed to load and/or play.
+	/// </summary>
+	public string ErrorMessage { get; } = string.Empty;
+
+	internal MediaFailedEventArgs(string errorMessage)
```
pictos, line 10: Do we want this to be `internal`? IMO this can be bad in terms of extensibility, if devs need to add more functions to their media element control.

brminnick: (I haven't reviewed the PR yet, just chiming in after seeing this notification)

If we do decide to not make the constructor public, I recommend changing `internal` to [`protected private`](https://learn.microsoft.com/dotnet/csharp/language-reference/keywords/private-protected) like we did for [`BaseCommand` in XCT](https://github.com/xamarin/XamarinCommunityToolkit/blob/b55322ac95e863347cc25a9b029b3a812e98682e/src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Internals/BaseCommand.shared.cs#L25-L29).

#### Background story

In XCT, I created `BaseCommand` to share logic between `AsyncCommand` and `AsyncValueCommand`, and I had to make `BaseCommand` a `public class` because `AsyncCommand` and `AsyncValueCommand` are both `public`. 

But, I didn't want us supporting `BaseCommand` as a public API because that is outside the scope of the `AsyncCommand` + `AsyncValueCommand` APIs. So I used `protected private` + [`[EditorBrowsable(EditorBrowsableState.Never)]`](https://github.com/xamarin/XamarinCommunityToolkit/blob/b55322ac95e863347cc25a9b029b3a812e98682e/src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Internals/BaseCommand.shared.cs#L12) to allow us to inherit from it internally, but it prevents devs from consuming it externally.

jfversluis: Yeah that all makes sense. I tend to work as most closed by default. Anything that is public is something we need to maintain and I'd rather be convinced after the fact on why someone wants this than to make it public and later on have to make a breaking change.

But this one makes sense and doesn't seem too bad. Thanks!

src/CommunityToolkit.Maui.MediaElement/CommunityToolkit.Maui.MediaElement.csproj
```diff
@@ -0,0 +1,53 @@
...
+
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">14.2</SupportedOSPlatformVersion>
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'maccatalyst'">14.0</SupportedOSPlatformVersion>
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">21.0</SupportedOSPlatformVersion>
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.17763.0</SupportedOSPlatformVersion>
+		<TargetPlatformMinVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.17763.0</TargetPlatformMinVersion>
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'tizen'">6.5</SupportedOSPlatformVersion>
+	</PropertyGroup>
+
+	<ItemGroup Condition="$(TargetFramework.StartsWith('Xamarin.iOS')) != true AND $(TargetFramework.StartsWith('net6.0-ios')) != true AND $(TargetFramework.StartsWith('net6.0-maccatalyst')) != true ">
```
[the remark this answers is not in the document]
jfversluis: Good call! The problem right now is that MediaElement is still on .NET 6 and the rest on .NET 7. But will see if I can update MediaElement to .NET 7 as well.

src/CommunityToolkit.Maui.MediaElement/EventArgs/SeekRequestedEventArgs.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace CommunityToolkit.Maui.MediaElement;
+
+class SeekRequestedEventArgs
```
pictos, line 3:
```suggestion
class SeekRequestedEventArgs : EventArgs
```

jfversluis: wth how did that happen 😂

brminnick (changes requested):
Thanks Gerald! This is looking really good!!

We'll need to add some Unit Tests. Since `CommunityToolkit.Maui.MediaElement` lives in its own project, let's create `CommunityToolkit.Maui.MediaElement.UnitTests` and also update the CI pipeline to run these new tests! 

And since we're adding `.UseMauiCommunityToolkitMediaElement()`, let's also add `CommunityToolkit.Maui.MediaElement.Analyzers` and `CommunityToolkit.Maui.MediaElement.CodeFixes` to generate a build error and its recommended intellisense fix when a user forgets to add it. You can basically copy/paste `CommunityToolkit.Maui.Analyzers` and `CommunityToolkit.Maui.CodeFixes`.

I also updated this PR to include Tizen which is causing the CI build to fail. This'll ensure that we don't accidentally ship it without Tizen support 👍.

I'll add the `do not merge` label for now until we've added the missing items:
- [ ] Tizen Support
- [ ] `CommunityToolkit.Maui.MediaElement.UnitTests`
- [x] `CommunityToolkit.Maui.MediaElement.Analyzers`
- [ ] `CommunityToolkit.Maui.MediaElement.CodeFixes`

I only tested the sample on iOS, but did find some bugs:
- On iOS, the `Load Local Reource` button in the sample app throws an Exception
- On iOS, the `Show playback controls` switch doesn't seem to change anything 
- On iOS, when the video loops, the `Position` value is not reset; it instead continues to count up:
    image

samples/CommunityToolkit.Maui.Sample/Pages/Views/MediaElement/PositionSlider.cs
```diff
@@ -0,0 +1,51 @@
+﻿namespace CommunityToolkit.Maui.Sample.Pages.Views;
+
+/// <summary>
+/// Added a custom slider as a workaround for a bug in the regular Slider. See: https://github.com/dotnet/maui/issues/12285
+/// Original implementation by David Britch: https://github.com/davidbritch/dotnet-maui-videoplayer/blob/main/src/VideoDemos/Controls/PositionSlider.cs
+/// </summary>
+public class PositionSlider : Slider
```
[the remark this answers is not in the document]
mhrastegari: I believe we had a TimeSpanTo Double converter in XCT! Can't we just use it here? because we used that converter with MCT's media element in our [project](https://github.com/mhrastegari/fx-files/blob/main/src/Client/App/Views/NativeVideoViewer.xaml#L71) and it worked perfectly with the regular Slider and there was no need to create a custom control and complexing stuff.

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
```diff
@@ -38,6 +40,7 @@ public static MauiApp CreateMauiApp()
 								})
 #endif
 								.UseMauiCommunityToolkitMarkup()
+								.UseMauiCommunityToolkitMediaElement()
```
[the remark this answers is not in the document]
pictos: @jfversluis if you can't implement it, please open an issue and assign it to me, then I can implement it

src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
```diff
@@ -0,0 +1,502 @@
...
+	internal void OnMediaOpened()
+	{
+		InitializeTimer();
+		eventManager.HandleEvent(this, EventArgs.Empty, nameof(MediaOpened));
+	}
+
+	/// <inheritdoc/>
+	protected override void OnBindingContextChanged()
+	{
+		if (Source != null)
```
pictos, line 368:
```suggestion
		if (Source is not null)
```

src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
```diff
@@ -0,0 +1,502 @@
...
+		{
+			SetInheritedBindingContext(Source, BindingContext);
+		}
+
+		base.OnBindingContextChanged();
+	}
+
+	static void OnAutoPlayPropertyChanged(BindableObject bindable, object oldValue, object newValue) =>
```
pictos, line 376: Just out of curiosity, why we aren't using mapper for those properties instead?

src/CommunityToolkit.Maui.MediaElement/MediaSource/MediaSource.shared.cs
```diff
@@ -0,0 +1,85 @@
...
+	{
+		add => weakEventManager.AddEventHandler(value);
+		remove => weakEventManager.RemoveEventHandler(value);
+	}
+
+	/// <summary>
+	/// An implicit operator to convert a string value into a <see cref="MediaSource"/>.
+	/// </summary>
+	/// <param name="source">Full path to a local file (starting with <c>file://</c>) or an absolute URI.</param>
+	[Preserve(Conditional = true)]
```
pictos, line 25: Not sure if this has an effect on the `One .NET` world since this is from Mono.
Anyway, since our lib isn't linker-safe we can remove it

src/CommunityToolkit.Maui.MediaElement/MediaSource/MediaSource.shared.cs
```diff
@@ -0,0 +1,85 @@
...
+	public static implicit operator MediaSource?(string? source) =>
+		Uri.TryCreate(source, UriKind.Absolute, out var uri) && uri.Scheme != "file"
+			? FromUri(uri)
+			: FromFile(source);
+
+	/// <summary>
+	/// An implicit operator to convert a <see cref="Uri"/> object into a <see cref="UriMediaSource"/>.
+	/// </summary>
+	/// <param name="uri">Absolute URI to load.</param>
+	[Preserve(Conditional = true)]
```
pictos, line 35:
```suggestion
```

src/CommunityToolkit.Maui.MediaElement/MediaSource/MediaSource.shared.cs
pictos:
```suggestion
```

src/CommunityToolkit.Maui.MediaElement/MediaSource/UriMediaSource.shared.cs
```diff
@@ -0,0 +1,49 @@
...
+	public static implicit operator UriMediaSource?(string uri) => (UriMediaSource?)FromUri(uri);
+
+	/// <summary>
+	/// An implicit operator to convert a <see cref="UriMediaSource"/> into a string value.
+	/// </summary>
+	/// <param name="uriMediaSource">A <see cref="UriMediaSource"/> instance to convert to a string value.</param>
+	public static implicit operator string?(UriMediaSource? uriMediaSource) => uriMediaSource?.Uri?.ToString();
+
+	static bool UriValueValidator(BindableObject bindable, object value) =>
+		value == null || ((Uri)value).IsAbsoluteUri;
```
pictos, line 45:
```suggestion
		value is null || ((Uri)value).IsAbsoluteUri;
```

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.macios.cs
```diff
@@ -0,0 +1,459 @@
...
+	NSObject? itemFailedToPlayToEndTimeObserver;
+	NSObject? playbackStalledObserver;
+	NSObject? errorObserver;
+	IDisposable? statusObserver;
+	IDisposable? volumeObserver;
+	IDisposable? rateObserver;
+	IDisposable? timeControlStatusObserver;
+	IDisposable? currentItemErrorObserver;
+	AVPlayerViewController? playerViewController;
+	AVPlayerItem? playerItem;
```
pictos, line 25: having those `private` will block devs to customize the MediaManager, I suggest to change it to `protected`

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.macios.cs
```diff
@@ -0,0 +1,459 @@
...
+using CoreFoundation;
+using CoreMedia;
+using Foundation;
+using Microsoft.Extensions.Logging;
+
+namespace CommunityToolkit.Maui.MediaElement;
+
+public partial class MediaManager : IDisposable
+{
+	const NSKeyValueObservingOptions valueObserverOptions =
```
pictos, line 12:
```suggestion
	protected const NSKeyValueObservingOptions valueObserverOptions =
```

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.macios.cs
```diff
@@ -0,0 +1,459 @@
...
+		}
+
+		currentItemErrorObserver?.Dispose();
+
+		player?.ReplaceCurrentItemWithPlayerItem(playerItem);
+
+		currentItemErrorObserver = playerItem?.AddObserver("error",
+			valueObserverOptions, (NSObservedChange change) =>
+		{
+			if (player?.CurrentItem?.Error != null)
```
pictos, line 162: can we invert this `if`?

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.macios.cs
```diff
@@ -0,0 +1,459 @@
...
+
+		mediaElement.CurrentStateChanged(newState);
+	}
+
+	void ErrorOccurred(object? sender, NSNotificationEventArgs args)
+	{
+		string message;
+
+		var error = player?.CurrentItem?.Error;
+		if (error != null)
```
pictos, line 404:
```suggestion
		if (error is not null)
```

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.shared.cs
```diff
@@ -0,0 +1,207 @@
...
+
+/// <summary>
+/// An class that acts as a manager for an <see cref="IMediaElement"/> instance.
+/// </summary>
+public partial class MediaManager
+{
+	readonly IMauiContext mauiContext;
+	readonly IMediaElement mediaElement;
+#if ANDROID || IOS || MACCATALYST || WINDOWS || TIZEN
+	PlatformMediaView? player;
```
pictos, line 23: Having those as `private` will hide from devs that want to customize the Manager, I suggest to change them to `protected`

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.windows.cs
```diff
@@ -0,0 +1,321 @@
...
+
+	/// <summary>
+	/// Creates the corresponding platform view of <see cref="MediaElement"/> on Windows.
+	/// </summary>
+	/// <returns>The platform native counterpart of <see cref="MediaElement"/>.</returns>
+	public PlatformMediaView CreatePlatformView()
+	{
+		player = new();
+		MediaPlayer mediaPlayer = new();
+		mediaPlayer.MediaOpened += MediaPlayer_MediaOpened;
```
pictos, line 24:
```suggestion
		mediaPlayer.MediaOpened += OnMediaPlayerMediaOpened;
```

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.windows.cs
```diff
@@ -0,0 +1,321 @@
...
+
+		player.SetMediaPlayer(mediaPlayer);
+
+		player.MediaPlayer.PlaybackSession.PlaybackStateChanged += PlaybackSession_PlaybackStateChanged;
+		player.MediaPlayer.PlaybackSession.SeekCompleted += PlaybackSession_SeekCompleted;
+		player.MediaPlayer.MediaFailed += MediaPlayer_MediaFailed;
+		player.MediaPlayer.MediaEnded += MediaPlayer_MediaEnded;
+		player.MediaPlayer.VolumeChanged += MediaPlayer_VolumeChanged;
```
pictos, line 32:
```suggestion
		player.MediaPlayer.PlaybackSession.PlaybackStateChanged += OnPlaybackSessionPlaybackStateChanged;
		player.MediaPlayer.PlaybackSession.SeekCompleted += OnPlaybackSessionSeekCompleted;
		player.MediaPlayer.MediaFailed += OnMediaPlayerMediaFailed;
		player.MediaPlayer.MediaEnded += OnMediaPlayerMediaEnded;
		player.MediaPlayer.VolumeChanged += OnMediaPlayerVolumeChanged;
```

pictos (changes requested):
Wow that's huge!
I just looked into the code, and here are my suggestions.

JoonghyunCho:
I have my Tizen implementation for MediaElement now :)

AnthonyIacono:
Just wanted to drop a comment that I am unable to see the playback controls on iOS, even with `ShowsPlaybackControls` set to `True`. Thanks for all your hard work so far!

jfversluis:
> Just wanted to drop a comment that I am unable to see the playback controls on iOS, even with `ShowsPlaybackControls` set to `True`. Thanks for all your hard work so far!

What version of iOS? What version of this package?

AnthonyIacono:
> > Just wanted to drop a comment that I am unable to see the playback controls on iOS, even with `ShowsPlaybackControls` set to `True`. Thanks for all your hard work so far!
> 
> What version of iOS? What version of this package?

iOS 16.2
Using these packages:

```
        <PackageReference Include="CommunityToolkit.Maui" Version="[1.0.0-build-667.84928]" />
        <PackageReference Include="CommunityToolkit.Maui.MediaElement" Version="[1.0.0-build-667.84928]" />
```

jfversluis:
That's strange because on iOS 16.2 and the latest code in this PR it works fine for me.

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.android.cs
```diff
@@ -272,20 +272,20 @@ public void OnVolumeChanged(float volume)
 
 		player.PlayWhenReady = mediaElement.AutoPlay;
 
-		if (mediaElement.Source is UriMediaSource)
+		if (mediaElement.Source is UriMediaSource uriMediaSource)
 		{
-			string uri = (mediaElement.Source as UriMediaSource)!.Uri!.AbsoluteUri;
-			if (!string.IsNullOrWhiteSpace(uri))
+			var uri = uriMediaSource.Uri;
+			if (!string.IsNullOrWhiteSpace(uri?.AbsolutePath))
```
jfversluis, line 276: @brminnick you've replaced AbsoluteUri with AbsolutePath here and other places 😬

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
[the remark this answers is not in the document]
jfversluis: @pictos I think I did it! But please review what I have committed, I have no idea what I'm doing!

samples/CommunityToolkit.Maui.Sample/Pages/Views/MediaElement/PositionSlider.cs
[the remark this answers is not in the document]
jfversluis: Yeah I need to look into this again. I still wasn't really happy about it. Still didn't work 100% as intended.

AnthonyIacono:
> That's strange because on iOS 16.2 and the latest code in this PR it works fine for me.

Have you tried when the `MediaElement` is inside a `StackLayout`? For example:
```
<StackLayout Orientation="Vertical" VerticalOptions="FillAndExpand">
    <mediaElement:MediaElement x:Name="MediaElement" WidthRequest="300" HeightRequest="300" ShowsPlaybackControls="True" />
</StackLayout>
```

I am setting the `Source` in the code-behind file.

pictos:
@AnthonyIacono this looks like something that will demand more time... Can you please open an issue or a discussion with a repro project?

AnthonyIacono:
@pictos I could, but before I do I just want to make sure that is what @jfversluis wishes. I fear that if I create a separate issue, it might just make it harder to track the changes.

pictos:
I believe it will be easier to solve that on an `issue` than here. Where the flow and focus are on the code-review. But I'm ok to wait for Gerald's input

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -70,7 +76,7 @@
     <RuntimeIdentifiers>maccatalyst-arm64;maccatalyst-x64</RuntimeIdentifiers>
   </PropertyGroup>
   
-  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net7.0-ios|AnyCPU'">
-    <CreatePackage>false</CreatePackage>
-  </PropertyGroup>
+  <ItemGroup Condition="$(TargetFramework.Contains('-windows'))">
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.2.221209.1" />
```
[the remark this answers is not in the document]
jfversluis: Yeah in theory that is true. At the time of starting this PR 1.2 was still in preview and that was the first version that had the `MediaPlayer`. That's why I had to bring in the version manually. The question is if we want to be dependant on the version that MAUI is referencing of we just want to make sure we always get the right version also in regard to potential breaking changes and such in the future?

jfversluis:
@AnthonyIacono With just this the playback controls show fine for me

```xml
<VerticalStackLayout>
        <media:MediaElement
            WidthRequest="300"
            HeightRequest="300"
                x:Name="mediaElement"
                AutoPlay="True"
                Source="https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4" />

    </VerticalStackLayout>
```

Can you please provide a reproduction where you don't see the playback controls where they should be visible?

src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
```diff
@@ -0,0 +1,446 @@
+﻿using System.ComponentModel;
+using CommunityToolkit.Maui.MediaElement.Converters;
+
+namespace CommunityToolkit.Maui.MediaElement;
+
+/// <summary>
+/// Represents an object that to render audio and video to the display.
+/// </summary>
+public class MediaElement : View, IMediaElement
```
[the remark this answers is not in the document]
jfversluis: Good point! On the other hand, people might be used to the MediaElement name by now and also search for that. Added a poll in our (private) Discord for a vote on this

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+	public sealed override ImmutableArray<string> FixableDiagnosticIds { get; } = ImmutableArray.Create(UseCommunityToolkitMediaElementInitializationAnalyzer.DiagnosticId);
+
+	public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+	public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+	{
+		var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+		var diagnostic = context.Diagnostics.First();
```
Youssef1313, line 31:
```suggestion
		var diagnostic = context.Diagnostics[0];
```

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Rename;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace CommunityToolkit.Maui.Analyzers;
+
+[ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider)), Shared]
+public class UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider : CodeFixProvider
```
Youssef1313, line 21:
```suggestion
public sealed class UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider : CodeFixProvider
```

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+
+	public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+	{
+		var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+		var diagnostic = context.Diagnostics.First();
+		var diagnosticSpan = diagnostic.Location.SourceSpan;
+
+		// Find the type declaration identified by the diagnostic.
+		var declaration = root?.FindToken(diagnosticSpan.Start).Parent?.AncestorsAndSelf().OfType<InvocationExpressionSyntax>().Last() ?? throw new InvalidOperationException();
```
Youssef1313, line 35: fyi, `root` is never null.

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+				QualifiedName(
+					IdentifierName("CommunityToolkit"),
+					IdentifierName("Maui")));
+
+		var newRoot = root.ReplaceNode(invocationExpression, updatedInvocationExpression);
+
+		if (newRoot is CompilationUnitSyntax compilationSyntax)
+		{
+			newRoot = compilationSyntax.AddUsings(mauiCommunityToolkitUsingStatement).NormalizeWhitespace();
+		}
```
Youssef1313, line 70: - This will add unnecessary using if the `using` is already a global using in another tree.
- normalizing the whole tree isn't a good idea.

Best is to let Roslyn do that via `AddImportsAnnotation`.

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+			diagnostic);
+	}
+
+	static async Task<Document> AddUseCommunityToolkit(Document document, InvocationExpressionSyntax invocationExpression, CancellationToken cancellationToken)
+	{
+		var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+		if (root is null)
+		{
+			return document;
+		}
```
Youssef1313, line 52: root was already calculated previously. You only need to pass it along.

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
Youssef1313, line 1: Curious why this is not in the existing analyzers project?

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes.csproj
```diff
@@ -0,0 +1,23 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <IsRoslynComponent>true</IsRoslynComponent>
+    <RootNamespace>CommunityToolkit.Maui.MediaElement.Analyzers</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.4.0" PrivateAssets="all" />
```
Youssef1313, line 11: Is 4.4 a requirement? This will not work in Visual Studio versions before 17.4.

src/CommunityToolkit.Maui.MediaElement.Analyzers/UseCommunityToolkitMediaElementInitializationAnalyzer.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace CommunityToolkit.Maui.MediaElement.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class UseCommunityToolkitMediaElementInitializationAnalyzer : DiagnosticAnalyzer
```
Youssef1313, line 11:
```suggestion
public sealed class UseCommunityToolkitMediaElementInitializationAnalyzer : DiagnosticAnalyzer
```

src/CommunityToolkit.Maui.MediaElement.Analyzers/UseCommunityToolkitMediaElementInitializationAnalyzer.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+	public const string DiagnosticId = "MCTME001";
```
Youssef1313, line 13: Tests?

src/CommunityToolkit.Maui.MediaElement.Analyzers/UseCommunityToolkitMediaElementInitializationAnalyzer.cs
```diff
@@ -0,0 +1,88 @@
...
+
+	static readonly DiagnosticDescriptor rule = new(DiagnosticId, title, messageFormat, category, DiagnosticSeverity.Error, isEnabledByDefault: true, description: description);
+
+	public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(rule);
+
+	public override void Initialize(AnalysisContext context)
+	{
+		context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+		context.EnableConcurrentExecution();
+		context.RegisterSyntaxNodeAction(AnalyzeNode, SyntaxKind.ExpressionStatement);
```
Youssef1313, line 28: I don't think it's a good idea to write this as a syntax analyzer.

Consider writing using IOperation.

src/CommunityToolkit.Maui.MediaElement.Analyzers/UseCommunityToolkitMediaElementInitializationAnalyzer.cs
jfversluis: Looks like you opened an issue for that. Outside of the scope of this PR.

src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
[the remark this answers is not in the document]
jfversluis: The people have spoken. Renamed to MediaView.

src/CommunityToolkit.Maui.MediaView.Analyzers/CommunityToolkit.Maui.MediaView.Analyzers.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IsPackable>false</IsPackable>
```
Youssef1313, line 5: I don't quite see how this ships to customers. Is there a missing packaging project?

The same applies to the existing `CommunityToolkit.Maui.Analyzers.csproj`

jfversluis: The assemblies are built in the pipeline and referenced as binaries and packed up through the other NuGet packable projects.

src/CommunityToolkit.Maui.MediaElement.Analyzers/UseCommunityToolkitMediaElementInitializationAnalyzer.cs
Youssef1313: The issue is for another analyzer (yes, I know they are very similar).

src/CommunityToolkit.Maui.MediaView.Analyzers/CommunityToolkit.Maui.MediaView.Analyzers.csproj
Youssef1313: @jfversluis Could you elaborate more if you don't mind?

All what I see in the CI pipeline is `dotnet build -c Release`, when/how exactly they are packed?

Youssef1313: Oh I see now.

```xml
    <None Include="..\CommunityToolkit.Maui.MediaView.Analyzers\bin\$(Configuration)\netstandard2.0\CommunityToolkit.Maui.MediaView.Analyzers.dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
    <None Include="..\CommunityToolkit.Maui.MediaView.Analyzers.CodeFixes\bin\$(Configuration)\netstandard2.0\CommunityToolkit.Maui.MediaView.Analyzers.CodeFixes.dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
```

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes.csproj
jfversluis: Tbh I think 17.4 is fine, if there is an issue, let's open a discussion and talk about it there. There is probably a reason for it.

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider.cs
jfversluis: Great suggestions, thanks. This will then be the same for the regular CodeFix project. Let's open an issue, or even better, a PR, for these fixes once this is merged.

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider.cs
jfversluis: Great suggestions, thanks. This will then be the same for the regular CodeFix project. Let's open an issue, or even better, a PR, for these fixes once this is merged.

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/UseCommunityToolkitMediaElementInitializationAnalyzerCodeFixProvider.cs
jfversluis: Great suggestions, thanks. This will then be the same for the regular CodeFix project. Let's open an issue, or even better, a PR, for these fixes once this is merged.

src/CommunityToolkit.Maui.MediaElement.Analyzers/UseCommunityToolkitMediaElementInitializationAnalyzer.cs
jfversluis: Great suggestions, thanks. This will then be the same for the regular CodeFix project. Let's open an issue, or even better, a PR, for these fixes once this is merged.

src/CommunityToolkit.Maui.MediaElement.Analyzers/UseCommunityToolkitMediaElementInitializationAnalyzer.cs
jfversluis: Great suggestions, thanks. This will then be the same for the regular CodeFix project. Let's open an issue, or even better, a PR, for these fixes once this is merged.

src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs
```diff
@@ -0,0 +1,134 @@
...
+	/// <summary>
+	/// Stops playing the currently playing media and resets the <see cref="Position"/>.
+	/// </summary>
+	void Stop();
+
+	/// <summary>
+	/// Triggers a <see cref="CurrentState"/> change.
+	/// </summary>
+	/// <param name="newState">The new state the <see cref="MediaElement"/> transitioned to.</param>
+	internal void CurrentStateChanged(MediaElementState newState);
```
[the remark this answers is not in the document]
jfversluis: The alternative is referencing the concrete implementation in the MediaManager which is not what we want?

src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
jfversluis: Old habits die hard...

src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes.csproj
jfversluis: Not sure I follow, this is also in the non-MediaView analyzers project

src/CommunityToolkit.Maui.MediaElement/Converters/MediaSourceConverter.cs
```diff
@@ -0,0 +1,45 @@
...
+	public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value)
+	{
+		var valueAsString = value.ToString();
+
+		if (string.IsNullOrWhiteSpace(valueAsString))
+		{
+			return null;
+		}
+
+		// TODO smarter detection of filesystem files/resource file/url/other
```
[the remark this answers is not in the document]
jfversluis: Removed the todo and added this for now: https://github.com/CommunityToolkit/Maui/pull/667/commits/f64b7fbeace8542d8c0eb6fc498500e56058be2d

How does that look?

src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
```diff
@@ -0,0 +1,502 @@
...
+	public static readonly BindableProperty VolumeProperty =
+		  BindableProperty.Create(nameof(Volume), typeof(double), typeof(MediaElement), 1.0,
+			  BindingMode.TwoWay, new BindableProperty.ValidateValueDelegate(ValidateVolume));
+
+	/// <summary>
+	/// Initializes a new instance of the <see cref="MediaElement"/> class.
+	/// </summary>
+	public MediaElement()
+	{
+		InitializeTimer();
```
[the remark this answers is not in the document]
jfversluis: Removed

samples/CommunityToolkit.Maui.Sample/Pages/Views/MediaElement/PositionSlider.cs
```diff
@@ -0,0 +1,51 @@
...
+		set { SetValue(PositionProperty, value); }
+	}
+
+	public PositionSlider()
+	{
+		PropertyChanged += (sender, args) =>
+		{
+			if (args.PropertyName == nameof(Value))
+			{
+				TimeSpan newPosition = TimeSpan.FromSeconds(Value);
```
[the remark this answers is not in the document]
jfversluis: Removed the `PositionSlider` entirely

src/CommunityToolkit.Maui.MediaPlayer/CommunityToolkit.Maui.MediaPlayer.csproj
```diff
@@ -0,0 +1,70 @@
...
+
+	<PropertyGroup>
+		<TargetFrameworks>net7.0;net7.0-android;net7.0-ios;net7.0-maccatalyst</TargetFrameworks>
+		<TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);net7.0-windows10.0.19041.0</TargetFrameworks>
+		<TargetFrameworks Condition="'$(IncludeTizenTargetFrameworks)' == 'true'">$(TargetFrameworks);net7.0-tizen</TargetFrameworks> 
+		<UseMaui>true</UseMaui>
+		<SingleProject>true</SingleProject>
+		<ImplicitUsings>enable</ImplicitUsings>
+    
+		<SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">14.2</SupportedOSPlatformVersion>
```
VladislavAntonyuk, line 11: Is it correct? In Core project we use version 11.0

ne0rrmatrix:
in toolkit:MediaPlayer ShouldShowPlaybackControls="True" blocks swipe gestures. Disabling playback controls remediates the issue.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
```diff
@@ -0,0 +1,324 @@
...
+using WinMediaSource = Windows.Media.Core.MediaSource;
+using WindowsMediaElement = Windows.Media.Playback.MediaPlayer;
+using CommunityToolkit.Maui.Core.Primitives;
+using CommunityToolkit.Maui.Views;
+
+namespace CommunityToolkit.Maui.Core.Views;
+
+partial class MediaManager : IDisposable
+{
+	readonly DisplayRequest displayRequest = new();
```
pictos, line 14:
```suggestion
	protected readonly DisplayRequest displayRequest = new();
```

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
```diff
@@ -0,0 +1,324 @@
...
+
+	// The requests to keep display active are cumulative, this bool makes sure it only gets requested once
+	bool displayActiveRequested;
```
pictos, line 17: not sure if this one should be `protected` or not... Thoughts?

src/CommunityToolkit.Maui.MediaElement/readme.txt
```diff
@@ -0,0 +1,27 @@
...
+In order to use the .NET MAUI Community Toolkit MediaElement you need to call the extension method in your `MauiProgram.cs` file as follows:
+
+```csharp
+using CommunityToolkit.Maui.MediaElement;
+
+public static MauiApp CreateMauiApp()
+{
+    var builder = MauiApp.CreateBuilder();
+
+    // Initialise the toolkit
```
pictos, line 14: can you update the readme following our updated one?
You can find it [here](https://github.com/CommunityToolkit/Maui#getting-started)

pictos (commented):
I've a couple of suggestions and questions... But this looks just fantastic❣️

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
```diff
@@ -0,0 +1,423 @@
...
+		mediaElement?.SeekCompleted();
+	}
+
+	/// <summary>
+	/// Occurs when ExoPlayer changes volume.
+	/// </summary>
+	/// <param name="volume">The new value for volume.</param>
+	/// <remarks>
+	/// This is part of the <see cref="IPlayer.IListener"/> implementation.
+	/// While this method does not seem to have any references, it's invoked at runtime.
```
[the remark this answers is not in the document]
jfversluis: So the reason I did this is because these actually do something so they're more discoverable. The ones that don't do anything but we need to have them because else Android will break are in the region below. I though this made most sense so that people don't spend time looking for the methods that actually have functionality and I added a comment to make it more clear 😅 I think I'll leave it like this for now, if it gets in the way or anything let's change it then.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
jfversluis: Hm yeah not sure about this one since I think it would be good to be in control here as the calls really need to match up. Let's keep it like this and if people do need it we can open it up then.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.tizen.cs
```diff
@@ -0,0 +1,340 @@
...
+	/// </summary>
+	/// <returns>The platform native counterpart of <see cref="MediaElement"/>.</returns>
+	public VideoView CreatePlatformView()
+	{
+		videoView = new VideoView()
+		{
+			WidthSpecification = LayoutParamPolicies.MatchParent,
+			HeightSpecification = LayoutParamPolicies.MatchParent
+		};
+		videoView.AddedToWindow += (s, e) =>
```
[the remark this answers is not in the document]
jfversluis: Is this what you mean?

```csharp
videoView.AddedToWindow += AddedToWindow;


void AddedToWindow(object? sender, EventArgs e)
{
	if (!isPlayerInitialized)
	{
		InitializePlayer(videoView);
		isPlayerInitialized = true;

		videoView.AddedToWindow -= AddedToWindow;
	}
}
```

jfversluis:
> in toolkit:MediaPlayer ShouldShowPlaybackControls="True" blocks swipe gestures. Disabling playback controls remediates the issue.

On which platform? Does that also happen on the platform without using .NET MAUI?

ne0rrmatrix:
> > in toolkit:MediaPlayer ShouldShowPlaybackControls="True" blocks swipe gestures. Disabling playback controls remediates the issue.
> 
> On which platform? Does that also happen on the platform without using .NET MAUI?

Android only. Works as expected in Windows.

jfversluis:
@ne0rrmatrix What I would be interested in is if this also happens on a Android project without .NET MAUI. I would assume it does. I can totally see how showing these controls will impact gestures. Either way, this will not be fixed with this merge. I can have a look at it later.

ne0rrmatrix:
> @ne0rrmatrix What I would be interested in is if this also happens on a Android project without .NET MAUI. I would assume it does. I can totally see how showing these controls will impact gestures. Either way, this will not be fixed with this merge. I can have a look at it later.

Sample repo to reproduce issue with comments showing mitigations.
[https://github.com/ne0rrmatrix/MediaPlayerBugSwipeGestures](url)
Sounds good. I have a minimal sample repo for this exact problem. I will investigate further. BTW how do we test this without using maui?

jfversluis:
@ne0rrmatrix make sure to open an issue for it because we won't be monitoring this PR after it's done.

You can create a .NET 6 Android app by doing `dotnet new android` and you'll want to use the ExoPlayer bindings that I'm also using: https://github.com/Baseflow/ExoPlayerXamarin

ne0rrmatrix:
> @ne0rrmatrix make sure to open an issue for it because we won't be monitoring this PR after it's done.
> 
> You can create a .NET 6 Android app by doing `dotnet new android` and you'll want to use the ExoPlayer bindings that I'm also using: https://github.com/Baseflow/ExoPlayerXamarin
 
I have started an issue so it can be tracked. TY for the suggestion. I have created an android app and added the exoplayer nuget package. I have the video player working. But now I need to add a main page and then navigate to player. Then add swipe gesture to test. That should keep me busy till after it gets patched. TY for the help so far. I enjoy the challenge. I am learning a lot and I will do my best to verify if the issue exists in android with just exoplayer and xamarin and nothing else. Now I have to go learn xamarin enough to do the test.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.tizen.cs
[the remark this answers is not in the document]
pictos: yes, but a small change

```csharp
videoView.AddedToWindow += AddedToWindow;


void AddedToWindow(object? sender, EventArgs e)
{
	if (!isPlayerInitialized)
	{
		InitializePlayer(videoView);
		isPlayerInitialized = true;
	}

          videoView.AddedToWindow -= AddedToWindow;
}
```

pictos (approved):

ne0rrmatrix:
> > @ne0rrmatrix make sure to open an issue for it because we won't be monitoring this PR after it's done.
> > You can create a .NET 6 Android app by doing `dotnet new android` and you'll want to use the ExoPlayer bindings that I'm also using: https://github.com/Baseflow/ExoPlayerXamarin
> 
> I have started an issue so it can be tracked. TY for the suggestion. I have created an android app and added the exoplayer nuget package. I have the video player working. But now I need to add a main page and then navigate to player. Then add swipe gesture to test. That should keep me busy till after it gets patched. TY for the help so far. I enjoy the challenge. I am learning a lot and I will do my best to verify if the issue exists in android with just exoplayer and xamarin and nothing else. Now I have to go learn xamarin enough to do the test.



> @ne0rrmatrix make sure to open an issue for it because we won't be monitoring this PR after it's done.
> 
> You can create a .NET 6 Android app by doing `dotnet new android` and you'll want to use the ExoPlayer bindings that I'm also using: https://github.com/Baseflow/ExoPlayerXamarin

I checked the ExoPlayer github and gestures was listed as a closed issue. It is blocked by player controls and this is intended behavior. The only way for it to work would be to add a custom handler to implement it. I am thinking of just adjusting how my app works to deal with this.  The media player that you are merging is not bugged. It is working as intended in android. I just wish i knew enough to implement the change myself. It is a new project for me now. I have to learn enough to implement the feature myself or hope someone here does it for me. I think doing it myself will be a good project for me to learn new things.

jfversluis:
@ne0rrmatrix awesome investigation, thank you so much and that was kind of what I expected. Thank you for letting me know!

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaPlayer.cs
```diff
@@ -0,0 +1,137 @@
+﻿using CommunityToolkit.Maui.Core.Primitives;
+using CommunityToolkit.Maui.Views;
+
+namespace CommunityToolkit.Maui.Core;
+
+/// <summary>
+/// With MediaElement you can play multimedia inside of your app.
+/// </summary>
+public interface IMediaElement : IView
```
VladislavAntonyuk, line 9: We need to rename the file to IMediaElement

Phantom-KNA:
Custom Headers
Hello this is my fist time, this is my proporsal for custom headers request in Android.
[MediaElement.shared.cs](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs)
`/// <summary>
	/// Backing store for the <see cref="CustomHeaders"/> property.
	/// </summary>
	public static readonly BindableProperty CustomHeadersProperty =
		  BindableProperty.Create(nameof(CustomHeaders), typeof(Dictionary<string, string>), typeof(MediaElement), null);`

`/// <inheritdoc cref="IMediaElement.CustomHeaders"/>
	public Dictionary<string, string> CustomHeaders
	{
		get => (Dictionary<string, string>)GetValue(CustomHeadersProperty);
		set => SetValue(CustomHeadersProperty, value);
	}`

[IMediaElement.cs](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs)

`/// <summary>
	/// Gets or sets headers custom headers
	/// </summary>
	Dictionary<string,string> CustomHeaders { get; set; }`

[MauiMediaElement.android.cs](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs)

`using Com.Google.Android.Exoplayer2.Source;
 using Com.Google.Android.Exoplayer2.Upstream;`

`if (MediaElement.Source is UriMediaSource uriMediaSource)
		{
			var uri = uriMediaSource.Uri;
			if (!string.IsNullOrWhiteSpace(uri?.AbsoluteUri))
			{
				if (MediaElement.CustomHeaders != null)
				{
					var httpDataSoureFactory = new DefaultHttpDataSource.Factory();
					httpDataSoureFactory.SetDefaultRequestProperties(MediaElement.CustomHeaders);
					httpDataSoureFactory.SetUserAgent("Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/115.0");
					var mediaSource = new DefaultMediaSourceFactory(httpDataSoureFactory).CreateMediaSource(MediaItem.FromUri(uri?.AbsoluteUri));
					Player.SetMediaSource(mediaSource);
					Player.Prepare();
					hasSetSource = true;
				}
				else
				{
					Player.SetMediaItem(MediaItem.FromUri(uri.AbsoluteUri));
					Player.Prepare();
					hasSetSource = true;
				}
			}
		}`

I test this implementation woking fine.
Maybe if Maui Team wants, i can add User Agent.
P.D: Sorry for my English.
@jfversluis

pictos:
@Phantom-KNA please open a discussion and add your proposal there. This PR is already close, thanks

PR #673, opened 2022-10-04
StateContainer

### LINKED ISSUE #120: [Proposal] StateLayout
# StateLayout

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation: https://github.com/MicrosoftDocs/CommunityToolkit/pull/155

## Summary
[summary]: #summary

Displaying a specific view when your app is in a specific state is a common pattern throughout any mobile app. Examples range from creating loading views to overlay on the screen, or on a subsection of the screen. Empty state views can be created for when there's no data to display, and error state views can be displayed when an error occurs.

## Detailed Design
[design]: #detailed-design

### StateLayout.shared.cs

```cs
public static class StateLayout
{
  public static readonly BindableProperty StateViewsProperty;
  public static readonly BindableProperty CurrentStateProperty;
  public static readonly BindableProperty CurrentCustomStateKeyProperty;
  public static readonly BindableProperty AnimateStateChangesProperty;
  
  public static IList<StateView>? GetStateViews(BindableObject b);
  public static void SetCurrentState(BindableObject b, LayoutState value);
  public static LayoutState GetCurrentState(BindableObject b);
  public static void SetCurrentCustomStateKey(BindableObject b, string value);
  public static string? GetCurrentCustomStateKey(BindableObject b);
  public static void SetAnimateStateChanges(BindableObject b, bool value);
  public static bool GetAnimateStateChanges(BindableObject b);
}
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             x:Class="MyLittleApp.MainPage">

     <Grid xct:StateLayout.CurrentState="{Binding CurrentState}">
      <xct:StateLayout.StateViews>
         <xct:StateView StateKey="Loading">
            <Grid BackgroundColor="White">
               <StackLayout VerticalOptions="Center" HorizontalOptions="Center">
                  <ActivityIndicator Color="#1abc9c" />
                  <Label Text="Loading..." HorizontalOptions="Center" />
               </StackLayout>
            </Grid>
         </xct:StateView>
      </xct:StateLayout.StateViews>      

     ...

  </Grid>

</ContentPage>
```

nicjay:
### Description of Change ###

 
Displaying a specific view when your app is in a specific state is a common pattern throughout any mobile app. Examples range from creating loading views to overlay on the screen, or on a subsection of the screen. Empty state views can be created for when there's no data to display, and error state views can be displayed when an error occurs.

Based on [StateLayout](https://learn.microsoft.com/en-us/xamarin/community-toolkit/views/statelayout) from the Xamarin Community Toolkit, originally created by Steven Thewissen. 

 ### Linked Issues ###
 

 - Fixes #120

 ### PR Checklist ###
 
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/155 

There are some changes since the original implementation/proposal. Summary:

- `StateLayout` renamed to `StateContainer`.
- Removed `StateView` wrapper, Views can now be states directly by defining a `StateKey` property.
- `CurrentState` is now a string binding.
- Removed `CurrentCustomStateKey`.
- Removed `RepeatCount` and `Template`.


src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayout.shared.cs
```diff
@@ -0,0 +1,143 @@
+﻿using CommunityToolkit.Maui.Core;
+
+namespace CommunityToolkit.Maui.Layouts;
+
+/// <summary>
+/// The <see cref="StateLayout"/> control enables the user to turn any layout element like a <see cref="Grid"/> or <see cref="StackLayout"/> into an individual state-aware element.
```
pictos, line 6: Let's not use the deprecated `StackLayout` here (;

```suggestion
/// The <see cref="StateLayout"/> control enables the user to turn any layout element like a <see cref="Grid"/> or <see cref="VerticalStackLayout"/> into an individual state-aware element.
```

samples/CommunityToolkit.Maui.Sample/ViewModels/Converters/StateToBooleanConverterViewModel.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System.Windows.Input;
 using CommunityToolkit.Maui.Converters;
+using CommunityToolkit.Maui.Core;
```
pictos, line 3: Do we need this?

src/CommunityToolkit.Maui.UnitTests/Converters/StateToBoolConverterTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System.ComponentModel;
 using System.Globalization;
 using CommunityToolkit.Maui.Converters;
+using CommunityToolkit.Maui.Core;
```
pictos, line 4: Do we need this?

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,326 @@
...
+
+				for (var i = 0; i < repeatCount; i++)
+				{
+					items.Add(i);
+				}
+
+				// We create a StackLayout to stack repeating items.
+				// It takes VerticalOptions and HorizontalOptions from the
+				// StateView to allow for more control over how it layouts.
+				var s = new StackLayout
```
pictos, line 134: Can we use the `VerticalStackLayout`?

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,326 @@
...
+
+/// <summary>
+/// StateLayout Controller
+/// </summary>
+public class StateLayoutController: IDisposable
+{
+	readonly WeakReference<Layout> layoutWeakReference;
+	bool layoutIsGrid;
+	LayoutState previousState = LayoutState.None;
+	IList<View> originalContent = Enumerable.Empty<View>().ToList();
```
pictos, line 15: This is a `private` field and we just use `List<T>` during the interaction, so we don't need to use the interface here.

```suggestion
	List<View> originalContent = Enumerable.Empty<View>().ToList();
```

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,326 @@
...
+			foreach (var item in layout.Children)
+			{
+				originalContent.Add((View)item);
+			}
+
+		}
+
+		var view = GetViewForState(state, customState);
+
+		if (view != null)
```
pictos, line 104: We use pattern matching to check for `null`. Please change it in the others as well
```suggestion
		if (view is not null)
```

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,326 @@
...
+
+			layout.Children.Clear();
+
+			var repeatCount = GetRepeatCount(state, customState);
+			var template = GetTemplate(state, customState);
+
+			if (template != null)
+			{
+				// We have a template we can use.
+				var items = new List<int>();
```
pictos, line 124: We already know the size of the `List`, so we can take advantage of that.
```suggestion
				var items = new List<int>(repeatCount);
```

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,326 @@
...
+					VerticalOptions = view.VerticalOptions,
+					HorizontalOptions = view.HorizontalOptions
+				};
+
+				// If the layout we're applying StateLayout to is a Grid,
+				// we want to have the StateLayout span the entire Grid surface.
+				// Otherwise it would just end up in row 0 : column 0.
+				if (layout is Grid grid)
+				{
+					if (grid.RowDefinitions.Any())
```
pictos, line 146: We don't need to use LINQ here.
```suggestion
					if (grid.RowDefinitions.Count > 0)
```

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,326 @@
...
+					{
+						Grid.SetRowSpan(s, grid.RowDefinitions.Count);
+					}
+
+
+					if (grid.ColumnDefinitions.Any())
```
pictos, line 152: We don't need to use LINQ here.
```suggestion
					if (grid.ColumnDefinitions.Count > 0)
```

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,326 @@
...
+					VerticalOptions = view.VerticalOptions,
+					HorizontalOptions = view.HorizontalOptions
+				};
+
+				// If the layout we're applying StateLayout to is a Grid,
+				// we want to have the StateLayout span the entire Grid surface.
+				// Otherwise it would just end up in row 0 : column 0.
+				if (layout is Grid grid)
+				{
+					if (grid.RowDefinitions.Any())
```
pictos, line 188: Let's use `Count > 0` here and in the next statement

pictos (changes requested):
@nicjay thanks for all this work! It's great! I've a few comments for you, let me know if have any questions.

Also, are you able to add a sample, that way we can test the feature?

Thanks again!

nicjay:
@pictos thanks for the review, I've made all updates per your requests. I also added a StateLayout example page to the sample app.

Regarding the `using CommunityToolkit.Maui.Core`, that's because I moved the `LayoutState` enum into the Primitives folder in Core. I saw some other common enums in there too...is that the right place for it? So far, it's used by all the new StateLayout stuff, StateToBoolConverter, and the associated samples and tests for each.

src/CommunityToolkit.Maui.UnitTests/Converters/StateToBoolConverterTests.cs
brminnick: Yup, because `public enum StateLayout` moved from `CommunityToolkit.Maui` to `CommunityToolkit.Maui.Core`

src/CommunityToolkit.Maui.UnitTests/Layouts/StateLayoutTests.cs
```diff
@@ -0,0 +1,119 @@
+﻿using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.Layouts;
+using CommunityToolkit.Maui.Views;
+using FluentAssertions;
+using Nito.AsyncEx;
+using Xunit;
+
+namespace CommunityToolkit.Maui.UnitTests.Layouts;
+
+public class StateLayoutTests : BaseTest
```
brminnick, line 10: This is a good start! Let's make sure we have a test that covers every public API in the new code: `StateLayout`, `StateLayoutController`, and `StateView`.

brminnick: Make sure you're also testing every edge case, eg What happens when `animationTokenSource` is canceled?

brminnick (commented):
Thanks @nicjay!

Just FYI - I made a couple tweaks:
- Move `StateView` to `CommunityToolkit.Maui.Views` namespace
- Make `StateLayoutController` an internal sealed class
  - I don't see a need for a .NET MAUI developer to need to access this directly; we can always make it `public` later without breaking changes
  - Making it `sealed` improves performance and reduces the `IDisposable` implementation
- Properly implement async/await for the following methods
  - `SwitchToTemplate(string customState, bool animate)`
    - Return `Task` instead of `void`
  - `SwitchToTemplate(LayoutState state, string? customState, bool animate)`
    - Return `Task` instead of `void`
  - `FadeLayoutChildren(Layout layout, bool shouldAnimate, bool isHidden, CancellationToken token)`
    - return `ValueTask` instead of `Task`
    - pass `CancellationToken` into `await Task.WhenAll().WaitAsync(token)` 
      - [see this thread](https://github.com/dotnet/runtime/issues/67392#issuecomment-1084944767) for more info on cancelling `Task.WhenAll()`
- Use `token.ThrowIfCancellationRequested()`
  - It's important to throw a `OperationCanceledException` when a `CancellationToken` instead of failing silently to allow the developer to handle the unexpected state in their app

## Name Change Proposal

One big change I'd like to discuss is changing the name from `StateLayout` to `BindableStateLayout` (or something similar).

This technically isn't a `Microsoft.Maui.Control.Layout` because it doesn't implement `ILayout`. Naming it `StateLayout` may confuse devs who try to use it like a typical `Layout` and try to add `Children` to it. 

I'll raise this to the team in our [monthly standup](https://www.youtube.com/watch?v=ZLwJlJPTKvY), tomorrow.

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,285 @@
...
+		return view?.Template;
+	}
+
+	View CreateItemView(LayoutState state, string? customState)
+	{
+		var view = StateViews.FirstOrDefault(x => (x.StateKey == state && state is not LayoutState.Custom) ||
+						(state is LayoutState.Custom && x.CustomStateKey == customState));
+
+		// TODO: This only allows for a repeatcount of 1.
+		// Internally in Xamarin.Forms we cannot add the same element to Children multiple times.
```
brminnick, line 266: I'm guessing this is a copy/pasted comment from the Xamarin.CommunityToolkit implementation. 

Can we do this ToDo before merging this PR?

src/CommunityToolkit.Maui.Core/Primitives/LayoutState.shared.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace CommunityToolkit.Maui.Core;
+
+/// <summary>
+/// Enum that represents possible layout states
+/// </summary>
+public enum LayoutState
```
VladislavAntonyuk, line 6: Do we need this class at all? can we assume the user always uses Custom state?

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,285 @@
...
+
+				for (var i = 0; i < repeatCount; i++)
+				{
+					items.Add(i);
+				}
+
+				// We create a VerticalStackLayout to stack repeating items.
+				// It takes VerticalOptions and HorizontalOptions from the
+				// StateView to allow for more control over how it layouts.
+				var s = new VerticalStackLayout
```
VladislavAntonyuk, line 125: could you please create a more valuable variable name

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,285 @@
...
+				{
+					Opacity = animate ? 0 : 1,
+					VerticalOptions = view.VerticalOptions,
+					HorizontalOptions = view.HorizontalOptions
+				};
+
+				// If the layout we're applying StateLayout to is a Grid,
+				// we want to have the StateLayout span the entire Grid surface.
+				// Otherwise it would just end up in row 0 : column 0.
+				if (layout is Grid grid)
```
VladislavAntonyuk, line 135: should we rely on Maui controls? I would prefer using the interface here: IGrid

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,285 @@
...
```
VladislavAntonyuk, line 176: seems like this code is the same as above. please extract it to remove duplication

src/CommunityToolkit.Maui.Core/Primitives/LayoutState.shared.cs
nicjay: It's probably not strictly necessary, just convenience. I do like having the preset enum built into the toolkit for common uses like Loading or Error, with Custom available for more advanced scenarios.

VladislavAntonyuk:
I have some idea-proposal about design to simplify the StateLayout.
1. Rename StateLayout to StateContainer
2. Replace `IList<StateView>` with `IList<IView>` so we can put any view inside it (no need to put a view inside another view)
3. Remove StateView
4. Create StateKey attached property for IView, so each view in StateContainer can describe its state

brminnick:
What if we change `IList<StateView>` to `IDictionary<LayoutState, IView>`?

This would allow us to still leverage the `LayoutState` enum and automatically change the displayed view whenever LayoutState changes. 

One limitation this would raise is it would limit us to only use one `LayoutState.CustomState`. But I imagine most devs don't use multiple custom states.

VladislavAntonyuk:
We still have Custom LayoutState, where a user needs to specify his own key. It is the same as a user choosing a predefined state like Save or Loading. I would prefer to remove this enum and use a string instead.

brminnick:
Just FYI - We aren't able to remove the LayoutState enum from the project because it's used by StateToLayoutConverter

nicjay:
Thanks guys, StateContainer it is. 

I love the idea of less view nesting. There's a lot to unpack:

I imagine a potential removal of StateView (and use IView directly) looks something like this?
```xaml
<VerticalStackLayout mct:StateContainer.CurrentState="{Binding CurrentState}">

    <mct:StateContainer.StateViews>

        <VerticalStackLayout mct:StateView.StateKey="Loading">
            <Label Text="Label of the First State" />
        </VerticalStackLayout>

        <HorizontalStackLayout mct:StateView.StateKey="Error">
            <Label Text="Label of the Second State" />
        </VerticalStackLayout>

        <Label mct:StateView.StateKey="Success" Text="I am a whole state on my own!" />

    </mct:StateContainer.StateViews>

    <Label Text="Default Content" />

</VerticalStackLayout>
```

- Where would the StateKey and other attached property live? Is there still a StateView (static) class containing these, or lumped into StateContainer, or elsewhere?
- I can see the case for custom string states only, but if LayoutState can't be removed, might as well keep using it. 
(Then again, is anyone actually using StateToBooleanConverter without StateLayout?)
- For RepeatCount functionality. XCT version relied on DataTemplates inside a Template attached property. I assume it could work the same but perhaps there's a better way.

VladislavAntonyuk:
The new design looks much cleaner now.
1. It's always hard to decide what is the right place for it. We don't want to have a lot of folders but also can't put everything in one. Let's for now keep it in Layout. We can move it later.
2. I would keep LayoutState in Converter. For user it doesn't matter if he uses enum or string. For us it's additional code that we have to support. So I propose assuming we always use LayoutState.Custom and StateKey is string. 
3. I don't know much about RepeatCount. Is it used for displaying the same state view multiple times? Is it required to be a DataTemplate?

nicjay:
## Update
I made some changes based on above discussion and @VladislavAntonyuk's suggestions. Usage is as in https://github.com/CommunityToolkit/Maui/pull/673#issuecomment-1270808062 and code logic is simplified quite a bit. 

- `StateLayout` renamed `StateContainer`
- `StateView` is now static, no longer inherits ContentView, and holds an attached StateKey string property for `View` classes. 
  - This means View classes can now be states directly instead of wrapping with a StateView.
  - StateView.cs is moved back into CommunityToolkit.Maui.Layouts for now. 
- `StateContainer` now exclusively uses custom  strings to determine state.
  - This allows infinite flexibility for states.
   - Null or empty string corresponds to the default 'none' state, aka the original content.
   - The `LayoutState` enum is currently not used at all and put back to its original location in StateToBooleanConverter.
- I didn't include the RepeatCount property in this iteration - see below.
- Using `IList<View>` instead of `IList<IView>` - the bindings didn't play nice with IView.

Let me know what you think!

## Naming
With the existence of BindableLayout, I can see a case for the original name "StateLayout" as a sort of cousin to BindableLayout. Both are a set of attached properties for Layout derived classes, and not being actual Layouts themselves. I'm fine with StateContainer, but it's food for thought. Also, does StateView still make sense for the attached view properties? StateItem?

## RepeatCount
I couldn't find a way around the requirement for a DataTemplate when using a RepeatCount property. The View itself cannot be repeated/cloned/duplicated as far as I tried. A child DataTemplate could, but then that just replicates BindableLayout.

In any case, even though XCT had it, should RepeatCount be dropped for MCT? It adds complexity and doesn't seem particularly related to the concept of state. Now with less StateView wrapper XAML, a user can do something like this easily enough, or handle their own BindableLayout.

```xaml
<VerticalStackLayout mct:StateContainer.StateKey="MyLoadingState">
    <controls:SkeletonLoaderControl />
    <controls:SkeletonLoaderControl />
    <controls:SkeletonLoaderControl />
</VerticalStackLayout>
```

src/CommunityToolkit.Maui/Layouts/StateLayout/StateLayoutController.shared.cs
nicjay: It seems the author ran into the same limitation cloning elements as I did - see RepeatCount discussion below. ToDO is removed for now.

brminnick:
Thanks @nicjay! Just a thought - maybe we remove the `DataTemplate` and `RepeatCount` for its v1 release?

These aren't really necessary for `StateContainer` to work. If anything, they seem like a bolted-on feature.

We could always implement them later without breaking changes if the we get demand from the community.

src/CommunityToolkit.Maui.UnitTests/Layouts/StateLayoutTests.cs
nicjay: Added a bunch more tests.

src/CommunityToolkit.Maui.Core/Primitives/LayoutState.shared.cs
nicjay: Going with a string StateKey (always custom) for now, we can see how it feels.

pictos (commented):
I really love the changes made here, the control is way simpler now.
I believe we're already ready to merge this! Here're my final suggestions.

src/CommunityToolkit.Maui/Layouts/StateContainer/StateContainer.shared.cs
```diff
@@ -0,0 +1,108 @@
...
+	static object ContainerControllerCreator(BindableObject bindable)
+	{
+		if (bindable is not Layout layoutView)
+		{
+			throw new InvalidOperationException($"Cannot create the {nameof(StateContainerController)}. The specified view '{bindable.GetType().FullName}' does not inherit Layout.");
+		}
+
+		return new StateContainerController(layoutView)
+		{
+			StateViews = GetStateViews(layoutView) ?? new List<View>()
```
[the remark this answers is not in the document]
nicjay: I removed the `new List<View>()`
1) GetStateViews already pulls a default value from the bindable property itself.
2) Controller is created only once anyway (per layout) on initial state change.

pictos (approved):
LGTM❣
Thanks @nicjay

VladislavAntonyuk (approved):

pictos:
@nicjay the next step is to create the docs and open a PR on our doc repo. Let me know if you have any question

nicjay:
@pictos Docs PR created!

brminnick (approved):
Thanks @nicjay!!!

bijington:
Sorry I haven't had a chance to check the docs yet but will try this week

nicjay:
Thanks all, glad to see this make it in 👍

sthewissen:
> Thanks all, glad to see this make it in 👍 

Thanks for taking my baby and giving it a new life 😀😘

CodingOctocat:
If I just want to show/hide the view, the best practice is IsVisible? StateContainer must be 2 and more StateView.StateKey?

PR #699, opened 2022-10-18
Create `IFileSaver.SaveAsync()` and `IFolderPicker.PickAsync()`

### LINKED ISSUE #634: [Proposal] Dialogs
# Dialogs

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [x] iOS Support
  * [x] Android Support
  * [x] macOS Support
  * [x] Windows Support
* [x] Unit Tests
* [x] Sample
* [x] Documentation

## Link to Discussion

Discussed in https://github.com/CommunityToolkit/Maui/discussions/381

## Summary
[summary]: #summary

This is a highly requested feature. It can include SaveFileDialog, FolderPickerDialog.
It is not a visual element. It’s closer to Toast and Snackbar

## Motivation
[motivation]: #motivation

Dialogs allow better integration of our application with the operating system. It can be useful for apps which needs integration with the file system (open files/directories, save files, etc)

## Detailed Design
[design]: #detailed-design

```csharp
public class FolderPicker
{
    static Task<Folder> PickAsync(string initialPath, CancellationToken cancellationToken);
    static Task<Folder> PickAsync(CancellationToken cancellationToken);
}

class Folder
{
    string Path {get;set;}
    string Name {get;set;}
}


public class SaveFileDialog
{
    static Task<bool> SaveAsync(string initialPath, bytes fileBytes, CancellationToken cancellationToken);
    static Task<bool> SaveAsync(bytes fileBytes, CancellationToken cancellationToken);
}
```

## Usage Syntax
[usage]: #usage-syntax

### C# Usage

```cs
var folder = FolderPicker.PickAsync("initial path", token);
var files = Directory.GetFiles(folder.Path);
for each (var file in files)
{
   ...
}

bool isSaved = SaveFileDialog.SaveAsync("initial path", fileBytes, token);
if (!isSaved){
   alert("file not saved");
}
```

-->

## Design

image

VladislavAntonyuk:
### Description of Change ###

Add SaveFileDialog and FolderPicker

 ### Linked Issues ###
 

 - Fixes #634 

 ### PR Checklist ###
 
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/180 


src/CommunityToolkit.Maui.Core/Extensions/UIViewExtensions.macios.cs
```diff
@@ -84,4 +84,30 @@ public static class UIViewExtensions
...
+		return UIApplication.SharedApplication.ConnectedScenes.ToArray()
+			.Select(x => x as UIWindowScene)
+			.FirstOrDefault()?
+			.Windows.FirstOrDefault(x => x.IsKeyWindow);
+	}
+	
+	/// <summary>
+	/// Get current UI view controller
+	/// </summary>
+	public static UIViewController? GetCurrentUIViewController()
```
pictos, line 102: why not use the `Microsoft.Maui.ApplicationModel.GetCurrentUIViewController` method?

src/CommunityToolkit.Maui.Core/Extensions/UIViewExtensions.macios.cs
```diff
@@ -84,4 +84,30 @@ public static class UIViewExtensions
 		UIDevice.CurrentDevice.CheckSystemVersion(11, 0)
 			? view.SafeAreaLayoutGuide.WidthAnchor
 			: view.WidthAnchor;
+
+	/// <summary>
+	/// Get Key Window
+	/// </summary>
+	public static UIWindow? GetKeyWindow()
```
pictos, line 91: Why not use `Microsoft.Maui.Platform.UIApplicationExtensions.GetKeyWindow` method?

pictos (commented):
Before you go too far...

src/CommunityToolkit.Maui.Core/Extensions/UIViewExtensions.macios.cs
VladislavAntonyuk: it is internal, I didn't know about it.

brminnick (commented):
Thanks Vlad!

Let's avoid using `Essentials` in the namespace to align with the fact that .NET MAUI dropped the `Essentials` namespace. 

If we align our namespaces closely to .NET MAUI, it will make it easier to promote the features from the Toolkit in to .NET MAUI.

I think using `namespace CommunityToolkit.Maui.Storage` would be good, because it  more-closely matches [.NET MAUI's `Microsoft.Maui.Storage.FilePicker` namespace](https://github.com/dotnet/maui/blob/c4d37c881db65f7dacba05e63613ac9b810a7ec1/src/Essentials/src/FilePicker/FilePicker.shared.cs#L10), but I'm totally open to suggestions!

src/CommunityToolkit.Maui.Core/Interfaces/IFolderPicker.shared.cs
```diff
@@ -0,0 +1,24 @@
...
+	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
+	/// <returns><see cref="Folder"/></returns>
+	Task<Folder?> PickAsync(string initialPath, CancellationToken cancellationToken);
+
+	/// <summary>
+	/// Allows picking a folder from the file system.
+	/// </summary>
+	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
+	/// <returns><see cref="Folder"/></returns>
+	Task<Folder?> PickAsync(CancellationToken cancellationToken);
```
brminnick, line 23: Let's return `Task<Folder>` (non-nullable) and throw an Exception if the folder cannot be returned.

This aligns more closely to .NET MAUI's `FilePicker` that does things like `throws throw new PlatformNotSupportedException`

src/CommunityToolkit.Maui.Core/Interfaces/IFolderPicker.shared.cs
```diff
@@ -0,0 +1,24 @@
...
+/// </summary>
+public interface IFolderPicker
+{
+	/// <summary>
+	/// Allows picking a folder from the file system.
+	/// </summary>
+	/// <param name="initialPath">Initial path</param>
+	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
+	/// <returns><see cref="Folder"/></returns>
+	Task<Folder?> PickAsync(string initialPath, CancellationToken cancellationToken);
```
brminnick, line 16: Let's return `Task<Folder>` (non-nullable) and throw an Exception if the folder cannot be returned.

This aligns more closely to .NET MAUI's `FilePicker` that does things like `throws throw new PlatformNotSupportedException`

For us, we can leverage `FileNotFoundException` when appropriate.

src/CommunityToolkit.Maui.Core/Interfaces/ISaveFileDialog.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+	/// <summary>
+	/// Allows selecting target folder and saving files to the file system.
+	/// </summary>
+	/// <param name="initialPath">Initial path</param>
+	/// <param name="fileName">File name with extension</param>
+	/// <param name="stream"><see cref="Stream"/></param>
+	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
+	/// <returns>True if file is saved, otherwise False</returns>
+	Task<bool> SaveAsync(string initialPath, string fileName, Stream stream, CancellationToken cancellationToken);
```
brminnick, line 16: Instead of returning `Task<bool>`, let's return `Task` and throw an Exception if the file is not saved.

We can create a new Exception like `FileSaveException` or something like that.

This aligns more closely to .NET MAUI's FilePicker that does things like `throws throw new PlatformNotSupportedException`.

src/CommunityToolkit.Maui.Core/Interfaces/ISaveFileDialog.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+
+	/// <summary>
+	/// Allows selecting target folder and saving files to the file system.
+	/// </summary>
+	/// <param name="fileName">File name with extension</param>
+	/// <param name="stream"><see cref="Stream"/></param>
+	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
+	/// <returns>True if file is saved, otherwise False</returns>
+	Task<bool> SaveAsync(string fileName, Stream stream, CancellationToken cancellationToken);
```
brminnick, line 25: Instead of returning `Task<bool>`, let's return `Task` and throw an Exception if the file is not saved.

We can create a new Exception like `FileSaveException` or something like that.

This aligns more closely to .NET MAUI's FilePicker that does things like `throws throw new PlatformNotSupportedException`.

src/CommunityToolkit.Maui.Core/Primitives/Folder.shared.cs
```diff
@@ -0,0 +1,17 @@
...
+
+/// <summary>
+/// Represents a folder in the file system.
+/// </summary>
+public class Folder
+{
+	/// <summary>
+	/// Folder path.
+	/// </summary>
+	public string Path { get; init; } = null!;
```
brminnick, line 11: Let's avoid the null-forging operator, `!`.

```suggestion
	public string Path { get; init; } = string.Empty;
```

src/CommunityToolkit.Maui.Core/Primitives/Folder.shared.cs
```diff
@@ -0,0 +1,17 @@
...
+
+	/// <summary>
+	/// Folder name.
+	/// </summary>
+	public string Name { get; init; } = null!;
```
brminnick, line 16: Let's avoid the null-forging operator, `!`.

```suggestion
	public string Name { get; init; } = string.Empty;
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+												or FileSelectionMode.FileSaveRoot)
+				{
+					// Add file names to the list if we are doing a file save or file open operation
+					directories.Add(file.Name);
+				}
+			}
+		}
+		catch (Exception)
+		{
+			// ignored
```
brminnick, line 229: Let's propagate Exceptions up to the developer consuming the library (eg `throw;`) to let them handle it instead of ignoring Exceptions.

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+	FileSave,
+	FolderChoose,
+	FileOpenRoot,
+	FileSaveRoot,
+	FolderChooseRoot
+}
+
+sealed class FileFolderDialog : IDisposable
+{
+	readonly Context? context;
```
brminnick, line 29: `context` will never be null because it is assigned in the constructor

```suggestion
	readonly Context context;
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+		this.context = context;
+		sdCardDirectory = Environment.RootDirectory.AbsolutePath;
+
+		try
+		{
+			sdCardDirectory = new File(sdCardDirectory).CanonicalPath;
+		}
+		catch (IOException)
+		{
+		}
```
brminnick, line 75: Why are we ignoring `IOException`? 

I'd prefer to propagate it up to the dev consuming FileFolderDialog so that they can be alerted to it and handle it properly for their app.

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+			{
+				result = this.directory;
+				autoResetEvent?.Set();
+			}
+		});
+
+		dialogBuilder.SetNegativeButton("Cancel", (_, _) => { });
+		dirsDialog = dialogBuilder.Create();
+
+		if (dirsDialog != null)
```
brminnick, line 169:
```suggestion
		if (dirsDialog is not null)
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+		Log.Debug("~~~~~", $"{nameof(directory)}={directory}");
+
+		directory = new File(directory).CanonicalPath;
+
+		this.directory = directory;
+		var subDirectories = GetDirectories(directory);
+
+		var dialogBuilder = CreateDirectoryChooserDialog(directory, subDirectories, (sender, args) =>
+		{
+			var mDirOld = this.directory;
```
brminnick, line 114: Let's choose a more descriptive variable name than `mDirOld`

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+			}
+			else
+			{
+				result = this.directory;
+				autoResetEvent?.Set();
+			}
+		});
+
+		dialogBuilder.SetNegativeButton("Cancel", (_, _) => { });
+		dirsDialog = dialogBuilder.Create();
```
brminnick, line 167: Let's choose a more descriptive variable name than `dirsDialog`

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+
+	public void Dispose()
+	{
+		autoResetEvent?.Dispose();
+		inputText?.Dispose();
+		titleView?.Dispose();
+		titleView1?.Dispose();
+	}
+
+	public async Task<string> GetFileOrDirectoryAsync(string directory, string defaultFileName = "default")
```
brminnick, line 86: Can we pass in a `CancellationToken`?

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+			var directoryFile = new File(dir);
+
+			// if directory is not the base sd card directory add ".." for going up one directory
+			if ((isRootSelectionMode || !directory.Equals(sdCardDirectory, StringComparison.Ordinal)) &&
+				!("/".Equals(directory, StringComparison.Ordinal)))
+			{
+				directories.Add("..");
+			}
+
+			Log.Debug("~~~~", "m_dir=" + directory);
```
brminnick, line 204: Can we remove these debug logs?

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+
+		var directoryFile = new File(directory);
+		while (!directoryFile.Exists() || !directoryFile.IsDirectory)
+		{
+			if (directoryFile.Parent is null)
+			{
+				throw new FileNotFoundException($"Could not locate {directoryFile}");
+			}
+
+			directoryFile = new File(directory);
```
brminnick, line 101: Can we remove these debug logs?

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+			Log.Debug("~~~~~", $"{nameof(directory)}={directory}");
```
brminnick, line 102: Can we remove these debug logs?

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+		}
+
+		Log.Debug("~~~~~", $"{nameof(directory)}={directory}");
```
brminnick, line 105: Can we remove these debug logs?

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+		var dialogBuilder = new AlertDialog.Builder(context);
+		////////////////////////////////////////////////
+		// Create title text showing file select type // 
+		////////////////////////////////////////////////
+		titleView1 = new TextView(context)
+		{
+			LayoutParameters = new ViewGroup.LayoutParams(ViewGroup.LayoutParams.MatchParent, ViewGroup.LayoutParams.WrapContent),
+
+			//m_titleView1.setTextAppearance(m_context, android.R.style.TextAppearance_Large);
+			//m_titleView1.setTextColor( m_context.getResources().getColor(android.R.color.black) );
```
brminnick, line 253: Can this comment be deleted?

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+	//////////////////////////////////////////////////////////////////////////////////////////////////////////
+	AlertDialog.Builder CreateDirectoryChooserDialog(string title,
+		List<string> listItems,
+		EventHandler<DialogClickEventArgs> onClickListener,
+		string defaultFileName = "default")
+	{
+		var dialogBuilder = new AlertDialog.Builder(context);
+		////////////////////////////////////////////////
+		// Create title text showing file select type // 
+		////////////////////////////////////////////////
```
brminnick, line 247: Let's trim this comment to one line. We don't use block comments like this anywhere else
```suggestion
		// Create title text showing file select type
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+			// ignored
+		}
+
+		directories.Sort();
+		return directories;
+	}
+
+	/////////////////////////////////////////////////////////////////////////////////////////////////////////
+	//////                                   START DIALOG DEFINITION                                    //////
+	//////////////////////////////////////////////////////////////////////////////////////////////////////////
```
brminnick, line 238: Let's replace this comment with something more descriptive/helpful, or delete it.

We don't use block comments like this anywhere else

```suggestion
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+
+		await Task.Run(() => autoResetEvent?.WaitOne());
+
+		return result ?? throw new FileNotFoundException($"Could not locate {directoryFile}");
+	}
+
+
+	static bool CreateSubDir(string newDir)
+	{
+		var newDirFile = new File(newDir);
```
brminnick, line 186: Let's avoid abbreviating variable names

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+
+	string directory = string.Empty;
+	ArrayAdapter<string>? listAdapter;
+	TextView? titleView;
+	TextView? titleView1;
+	string selectedFileName;
+
+	//////////////////////////////////////////////////////
+	// Callback interface for selected directory
+	//////////////////////////////////////////////////////
```
brminnick, line 46: Let's either replace this comment with something more descriptive (like an XML comment) or delete it. 

We don't use block comments like this anywhere else in the code base.

```suggestion
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+		{
+			Orientation = Orientation.Vertical
+		};
+		titleLayout1.AddView(titleView1);
+
+		if (fileSelectionMode is FileSelectionMode.FileSave)
+		{
+			///////////////////////////////
+			// Create New Folder Button  //
+			///////////////////////////////
```
brminnick, line 282: This comment isn't helpful; `new Button` is pretty obvious that we're creating a button
```suggestion
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+					})
+					?.SetNegativeButton("Cancel", (_, _) => { })
+					?.Show();
+			};
+			titleLayout1.AddView(newDirButton);
+		}
+
+		/////////////////////////////////////////////////////
+		// Create View with folder path and entry text box // 
+		/////////////////////////////////////////////////////
```
brminnick, line 320:
```suggestion
		// Create View with folder path and entry text box
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,435 @@
...
+			{
+				Text = defaultFileName
+			};
+
+			titleLayout.AddView(inputText);
+		}
+
+		//////////////////////////////////////////
+		// Set Views and Finish Dialog builder  //
+		//////////////////////////////////////////
```
brminnick, line 366:
```suggestion
		// Set Views and Finish Dialog builder
```

brminnick (changes requested):
Thanks Vlad!! 

What do you think about propagating more Exceptions up to the developer? This PR is using a lot of try/catch blocks that swallow Exceptions like this:
```
try
{
    // ...
}
catch (IOException)
{
    // nothing in the catch block
}
```

I've left more-specific comments in-line in my review.

JoonghyunCho:
@VladislavAntonyuk Thanks for the Tizen files! I'm working on filling those up 😄

JoonghyunCho:
Just added Tizen implementation. I tried to keep align with the existing android implementation.

<image src="https://user-images.githubusercontent.com/14328614/203915687-11767b28-0e44-4a22-a60a-3381bbb9d54f.gif" width="250" /> <image src="https://user-images.githubusercontent.com/14328614/203915701-7521e56e-e38a-43e4-9836-11d627965da8.gif" width="250" />

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,397 @@
...
+		{
+			inputText.Text = selectedFileName;
+		}
+	}
+
+	ArrayAdapter<string> CreateListAdapter(List<string>? items)
+	{
+		var adapter = new SimpleArrayAdapter(context ?? throw new InvalidOperationException(),
+			Android.Resource.Layout.SelectDialogItem,
+			Android.Resource.Id.Text1, items ?? Array.Empty<string>().ToList());
```
pictos, line 369: I would like to remove the `.ToList()`, I can see two solutions, one is changing the argument to be `IList` as you can see in the snippet below
```suggestion
	ArrayAdapter<string> CreateListAdapter(IList<string>? items)
	{
		var adapter = new SimpleArrayAdapter(context ?? throw new InvalidOperationException(),
			Android.Resource.Layout.SelectDialogItem,
			Android.Resource.Id.Text1, items ?? Array.Empty<string>());
```

and other solution will be to create a static empty List<string> and use that instead. That way we don't allocate any memory.
```
|         Method |       Mean |     Error |    StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
|--------------- |-----------:|----------:|----------:|------:|-------:|----------:|------------:|
|    WithNoCache | 17.2207 ns | 0.0703 ns | 0.0587 ns |  1.00 | 0.0051 |      32 B |        1.00 |
|  WithCacheList |  0.5017 ns | 0.0142 ns | 0.0132 ns |  0.03 |      - |         - |        0.00 |
| WithCacheArray |  0.5133 ns | 0.0098 ns | 0.0092 ns |  0.03 |      - |         - |        0.00 |
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,397 @@
...
+		return adapter;
+	}
+
+	class SimpleArrayAdapter : ArrayAdapter<string>
```
pictos, line 373:
```suggestion
	sealed class SimpleArrayAdapter : ArrayAdapter<string>
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,397 @@
...
+		{
+		}
+
+		public override View GetView(int position, View? convertView, ViewGroup parent)
+		{
+			var v = base.GetView(position, convertView, parent);
+			if (v is TextView textView)
+			{
+				// Enable list item (directory) text wrapping
+				if (textView.LayoutParameters != null)
```
pictos, line 386:
```suggestion
				if (textView.LayoutParameters is not null)
```

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,397 @@
...
+	{
+		public SimpleArrayAdapter(Context context, int resource, int textViewResourceId, IList<string> objects) : base(
+			context, resource, textViewResourceId, objects)
+		{
+		}
+
+		public override View GetView(int position, View? convertView, ViewGroup parent)
+		{
+			var v = base.GetView(position, convertView, parent);
+			if (v is TextView textView)
```
pictos, line 383: can we invert this `if`, for better readability

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,390 @@
...
+using ScrollView = Tizen.UIExtensions.NUI.ScrollView;
+using Shadow = Tizen.NUI.Shadow;
+using HorizontalAlignment = Tizen.NUI.HorizontalAlignment;
+using VerticalAlignment = Tizen.NUI.VerticalAlignment;
+using View = Tizen.NUI.BaseComponents.View;
+using Window = Tizen.NUI.Window;
+
+namespace CommunityToolkit.Maui.Storage;
+
+enum FileSelectionMode
```
pictos, line 21: This looks like to be the same enum from Android, can this be at `shared` lvl?

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,390 @@
...
+
+		foreach(var view in directoyViews)
+		{
+			directoryScrollView.ContentContainer.Remove(view);
+		}
+		directoyViews.Clear();
+
+		var listItems = GetDirectories(path);
+
+		if (listItems != null)
```
pictos, line 302:
```suggestion
		if (listItems is not null)
```

pictos: can we invert this `if`?

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,390 @@
...
+			directories.Add(previousDirectorySymbol);
+		}
+
+		if (!Directory.Exists(directoryPath) || !IsDirectory(directoryPath))
+		{
+			return directories;
+		}
+
+		string[] allFilesAndDirectories = Directory.GetDirectories(directoryPath);
+		allFilesAndDirectories = allFilesAndDirectories.Concat(Directory.GetFiles(directoryPath)).ToArray();
```
pictos, line 373: Do we need to call `ToArray` here?

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,390 @@
...
+		return false;
+	}
+
+	static bool CreateSubDirectory(string newDirectory)
+	{
+		Directory.CreateDirectory(newDirectory);
+		return Directory.Exists(newDirectory);
+	}
+
+	List<string> GetDirectories(string path)
```
pictos, line 357: can we return `IEnumerable<string>` instead?

src/CommunityToolkit.Maui/Essentials/FolderPicker/FolderPickerImplementation.android.cs
```diff
@@ -0,0 +1,41 @@
...
+	/// <inheritdoc />
+	public Task<Folder> PickAsync(CancellationToken cancellationToken)
+	{
+		return PickAsync(GetExternalDirectory(), cancellationToken);
+	}
+
+	static string GetExternalDirectory()
+	{
+		return Platform.CurrentActivity?.GetExternalFilesDir(null)?.ParentFile?.ParentFile?.ParentFile?.ParentFile?.AbsolutePath
+				?? "/storage/emulated/0";
```
pictos, line 39: this `"/storage/emulated/0"` is always safe? Would be better to ask the user to provide a default path?

src/CommunityToolkit.Maui/Essentials/FolderPicker/FolderPickerImplementation.macios.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace CommunityToolkit.Maui.Storage;
+
+/// <inheritdoc />
+public class FolderPickerImplementation : IFolderPicker
+{
+	/// <inheritdoc />
+	public async Task<Folder> PickAsync(string initialPath, CancellationToken cancellationToken)
+	{
+		var documentPickerViewController = new UIDocumentPickerViewController(new[] { UTTypes.Folder })
```
pictos, line 15: can we cache this array, instead of creating a new one everytime?

src/CommunityToolkit.Maui/Essentials/FolderPicker/FolderPickerImplementation.macios.cs
```diff
@@ -0,0 +1,50 @@
...
+		};
+
+		documentPickerViewController.WasCancelled += (s, e) =>
+		{
+			taskCompetedSource.SetException(new FolderPickerException("Operation cancelled"));
+		};
+
+		currentViewController?.PresentViewController(documentPickerViewController, true, null);
+
+		return await taskCompetedSource.Task;
```
pictos, line 42: can we just return the `Task` instead of awaiting it here?

src/CommunityToolkit.Maui/Essentials/FolderPicker/FolderPickerImplementation.macios.cs
```diff
@@ -0,0 +1,50 @@
...
+using UniformTypeIdentifiers;
+
+namespace CommunityToolkit.Maui.Storage;
+
+/// <inheritdoc />
+public class FolderPickerImplementation : IFolderPicker
+{
+	/// <inheritdoc />
+	public async Task<Folder> PickAsync(string initialPath, CancellationToken cancellationToken)
+	{
```
pictos, line 14: On iOS we don't need to ask for permission?

src/CommunityToolkit.Maui/Essentials/FolderPicker/FolderPickerImplementation.tizen.cs
```diff
@@ -0,0 +1,46 @@
...
+	{
+		string? externalDirectory = null;
+		foreach (var storage in Tizen.System.StorageManager.Storages)
+		{
+			if (storage.StorageType == Tizen.System.StorageArea.External)
+			{
+				externalDirectory = storage.RootDirectory;
+			}
+		}
+		return externalDirectory ?? "/home/owner/media/";
```
pictos, line 44: Is "/home/owner/media/" path always valid?

src/CommunityToolkit.Maui/Essentials/SaveFileDialog/SaveFileDialogImplementation.android.cs
```diff
@@ -0,0 +1,42 @@
...
+		var currentActivity = Platform.CurrentActivity ?? throw new InvalidOperationException($"{nameof(Platform.CurrentActivity)} cannot be null");
+		var dialog = new FileFolderDialog(currentActivity, FileSelectionMode.FileSave, fileName);
+		var path = await dialog.GetFileOrDirectoryAsync(GetExternalDirectory(), cancellationToken: cancellationToken);
+
+		if (string.IsNullOrEmpty(path))
+		{
+			throw new FileSaveException("Path doesn't exist");
+		}
+
+		await WriteStream(stream, path, cancellationToken);
```
pictos, line 28:
```suggestion
		await WriteStream(stream, path, cancellationToken).ConfigureAwait(false);
```

src/CommunityToolkit.Maui/Essentials/SaveFileDialog/SaveFileDialogImplementation.shared.cs
```diff
@@ -0,0 +1,21 @@
+namespace CommunityToolkit.Maui.Storage;
+
+public partial class SaveFileDialogImplementation
+{
+	static async Task WriteStream(Stream stream, string filePath, CancellationToken cancellationToken)
+	{
+		await using var fileStream = new FileStream(filePath, FileMode.OpenOrCreate);
```
pictos, line 7:
```suggestion
		await using var fileStream = new FileStream(filePath, FileMode.OpenOrCreate).ConfigureAwait(false);
```

src/CommunityToolkit.Maui/Essentials/SaveFileDialog/SaveFileDialogImplementation.tizen.cs
```diff
@@ -0,0 +1,46 @@
...
+
+		var dialog = new FileFolderDialog(FileSelectionMode.FileSave, initialPath, fileName: fileName, cancellationToken: cancellationToken);
+		var path = await dialog.Open();
+
+		if (string.IsNullOrEmpty(path))
+		{
+			throw new FileSaveException("Path doesn't exist");
+		}
+
+		await WriteStream(stream, path, cancellationToken);
```
pictos, line 25:
```suggestion
		await WriteStream(stream, path, cancellationToken).ConfigureAwait(false);
```

src/CommunityToolkit.Maui/Essentials/SaveFileDialog/SaveFileDialogImplementation.windows.cs
```diff
@@ -0,0 +1,41 @@
...
+		savePicker.SuggestedFileName = GetFileName(fileName);
+
+		var file = await savePicker.PickSaveFileAsync();
+
+		if (string.IsNullOrEmpty(file?.Path))
+		{
+			throw new FileSaveException("Path doesn't exist");
+		}
+
+		await WriteStream(stream, file.Path, cancellationToken);
```
pictos, line 33:
```suggestion
		await WriteStream(stream, file.Path, cancellationToken).ConfigureAwait(false);
```

pictos (changes requested):
@VladislavAntonyuk thanks for moving this forward, I've a couple of general questions/suggestions, they are:

1. Why this implementation isn't on `Core` project? I couldn't find any `Microsoft.Maui.Controls` reference on your implementation, may I miss something?
2. I see that on some platforms you've to create the dialog and the strings, but on platforms that you don't need that like Windows, the text on the dialog will be localized and your implementation don't, would be great if you can _fix_ that.
3. I see a lot of events using lambdas, which can cause a lot of memory leaks, since (at least for me) is very hard to say if the objects in the mix have the same lifetime, so to be safe I recommend you use methods or local functions and make sure that you unsubscribe those events.
4. About the `Directory` logic, can we use the System.IO at .net6 level, that way we can have one implementation to rule all platforms, making it easier to maintain.

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
```diff
@@ -0,0 +1,397 @@
...
+using Button = Android.Widget.Button;
+using Color = Android.Graphics.Color;
+using Environment = Android.OS.Environment;
+using File = Java.IO.File;
+using IOException = Java.IO.IOException;
+using View = Android.Views.View;
+
+namespace CommunityToolkit.Maui.Storage;
+
+enum FileSelectionMode
```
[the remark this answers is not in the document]
VladislavAntonyuk: It is now used only in Tizen file. we can keep it there

src/CommunityToolkit.Maui/Essentials/FolderPicker/FolderPickerImplementation.tizen.cs
VladislavAntonyuk: I suppose yes. it is the default emulated path, like /storage/emulated/0 on Android

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.tizen.cs
VladislavAntonyuk: Android implementation is reworked from scratch. Now this enum is only used in this file

brminnick: Oooh! You could maybe use the new `file` scope, introduced in C# 11 https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-11.0/file-local-types

src/CommunityToolkit.Maui/Essentials/FolderPicker/FolderPickerImplementation.android.cs
VladislavAntonyuk: it is only used in case user wants to use default location (see line 44). We also have a method where user can set his own path

src/CommunityToolkit.Maui/Essentials/FolderPicker/FolderPickerImplementation.macios.cs
VladislavAntonyuk: iOS file storage is protected by default. we work in sandbox, so path permission is not needed. Also macOS asks for it automatically when call controller

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.android.cs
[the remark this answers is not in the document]
pictos: I prefer to move it to a separated file, it's hard to ask external contributors to follow a pattern if we don't have one. Even if this Tizen only, we should move it to a `primitive` folders or something like that

JoonghyunCho:
Let me test on some different Tizen devices to see if it is better for Tizen to switch dialogs to system program like your Android updates, or just to stay as is.

JoonghyunCho:
> Let me test on some different Tizen devices to see if it is better for Tizen to switch dialogs to system program like your Android updates, or just to stay as is.

Decided to stay as is to cover as many device types as possible.

src/CommunityToolkit.Maui/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -44,28 +33,36 @@ sealed class FileFolderDialog : Popup<string>, IDisposable
...
+				externalDirectory = storage.RootDirectory;
+				break;
+			}
+			else if (storage.StorageType == Tizen.System.StorageArea.Internal)
+			{
+				externalDirectory = storage.RootDirectory;
+			}
 		}
+
+		return externalDirectory!;
```
pictos, line 52: we don't use the `!` operator. here you can do something like
```suggestion
		return externalDirectory ?? throw new InvalidOperationException("Huston we have a problem!");
```

Directory.Build.props
```diff
@@ -20,7 +20,7 @@
   <ItemGroup>
     <InternalsVisibleTo Include="CommunityToolkit.Maui.UnitTests" />
   </ItemGroup>
-  <PropertyGroup Condition="'$(TF_BUILD)' == 'true' and $([MSBuild]::IsOSPlatform('windows')) == 'true'">
```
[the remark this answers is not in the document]
VladislavAntonyuk: I would suggest removing this condition, otherwise, the developer will have a failed build on CI but a green build locally, which may be confusing.
as we only need to install the script once, I don't think it is an issue.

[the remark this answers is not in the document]
JoonghyunCho: That is weird. Is the CI build failed with including the condition? I expected Tizen to be included in the CI build and excluded on the local build.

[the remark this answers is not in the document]
JoonghyunCho: Looks like the build will be ok with the condition in both CI and local.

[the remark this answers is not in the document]
VladislavAntonyuk: If you miss Tizen file on local env, CI will fail.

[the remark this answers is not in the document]
JoonghyunCho: Oh, you meant when you add a new features? All right, I'm Ok if other members are not skeptical. 😂

Directory.Build.props
```diff
@@ -3,8 +3,7 @@
   <PropertyGroup>
     <!-- Fixes https://github.com/dotnet/maui/pull/12114 -->
     <PublishReadyToRun>false</PublishReadyToRun>
-
-    <LangVersion>latest</LangVersion>
+    <LangVersion>preview</LangVersion>
```
pictos, line 6: can this break our users somehow? I would suggest to keep it at latest, and open an proposal for this change

src/CommunityToolkit.Maui.UnitTests/Mocks/FolderPickerImplementationMock.cs
```diff
@@ -0,0 +1,22 @@
+﻿using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.Core.Primitives;
+using CommunityToolkit.Maui.Storage;
+
+namespace CommunityToolkit.Maui.UnitTests.Mocks;
+class FolderPickerImplementationMock : IFolderPicker
+{
+	public Task<Folder> PickAsync(string initialPath, CancellationToken cancellationToken)
```
pictos, line 8: Can we change it to be ValueTask?

src/CommunityToolkit.Maui.UnitTests/Mocks/SaveFileDialogImplementationMock.cs
```diff
@@ -0,0 +1,19 @@
+﻿using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.Storage;
+
+namespace CommunityToolkit.Maui.UnitTests.Mocks;
+
+class SaveFileDialogImplementationMock : ISaveFileDialog
+{
+	public Task<string> SaveAsync(string initialPath, string fileName, Stream stream, CancellationToken cancellationToken)
```
pictos, line 8: Can we change this to be ValueTask?

Directory.Build.props
VladislavAntonyuk: LangVersion latest uses c# 10, but we have features from c# 11

pictos: C#11 is already stable

VladislavAntonyuk:
> @VladislavAntonyuk, friendly reminder about those questions that I did for my first review:
> 
> 1.Why this implementation isn't on Core project? I couldn't find any Microsoft.Maui.Controls reference on your implementation, may I miss something?
2. I see that on some platforms you've to create the dialog and the strings, but on platforms that you don't need that like Windows, the text on the dialog will be localized and your implementation don't, would be great if you can fix that.
3.  I see a lot of events using lambdas, which can cause a lot of memory leaks, since (at least for me) is very hard to say if the objects in the mix have the same lifetime, so to be safe I recommend you use methods or local functions and make sure that you unsubscribe those events. 
4. About the Directory logic, can we use the System.IO at .net6 level, that way we can have one implementation to rule all platforms, making it easier to maintain.

1. Fixed
2. Fixed. Created public methods, that should allow users to set their values. It is used only on Tizen.
3. Fixed
4. Fixed.

brminnick (changes requested):
Thanks Vlad! 

My review just has a couple suggestions, the biggest refactor is around returning `Task` instead of `ValueTask` for `SaveAsync()` and `PickAsync` since all of the platform-specific implementations are using `async`/`await`.

### Possible Naming Change

I want to get your thoughts on changing the names:
- `FolderPicker` -> `Folder`
- `SaveFileDialog` -> `File`

With these new names, we'd have `Folder.PickAsync()` and `File.SaveAsync()` which still make sense. Plus, it would create the opportunity for additional methods in the future like `Folder.DeleteAsync()` and `File.DeleteAsync()`.

samples/CommunityToolkit.Maui.Sample/Platforms/Android/AndroidManifest.xml
```diff
@@ -2,4 +2,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 	<application android:allowBackup="true" android:icon="@mipmap/appicon" android:roundIcon="@mipmap/appicon_round" android:supportsRtl="true"></application>
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+	<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
+	<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
```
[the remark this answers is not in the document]
pictos: This isn't a good practice. Because if the app doesn't use this feature and reference our lib this permission will be added to their app, and can be refused by app stores.

src/CommunityToolkit.Maui.Core/Essentials/FolderPicker/FolderPickerImplementation.android.cs
```diff
@@ -0,0 +1,64 @@
...
+using CommunityToolkit.Maui.Core.Primitives;
+using Microsoft.Maui.ApplicationModel;
+using AndroidUri = Android.Net.Uri;
+
+namespace CommunityToolkit.Maui.Storage;
+
+/// <inheritdoc />
+public class FolderPickerImplementation : IFolderPicker
+{
+	const int requestCodeFolderPicker = 12345;
```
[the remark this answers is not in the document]
pictos: This value is an `id` that we can use to check from where the request came, if you override the `OnActivityResult` method. As you can see [here](https://github.com/xamarin/Essentials/blob/main/Xamarin.Essentials/Platform/Platform.android.cs#L32-L35) we need this value when we want a value back from an Intent or Activity.

a sample of [usage](https://stackoverflow.com/a/14148838)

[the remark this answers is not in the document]
pictos: Also, as Essential does, I recommend us to set a range for all `requestCode`, something starting from 2000 ~ 3000, and not choosing random numbers

VladislavAntonyuk:
> Thanks Vlad! 
> 
> 
> 
> My review just has a couple suggestions, the biggest refactor is around returning `Task` instead of `ValueTask` for `SaveAsync()` and `PickAsync` since all of the platform-specific implementations are using `async`/`await`.
> 
> 
> 
> ### Possible Naming Change
> 
> 
> 
> I want to get your thoughts on changing the names:
> 
> - `FolderPicker` -> `Folder`
> 
> - `SaveFileDialog` -> `File`
> 
> 
> 
> With these new names, we'd have `Folder.PickAsync()` and `File.SaveAsync()` which still make sense. Plus, it would create the opportunity for additional methods in the future like `Folder.DeleteAsync()` and `File.DeleteAsync()`.

I suggest keeping the current names. First it follows existing FilePicker from MAUI Essentials, Second we have a class called Folder. We'll need to move all methods to that class. Folder.PickAsync makes sense in case you do it on defined folder. But the idea is about selecting folder. For that we need a folder picker.

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,399 @@
+﻿using Microsoft.Maui.Controls.Compatibility.Platform.Tizen;
```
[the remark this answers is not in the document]
VladislavAntonyuk: removed

samples/CommunityToolkit.Maui.Sample/Platforms/Tizen/tizen-manifest.xml
```diff
@@ -1,15 +1,18 @@
...
+        <label>maui-application-title-placeholder</label>
+        <icon>maui-appicon-placeholder</icon>
+        <metadata key="http://tizen.org/metadata/prefer_dotnet_aot" value="true" />
+        <splash-screens />
+    </ui-application>
+    <shortcut-list />
+    <privileges>
+        <privilege>http://tizen.org/privilege/internet</privilege>
+        <privilege>http://tizen.org/privilege/mediastorage</privilege>
+        <privilege>http://tizen.org/privilege/externalstorage</privilege>
```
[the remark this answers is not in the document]
VladislavAntonyuk: This isn't a good practice. Because if the app doesn't use this feature and reference our lib this permission will be added to their app, and can be refused by app stores.

demooooooo:
Hi, I am using PickFolder for selecting android folders. This is all working fine. But when I get the path of the file inside the folder, it reports an error that Access to the path '/storage/emulated/0/xxx' is denied. my code to read the file is as follows
`var pickedFolder = await _folderPicker.PickFolder();
var files = Directory.GetFiles(pickedFolder.Path);
           foreach (var item in files)
            {
            }`

VladislavAntonyuk:
@demooooooo Hi, .NET Directory class has a limited number of features that work on all platforms. With Android 33 Google implemented a "Virtual path", so you cannot simply traverse all files and folders. You should use Android classes to work with files and folders.
See example https://github.com/CommunityToolkit/Maui/blob/83e30c9870b7a1daa26f5fe5966dbc76d9f7ad01/src/CommunityToolkit.Maui.Core/Essentials/SaveFileDialog/SaveFileDialogImplementation.android.cs#L74 and https://github.com/VladislavAntonyuk/MauiSamples/blob/main/MauiPaint/Platforms/Android/Services/FileDialog.cs

brminnick:
Hey Vlad! After giving it some thought, let's change `ISaveFileDialog` to `IFileSaver` and merge this PR. `IFileSaver` better aligns to our `IFolderPicker` and .NET MAUI's `IFilePicker`.

If/when this feature gets promoted into .NET MAUI, the team would likely move `IFileSaver.SaveAsync()` into their existing interface, `IFilePicker.SaveAsync()`, but we can save that discussion for a future day.

Thanks again! And don't forget to update the `IFileSaver` text in the sample app and the docs 👍

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,398 @@
...
+using ScrollView = Tizen.UIExtensions.NUI.ScrollView;
+using Shadow = Tizen.NUI.Shadow;
+using HorizontalAlignment = Tizen.NUI.HorizontalAlignment;
+using VerticalAlignment = Tizen.NUI.VerticalAlignment;
+using View = Tizen.NUI.BaseComponents.View;
+using Window = Tizen.NUI.Window;
+
+namespace CommunityToolkit.Maui.Storage;
+
+sealed class FileFolderDialog : Popup<string>
```
brminnick, line 19: Hey @JoonghyunCho! Could you confirm that `FileFolderDialog.tizen.cs` only contains Tizen UI Code and doesn't contain any .NET MAUI UI code. I.e., It doesn't contain any code  from the `Microsoft.Maui.Controls` namespace like `Microsoft.Maui.Controls.Button`.

I think we're good, but I just want to double-check that we're not accidentally putting .NET MAUI UI code in `CommunityToolit.Maui.Core`: 
https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#where-to-make-your-changes

[](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#where-to-make-your-changes)

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,398 @@
...
+			var newFolderButton = new Button
+			{
+				Focusable = true,
+				Text = "New Folder",
+				TextColor = TColor.Black,
+				BackgroundColor = TColor.Transparent.ToNative(),
+				Margin = new Extents(margin1, margin1, 0, 0),
+				WidthSpecification = LayoutParamPolicies.MatchParent
+			};
+			newFolderButton.Clicked += async (sender, args) =>
```
pictos, line 165: possible memory leak here. The `content` and this button will be alive for how long? Looking is hard to tell the life-time of this object, so would be good to unsubscribe this event when it's no needed anymore.

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,398 @@
...
+			{
+				try
+				{
+					var newDirectoryName = await new PromptPopup("New Folder Name", "").Open();
```
pictos, line 169: Is this some title? If so isn't localized and will not be aligned with other platforms that have the dialog out of the box, like Windows

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,398 @@
...
+					if (CreateSubDirectory(newDirectoryPath + newDirectoryName))
+					{
+						UpdateDirectoryScrollView(selectedPath);
+					}
+				}
+				catch (Exception ex)
+				{
+					new Tizen.Applications.ToastMessage
+					{
+						Message = $"Failed to create new folder, {ex.Message}"
```
pictos, line 185: This isn't localized

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,398 @@
...
+		};
+		cancelButton.TextLabel.PixelSize = 15d.ToPixel();
+		cancelButton.SizeWidth = cancelButton.TextLabel.NaturalSize.Width + 15d.ToPixel() * 2;
+		cancelButton.Clicked += (s, e) => SendCancel();
+		hlayout.Add(cancelButton);
+
+		var okButton = new Button
+		{
+			Focusable = true,
+			Text = isFileSelectionMode ? "Save" : "OK",
```
pictos, line 236: This string isn't localized... And there're others that I'll not comment to not spam the review

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,398 @@
...
+
+		foreach(var view in directoyViews)
+		{
+			directoryScrollView.ContentContainer.Remove(view);
+		}
+		directoyViews.Clear();
+
+		var listItems = GetDirectories(path);
+
+		if (listItems is not null)
```
pictos, line 325: can we invert this if, to better readability

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
```diff
@@ -0,0 +1,398 @@
...
+				directoryScrollView.ContentContainer.Add(itemLabel);
+				directoyViews.Add(itemLabel);
+			}
+			directoryScrollView.SizeHeight = 30d.ToPixel() * Math.Min(listItems.Count(), 5);
+		}
+	}
+
+	IEnumerable<string> GetDirectories(string path)
+	{
+		var directories = new List<string>();
```
pictos, line 367: instead of creating a collection, can we use the `yield return`?

src/CommunityToolkit.Maui.Core/Essentials/FolderPicker/FolderPickerImplementation.android.cs
```diff
@@ -0,0 +1,63 @@
...
+
+	static string EnsurePhysicalPath(AndroidUri? uri)
+	{
+		if (uri is null)
+		{
+			throw new FolderPickerException("Path is not selected.");
+		}
+		
+		const string uriSchemeFolder = "content";
+		if (uri.Scheme != null && uri.Scheme.Equals(uriSchemeFolder, StringComparison.OrdinalIgnoreCase))
```
pictos, line 55:
```suggestion
		if (uri.Scheme is not null && uri.Scheme.Equals(uriSchemeFolder, StringComparison.OrdinalIgnoreCase))
```

src/CommunityToolkit.Maui.Core/Essentials/FolderPicker/FolderPickerImplementation.windows.cs
```diff
@@ -0,0 +1,42 @@
...
+		WinRT.Interop.InitializeWithWindow.Initialize(folderPicker, Process.GetCurrentProcess().MainWindowHandle);
+		folderPicker.FileTypeFilter.Add("*");
+		var folderPickerOperation = folderPicker.PickSingleFolderAsync();
+
+		void CancelFolderPickerOperation()
+		{
+			folderPickerOperation.Cancel();
+		}
+
+		cancellationToken.Register(CancelFolderPickerOperation);
```
pictos, line 27: We need to Dispose of the cancellationTokenRegistration, the return type of `Register` method
```suggestion
		var ctsRegistration = cancellationToken.Register(CancelFolderPickerOperation);
		//...
		ctsRegistration.Dispose();
```

src/CommunityToolkit.Maui.Core/Essentials/SaveFileDialog/SaveFileDialogImplementation.android.cs
```diff
@@ -0,0 +1,88 @@
...
+
+	static AndroidUri EnsurePhysicalPath(AndroidUri? uri)
+	{
+		if (uri is null)
+		{
+			throw new FolderPickerException("Path is not selected.");
+		}
+
+		const string uriSchemeFolder = "content";
+		if (uri.Scheme != null && uri.Scheme.Equals(uriSchemeFolder, StringComparison.OrdinalIgnoreCase))
```
pictos, line 64:
```suggestion
		if (uri.Scheme is not null && uri.Scheme.Equals(uriSchemeFolder, StringComparison.OrdinalIgnoreCase))
```

src/CommunityToolkit.Maui.Core/Essentials/SaveFileDialog/SaveFileDialogImplementation.macios.cs
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	/// <inheritdoc/>
+	public Task<string> SaveAsync(string fileName, Stream stream, CancellationToken cancellationToken)
+	{
+		return SaveAsync("/", fileName, stream, cancellationToken);
+	}
+
+	/// <inheritdoc />
+	public void Dispose()
```
pictos, line 37: who calls the Dispose? If it's called when the service is disposed those events will be subscribed to more than once.

src/CommunityToolkit.Maui.Core/Essentials/SaveFileDialog/SaveFileDialogImplementation.windows.cs
```diff
@@ -0,0 +1,48 @@
...
+		savePicker.SuggestedFileName = GetFileName(fileName);
+
+		var filePickerOperation = savePicker.PickSaveFileAsync();
+
+		void CancelFilePickerOperation()
+		{
+			filePickerOperation.Cancel();
+		}
+
+		cancellationToken.Register(CancelFilePickerOperation);
```
pictos, line 32: We need to dispose it

pictos (changes requested):
After another round of reviews, I have those suggestions

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
JoonghyunCho: Thanks for the double check! No Maui code is included in the `FileFolderDialog.tizen.cs`. It only contains Tizen UI code (Tizen.NUI).

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
JoonghyunCho: Tizen doesn't have the way to get the localized text from the system that's commonly available on all target device profiles... I think I will think about how I can handle this. Or maybe providing the methods or properties to get the text.

VladislavAntonyuk: I added it for 2 strings: Save and OK.
I haven't found a better solution rather then creating a static method to setup values

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
JoonghyunCho: Yes, I've wanted directories to be sorted, but I don't see it is necessary now.

src/CommunityToolkit.Maui.Core/Essentials/FileFolderDialog.tizen.cs
JoonghyunCho: sure

dersia:
I just saw the discussion about this issue in the communitytoolkit standup video on YouTube and to me this feels a bit weird. This is about a dialog and to me it does not make sense at all that a dialog is responsible for saving files. The dialog should give the UI and return the chosen path to me and saving/opening/deleting etc should happen in a different `FileManger` or just using the bcl. Same for the folder picker, I understand the point of @VladislavAntonyuk that you pick a folder and therefor picker makes sense, but it is also a dialog and therefor I think it should also be called `FolderPickerDialog`.

But this is just me, maybe I am to puristic about this kind of things 😅

VladislavAntonyuk:
@dersia thank you for your feedback. FolderPicker is chosen because .NET MAUI already has a FilePicker.
SaveFileDialog is renamed to FileSaver. When user clicks Save button in dialog, I expect to receive a path to the saved file (with its content - no other actions required). Assume Save operation takes 1 minute. User clicked Save button. Then he tries to make changes to the file. But it is not saved yet. Fix me if I am wrong but The result of the operation should already contain a saved file.

dersia:
@VladislavAntonyuk thanks for your reply! 

I totally understand your point, but from a DEV perspective, a dialog is a dialog, UI element that gives me an overlay experience over the app, and I see a lot of value in such an UI that is probably extendable by the DEV. However doing file operations like saving, opening (reading) and so on, should in my opionon be something that is done by non UI tech, because I might want that experience by its own, without presenting the user an UI. So taking the case that you are describing, I would rather expect some sort of `FileManager` or such, that gives me the operations for saving, opening and probably SaveAndOoenAsync (I am not sure what I would need that for because for the saving I must have had the content already, right?, but maybe there are use cases I can't think of right now). The responsibility of UI presentation and the file operations are mixed in this. It would be up to be so show a loading spinner until the file is saved and refetched from the io.
The dialog cloud however provide the loading spinner and give the dev a callback function that is called and the spinner is show as part of the dialog UI until the callback finishes (successfully or not), this could also give the dev the opportunity to display an nice looking error UI in case saving or opening failed.

Think of a downloading and save to file scenario. I would present the user a file save dialog, the user gets to choose where to save that file and I download and save the file. Afterwards I would like to directly open the files content, or just use am image and set the image source from the returned path. But this could also be the case for a temporary file, where I would like to have the `SaveAndOpenAsync` method without an dialog or the user to choose. I see value in both things being separate.

With that said, I prefer `IFileSaver` over `IFileDialog` but honestly from a devs perspective I would rather look for something called dialog if I wanted an UI experience that is an dialog.

I don't this feature to be delayed, but I think it makes much more sense to take the time and rethink this and split dialogs and functionality. 😊

brminnick:
> from a DEV perspective, a dialog is a dialog

@dersia To clarify - we're not creating any UI; we're not creating any dialogs. 

This PR creates two platform-specific methods, `IFolderPicker.PickAsync()` and `IFileSaver.SaveAsync()`, that enable a developer to pick a folder / save a file using the native iOS/Android/MacCatalyst/Windows/Tizen user interfaces.

In other words, we are not creating any .NET MAUI UI controls.

This PR is similar to [.NET MAUI's `Microsoft.Maui.Storage.IFilePicker`](https://github.com/dotnet/maui/blob/34a8eb02a6921d9883632063eb3e364735330149/src/Essentials/src/FilePicker/FilePicker.shared.cs#L16-L45) which is a platform-specific method that allows the developer to use the native file system to select a file.

I encourage you to check out the [`FileSaver` sample](https://github.com/CommunityToolkit/Maui/pull/699/files#diff-f2e4634fd60ec3ab3fcbb73c59459fba343979b126b97b515617f2bc700cbe61) and [`FolderPicker` sample](https://github.com/CommunityToolkit/Maui/pull/699/files#diff-563d420036113e36e74358330453c0e238ac8d558a79ea0fca5920c604ff6683) that Vlad has added to `CommunityToolkit.Maui.Sample` in this PR to experience and understand the usage of these methods. I also encourage you to watch our monthly standup which are live-streamed to the .NET Foundation YouTube channel to get further insight into this feature as well as other features we are working on: https://github.com/CommunityToolkit/Maui/wiki/Monthly-Standups#january-2023

dersia:
Hey @brminnick, thanks for the clarification.

In this case it all makes much more sense. I understood, that there is some UI involved and tgere for the dialog made sense. But if this is just bringing up the native UI, then this sound much more like a `IFileSaver`. 👍🏿

brminnick (approved):
Thanks Vlad!!! Approved!

brminnick (approved):

marcojak:
Congratulations of the approval.
Do you have an ETA for the new Release on nuget?

brminnick:
Thanks! No exact date is scheduled yet. 

We'll likely push out a new release next week pending the progress of https://github.com/CommunityToolkit/Maui/pull/667.

j03ybh:
Hey @brminnick ,

I am absolutely super excited for this feature. About one month ago we could expect a release depending on the progress of #667 , and in that item I could not see if it was already done or not. 

Is this feature already released, or do you have yet another confirmation/estimation of when we may expect it?

Many thanks in advance!

jfversluis:
@j03ybh these features are available today in v4.0.0! In fact, I just release a video about one of them:
https://youtu.be/Q9T-dRYq3Ps

j03ybh:
@jfversluis , thank you a bunch!

j03ybh:
@jfversluis, 

I tried to download the v4.0.0 nuget, however, it seems it is only compatible with .net7.0. I want to use it in a .NET MAUI Blazor app, and this can only be created with .NET 6.0. Is there a reason v4.0.0 is not compatible with .NET 6? Is this planned to be supported as well?

VladislavAntonyuk:
@j03ybn the last version with .net 6 was 2.0.0
Starting with version 3.0.0 we use only .net 7

Maui Blazor works in .net 7 as well.

j03ybh:
@VladislavAntonyuk, 

It is my first time rolling with the Community Kit! I did not know it was only supporting .NET  7 since such a while. 

Thanks for pointing out Maui Blazor is also supported, I think I just need to update my Visual Studio to the latest version to be able to choose the .net 7 template, or otherwise I will have to use CLI. 

Thanks for answering a likely silly question then!

PR #751, opened 2022-11-10
Fix Expander in CollectionView on Apple

### LINKED ISSUE #742: [BUG] Add `CollectionView` and `ListView` Support to `Expander`

### Current Behavior

When `Expander` is used inside of a `CollectionView` or a `ListView`, a `NotSupportedException` is thrown:
https://github.com/CommunityToolkit/Maui/blob/1ea488c9be7aac234af32e1ab5b014d6a52a5ff0/src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs#L197-L202

### Expected Behavior

When an `Expander` is expanded, its parent controls should re-measure their size and layouts.

Here is a incomplete sample at how we could approach solving this problem:

```cs
if (element is ListView listView)
{
  foreach (var child in listView.AllChildren.OfType<Cell>())
  {
	  child.ForceUpdateSize();
  }
  listView.InvalidateMeasureInternal(Microsoft.Maui.Controls.Internals.InvalidationTrigger.MeasureChanged);
}
else if (element is CollectionView collectionView)
{
  collectionView.InvalidateMeasureInternal(Microsoft.Maui.Controls.Internals.InvalidationTrigger.MeasureChanged);
}
```

### Steps To Reproduce

1. Download/Clone the [Expander-ItemsView-NotSupportedException branch in `CommunityToolkit.Maui`](https://github.com/CommunityToolkit/Maui/tree/Expander-ItemsView-NotSupportedException)
2. In Visual Studio, open `CommunityToolkit.Maui.Samples.sln`
3. Navigate to the [`ExpanderPage`](https://github.com/CommunityToolkit/Maui/blob/d93f2aa8cda296de2ae39e9c294b891065c7a6bc/samples/CommunityToolkit.Maui.Sample/Pages/Views/Expander/ExpanderPage.xaml#L51-L96)
4. Confirm the app crashes when the ExpanderPage opens

https://github.com/CommunityToolkit/Maui/blob/d93f2aa8cda296de2ae39e9c294b891065c7a6bc/samples/CommunityToolkit.Maui.Sample/Pages/Views/Expander/ExpanderPage.xaml#L51-L96

### Environment

```markdown
- .NET MAUI CommunityToolkit: v1.4.0
- OS: All
- .NET MAUI: 6.0.312
```

### Anything else?

This is a known-issue, documented in the .NET MAUI Community Toolkit documentation: https://learn.microsoft.com/dotnet/communitytoolkit/maui/views/expander

VladislavAntonyuk:
### Description of Change ###

Set valid size and invalidate layout of CollectionView

 ### Linked Issues ###
 

 - Fixes #742 

 ### PR Checklist ###
 

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
```diff
@@ -237,6 +212,45 @@ void SetHeaderGestures(in IView header)
...
+				{
+					layout.EstimatedItemSize = new CoreGraphics.CGSize(size.Width, size.Height);
+					layout.ItemSize = layout.EstimatedItemSize;
+					await Task.Delay(500);
+					layout.InvalidateLayout();
+				}
+
+				collectionView.InvalidateMeasureInternal(Microsoft.Maui.Controls.Internals.InvalidationTrigger.MeasureChanged);
+			}
+#endif
```
inforithmics, line 248: Isn't there a code path needed for Android Windows
#elif ANDROID ....
#elif WINDOWS ....

VladislavAntonyuk: other platforms seems to work correct without this additional recalculation

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
```diff
@@ -237,6 +212,45 @@ void SetHeaderGestures(in IView header)
...
+		{
+			if (element.Parent is ListView listView)
+			{
+				(element as Cell)?.ForceUpdateSize();
+			}
+#if IOS || MACCATALYST
+			else if (element is CollectionView collectionView)
+			{
+				var handler = collectionView.Handler as Microsoft.Maui.Controls.Handlers.Items.CollectionViewHandler;
+				var controller = handler?.GetType().BaseType?.BaseType?.BaseType?.BaseType?.BaseType?.GetProperty("Controller", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.GetProperty);
```
pictos, line 236: IMHO this doesn't look very solid. I would say the best solution is to work with someone .NET Maui team and try to expose the property that we need or see if there's other way to grab this information

VladislavAntonyuk: It is protected: `protected ItemsViewController<TItemsView> Controller { get; }`
image

pictos: So let's talk with the team about it. Any chance that you can send the link to the Maui code line that we need to have access to?

VladislavAntonyuk: src/Controls/src/Core/Handlers/Items/ItemsViewHandler.iOS.cs

https://github.com/dotnet/maui/blob/05ad340fd6b5e8d673da21642c4bdc178f890554/src/Controls/src/Core/Handlers/Items/ItemsViewHandler.iOS.cs#L37-L38

brminnick: @pictos / @VladislavAntonyuk Any progress on getting the needed code promoted to `public` or `internal` via the .NET MAUI engineering team?

pictos: no progress so far

pictos:
I tried on my end, and that didn't fix for maccatalyst, as you can see in the video below

https://user-images.githubusercontent.com/20712372/209261513-f835e72d-fb02-4141-b5f1-e5d2facbcdc4.mp4

VladislavAntonyuk:
Thank you @pictos. Seems like Task.Delay before InvalidateLayout was not a good solution, that works on all devices. I reworked it. Please check if it works for you

rmarinho:
We merged the change on MAUI

VladislavAntonyuk:
> We merged the change on MAUI

Thank you @rmarinho! Will it be included in the next release?

rmarinho:
I don't think so, unless we backport the fix to net7

pictos:
@VladislavAntonyuk before moving with this and the other PR that I opened, the fix didn't work well for ListView on maccatalyst

VladislavAntonyuk:
@pictos the fix is only for CollectionView. I also saw back merge of .NET MAUI to avoid reflection. 
I believe we should wait for that release to avoid the workarounds

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
```diff
@@ -132,12 +130,42 @@ public ExpandDirection Direction
 
 	Grid ContentGrid => (Grid)base.Content;
 
-	/// <inheritdoc/>
-	protected override async void OnParentChanged()
+	/// <summary>
+	/// Overridable method that executes each time the <see cref="Header"/> is tapped
+	/// </summary>
+	/// <param name="tappedEventArgs">The <see cref="TappedEventArgs"/> provided by <see cref="Expander.tapGestureRecognizer"/>.</param>
+	public virtual void OnHeaderTapped(TappedEventArgs tappedEventArgs)
```
brminnick, line 137: @pictos @VladislavAntonyuk 

FYI - I simplified the code from #890 a bit, removing the `public static Action`, replacing it with `public virtual void OnHeaderTapped()`.

I configured Expander to [run `OnHeaderTapped()` every time `Expander.tapGestureRecognizer` is triggered](https://github.com/CommunityToolkit/Maui/pull/751/files#diff-a7933be10a281abb7a9062799e5eb3b5e633518aaff6326c9a8d1b4adc829a6bR251).

This ensures that `Expander` supports `CollectionView` by default, and it allows users to override this method, implementing their own logic if they'd like. It also allows us to one day remove the code inside `OnHeaderTapped()` if/when .NET MAUI fixes the bug on their end without causing any breaking changes to MCT devs.

brminnick: We should also add a Unit Test for this new method since it is `public`, if we can

VladislavAntonyuk: With public virtual method user needs to inherit from Expander and create his own control.
With public static property he only need to create an action and assign it.
Public static property seems to be a simpler solution from user perspective. We can mark that property obsolete and remove it with major release.

brminnick: > With public virtual method user needs to inherit from Expander and create his own control. With public static property he only need to create an action and assign it. Public static property seems to be a simpler solution from user perspective

True, but let's zoom out a bit: How many developers are actually going to change the default behavior? My guess is not many, if any. 

If a developer is going to go through the trouble of changing the default behavior of `Expander`, I doubt they'll mind creating a custom Expander class. It doesn't require many lines of code:

```cs
class CustomExpander : Expander
{
  public virtual void OnHeaderTapped(TappedEventArgs tappedEventArgs)
  {
    base.OnHeaderTapped(tappedEventArgs);
    
    // custom code
  }
}
```

> We can mark that property obsolete and remove it with major release

My goal is to avoid us purposely adding a property that we knowingly will `[Obsolete]` later because `[Obsolete]` code adds technical debt and becomes a headache to manage in the code + the docs.

tl;dr - I prefer that we be selfish and avoid purposely adding technical debt, even if it requires a small bit more work for the developer to create a new subclass to override the default behavior

pictos: @brminnick, I would recommend keeping the `Action` approach, take a sample of how you can customize the PlatformView on .NET MAUI `EntryHandler.PlatformFactoryView = (handler) => return PlatformView;`, the idea .NET MAUI is to avoid inheritance and have external points of extensibility.

pictos: > My goal is to avoid us purposely adding a property that we knowingly will [Obsolete] later because [Obsolete] code adds technical debt and becomes a headache to manage in the codebase + the docs.

We don't need to Obsolete that, we can make it available for anyone that want to influence the Expand behavior, we will need to Obsolete the default methods when that be fixed.

brminnick: Let's again zoom out: Why are we even adding a public API? Are devs demanding this? 

If not, let's make `OnHeaderTapped` a private method.

I don't think many, if any, developers are going to modify the default functionality.

pictos: Sorry, but I still can't see your point be an argument against the usage of an Action instead of `virtual` method.

brminnick: My question is Why are we adding this new API? Are developers asking for this? Do we have an open Issue for this?

I understand it's a workaround to adding Expander support to CollectionView + ListView. But, this PR is created to _fix_ Expander support for CollectionView + ListView, right?

So once this PR is merged, we won't need this new API, right?

pictos: This PR, IMHO, doesn't **fix** the issue, it provides a workaround since the fix should be in the .NET MAUI. We merged a PR on MAUI to improve the workaround (for .net 8), making it possible to get access to the members that we need without reflection.
So, since this is a workaround devs can opt-in to use ours or create their own implementation, with that we don't have the _obligation_ to support all possible scenarios (I would imagine there's at least one scenario that our workaround will fail) so let's say we provide a fix for 80% of the cases and the other 20% will have an easy way to implement their own fix.

This new API (Action) is a point of extensibility that devs can use to implement their own delegate in other to control that behavior, the inspiration come from [here](https://github.com/dotnet/maui/blob/3bae70081e3bb2639cf558e8351454368ac811c5/src/Core/src/Handlers/View/ViewHandlerOfT.cs#L58), you can see the use at line 72 of the same file. The idea is to have a place that devs can customize without need to extend the control, what's normal on XF but not on .NET MAUI

brminnick: Ah - got it. 

Thanks man - I remember the conversation from our standup now. 

I'll clean up my mess in the morning 👍

brminnick: ✅ `public virtual void` reverted to `void`
✅ `Action` added with updated XML documentation

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
```diff
@@ -66,14 +67,23 @@ public Expander()
...
 	}
 
+	/// <summary>
+	/// The Action that fires when <see cref="Header"/> is tapped.
+	/// By default, this <see cref="Action"/> runs <see cref="ResizeExpanderInItemsView(TappedEventArgs)"/>.
+	/// </summary>
+	/// <remarks>
+	/// Warning: Overriding this <see cref="Action"/> may cause <see cref="Expander"/> to work improperly when placed inside of a <see cref="CollectionView"/> and placed inside of a <see cref="ListView"/>.
+	/// </remarks>
+	public Action<TappedEventArgs>? HandleHeaderTapped { get; set; }
```
pictos, line 85: this should be static or the dev will need to implement it for each instance of `Expander`
see 
- https://github.com/CommunityToolkit/Maui/blob/58240c8109f8d1bfe8313ddb4dc81898671ee728/src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs#L255

brminnick: True, but I think it's best to keep it non-static. 

If a dev is using multiple `Expander`s in their app, they may not want to override the behavior the same way for each. 

And if a dev_does want to implement the same behavior for every `Expander`, they can subclass it like so:

```cs
public class CustomExpander : Expander
{
  public CustomExpander()
  {
    this.HandleHeaderTapped = CustomHeaderTappedMethod;
  }
  
  void CustomHeaderTappedMethod(TappedEventArgs args)
  {
    // ...
  }
}
```

pictos: @brminnick  Can you double-check if the sample is still working? Since we use a collection, I believe we will create one per cell.

brminnick: Of course! 

Yup, here is `Expander` inside of a `CollectionView`, working on iOS + MacCatalyst:
image

src/CommunityToolkit.Maui/Views/Expander/Expander.macios.cs
```diff
@@ -0,0 +1,55 @@
...
+			if (clickedCell is null)
+			{
+				return;
+			}
+
+			for (int i = 0; i < cells.Length; i++)
+			{
+				var cell = cells[i];
+
+				if (i > 0)
```
pictos, line 31: is there a change to `i` be less than zero?

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
```diff
@@ -233,8 +212,48 @@ void HandleDirectionChanged(ExpandDirection expandDirection)
...
+		}
+
+		Element element = this;
+		var size = IsExpanded
+					? Measure(double.PositiveInfinity, double.PositiveInfinity, MeasureFlags.IncludeMargins).Request
+					: Header.Measure(double.PositiveInfinity, double.PositiveInfinity);
+
+		while (element is not null)
+		{
+			if (element.Parent is ListView && element is Cell cell)
```
pictos, line 239: @VladislavAntonyuk you mentioned [here](https://github.com/CommunityToolkit/Maui/pull/751#issuecomment-1380883728) this PR just fix for `CollectionView`, so why do we have a `ListView` handle here?

pictos (changes requested):
I've two suggestions. I just focused on the `Expander` related changed

src/CommunityToolkit.Maui/Views/Expander/Expander.macios.cs
VladislavAntonyuk: i can be equal to 0

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
VladislavAntonyuk: because for listview we still need to throw an unsupported exception on Apple.
But on other platforms, we can use it with ForceUpdateSize().

pictos (approved):

brminnick:
@VladislavAntonyuk Can you update the docs, changing it to "Expander is not supported inside `ListView`" now that only ListView is not supported?

image

AnthonyNjuguna:
This doesn't work if the collection has a header view. 

`var clickedCell = GetCellByPoint(cells, new CGPoint(tapLocation.Value.X, tapLocation.Value.Y));` 

the clickedCell always returns null.

pictos:
@AnthonyNjuguna can you please open a new bug issue?

PR #763, opened 2022-11-14
Use `[GeneratedRegex]`

brminnick:
New in .NET 7.0 is [.NET regular expression source generators](https://learn.microsoft.com/dotnet/standard/base-types/regular-expression-source-generators). 

Using `[GeneratedRegex]` uses source generators to bring large performance improvements, as noted in the [documentation](https://learn.microsoft.com/dotnet/standard/base-types/regular-expression-source-generators):
> But as can be seen, it's not just doing new Regex(...). Rather, the source generator is emitting as C# code a custom Regex-derived implementation with logic similar to what RegexOptions.Compiled emits in IL. You get all the throughput performance benefits of RegexOptions.Compiled (more, in fact) and the start-up benefits of Regex.CompileToAssembly, but without the complexity of CompileToAssembly. The source that's emitted is part of your project, which means it's also easily viewable and debuggable.

### Additional Info
This PR also adds the `RegexOptions.Compiled` flag. As noted by [Jonathan Peppers at .NET Conf](https://www.youtube.com/watch?v=J7bWH9LyaoU), the `RegexOptions.Compiled` flag previously was unused by Mono prior to .NET 7.0

### DIFF
--- a/src/CommunityToolkit.Maui/Behaviors/Validators/EmailValidationBehavior.shared.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/Validators/EmailValidationBehavior.shared.cs
@@ -9,7 +9,7 @@ namespace CommunityToolkit.Maui.Behaviors;
 /// The validation is achieved through a regular expression that is used to verify whether or not the text input is a valid e-mail address.
 /// It can be overridden to customize the validation through the properties it inherits from <see cref="ValidationBehavior"/>.
 /// </summary>
-public class EmailValidationBehavior : TextValidationBehavior
+public partial class EmailValidationBehavior : TextValidationBehavior
 {
 	/// <inheritdoc /> 
 	protected override async ValueTask<bool> ValidateAsync(string? value, CancellationToken token)
@@ -72,7 +72,7 @@ public class EmailValidationBehavior : TextValidationBehavior
 		try
 		{
 			// Normalize the domain
-			email = Regex.Replace(email, @"(@)(.+)$", DomainMapper, RegexOptions.None, TimeSpan.FromMilliseconds(200));
+			email = NormalizeDomainRegex().Replace(email, DomainMapper);
 		}
 		catch (RegexMatchTimeoutException)
 		{
@@ -85,9 +85,7 @@ public class EmailValidationBehavior : TextValidationBehavior
 
 		try
 		{
-			return Regex.IsMatch(email,
-				@"^[^@\s]+@[^@\s]+\.[^@\s]+$",
-				RegexOptions.IgnoreCase, TimeSpan.FromMilliseconds(250));
+			return ValidEmailRegex().IsMatch(email);
 		}
 		catch (RegexMatchTimeoutException)
 		{
@@ -104,7 +102,7 @@ public class EmailValidationBehavior : TextValidationBehavior
 			string domainName = idn.GetAscii(match.Groups[2].Value);
 
 			if (domainName.All(x => char.IsDigit(x) || x is '.')
-				&& !Regex.IsMatch(domainName, @"^([1-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])(\.([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])){3}$"))
+				&& !ValidIpv4Regex().IsMatch(domainName))
 			{
 				throw new ArgumentException("Invalid IPv4 Address.");
 			}
@@ -117,4 +115,13 @@ public class EmailValidationBehavior : TextValidationBehavior
 			return match.Groups[1].Value + domainName;
 		}
 	}
+
+	[GeneratedRegex(@"^([1-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])(\.([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])){3}$", RegexOptions.None, 250)]
+	private static partial Regex ValidIpv4Regex();
+
+	[GeneratedRegex(@"^[^@\s]+@[^@\s]+\.[^@\s]+$", RegexOptions.IgnoreCase, 250)]
+	private static partial Regex ValidEmailRegex();
+
+	[GeneratedRegex(@"(@)(.+)$", RegexOptions.None, 250)]
+	private static partial Regex NormalizeDomainRegex();
 }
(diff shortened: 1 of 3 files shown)


src/CommunityToolkit.Maui/Behaviors/Validators/EmailValidationBehavior.shared.cs
```diff
@@ -117,4 +115,13 @@ static string DomainMapper(Match match)
 			return match.Groups[1].Value + domainName;
 		}
 	}
+
+	[GeneratedRegex(@"^([1-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])(\.([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])){3}$", RegexOptions.None, 250)]
+	private static partial Regex ValidIpv4Regex();
```
VladislavAntonyuk, line 120: We don't use private modifiers explicitly, is it required?

src/CommunityToolkit.Maui/Converters/MathExpressionConverter/MathExpression.shared.cs
```diff
@@ -127,15 +123,12 @@ public double Calculate()
 		return stack.Pop();
 	}
 
+	[GeneratedRegex(@"(?<!\d)\-?(?:\d+\.\d+|\d+)|\+|\-|\/|\*|\(|\)|\^|\%|\,|\w+")]
+	private static partial Regex ValidExpressionRegex();
```
VladislavAntonyuk, line 127: Maybe we should get a more meaningful name like MathExpressionRegexPattern?

src/CommunityToolkit.Maui/Behaviors/Validators/EmailValidationBehavior.shared.cs
```diff
@@ -117,4 +115,13 @@ static string DomainMapper(Match match)
...
+
+	[GeneratedRegex(@"^[^@\s]+@[^@\s]+\.[^@\s]+$", RegexOptions.IgnoreCase, 250)]
+	private static partial Regex ValidEmailRegex();
```
VladislavAntonyuk, line 123: Do we need a valid prefix?

src/CommunityToolkit.Maui/Behaviors/Validators/EmailValidationBehavior.shared.cs
pictos: I would say this (and others Regex methods) should be `protected` since this class can be `inherited`

brminnick: > We don't use private modifiers explicitly, is it required?

Yup! For some reason, partial methods explicitly require modifiers even if the default is private. 🤷‍♂️

src/CommunityToolkit.Maui/Converters/MathExpressionConverter/MathExpression.shared.cs
brminnick: Agreed! To be honest, I named it this because I couldn't actually figure out what it does. (I don't really know much about reading/writing RegEx)

src/CommunityToolkit.Maui/Behaviors/Validators/EmailValidationBehavior.shared.cs
pictos: If I'm not wrong the only one that you don't need to add the access modifiers is when the return type is `void`. If I remember well, the lang. the team did it to not break old partial methods, so if your method returns something other than `void` the modifiers are required.

brminnick: > should be protected since this class can be inherited
Sure! I'm not against it.

The current `Regex`s are `private`, not `protected`. So this wouldn't be a breaking change, but we would want to add unit tests around the newly `protected Regex` methods to ensure that we don't accidentally break anything in future PRs.

pictos: Since the user can extend from `EmailValidationBehavior` I can see value in changing those `Regex` methods to be `protected`, that way the user can reuse then to perform more checks... Otherwise, they can't reuse anything from the base class.

Also, I'm fine if you aren't comfortable to do this right now, and we can wait someone from the community to request this change.

brminnick: Let's merge this PR and then I'll open a new PR to promote the `[GeneratedRegex]` methods to protected

src/CommunityToolkit.Maui/Converters/MathExpressionConverter/MathExpression.shared.cs
brminnick:
```suggestion
	private static partial Regex MathExpressionRegexPattern();
```

src/CommunityToolkit.Maui/Behaviors/Validators/EmailValidationBehavior.shared.cs
brminnick: What do you mean? Could you share an example?

src/CommunityToolkit.Maui/Behaviors/Validators/EmailValidationBehavior.shared.cs
brminnick: > Let's merge this PR and then I'll open a new PR to promote the `[GeneratedRegex]` methods to protected

Here's the branch where I've begun the work: https://github.com/CommunityToolkit/Maui/tree/Promote-%60-GeneratedRegex%5D%60-Methods-to-%60protected%60

src/CommunityToolkit.Maui/Behaviors/Validators/EmailValidationBehavior.shared.cs
VladislavAntonyuk: Just EmailRegex without Valid

VladislavAntonyuk: Because if we have “valid” email reflex there should be invalid email reflex which doesn’t make sense. So I suggest rename it to EmailRegex.

brminnick: ✅ Done [ValidEmailRegex -> EmailRegex](https://github.com/CommunityToolkit/Maui/pull/763/commits/e8042ed698ca632466d0f2b33752d9ccd14327cb)

VladislavAntonyuk (approved):

PR #787, opened 2022-11-26
Fix MediaElement Androids' player.Duration OverflowException and PlaybackState

Mephisztoe:
When streaming from an URL it is possible that ExoPlayer does not receive the information necessary to set the duration of a title. 
If this is the case, the ExoPlayers' duration property is set to TIME_UNSET which is a constant defined as Long.MIN_VALUE +1 (-9223372036854775807). 

However, the current MediaManager implementation for Android defined in **MediaManager.android.cs** handles the **OnPlayerStateChanged** event by setting the mediaElements' duration property if the playback state equals StateReady. It does so by converting the underlying players' duration property to  a TimeSpan.

In case ExoPlayers' duration property equals TIME_UNSET in that moment, the call to TimeSpan.FromMilliseconds results in an OverflowException.

More specifically:
**System.OverflowException:** 'TimeSpan overflowed because the duration is too long.'

This can be prevented by checking the duration beforehand:

`mediaElement.Duration = TimeSpan.FromMilliseconds(player.Duration == C.TimeUnset ? 0 : player.Duration);`

### DIFF
--- a/src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.android.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.android.cs
@@ -162,7 +162,7 @@ public partial class MediaManager : Java.Lang.Object, IPlayer.IListener
 		switch ((PlaybackStateCode)player.PlaybackState)
 		{
 			case PlaybackStateCode.Playing:
-				videoStatus = MediaElementState.Playing;
+				videoStatus = player.PlayWhenReady ? MediaElementState.Playing : MediaElementState.Paused;
 				break;
 
 			case PlaybackStateCode.Paused:
--- a/src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.android.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.android.cs
@@ -210,7 +210,8 @@ public partial class MediaManager : Java.Lang.Object, IPlayer.IListener
 
 		if (playbackState == IPlayer.StateReady)
 		{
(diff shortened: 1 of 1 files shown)


src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.android.cs
```diff
@@ -210,7 +210,8 @@ public void OnPlayerStateChanged(bool playWhenReady, int playbackState)
 
 		if (playbackState == IPlayer.StateReady)
 		{
-			mediaElement.Duration = TimeSpan.FromMilliseconds(player.Duration);
+			mediaElement.Duration = TimeSpan.FromMilliseconds(
+				player.Duration == C.TimeUnset ? 0 : player.Duration);
```
pictos, line 214: can we extract it to be an extension method? I can see value in doing that, since this is something very particular to exoplayer

pictos (commented):
Thanks @Mephisztoe for this❣️
Can you also assing the CLA?

src/CommunityToolkit.Maui.MediaElement/PlatformView/MediaManager.android.cs
jfversluis: Also, why not make it a little bit more fool proof and not only check for that specific value? I would assume anything less than 0 would be invalid as a duration

jfversluis:
I think the CLA check look OK now, so I'm going to assume it is

jfversluis:
I'll merge this for now and fix the feedback in the "main" branch so we don't keep reviewing very similar things across different branches and lose overview.

Thanks for this help!

PR #793, opened 2022-11-29
Update ToastDuration options documentation with actual duration

jfversluis:
While creating a video for Toast I noticed that this enum only specified "short" and "long" which isn't very meaningful to the user. Added the actual length to the description for this.

### DIFF
--- a/src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
+++ b/src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
@@ -22,12 +22,12 @@ public interface IToast : IAlert
 public enum ToastDuration
 {
 	/// <summary>
-	/// Displays Toast for a short time
+	/// Displays Toast for a short time (~2 seconds).
 	/// </summary>
 	Short,
 
 	/// <summary>
-	/// Displays Toast for a long time
+	/// Displays Toast for a long time (~3.5 seconds).
 	/// </summary>
 	Long
 }


src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
```diff
@@ -22,12 +22,12 @@ public interface IToast : IAlert
 public enum ToastDuration
 {
 	/// <summary>
-	/// Displays Toast for a short time
+	/// Displays Toast for a short time (~2 seconds).
```
VladislavAntonyuk, line 25: I would prefer not to use explicit value as it may be internally changed. On Android it is an enum Short and Long

jfversluis: Sure, but from the consumer side... How do I know what is short and long? Maybe it's just me, but I would love to have some indication how long it is. And yes, we would have to update as values change, isn't that always the case with comments?

VladislavAntonyuk (approved):
Just keep in mind we'll need to monitor duration of toast with each Android update.

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui.Core/Interfaces/IToast.shared.cs
VladislavAntonyuk: Comments are usually the place developers (maybe that's only me) forget to update.

jfversluis: You're absolutely right! And that will probably happen and someone will open a bug for it ;)

Thanks!

PR #801, opened 2022-12-01
Cancels silently without exception, because unhandled exceptions cras…

MrDeej:
This is a fix so that you dont crash a WinUI Maui App when you try to cycle trough the states so fast that it needs to use cancellationtoken.

### DIFF
--- a/src/CommunityToolkit.Maui.UnitTests/Layouts/StateContainerTests.cs
+++ b/src/CommunityToolkit.Maui.UnitTests/Layouts/StateContainerTests.cs
@@ -125,24 +125,6 @@ public class StateContainerTests : BaseTest
 	}
 
 	[Fact]
-	public async Task Controller_CanceledSwitchToStateThrowsException()
-	{
-		var cts = new CancellationTokenSource();
-		cts.Cancel();
-
-		await Assert.ThrowsAsync<OperationCanceledException>(() => controller.SwitchToState(StateKey.Loading, false, cts.Token));
-	}
-
-	[Fact]
-	public async Task Controller_CanceledSwitchToContentThrowsException()
-	{
-		var cts = new CancellationTokenSource();
-		cts.Cancel();
-
-		await Assert.ThrowsAsync<OperationCanceledException>(() => controller.SwitchToContent(false, cts.Token));
-	}
-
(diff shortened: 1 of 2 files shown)


brminnick:
Hey @MrDeej! We intentionally throw the exception to allow the developer to wrap `SwitchToContent` in their own try/catch block and handle the `TaskCanceledException` themselves. They may have a scenario where they do something different based on the token expiring or being canceled.

I don't see an open Issue for this. Are you saying that devs on WinUI can't catch this exception inside of a try/catch block?

MrDeej:
> Hey @MrDeej! We intentionally throw the exception to allow the developer to wrap `SwitchToContent` in their own try/catch block and handle the `TaskCanceledException` themselves. They may have a scenario where they do something different based on the token expiring or being canceled.
> 
> I don't see an open Issue for this. Are you saying that devs on WinUI can't catch this exception inside of a try/catch block?

Yeah, no you are right. I should rather just do that in the UI part of the app.

Generally I do think that exception throwing is bad practice, specially since token.IsCancellationRequested exits.
 But I think I was a little bit fast here, app crash is a red flag for me

pictos:
@MrDeej not throwing an exception when the cancelation is requested is an anti-pattern. We should throw the exception to:
1. Stop all the process chains;
2. Have a clear way that the consumer will know that operation was canceled and handle that. If we just return from the process chain, the consumer can think the _result_ is in a valid state, which is not.

I recommend this blog post series around this [topic](https://blog.stephencleary.com/2022/02/cancellation-1-overview.html).

With that I'll close this PR, please feel free to open an issue if you think that our implementation is wrong and in the future, we can reopen this PR if needed. Thanks for taking time to contribute with us

MrDeej:
@pictos 
Now that i look into it.

The StateContainer binds via a dependencyproperty, which means that all the logic for exception and cancellationtoken is within the scope of the control. 

And it is no reasonable way to catch the exception (neither on android/ios or windows), if I understand this correctly. 

Meaning that actually, this makes my original suggestion the only way.

If you agree to this I can reopen a fork, and fix this again.

pictos:
@MrDeej please, open an issue first following our template, that way we can keep track on that

PR #791, opened 2022-11-29
Make `TextColorTo` Generated Classes for .NET MAUI Controls Internal

brminnick:
### Description of Change ###

This PR changes generated classes for .NET MAUI controls, i.e.  code generated for `Microsoft.Maui.Controls.*` by `TextColorTo`, from `public` to `internal`.

The bug in https://github.com/CommunityToolkit/Maui.Markup/issues/158 is caused by two matching public classes being generated in both the .NET MAUI library (eg `net7.0-ios`) and the referenced class library (e.g. `net7.0`).

```
📦MyMauiSolution
 ┣ 📂MyMauiApp
    - <TargetFrameworks>net7.0-android;net7.0-ios;net7.0-maccatalyst</TargetFrameworks>
 ┗ 📂MyLibrary
    - <TargetFramework>net7.0</TargetFramework>
```

 ### Linked Issues ###
 

 - Fixes https://github.com/CommunityToolkit/Maui/discussions/756#discussioncomment-4125731

 ### PR Checklist ###
 
 - [x] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [x] Has tests (if omitted, state reason in description)
 - [x] Has samples (if omitted, state reason in description)
 - [x] Rebased on top of `main` at time of PR
 - [x] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui.Markup/blob/main/CONTRIBUTING.md#contributing-code---best-practices)

PR #813, opened 2022-12-06
"TextColorTo" Extension Methods Access Modifier Tests

loop8ack:
### Description of Change ###

This PR adds two tests for the access modifiers of the `TextColorTo` extension methods.
@brminnick asked me to write these tests. More about this in his PR: https://github.com/CommunityToolkit/Maui/pull/791

`AccessModifierForMauiControlsShouldNotBePublic`
If the control is a predefined Maui control, the generated class should be `internal`.
This prevents conflicts with multiple projects, since the same classes are generated in the referencing projects as well.

`AccessModifierForCustomControlsShouldMatchTheControl`
If the control is a self-defined control, the generated class should have the same access modifier as the control.
This allows the extension methods for publicly available controls to be used in other projects that do not generate the same code.

The first test fails for main branch because the generated code for Maui controls is always `public`.
When I use the code from @brminnick, the second test fails because the code for self-defined controls is always `internal`.
Both tests pass if the `public` is changed to `internal` in the [TextColorToGenerator.cs in line 60](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs#L60).

 ### Linked Issues ###

 - https://github.com/CommunityToolkit/Maui/discussions/756#discussioncomment-4125731

 ### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
+++ b/src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
@@ -120,7 +120,7 @@ using Microsoft.Maui.Graphics;
 
 namespace " + textStyleClass.Namespace + @";
 
-" + textStyleClass.ClassAcessModifier + @" static partial class ColorAnimationExtensions_" + textStyleClass.ClassName + @"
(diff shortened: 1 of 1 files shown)


brminnick:
Thanks @loop8ack!!

PR #815, opened 2022-12-08
Fix the Incremental behavior of our Source Generator

### LINKED ISSUE #785: [BUG] Very slow Intellisense & Code Completion in MAUI projects
---
**Issue moved from dotnet/maui#11592**
- Please respond to @vsfeedback.

---

_From @vsfeedback on Wednesday, November 23, 2022 7:37:51 PM_

EDIT: Adding context to this issue:

It looks like the source generator in MAUI Community Toolkit is causing slowdowns with Roslyn.

[TextColorToGenerator](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs)

The suggestion is to move to using `ForAttributeWithMetadataName` to create the analyzer which would mean creating an attribute and decorating any types you want the generator to run on with it.

This may reduce the usefulness of the generator towards the original goal, but I think short of making this change, perhaps we need to consider removing the generator, or otherwise making it opt-in somehow via msbuild property or something else.

--------

_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Very-slow-Intellisense--Code-Completion/10198504)._

---
[severity:It's more difficult to complete my work]
Working in MAUI projects within Visual Studio is extremely slow.
It already occurred in 17.3.* but even after upgrading to 17.4 I can wait ~ 10 seconds before code completoin or INtelliSense kick in.
Adding missing using-statements take up minutes now!

This does not happen in Console appplications, Xamarin projects, etc., *only* in MAUI solutions.

---
#### Feedback Bot on 11/14/2022, 00:42 AM: 

(private comment, text removed)
#### Roi Chen -MSFT [MSFT] on 11/15/2022, 01:14 AM: 

(private comment, text removed)
#### Schmidt, Helmut on 11/16/2022, 11:14 AM: 

(private comment, text removed)
#### Feedback Bot on 11/17/2022, 04:46 AM: 

(private comment, text removed)
#### Cyrus Najmabadi [MSFT] on 11/23/2022, 01:36 PM: 

(private comment, text removed)

---
### Original Solutions
(no solutions)

pictos:
### Description of Change ###

This PR fixes our SG, so it will re-generate code just when needed.

 ### Linked Issues ###
 

 - Fixes #785 

 ### PR Checklist ###
 

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -18,97 +26,80 @@ class TextColorToGenerator : IIncrementalGenerator
...
-				textStyleClassList.Add((declarationSymbol.Name, accessModifier, nameSpace, declarationSymbol.TypeArguments.GetGenericTypeArgumentsString(), declarationSymbol.GetGenericTypeConstraintsAsString()));
-			}
-		}
+		// Here we Collect all the Classes candidates from the first pipeline
+		// Then we merge then with the Maui.Controls that implements the desired interfaces
+		// Then we make sure they are unique and the user control doesn't inherit from any Maui control that implements the desired interface already
+		// Then we transform the ISymbol to be a type that we can compare and preserve the Incremental behavior of this Source Generator
+		var inputs = userGeneratedClassesProvider.Collect()
+			.Combine(mauiControlsAssemblySymbolProvider)
+			.SelectMany(static (x, _) => Deduplicate(x.Left!, x.Right).ToImmutableArray())
```
pictos, line 89: I'm using the `!` here, because for some reason the roslyn didn't see that I filtered for not null values, at line `62`

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -18,97 +26,80 @@ class TextColorToGenerator : IIncrementalGenerator
 
 	public void Initialize(IncrementalGeneratorInitializationContext context)
 	{
+		if (!Debugger.IsAttached)
```
VladislavAntonyuk, line 29: Please remove

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -18,97 +26,80 @@ class TextColorToGenerator : IIncrementalGenerator
...
-					var diag = Diagnostic.Create(TextColorToDiagnostic.InvalidModifierAccess, Location.None, declarationSymbol.Name);
-					context.ReportDiagnostic(diag);
-					continue;
-				}
 
-				textStyleClassList.Add((declarationSymbol.Name, accessModifier, nameSpace, declarationSymbol.TypeArguments.GetGenericTypeArgumentsString(), declarationSymbol.GetGenericTypeConstraintsAsString()));
-			}
-		}
+		// Here we Collect all the Classes candidates from the first pipeline
+		// Then we merge then with the Maui.Controls that implements the desired interfaces
```
VladislavAntonyuk, line 84:
```suggestion
		// Then we merge them with the Maui.Controls that implements the desired interfaces
```

src/CommunityToolkit.Maui.SourceGenerators/Helpers/IsExternalInit.cs
```diff
@@ -0,0 +1,66 @@
+﻿// <auto-generated>
+//   This code file has automatically been added by the "IsExternalInit" NuGet package (https://www.nuget.org/packages/IsExternalInit).
+//   Please see https://github.com/manuelroemer/IsExternalInit for more information.
+//
+//   IMPORTANT:
+//   DO NOT DELETE THIS FILE if you are using a "packages.config" file to manage your NuGet references.
```
VladislavAntonyuk, line 6: Can we remove this file and add it to gitignore?

pictos: No, this file is needed in order to use `records` in `netstandard`

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -18,97 +26,80 @@ class TextColorToGenerator : IIncrementalGenerator
...
-		foreach (var textStyleClass in textStyleClassList)
-		{
-			var textColorToBuilder = @"
+		context.RegisterSourceOutput(inputs, Execution);
+	}
+
+	static void Execution(SourceProductionContext context, ModelToGenerate textStyleClass)
+	{
+
+		var textColorToBuilder = $@"
```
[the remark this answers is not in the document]
pictos: We can change that in the future, this PR is just to fix the SG, so I didn't want to spend time converting it to string literals, if you see we concatenate the strings, this would add more work and more things to test. Feel free to open a discussion around that and if you want to submit a PR doing that (after we approve the discussion) will be awesome

src/CommunityToolkit.Maui.SourceGenerators/Helpers/IsExternalInit.cs
VladislavAntonyuk: But it is autogenerated by the package: IsExternalInit. However I do not see where it is used

pictos: But we don't use the package, we just have this file on our own. Without this file, we can't compile the SG when using `record`, so we don't use this directly. This explanation was added to the summary of the method.
You can find more info [here](https://stackoverflow.com/questions/67687078/can-we-use-records-in-c-sharp-8-0)

brminnick: Yup, this is a known-issue in .NET Standard 2.0. 

I had to add this class to all of my Xamarin.Forms apps where I used `record` until .NET Standard 2.1 came out.

pictos:
@brminnick did you test if the SG keeps its Incremental behavior?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
brminnick: I'll keep poking around to try and find another way to avoid `!`. But if I can't figure out something by the end of the day today, we can merge it and create a Bug Report noting it so that we don't forget.

Perhaps we could ping the Roslyn team on this to add some [Nullable static analyzers](https://learn.microsoft.com/dotnet/csharp/language-reference/attributes/nullable-analysis) to help other devs avoid this too.

brminnick:
> did you test if the SG keeps its Incremental behavior?

Good point! I did not. 

What are the steps to verify this? I haven't tested anything like this before 😅 

Do we need to use a reference type (`record` is a reference type whereas `ValueTuple` is a value-type) for the incremental behavior to work? If so, I'm happy to create a read-only class to replace the `record`.

pictos:
@brminnick to test it:
1. grab the comment at line `15` and paste it at line [`99`](https://github.com/CommunityToolkit/Maui/blob/4cb2bc88af40fa11848db2ad65ff55dbc9b2ce88/src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs#L99).
2. When you do that rebuild the SG project and reopen VS (for some reason if you don't reopen the changes will not apply)
3. Build the sample app and go into the generated files (inside the Analyzer view on VS)
4. Open any one of those files and look for the `DateTime` that will be printed there
5. keep that file open and change any file in the sample project (App.xaml.cs or any other) and see if the value changes
6. If the value changes it's broken if not it's fine

> Do we need to use a reference type (record is a reference type whereas ValueTuple is a value-type) for the incremental behavior to work? If so, I'm happy to create a read-only class to replace the record.

To keep the incremental behavior we should pass models that can be compared, so `records` are built in a way that we can compare them and check if the `oldValue` is the same as the `newValue`. Now the `ISymbol`, for example, can't be compared so the SG will regenerate the code everytime that you change a file in the project.

brminnick:
Ok, I'm convinced! Re-added the `record` 👍

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
brminnick: This is an old way to debug Source Generators that's no longer required.

Here's how to debug Source Generators using a Debug Launch UI Profile in VS 2022: https://github.com/JoanComasFdz/dotnet-how-to-debug-source-generator-vs2022

As a demonstration, here's the debugger hitting a break point in the Source Generator without the `Debugger.Launch()` logic:

image

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
brminnick: Fixed. 

I just passed through the nullable `INamedTypeSymbol?` to the `Deduplicate()` method, and do the null check there.

```cs
static IEnumerable<INamedTypeSymbol> Deduplicate(ImmutableArray<INamedTypeSymbol?> left, IEnumerable<INamedTypeSymbol> right)
{
  foreach (var leftItem in left)
  {
	  if (leftItem is null)
	  {
		  continue;
	  }
      // ...
}
```

brminnick (approved):
Thanks @pictos!!!

I'll defer to you to merge this. 

Could you also double-check that it is still working in an incremental fashion after I mucked things up (and fixed them, putting back the `record`) before merging it tonight?

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -21,91 +28,66 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
-
-		if (textStyleSymbol is null || iAnimatableSymbol is null)
-		{
-			var diag = Diagnostic.Create(TextColorToDiagnostic.MauiReferenceIsMissing, Location.None);
-			context.ReportDiagnostic(diag);
-			return;
-		}
+				if (iTextStyleInterfaceSymbol is null || iAnimatableInterfaceSymbol is null)
+				{
+					throw new Exception("There's no .NET MAUI referenced in the project.");
```
Youssef1313, line 40: I don't think throwing an exception is correct if it's not caught and transformed to a diagnostic.

src/CommunityToolkit.Maui.SourceGenerators/Generators/TextColorToGenerator.cs
```diff
@@ -21,91 +28,66 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+							&& classSymbol.AllInterfaces.Contains(iTextStyleInterfaceSymbol, SymbolEqualityComparer.Default)))
+				{
+					return null;
+				}
 
-		foreach (var namedTypeSymbol in mauiTextStyleImplementors)
-		{
-			textStyleClassList.Add((namedTypeSymbol.Name, "internal", namedTypeSymbol.ContainingNamespace.ToDisplayString(), namedTypeSymbol.TypeArguments.GetGenericTypeArgumentsString(), namedTypeSymbol.GetGenericTypeConstraintsAsString()));
-		}
+				return classSymbol;
```
Youssef1313, line 53: Using symbols as part of generator pipeline **must** be avoided. This can break incrementality.

PR #826, opened 2022-12-13
Close Needs Reproduction issues after 5 days

VladislavAntonyuk:
We want to encourage people to provide reproduction links.

It will help us avoid stale issues and keep our backlog cleaner.


brminnick:
Thanks Vlad! How do we test this before merging it?

VladislavAntonyuk:
I  copied it from .NET MAUI repo, I assume it should work.
At least it won't close issues in the next 5 days, furthermore, it will add a comment in 3 days.

PR #835, opened 2022-12-19
- Fixed 822

### LINKED ISSUE #822: [Bug] InvalidOperationException on iOS when opening Popup from Modal dialog with NavigationPage

### Current Behavior

`this.ShowPopup` throws `System.InvalidOperationException` (The Parent must be of type Microsoft.Maui.Handlers.PageHandler) when trying to open a popup from a Modal dialog with NavigationPage on iOS.

### Expected Behavior

Display the popup

### Steps To Reproduce

1. Open and run on iOS simulator from reproduction repository
2. Tap "Open Modal Dialog"
3. Tap "Open Popup" and it is crashing with the exception described

https://github.com/eichbaum/maui-popupdemo-bug

### Environment

```markdown
- .NET MAUI CommunityToolkit: 3.1.0
- OS: net7.0-ios Microsoft.iOS 16.0.1478
- .NET MAUI: 7.0.49
```

### Anything else?

Works fine on Android.

jaigupta91:
### Description of Change ###
Fixed InvalidOperationException on iOS when opening Popup from Modal dialog with NavigationPage

 ### Linked Issues ###
 

 - Fixes #822


src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
```diff
@@ -66,9 +66,9 @@ public override void ViewWillDisappear(bool animated)
 	[MemberNotNull(nameof(VirtualView), nameof(ViewController))]
 	public void SetElement(IPopup element)
 	{
-		if (element.Parent?.Handler is not PageHandler)
+		if (element.Parent?.Handler is null)
```
pictos, line 69: Our popup relies on `PageHandler` or `Page`, so we can't use the `null` check here, because in this case, we will allow devs to pass an `EntryHandler` for example.

jaigupta91:
```suggestion
		if (element.Parent?.Handler is not PageHandler && element.Parent?.Handler is not UIViewController)
		{
			throw new InvalidOperationException($"The {nameof(element.Parent)} must be of type {typeof(PageHandler)} or {typeof(UIViewController)}.");
		}
```
Please let me know if it is correct. After your opinion, will be pushing the code.

pictos: I'll try here and see how that goes. Thanks

bijington:
Can this be closed based on #934 merging?

jaigupta91:
I think, Yes

PR #862, opened 2023-01-03
Do not use shared default value for LinesProperty

Ghostbird:
### Description of Change ###

I've change the `LinesProperty` of the `DrawingView` to use a `defaultValueCreator` delegate instead of a `defaultValue`. The default value is a reference `new ObservableCollection<IDrawingLine>()` in a static create function. Hence it was a collection shared over all `DrawingView`s in a MAUI app.

These kind of situations are AFAIK _exactly_ why the `defaultValueCreator` delegate exists.

I haven't provided samples, but I did add a unittest to verify that `new DrawingView().Lines` no longer share a reference.

I haven't changed any documentation, as it doesn't conflict. The documented use coincidentally worked, because it uses either a `Lines = { Binding MyLines }` in XAML and initialises with a `Lines = new ObservableCollection<IDrawingLine>()` in C#. I think the approach where by default a new collection is initialised with every instance of the component, is the desirable outcome.

 ### Linked Issues ###
 

 - Fixes #794 , #871

 ### PR Checklist ###
 
 - ~Has samples (if omitted, state reason in description)~
 - ~Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/XYZ~ 

### DIFF
--- a/src/CommunityToolkit.Maui.UnitTests/Views/DrawingView/DrawingViewTests.cs
+++ b/src/CommunityToolkit.Maui.UnitTests/Views/DrawingView/DrawingViewTests.cs
@@ -38,6 +38,12 @@ public class DrawingViewTests : BaseHandlerTest
 	}
 
 	[Fact]
+	public void DefaultLinesShouldNotBeShared()
+	{
+		new Maui.Views.DrawingView().Lines.Should().NotBeSameAs(new Maui.Views.DrawingView().Lines);
+	}
+
+	[Fact]
 	public void OnLinesCollectionChangedHandlerIsCalled()
 	{
 		var drawingViewHandler = CreateViewHandler<MockDrawingViewHandler>(drawingView);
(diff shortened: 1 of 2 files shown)


Ghostbird:
Since this came in through _Discussion_ rather than _Issues_, should an issue be created so that it can be approved? Or can someone with the appropriate rights simply sign off on this?

VladislavAntonyuk (approved):

VladislavAntonyuk:
@Ghostbird Thank you for the fix. Could you please create an issue from the discussion, so we follow our flow? I will approve and merge this PR after that.

Ghostbird:
> @Ghostbird Thank you for the fix. Could you please create an issue from the discussion, so we follow our flow? I will approve and merge this PR after that.

I've written it. The issue description is a bit odd, since it's written "the wrong way around", from a technical perspective where everything about the underlying technical cause is already known. Normally an issue would be written from the user side and describe behaviour. Since it's largely administrative, I guess it doesn't matter that much.

Issue #909, opened 2023-01-19
[BUG] Wrong version for Nightly builds


twsl:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [X] I have read the "Reporting a bug" section on Contributing file.

### Current Behavior

As shown in https://github.com/CommunityToolkit/Maui/discussions/852, the regular release on nuget is tagged with version 3.1.0, the nightly builds are tagged as 1.0.0-preview. Which means it always shows the regular build as latest version and causes issues with other libraries that require CommunityToolkit >= 3.0.0

### Expected Behavior

Tag nightly library accordingly

### Steps To Reproduce

1. Add nuget feed
2. try to update

### Link to public reproduction project repository

https://github.com/twsl/Maui/tree/feature/version

### Environment

```markdown
- .NET MAUI CommunityToolkit:3.1.0
- OS: Windows 11
- .NET MAUI: 7
```

twsl:
A simple fix has been proposed because azure pipelines offer no way for me to test the code locally

vhugogarcia:
@brminnick I believe this can maybe be moved to a discussion about the versioning strategy first, and so everybody can share ideas for nightly builds, etc. I noticed the PR was closed and not anymore movement there. 

So, I think we can also close this issue.

Regards,
Victor

VladislavAntonyuk:
@vhugogarcia we had an open PR. For nightly builds for simplicity we decided to use version 99 to ensure it is higher than our regular version

bijington:
I did also have a branch working with GitVersion but we didn't work out the tagging approach

PR #910, opened 2023-01-19
Update Base Version in azure-pipelines.yml

twsl:
### Description of Change ###

 
This is a quick and dirty fix for the issue regarding the nightly versions by raising the semantic base version.

 ### Linked Issues ###
 

 - Fixes https://github.com/CommunityToolkit/Maui/issues/909

 ### PR Checklist ###
 

azure-pipelines.yml
```diff
@@ -1,5 +1,5 @@
 variables:
-  CurrentSemanticVersionBase: '1.0.0'
+  CurrentSemanticVersionBase: '3.1.0'
```
VladislavAntonyuk, line 2: should it be (current version + 1), so 3.2.0 at the moment of writing?

VladislavAntonyuk (dismissed):

brminnick:
Is there a way to automatically set its version number based on the version number of the latest release?

bijington:
`git describe` might get you some of the way there...

https://git-scm.com/docs/git-describe

Sadly I don't know how we would determine the type of change and ultimately the change in version component. 

The best long term solution is to use a tool like GitVersion that just does it for us

VladislavAntonyuk:
I think we can set the version number to 99.99.99.99, in that case it will always be a latest version, if user wants to use nightly builds

brminnick:
> I think we can set the version number to 99.99.99.99

Works for me!

twsl:
as proposed in the discussion, ideally something like [GitVersion](https://github.com/GitTools/GitVersion) would be used to automatically calculate the semver. Just using the same version number for everything might cause other problems in the future but could work for now

VladislavAntonyuk:
as our next release is 4.0.0, let's for now use 4.1.0 and see how it works.
Ideally, we can use GitVersion as suggested.
I do not recommend use 99.0.0.00 in case somebody limits the latest version (>3.1.0 && <5.0.0)

twsl:
Can someone else approve this so it gets merged?

VladislavAntonyuk:
@brminnick, @jfversluis are you ok with the changes?

brminnick:
If this is not urgent, I prefer that we implement a permanent solution that references the latest release number and works for both `CommunityToolkit.Maui` and `CommunityToolkit.Maui.MediaElement`. 

This is a pretty low-effort PR that just delays the problem and creates more work for us in the future; we'll have to open a PR to bump this version number after every release.

VladislavAntonyuk:
@twsl Could you please investigate how we can use GitVersion solution?

twsl:
I can try, the issue with that is that I cannot test the azure pipelines locally

VladislavAntonyuk:
@twsl don't worry about it. If the build passes, you can find the version number in the GitHub actions result:
image

twsl:
@bijington I'll give it a try now

bijington:
> @bijington Feel free to tackle this. I did a quick test and just adding the nuget to the project causes issues and fails to build. Not sure if this requires https://github.com/GitTools/GitVersion/issues/3334. Adding the azure pipelines task could also be a solution tho

Thanks for having a look. I'll look into that issue and hope that we can workaround it

twsl:
@dotnet-policy-service agree

bijington:
@jfversluis this reminds me we need to have a chat about GitVersion

PR #918, opened 2023-01-22
Add MediaElement.Aspect Functionality

jfversluis:
Adds the aspect functionality for MediaElement.

## Public API changes

Adds
```csharp
IMediaElement.Aspect { get; }
MediaElement.Aspect { get; set }
public static void MapAspect(object handler, MediaElement MediaElement);
protected virtual partial void PlatformUpdateAspect();
```

Additionally updates the MediaElement sample page a bit so that it's a bit cleaner.


VladislavAntonyuk (approved):

src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs
```diff
@@ -8,6 +8,12 @@ namespace CommunityToolkit.Maui.Core;
 /// </summary>
 public interface IMediaElement : IView
 {
+	/// <summary>
+	/// Gets or sets the media aspect.
+	/// </summary>
+	/// <remarks>Not functional for non-visual media.</remarks>
+	Aspect Aspect { get; set; }
```
pictos, line 15: If the Platform layer will not `set` the aspect, it should be readonly, from what I saw in the code... The platform layer just reads this value and applies it. If I missed something, please let me know

```suggestion
	Aspect Aspect { get; }
```

pictos (changes requested):
Just a small design detail

pictos (approved):
Looks great!

PR #930, opened 2023-01-26
Update Xcode version in build pipeline

jfversluis:


azure-pipelines.yml
```diff
@@ -20,8 +20,8 @@ variables:
   PathToCommunityToolkitMediaElementAnalyzersCodeFixCsproj: 'src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes.csproj'
   PathToCommunityToolkitAnalyzersUnitTestCsproj: 'src/CommunityToolkit.Maui.Analyzers.UnitTests/CommunityToolkit.Maui.Analyzers.UnitTests.csproj'
   DotNetMauiRollbackFile: 'https://maui.blob.core.windows.net/metadata/rollbacks/7.0.49.json'
-  Latest_Xcode_Version: '14.1.0'
-  CommunityToolkit_Xcode_Version: '14.0.1'
+  Latest_Xcode_Version: '14.2.0'
+  CommunityToolkit_Xcode_Version: '14.2.0'
```
brminnick, line 24: We need to keep the Community Toolkit 
Xcode version pinned to the same version of Xcode that was supported when .NET v7.0.100 shipped. 

I don't believe net-ios supported Xcode 14.2.0 when .NET v7.0.100 shipped

brminnick: We can bump this Xcode version when we bump our .NET version

jfversluis: We might need to add some kind of comment on why that is I guess then because I have no clue why there are two separate versions in here. It seems a bit weird 😅

brminnick:
Hey Gerald! Can you add some context to this PR? Do we need to bump the Xcode version for a certain reason?

jfversluis:
This [build](https://dev.azure.com/dotnet/CommunityToolkit/_build/results?buildId=86441&view=logs&jobId=7f63e264-4998-5012-ab3d-c5232e82c0a0) reports:

> ILLINK : error MT0180: This version of Microsoft.iOS requires the iOS 16.2 SDK (shipped with Xcode 14.2). Either upgrade Xcode to get the required header files or set the managed linker behaviour to Link Framework SDKs Only in your project's iOS Build Options > Linker Behavior (to try to avoid the new APIs). [/Users/runner/work/1/s/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj::TargetFramework=net7.0-ios]
ILLINK : error MT2301: The linker step 'Setup' failed during processing: This version of Microsoft.iOS requires the iOS 16.2 SDK (shipped with Xcode 14.2). Either upgrade Xcode to get the required header files or set the managed linker behaviour to Link Framework SDKs Only in your project's iOS Build Options > Linker Behavior (to try to avoid the new APIs). [/Users/runner/work/1/s/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj::TargetFramework=net7.0-ios]

So I was testing if using Xcode 14.2 will fix this or if we need to pin some other version, because something seems to have updated causing iOS builds to not work anymore.

azure-pipelines.yml
brminnick: One should be the latest (the most recent) supported release of Xcode. 

The other is the version of Xcode that we use to build the toolkit.

VladislavAntonyuk:
We can use any version, because XCode doesn't used in build machine ( we build on windows)

jfversluis:
Changed it back! I see the difference is the version we use for the library and the sample. Didn't see that at the first glance so we might make that a bit more clear maybe but fine for now.

brminnick:
Sounds good! Do you want to add a comment or rename a variable here in this PR before we merge it?

jfversluis:
Done!

brminnick (approved):

PR #953, opened 2023-02-01
Implement Mediaelement ShouldMute

jfversluis:
### Description of Change ###

Implements muting functionality for `MediaElement`

 ### Linked Issues ###
 

 - Fixes #950

 ### PR Checklist ###
 
 - [x] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal) (sort of)
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/206

### DIFF
--- a/src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs
@@ -61,7 +61,7 @@ public interface IMediaElement : IView
 	/// <summary>
 	/// Gets or sets if audio should be muted.
 	/// </summary>
-	public bool ShouldMute { get; set; }
+	bool ShouldMute { get; set; }
 
 	/// <summary>
 	/// Gets or sets whether the player should show the platform playback controls.
--- a/src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs
@@ -59,6 +59,11 @@ public interface IMediaElement : IView
 	bool ShouldKeepScreenOn { get; set; }
 
 	/// <summary>
+	/// Gets or sets if audio should be muted.
+	/// </summary>
+	public bool ShouldMute { get; set; }
+
+	/// <summary>
 	/// Gets or sets whether the player should show the platform playback controls.
 	/// </summary>
 	bool ShouldShowPlaybackControls { get; set; }
--- a/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
@@ -421,7 +421,7 @@ public partial class MediaManager : Java.Lang.Object, IPlayer.IListener
 
 		// We're going to muted state, capture current volume first
 		// so we can restore later
-		if (MediaElement.ShouldMute == true)
+		if (MediaElement.ShouldMute)
 		{
 			volumeBeforeMute = Player.Volume;
 		}
--- a/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
@@ -16,6 +16,7 @@ namespace CommunityToolkit.Maui.Core.Views;
 public partial class MediaManager : Java.Lang.Object, IPlayer.IListener
 {
 	double previousSpeed = -1;
+	float volumeBeforeMute = 1;
 
 	/// <summary>
 	/// The platform native counterpart of <see cref="MediaElement"/>.
@@ -411,7 +412,6 @@ public partial class MediaManager : Java.Lang.Object, IPlayer.IListener
 		PlayerView.KeepScreenOn = MediaElement.ShouldKeepScreenOn;
 	}
 
-	float volumeBeforeMute = 1;
 	protected virtual partial void PlatformUpdateShouldMute()
 	{
 		if (Player is null || MediaElement is null)
--- a/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
@@ -411,6 +411,24 @@ public partial class MediaManager : Java.Lang.Object, IPlayer.IListener
 		PlayerView.KeepScreenOn = MediaElement.ShouldKeepScreenOn;
 	}
 
+	float volumeBeforeMute = 1;
+	protected virtual partial void PlatformUpdateShouldMute()
+	{
+		if (Player is null || MediaElement is null)
+		{
+			return;
+		}
+
+		// We're going to muted state, capture current volume first
+		// so we can restore later
+		if (MediaElement.ShouldMute == true)
+		{
+			volumeBeforeMute = Player.Volume;
+		}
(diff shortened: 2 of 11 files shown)


src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
```diff
@@ -411,6 +412,23 @@ public void OnVolumeChanged(float volume)
...
+		}
+
+		// We're going to muted state, capture current volume first
+		// so we can restore later
+		if (MediaElement.ShouldMute == true)
+		{
+			volumeBeforeMute = Player.Volume;
+		}
+
+		Player.Volume = MediaElement.ShouldMute ? 0 : volumeBeforeMute;
```
VladislavAntonyuk, line 429: can the Volume be changed while on mute?
like this:
```
MediaElement.Volume = 50; // volume 50, volumeBeforeMove 1
MediaElement.ShouldMute = true; // volume 0, volumeBeforeMove 50
MediaElement.Volume = 100; // volume 100, volumeBeforeMove 50
MediaElement.ShouldMute = false; // volume 50, volumeBeforeMove 50

Expected volume 100

Comment (jfversluis) [reply src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs:429]: Hmm yeah I guess. I hate it when I am implementing features that should be available in these controls... :P I will think about this some more. Thanks!

Comment (jfversluis) [reply src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs:429]: Oh so... I think the solution is actually the other way around. If you start updating the volume it will probably unmute. So the solution here is when the volume is updated, the `ShouldMute` could become false again I guess.

Comment (VladislavAntonyuk) [reply src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs:429]: You can update volumeBeforeMute when Volume changes, except this line of code

File: src/CommunityToolkit.Maui.MediaElement/Interfaces/IMediaElement.cs [hunk 2]
```diff
@@ -58,6 +58,11 @@ public interface IMediaElement : IView
 	/// <remarks>If media is paused, stopped or has completed playing, the display will turn off.</remarks>
 	bool ShouldKeepScreenOn { get; set; }
 
+	/// <summary>
+	/// Gets or sets if audio should be muted.
+	/// </summary>
+	bool ShouldMute { get; set; }
```
brminnick, line 64: Let's add a `<remarks` here about how this impacts volume control.

Something like this (caveat: I'm unsure if this statement is accurate to how we handle `Volume` vs `ShouldMute`)

```suggestion
	/// <summary>
	/// Gets or sets if audio should be muted.
	/// </summary>
	/// <remarks>When <see cref="ShouldMute"/> is <see langword="true" />, the value of <see cref="Volume"/> is ignored.</remarks>
	bool ShouldMute { get; set; }
```

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.shared.cs
```diff
@@ -120,6 +120,14 @@ public void UpdateShouldKeepScreenOn()
 		PlatformUpdateShouldKeepScreenOn();
 	}
 
+	/// <summary>
+	/// Update whether of not the audio should be muted.
```
brminnick, line 124:
```suggestion
	/// Update whether or not the audio should be muted.
```

brminnick (changes requested):
Thanks Gerald! Just a small request to the XML Comments to help the developer understand the relationship between `ShouldMute` and `Volume`.

(I can't add a comment on the PR for `Volume` because it is unchanged in the PR, so I'll add it here)
Let's add a `///<remark>` to `Volume` that explains its relationship with `ShouldMute`.

Something like this (caveat: I'm unsure if this statement is accurate to how we handle `Volume` vs `ShouldMute`)

```xml
	/// <summary>
	/// Gets or sets the volume of the audio for the media.
	/// </summary>
	/// <remarks>
	/// A value of 1 means full volume, 0 is silence.
	/// <br/>
	/// When <see cref="ShouldMute"/> is <see langword="true" />, <see cref="Volume"/> is ignored.</remarks>
	/// </remarks>
	double Volume { get; set; }
```

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
brminnick: Whichever path we move forward with, let's ensure our implementation is consistent across iOS, Android, macOS, Windows + Tizen

Ie `MediaElement` shouldn't unmute when the volume changes on Android and not unmute when the volume changes on iOS

jfversluis:
@brminnick definitely something worth adding, but I don't think this comment is how it will/should work. On iOS/macOS and Windows when you mute and you change the volume it unmutes. I think we should mimic that behavior for Android as well.

brminnick:
> I think we should mimic that behavior for Android as well
Absolutely agree! Which ever path we move forward with should be consistent across each platform, and we should add that as a `<remark>` in our XML Comments for `Volume` and `ShouldMute` so that devs (hopefully) see it in their intellisense 🙌

jfversluis:
OK so here is where we landed, not sure if I'm entirely happy (yet).

For all platforms you can now mute and when muted whenever you change the volume through the `Volume` property. However, if you change the volume through the platform controls (where available) it will automatically unmute.

It kind of maybe feels natural because in the latter case that will always happen by a user and is the behavior on the platform. This means you can programatically set the `Volume` without unmuting too. So it might not be all bad. This also places the power of unmuting on a volume change by the developer.

Thoughts? Should we make it all aligned and also unmute when the value in `Volume` changes through code?

vhugogarcia:
> OK so here is where we landed, not sure if I'm entirely happy (yet).
> 
> For all platforms you can now mute and when muted whenever you change the volume through the `Volume` property. However, if you change the volume through the platform controls (where available) it will automatically unmute.
> 
> It kind of maybe feels natural because in the latter case that will always happen by a user and is the behavior on the platform. This means you can programatically set the `Volume` without unmuting too. So it might not be all bad. This also places the power of unmuting on a volume change by the developer.
> 
> Thoughts? Should we make it all aligned and also unmute when the value in `Volume` changes through code?

I think it is good and feels natural. When I play a video or audio if I mute it and then I volume up the mute should be discarded and volume up should have priority. 

Just one minor thing , if I have set volume 50% then I mute the video or audio and then I unmute it, the volume should stay in the same % of volume it was before.

Other than that it looks great! btw, I will take a look at this media element soon in a real app! thanks @jfversluis and team for bringing this real!

jfversluis:
> Just one minor thing , if I have set volume 50% then I mute the video or audio and then I unmute it, the volume should stay in the same % of volume it was before.

Do you have any reason to believe this is not happening right now? Because that is how it should work

jfversluis:
I think we're all on the same page with this one! I'm going to go ahead and merge this.

borrmann:
<toolkit:MediaElement IsVisible="false"
                                  x:Name="mediaElement"
                                  Volume="1"
                                  ShouldMute="False"                                  
                                  ShouldLoopPlayback="False"
                                  Source="{Binding MediaSource}"
                                  ShouldAutoPlay="True" />

On iOS, using these properties the sound currently stays muted when the user changes the sound using physical device buttons. When using headphones it works, but I assume this is because the phone isn't (fully ?) muted in this case. With headphones, keyboards sounds remain silent but videos on other apps play out loud automatically as well.

PR #1015, opened 2023-02-20
Perform MediaElement Seek on main thread for WinUI

### LINKED ISSUE #1012: [BUG] Media Element Seek not firing

### Current Behavior

Media Element fails to seek playback positions under conditions where windows or android works fine.
Examples include:
-if using mediaElement.MediaOpened it will not fire under any situation.
-if you are using StateChanged it will fire but only if you check its value before trying to seek.

I have no idea why the behavior in iOS is so different from Windows and Android? Is this intended to work this way?
I have include3d sample repo showing what works does not work. Notes in app with what you have to do to make iOS seek.

### Expected Behavior

Media element behavior should be the same across devices. Or please document differences for iOS.

### Steps To Reproduce

1. Open included solution and run it.
2. Or create new maui solution and add media element package.
3. Add an event to seek an arbitrary position at start of video
4. Set event to fire when opening video.

https://github.com/ne0rrmatrix/MediaElementIOSBug

### Environment

```markdown
- .NET MAUI CommunityToolkit: 3.1.0
- OS: iOS
- .NET MAUI: 7.0
Packages used:

 <ItemGroup>
	  <PackageReference Include="CommunityToolkit.Maui" Version="1.0.0-preview3840" />
	  <PackageReference Include="CommunityToolkit.Maui.MediaElement" Version="1.0.0-preview3840" />
	  <PackageReference Include="CommunityToolkit.MVVM" Version="8.1.0" />
</ItemGroup>
```
```


### Anything else?

In code there is commented out code that demonstrates a work around. Is media element behavior intended to be different across platforms? And if so can it be documented?

jfversluis:
### Description of Change ###

This wraps the check if the media can seek and do the actual seek in a `MainThread.BeginInvokeOnMainThread` call. Not doing so could cause an NRE, see linked issue.

 ### Linked Issues ###

 - Fixes #1012

 ### PR Checklist ###

 - [ ] Documentation created or updated: NA


brminnick:
Do the other platforms automatically marshall to the UI Thread? What happens when the user tries to seek on iOS, Android + Tizen?

ne0rrmatrix:
> Do the other platforms automatically marshall to the UI Thread? What happens when the user tries to seek on iOS, Android + Tizen?

On IOS It behaves differently and as expected. On Android I'm not sure. Windows CTD on certain requests because Main thread is not specified.

brminnick:
Thanks for following up, Gerald! 

I have a few concerns about using `MainThread.BeginInvokeOnMainThread`
- We may marshall to the MainThread when not needed 
  - This would cause a small performance impact when the current thread is already the Main Thread
- We are not `await`'ing `BeginInvokeOnMainThread`
  -  The `SeekTo()` method will always return before `Player.MediaPlayer.Position` is changed
  - This may cause a race condition for methods consuming `SeekTo()` which would incorrectly assume that the seek has completed when `SeekTo()` returns
- We are using `MainThread` instead of `IDispatcher`
  - `IDispatcher` basically does the same thing
  - `IDispatcher` allows more flexibility since it is an interface
  - `IDispatcher` can be used with Dependency Injection, allowing users to inject their own implementation if required (they wouldn't be able to do this with `MainThread`

I  took a quick stab at addressing the above concerns and opened up PR #1121. Check it out and let me know what you think!

pictos:
I would say the change is fine, just pointing out one item on your list

> We may marshall to the MainThread when not needed
> This would cause a small performance impact when the current thread is already the Main Thread

The MainThread checks if the code needs to change the context, if don't it just executes the method, as you can see [here](https://github.com/dotnet/maui/blob/main/src/Essentials/src/MainThread/MainThread.shared.cs#L22-L32)

brminnick:
> The MainThread checks if the code needs to change the context, if don't it just executes the method, as you can see [here](https://github.com/dotnet/maui/blob/main/src/Essentials/src/MainThread/MainThread.shared.cs#L22-L32)

Oh cool! Good to know!!

ne0rrmatrix (dismissed):
This fixes the issue I was having with threading crashes on Windows.

brminnick (approved):

PR #1050, opened 2023-03-01
FileSaver does not truncate existing files

### LINKED ISSUE #1049: [BUG] FileSaver Leaves junk data at the end of existing files

### Current Behavior

- Ask the file saver to save over a file that already exists and is longer than the data to be saved
- Use FileSaver to save the new data
- Data from the original file will be present at the end of the newly written file.

This is caused by the use of FileMode.OpenOrCreate which does not truncate existing files in CommunityToolkit.Maui.Core\Essentials\FileSaver\FileSaverImplementation.shared.cs:

await using var fileStream = new FileStream(filePath, FileMode.OpenOrCreate);

### Expected Behavior

Only the new data being written to the file should be in the file. Data from a pre-existing file that was longer than the new file should not be present.

### Steps To Reproduce

- Ask the file saver to save over a file that already exists and is longer than the data to be saved
- Use FileSaver to save the new data
- Data from the original file will be present at the end of the newly written file.

https://github.com/farmergreg/maui-filesaver-bug-demo

### Environment

```markdown
- OS: Windows 10 and 11
- .NET MAUI: github main branch pulled on 2023-03-01
```

### Anything else?

I will submit a PR with a proposed fix.

farmergreg:
### Description of Change ###

FileMode.OpenOrCreate does not automatically truncate files.

This fix truncates files to ensure that junk data does not appear at the end of a newly saved file if the data being written is shorter than data present in a previously existing file.

 ### Linked Issues ###
 - Fixes #1049

### DIFF
--- a/src/CommunityToolkit.Maui.Core/Essentials/FileSaver/FileSaverImplementation.shared.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/FileSaver/FileSaverImplementation.shared.cs
@@ -35,6 +35,7 @@ public sealed partial class FileSaverImplementation
 	static async Task WriteStream(Stream stream, string filePath, CancellationToken cancellationToken)
 	{
 		await using var fileStream = new FileStream(filePath, FileMode.OpenOrCreate);
+		fileStream.SetLength(0);
 		stream.Seek(0, SeekOrigin.Begin);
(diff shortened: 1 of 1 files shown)


VladislavAntonyuk:
@ccmcdaniel it is not released yet

Issue #86, opened 2021-09-26
[Proposal] TouchEffect


brminnick:
# TouchEffect

* [x] Proposed
* [x] Prototype
* [x] Implementation
  * [ ] iOS Support
  * [x] Android Support
  * [ ] macOS Support
  * [ ] Windows Support
* [ ] Unit Tests: Not Started
* [ ] Sample: Not Started
* [ ] Documentation: Not Started

## Summary
[summary]: #summary

Adds additional touch events to any VisualElement

## Detailed Design
[design]: #detailed-design

### TouchEffect.shared.cs

```cs
public class TouchEffect : RoutingEffect
{
  public event EventHandler<TouchStatusChangedEventArgs> StatusChanged;
  public event EventHandler<TouchStateChangedEventArgs> StateChanged;
  public event EventHandler<TouchInteractionStatusChangedEventArgs> InteractionStatusChanged;
  public event EventHandler<HoverStatusChangedEventArgs> HoverStatusChanged;
  public event EventHandler<HoverStateChangedEventArgs> HoverStateChanged;
  public event EventHandler<TouchCompletedEventArgs> Completed;
  public event EventHandler<LongPressCompletedEventArgs> LongPressCompleted;
  
  public static readonly BindableProperty IsAvailableProperty;
  public static readonly BindableProperty ShouldMakeChildrenInputTransparentProperty;
  public static readonly BindableProperty CommandProperty;
  public static readonly BindableProperty LongPressCommandProperty;
  public static readonly BindableProperty CommandParameterProperty;
  public static readonly BindableProperty LongPressCommandParameterProperty;
  public static readonly BindableProperty LongPressDurationProperty;
  public static readonly BindableProperty StatusProperty;
  public static readonly BindableProperty StateProperty;
  public static readonly BindableProperty InteractionStatusProperty;
  public static readonly BindableProperty HoverStatusProperty;
  public static readonly BindableProperty HoverStateProperty;
  public static readonly BindableProperty NormalBackgroundColorProperty;
  public static readonly BindableProperty HoveredBackgroundColorProperty;
  public static readonly BindableProperty PressedBackgroundColorProperty;
  public static readonly BindableProperty NormalOpacityProperty;
  public static readonly BindableProperty HoveredOpacityProperty;
  public static readonly BindableProperty PressedOpacityProperty;
  public static readonly BindableProperty NormalScaleProperty;
  public static readonly BindableProperty HoveredScaleProperty;
[...]

PR #1063, opened 2023-03-04
[WiP] Touch effect

pictos:
This is a draft PR for TouchEffect, it's a port of the XCT version. This isn't in a good shape yet, just made the code transition between forms and Maui, but it's missing the Apple implementation.

A side note, the Windows version doesn't work in the normal way, need to set the `NativeAnimation` to `true`, I believe that's because Windows doesn't handle _touch_ and I don't have a touch monitor to test.

I skipped the summary docs, if you try to run it locally the build will fail, but you can change the severity in order to run it locally


brminnick:
Thanks @pictos!

Assigning to @kphillpotts who is taking the lead on #86

AdamEssenmacher:
The XCT version has received a critical bugfix on iOS. A MAUI port should be sure to pick up the fix.

https://github.com/xamarin/XamarinCommunityToolkit/pull/1954

brminnick:
Closing: https://github.com/CommunityToolkit/Maui/pull/1673

Issue #978, opened 2023-02-09
[Proposal] Add a behavior (or attached property) to allow users the ability to close the soft keyboard when clicking outside of a view


PureWeen:
### Feature name

SoftKeyboardBehavior

### Link to discussion

https://github.com/CommunityToolkit/Maui/discussions/967

### Progress tracker

- [ ] Android Implementation
- [ ] iOS Implementation
- [ ] MacCatalyst Implementation
- [ ] Windows Implementation
- [ ] Tizen Implementation
- [ ] Unit Tests
- [ ] Samples
- [ ] Documentation

### Summary

There are a number of users that want to be able to tap into the nothingness and close the soft keyboard. 

https://github.com/dotnet/maui/issues/12002#issuecomment-1385826740

We didn't copy this behavior over from XF because the mechanism used to implement this behavior was to make the `PageRenderer` focusable/clickable which wreaked havoc on accessibility. 

I have a proposal here for some APIs we could add to MAUI
https://github.com/dotnet/maui/issues/12003
https://github.com/dotnet/maui/issues/11983

I feel like this type of behavior should live inside the community toolkit, so it can be refined and added to out of band from the MAUI. We can release these initial features inside a NET 7 version of the community toolkit. Since this is added on behavior it seems like it would survive and iterate better inside the toolkit vs inside the core code. 

The core MAUI code just needs to provide the APIs necessary to enable this type of feature. 

I already have a behavior built here
https://github.com/PureWeen/ShanedlerSamples/tree/main/ShanedlerSamples/Library/TapToCloseBehavior
and the iOS code can be easily removed from MAUI core and put into a behavior as well. 

```XAML
            <Entry Placeholder="Text Field" Text="Starting text" WidthRequest="300">
                <Entry.Behaviors>
                    <local:SoftKeyboardBehavior CloseOnTapOutsideBounds="true"></local:SoftKeyboardBehavior>
                </Entry.Behaviors>
            </Entry>
```

### Motivation

Being able to close the soft keyboard when you tap off of an entry is a use case that a number of users really want. 
https://github.com/dotnet/maui/issues/12002

The implementation of this feature in Android was done in such a way that broke accessibility. 

This type of behavior is an add-on to the platform not something that we should adopt by default in core. 

Even on the iOS side this behavior can have negative impacts on accessibility. 
https://github.com/dotnet/maui/pull/10948

[...]

PR #1064, opened 2023-03-05
KeyboardExtensions

### LINKED ISSUE #1055: [Proposal] Add a KeyboardManager for interacting with SoftInputKeyboard device
### Feature name

KeyboardManager

### Link to discussion

https://github.com/CommunityToolkit/Maui/discussions/969

### Progress tracker

- [x] Android Implementation
- [x] iOS Implementation
- [ ] Windows Implementation
- [ ] Tizen Implementation
- [ ] Unit Tests
- [x] Samples
- [x] Documentation

### Summary

Currently MAUI lacks an API for interacting with the soft keyboard. This lack of API caused a lot of implicit behavior to get added to Xamarin.Forms. Because it's impossible to correctly cover how every user might want the keyboard to react we should add APIs to empower the developer to make these choices themselves

### Motivation

This will allow developers to hide/show and be alerted when the softkeyboard is opened. This will give developers the power to define how they want the keyboard to react vs us just making guesses and only being right for a subset of the users.

### Detailed Design

The changes will most likely match up with the code [here](https://github.com/PureWeen/ShanedlerSamples/tree/main/ShanedlerSamples/Library/Keyboard). 

### Usage Syntax

I realize it might seem odd that you need to pass a target view in but you are required to indicate the target of the keyboard when you open the soft input. This way the platform can focus that control and notify the user. 

```C#
KeyboardManager.Show(targetView)
KeyboardManager.Hide(targetView)
```

### Drawbacks

_No response_

### Alternatives

_No response_

### Unresolved Questions

_No response_

PureWeen:
### Description of Change ###

 

 ### Linked Issues ###

 - Fixes #1055

 ### PR Checklist ###
 
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/248


samples/CommunityToolkit.Maui.Sample/Pages/Extensions/KeyboardManagerExtensionsPage.xaml
```diff
@@ -0,0 +1,18 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    xmlns:pages="clr-namespace:CommunityToolkit.Maui.Sample.Pages"
+    x:Class="CommunityToolkit.Maui.Sample.Pages.Extensions.KeyboardManagerExtensionsPage"
+    Title="KeyboardManagerExtensionsPage"
+    xmlns:vm="clr-namespace:CommunityToolkit.Maui.Sample.ViewModels.Extensions"
+    x:TypeArguments="vm:KeyboardManagerExtensionsViewModel"
+    x:DataType="vm:KeyboardManagerExtensionsViewModel">
+    <pages:BasePage.Content>
+        <VerticalStackLayout Spacing="12">
+            <Entry Text="Hello" x:Name="entry"></Entry>
```
VladislavAntonyuk, line 13: Please add label describing the feature.

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardManager.android.cs
```diff
@@ -0,0 +1,73 @@
...
+	{
+		if (inputView?.Context == null)
+		{
+			throw new ArgumentNullException(nameof(inputView), "Must be set before the keyboard can be hidden.");
+		}
+
+		var focusedView = inputView.Context?.GetActivity()?.Window?.CurrentFocus;
+		AView tokenView = focusedView ?? inputView;
+
+		using (var inputMethodManager = (InputMethodManager)tokenView.Context?.GetSystemService(Context.InputMethodService)!)
```
VladislavAntonyuk, line 23: Please don't use null forgiving operator.

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardManager.android.cs
```diff
@@ -0,0 +1,73 @@
...
+		{
+			var windowToken = tokenView.WindowToken;
+			if (windowToken != null && inputMethodManager != null)
```
VladislavAntonyuk, line 26:
```suggestion
			if (windowToken is not null && inputMethodManager is not null)
```

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardManager.android.cs
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	static void ShowKeyboard(this TextView inputView)
+	{
+		if (inputView?.Context == null)
+		{
+			throw new ArgumentNullException(nameof(inputView), "Must be set before the keyboard can be shown.");
+		}
+
+		using (var inputMethodManager = (InputMethodManager)inputView.Context.GetSystemService(Context.InputMethodService)!)
```
VladislavAntonyuk, line 40: Please don't use null forgiving operator

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardManager.android.cs
```diff
@@ -0,0 +1,73 @@
...
+			if (windowToken != null && inputMethodManager != null)
+			{
+				inputMethodManager.HideSoftInputFromWindow(windowToken, HideSoftInputFlags.None);
+			}
+		}
+	}
+
+	static void ShowKeyboard(this TextView inputView)
+	{
+		if (inputView?.Context == null)
```
VladislavAntonyuk, line 35:
```suggestion
		if (inputView?.Context is null)
```

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardManager.android.cs
```diff
@@ -0,0 +1,73 @@
...
+			case ViewGroup viewGroup:
+				viewGroup.GetFirstChildOfType<TextView>()?.ShowKeyboard();
+				break;
+		}
+	}
+
+	static bool IsSoftKeyboardVisible(this AView view)
+	{
+		var insets = ViewCompat.GetRootWindowInsets(view);
+		if (insets == null)
```
VladislavAntonyuk, line 65:
```suggestion
		if (insets is null)
```

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardManager.ios.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using UIKit;
+
+namespace CommunityToolkit.Maui.Core.Platform;
+
+public static partial class KeyboardManager
```
VladislavAntonyuk, line 6: Please check this file is included. We combine iOS and macOS with *.macios.cs files

samples/CommunityToolkit.Maui.Sample/Pages/Extensions/KeyboardExtensionsPage.xaml
```diff
@@ -0,0 +1,29 @@
...
+    xmlns:pages="clr-namespace:CommunityToolkit.Maui.Sample.Pages"
+    x:Class="CommunityToolkit.Maui.Sample.Pages.Extensions.KeyboardExtensionsPage"
+    Title="KeyboardExtensionsPage"
+    xmlns:vm="clr-namespace:CommunityToolkit.Maui.Sample.ViewModels.Extensions"
+    x:TypeArguments="vm:KeyboardExtensionsViewModel"
+    x:DataType="vm:KeyboardExtensionsViewModel">
+    <pages:BasePage.Content>
+        <ScrollView>
+            <VerticalStackLayout Spacing="12">
+                <Label Text="The Keyboard extension method provides an easy way to open and close the SoftInput Keyboard device on platfomrs that support it." />
```
rachelkang, line 14:
```suggestion
                <Label Text="The Keyboard extension method provides an easy way to open and close the SoftInput Keyboard device on platforms that support it." />
```

samples/CommunityToolkit.Maui.Sample/ViewModels/Extensions/KeyboardExtensionsViewModel.cs
```diff
@@ -0,0 +1,70 @@
...
+	public KeyboardExtensionsViewModel()
+	{
+		ShowKeyboard = new Command<ITextInput>(OnShowKeyboard);
+		HideKeyboard = new Command<ITextInput>(OnHideKeyboard);
+		IsKeyboardShowing = new Command<ITextInput>(OnIsKeyboardShowing);
+	}
+
+	void OnIsKeyboardShowing(ITextInput view)
+	{
+		if (!view.IsSoftKeyboardShowing())
```
VladislavAntonyuk, line 22: Inverse if

samples/CommunityToolkit.Maui.Sample/ViewModels/Extensions/KeyboardExtensionsViewModel.cs
```diff
@@ -0,0 +1,70 @@
...
+		{
+			KeyboardShowingText = $"Soft Input Is Currently Hidden";
+		}
+		else
+		{
+			KeyboardShowingText = $"Soft Input Is Currently Showing";
+		}
+	}
+
+	async void OnHideKeyboard(ITextInput view)
```
VladislavAntonyuk, line 32: Use RelayCommand attribute and convert to async Task

samples/CommunityToolkit.Maui.Sample/ViewModels/Extensions/KeyboardExtensionsViewModel.cs
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	[ObservableProperty]
+	string? keyboardShowingText;
+
+	[ObservableProperty]
+	string? operationResult;
+
+	[ObservableProperty]
+	Command<ITextInput>? showKeyboard;
```
VladislavAntonyuk, line 63: Use RelatCommand attribute on method. The field can be removed

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardExtensions.net.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Diagnostics;
+using System.Reflection.Metadata;
+
+namespace CommunityToolkit.Maui.Core.Platform;
+
+public static partial class KeyboardExtensions
+{
+	static bool HideKeyboard(this object _)
+	{
+		return false;
```
VladislavAntonyuk, line 11: Should we return false or throw exception similar to macOS?

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardExtensions.shared.cs
```diff
@@ -0,0 +1,126 @@
...
+		return Task.FromResult(HideKeyboard(platformView));
+	}
+
+	/// <summary>
+	/// If a soft input device is currently hiding, this will attempt to show it.
+	/// </summary>
+	/// <param name="targetView"></param>
+	/// <returns>
+	/// Returns <c>true</c> if the platform was able to show the soft input device.</returns>
+	public static Task<bool> ShowKeyboardAsync(this ITextInput targetView)
```
VladislavAntonyuk, line 80: We should pass cancellation tokens to all Async methods

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardExtensions.net.cs
brminnick: Good catch. Let's throw a `NotSupportedException`

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardExtensions.shared.cs
```diff
@@ -0,0 +1,126 @@
...
+	/// <returns>
+	/// Returns <c>true</c> if the platform was able to hide the soft input device.</returns>
+	public static Task<bool> HideKeyboardAsync(this ITextInput targetView)
+	{
+		if (!targetView.TryGetPlatformView(
+			out var platformView,
+			out _,
+			out _))
+		{
+			return Task.FromResult(false);
```
pictos, line 68: Here and in other places I can see a lot of hot paths, so what do you think to change the return type to `ValueTask`?

pictos (commented):
I've one suggestion

samples/CommunityToolkit.Maui.Sample/ViewModels/Extensions/KeyboardExtensionsViewModel.cs
PureWeen: ohhh fancy!!

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardExtensions.shared.cs
PureWeen: I haven't used `ValueTask` much.

Thoughts from this post?
https://stackoverflow.com/a/43003779

> As such, the default choice for any asynchronous method should be to return a Task or Task<TResult>. Only if performance analysis proves it worthwhile should a ValueTask<TResult> be used instead of Task<TResult>.

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardExtensions.shared.cs
```diff
@@ -0,0 +1,128 @@
...
+
+	/// <summary>
+	/// If a soft input device is currently hiding, this will attempt to show it.
+	/// </summary>
+	/// <param name="targetView"></param>
+	/// <param name="token">Cancellation token</param>
+	/// <returns>
+	/// Returns <c>true</c> if the platform was able to show the soft input device.</returns>
+	public static Task<bool> ShowKeyboardAsync(this ITextInput targetView, CancellationToken token)
+	{
```
VladislavAntonyuk, line 83:
```suggestion
	{
	    token.ThrowIfCancellationRequested();
```

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardExtensions.shared.cs
```diff
@@ -0,0 +1,128 @@
...
+
+	/// <summary>
+	/// If a soft input device is currently showing, this will attempt to hide it.
+	/// </summary>
+	/// <param name="targetView"></param>
+	/// <param name="token">Cancellation token</param>
+	/// <returns>
+	/// Returns <c>true</c> if the platform was able to hide the soft input device.</returns>
+	public static Task<bool> HideKeyboardAsync(this ITextInput targetView, CancellationToken token)
+	{
```
VladislavAntonyuk, line 63:
```suggestion
	{
	    token.ThrowIfCancellationRequested();
```

VladislavAntonyuk (dismissed):

src/CommunityToolkit.Maui.Core/Platform/KeyboardManager/KeyboardExtensions.shared.cs
brminnick: Here's an easy rule of thumb for `ValueTask`:
If the hot-path of your method not require/use the `await` keyword, you should return `ValueTask` / `ValueTask<T>` instead of `Task` / `Task<T>`

PureWeen: What if a few months from now we do require/use the `await` keyboard? do you just break the API? or do you still just return it via `ValueTask` with the result?

My rule of thumb is if someone with the first name "Stephen" says to do something I just do that thing. 

That being said, I'm fine here with ValueTask. I don't have a strong argument for one vs the other.

brminnick: It is not a breaking change to move from `ValueTask` to `Task`.

It is a breaking change to move from `Task` to `ValueTask`, albeit it would break very few (if any) users. The biggest difference in using `ValueTask` is that you cannot re-use it and pass it around like you can with a `Task`; most users don't do advanced things like that with `Task` though.

pictos: Well, we don't need to break it. So, the `ValueTask` can be used in async scenarios. The thing is, if the code resolves synchronously, we don't allocate anything to the work. Now if we need to complete it asynchronously then we will allocate a `Task` and that will work as normal `Task`. Also, if the user needs/want a `Task` behavior here, the dev can use `.AsTask()` extension method.

rachelkang:
Just did some initial testing - ran the sample on Windows and Android, and I'm not sure if the behavior I'm seeing is what we're expecting. (From what I've looked of the code so far (haven't looked at it all yet), this doesn't seem like the intended behavior)

1. On Windows only, clicking on "Check Keyboard State" would hide the keyboard. I'm not sure why this is happening.
2. On both Windows (touch mode) and Android, clicking on "Show Keyboard" does successfully show the keyboard. However, it will ALWAYS focus on the first entry, even if the keyboard is already focused and open on another entry. If a control already has focus, I think we should ensure the keyboard will open on that currently focused control. (perhaps even consider making it open on the last focused control if nothing has focus, but I wouldn't be surprised if that actually unleashed a can of worms so maybe not)

PureWeen:
> Just did some initial testing - ran the sample on Windows and Android, and I'm not sure if the behavior I'm seeing is what we're expecting. (From what I've looked of the code so far (haven't looked at it all yet), this doesn't seem like the intended behavior)
> 
> 1. On Windows only, clicking on "Check Keyboard State" would hide the keyboard. I'm not sure why this is happening.



> 2. On both Windows (touch mode) and Android, clicking on "Show Keyboard" does successfully show the keyboard. However, it will ALWAYS focus on the first entry, even if the keyboard is already focused and open on another entry. If a control already has focus, I think we should ensure the keyboard will open on that currently focused control. (perhaps even consider making it open on the last focused control if nothing has focus, but I wouldn't be surprised if that actually unleashed a can of worms so maybe not)

The sample is specifically calling ShowKeyboard on that first entry. Because the user is required to indicate the target for the keyboard. In theory we could add a HideKeyboard that will generically hide but I don't think we could add a ShowKeyboard that generically opens. On Android 31+ and iOS when they keyboard is closed nothing has focus so we can't just guess what to give focus to. 

[KeyboardExtensions by PureWeen · Pull Request #1064 · CommunityToolkit/Maui (github.com)](https://github.com/CommunityToolkit/Maui/pull/1064/files#diff-68cc9e07481c1c5b27dd3c05e2b7e6d14b672f745334713fa9f5ddf76c92095aR47)

rachelkang:
1. Yeah, we'll need to look into it more!
2. Oops, totally missed that part. That makes perfect sense / sgtm!

brminnick:
Just FYI - when reviewing this PR today I found a bug in our `Directory.Build.targets` file and have opened a fix, https://github.com/CommunityToolkit/Maui/pull/1081.

Long story short, `KeyboardExtensions.macos.cs` is currently being ignored by the compiler (it will be fixed once #1081 is merged and this PR is updated from `main`).

rachelkang:
1. This is also happening on iOS, as tested on iPad. So on both Windows and iOS, checking the keyboard state hides the keyboard.

image

2. On iOS, I'm seeing that while "Hide Keyboard" succeeds to hide the keyboard, the label says "Hide Failed"

image

JoonghyunCho:
Added Tizen implementation :)
<image src="https://user-images.githubusercontent.com/14328614/224652491-882f4293-8516-4367-b68b-6e7769722bf9.gif"  width="500" />

PureWeen:
> 1. This is also happening on iOS, as tested on iPad. So on both Windows and iOS, checking the keyboard state hides the keyboard.
> 
> image
> 
> 2. On iOS, I'm seeing that while "Hide Keyboard" succeeds to hide the keyboard, the label says "Hide Failed"
> 
> image

Yea :-/ this is because of the automatically closing `TapGesture` that gets added to the underlying page. I added some code to remove that tap gesture with a comment here https://github.com/CommunityToolkit/Maui/pull/1064/files#diff-30341c4d54048fcf5f19b98431e6a7fbdf7b21ab928538aa17ff335961a6f667R16-R34.

samples/CommunityToolkit.Maui.Sample/Pages/Extensions/KeyboardExtensionsPage.xaml.cs
```diff
@@ -0,0 +1,37 @@
...
+	{
+#if IOS || MACCATALYST
+		// Currently .NET MAUI will auto close the keyboard on iOS if you click outside of the entry
+		// This causes the examples on this page to behave oddly, because the keybord auto closes if the user
+		// taps the "hide keyboard" or "check keyboard status" buttons.
+		// We're going to remove this as the default behavior in .NET 8 and then users can re-enable it via
+		// https://github.com/CommunityToolkit/Maui/issues/978
+		// This code removes the TapGestureRecognizer used to detect and close the keyboard so that we can accurately
+		// validate that these APIs work on iOS
+		if (this?.Handler?.PlatformView is UIKit.UIView uiView &&
```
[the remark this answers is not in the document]
PureWeen: That's my plan with this Issue https://github.com/CommunityToolkit/Maui/issues/978

Once we merge this issue, then I'll look at creating the behavior/effect/other that lets users disable this.

pictos:
@PureWeen is this PR ready for review?

jfversluis:
@pictos marked as ready for review and talked to Shane for the docs, so ready to go!

src/CommunityToolkit.Maui.Core/Platform/KeyboardExtensions/KeyboardExtensions.shared.cs
```diff
@@ -0,0 +1,132 @@
...
+
+#if IOS || MACCATALYST
+using PlatformView = UIKit.UIView;
+#elif ANDROID
+using PlatformView = Android.Views.View;
+#elif WINDOWS
+using PlatformView = Microsoft.UI.Xaml.FrameworkElement;
+#elif TIZEN
+using PlatformView = Tizen.NUI.BaseComponents.View;
+#elif (NETSTANDARD || !PLATFORM) || (NET6_0_OR_GREATER && !IOS && !ANDROID && !TIZEN)
```
VladislavAntonyuk, line 12: Can we use #else ?

VladislavAntonyuk (approved):

pictos (approved):

al3xmeister:
How is this meant to be working? The documentation here makes no comment on the token functionality https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.itextinput.keyboard?view=net-maui-7.0
It would appear replicating the sample code with a command executed from the same control, i.e. the return command of an entry throws a : 'Could not resolve type with token in assembly 'CommunityToolkit.Maui.Core, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null')'.
Could someone please update?

brminnick:
@al3xmeister `KeyboardExtensions` provides extension methods for `ITextInput`, for example;

‘’’cs
entry.HideKeyboardAsync();
‘’’

You can find out documentation here: https://learn.microsoft.com/dotnet/communitytoolkit/maui/extensions/keyboard-extensions

al3xmeister:
Thanks, but doesnt answer my question. The microsoft documentation makes no reference to a cancellation token. It is a required parameter to HideKeyboardAsync();. I presume it's likely after activity.CurrentFocus?.WindowToken, however you can't access it from the return command of an entry. The sample does it from a button. Could this be updated or revisited?

jfversluis:
This video will be made public later today that goes over it: https://youtu.be/bIsGFx7jGcM

If you don't plan on using the cancellation token you can pass in `default`. Nonetheless, we should fix this up in the docs.

PR #1121, opened 2023-03-30
Promote `MediaManager.PlatformSeek()` from `void` to `ValueTask`

brminnick:
### Description of Change ###

This PR is a recommendation to update the existing PR #1015.

### Additional Information

This PR also includes formatting fixes to `MediaManager.Tizen.cs`:
- Use `or` to evaluate multiple `enum` states
- Consolidate methods, keeping `public` methods grouped together, `protected` methods grouped together, etc


ne0rrmatrix:
I am currently using device specific code to seek when loading media element. On IOS it fails to seek if I do not use mediaelement.StateChanged with null check. On Android/WinUi I am using mediaelement.MediaOpened to start seeking. If I try to do anything else it fails. I could of course just use on Loaded but I thought I would mention this here in case if affects how you move forward. As to seeking on MainThread that is an ongoing issue I have device specific code all over the place.

PR #1126, opened 2023-04-03
BadgeCounter #807

### LINKED ISSUE #807: [Proposal] AppIcon Badge
### Feature name

AppIconBadge

### Link to discussion

https://github.com/CommunityToolkit/Maui/discussions/734

### Progress tracker

- [ ] Android Implementation
- [ ] iOS Implementation
- [ ] Windows Implementation
- [ ] Tizen Implementation
- [ ] Unit Tests
- [ ] Samples
- [ ] Documentation

### Summary

Screenshot 2022-11-07 at 13 46 05

Notify the user about some action in the application.

Step by step with output screenshots at https://vladislavantonyuk.azurewebsites.net/articles/AppIcon-Badge-with-.NET-MAUI

### Motivation

Better user experience and app interaction with OS.

https://developer.apple.com/documentation/uikit/uiapplication/1622918-applicationiconbadgenumber
https://github.com/leolin310148/ShortcutBadger</div>

### Detailed Design

```csharp
public interface INotificationCounter
{
	Task<bool> RequestPermission();
	void SetNotificationCount(int count);
}
```

### Usage Syntax

```markdown
var isGranted = await NotificationCounter.Default.RequestPermission();
if (isGranted)
    NotificationCounter.Default.SetNotificationCount(count);
```

### Drawbacks

Android implementation requires the support of many launchers. As part of this feature, I suggest creating a base class and implementation for some main providers, because we can't support all launchers

### Alternatives

_No response_

### Unresolved Questions

_No response_

VladislavAntonyuk:
### Description of Change ###

Adds Badge Counter

 ### Linked Issues ###
 

 - Fixes #807 

 ### PR Checklist ###
 
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/298 


src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
```diff
@@ -56,4 +56,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net7.0-android'">
+    <PackageReference Include="Xamarin.Android.ShortcutBadger" Version="1.2.0" />
+  </ItemGroup>
```
brminnick, line 62: Let's avoid adding a NuGet dependency to `CommunityToolkit.Maui.Core`

We should also avoid taking a dependency on a non-Microsoft library. (Albeit, we did take a dependency on `Xam.Plugins.Android.ExoPlayer` for `CommunityToolkit.Maui.MediaElement`).

Perhaps you could reach out to this library owner to see if we could leverage his implementation for the toolkit?

src/CommunityToolkit.Maui.Core/Essentials/BadgeCounter/BadgeCounter.shared.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        Default.SetBadgeCount(count);
+    }
+
+    /// <inheritdoc cref="IBadgeCounter" />
+	public static IBadgeCounter Default =>
+        defaultImplementation ??= new BadgeCounterImplementation();
+
+    internal static void SetDefault(IBadgeCounter implementation) =>
+        defaultImplementation = implementation;
```
brminnick, line 19: Let's use `Lazy<T>` to ensure thread safety.

Let's change the class name to `Badge`.

Let's use `SetCount()` now that the class name is `Badge`.

```suggestion
	static Lazy<IBadge> defaultImplementation = new(new BadgeImplementation());

	/// <inheritdoc cref="IBadge.SetCount" />
	public static void SetCount(int count)
	{
		Default.SetCount(count);
	}

	/// <inheritdoc cref="IBadge" />
	public static IBadge Default => defaultImplementation.Value;

	internal static void SetDefault(IBadge implementation) =>
		defaultImplementation = new(implementation);
```

brminnick (changes requested):
Thanks Vlad! 

Let's avoid adding a NuGet Dependency to `CommunityToolkti.Maui.Core`. Let's see if we can instead leverage that author's implementation. 

I'd also like to rename the namespace to `CommunityToolkit.Maui.ApplicationModel` to align with .NET MAUI and rename the class to `Badge` so that we could add additional methods in the future (eg `Clear()`).

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
```diff
@@ -193,7 +193,7 @@ static void RegisterEssentials(in IServiceCollection services)
 		services.AddSingleton<IDeviceInfo>(DeviceInfo.Current);
 		services.AddSingleton<IFileSaver>(FileSaver.Default);
 		services.AddSingleton<IFolderPicker>(FolderPicker.Default);
-		services.AddSingleton<IBadgeCounter>(BadgeCounter.BadgeCounter.Default);
+		services.AddSingleton<IBadge>(Badge.Default);
 		services.AddSingleton<ISpeechToText>(SpeechToText.Default);
 		services.AddSingleton<ITextToSpeech>(TextToSpeech.Default);
```
pictos: Wondering if we can do better. What do you think to add the default implementation on our `UseCommunityToolkit`? If the user wants a different implementation, they can override it. Of course this change should be in another PR, but this come to my mind when I saw this change

src/CommunityToolkit.Maui.Core/CommunityToolkit.Maui.Core.csproj
LennoxP90: maybe reach out to alexrainman, as his does not just encapsulate another native library
https://github.com/alexrainman/Badge

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
VladislavAntonyuk: Pros of this approach:
- User writes less code and get everything out of the box
Cons of this approach:
- Create and register instances that are not used in my project at all
- it differs from .NET MAUI implementation that also doesn't register essentials (probably because it affects performance).

pictos: > Create and register instances that are not used in my project at all

Aren't those lazy generated? So they will be mapped, but just instantiated when the user calls it for the first time?

> it differs from .NET MAUI implementation that also doesn't register essentials (probably because it affects performance).

Not sure why maui doesn't do that, maybe they didn't think about it?

Also, our lib isn't linker safe, so even if the dev doesn't use the code, it will be package by default

VladislavAntonyuk: Default creates instance immediately. We don't use Lazy, but probably should.
Do you know what should we do to make it linker friendly?

pictos: I'll create a discussion then, since this is out of the scope of this PR.

> Do you know what should we do to make it linker friendly?

Yes, I need to find the reference... When Maui was in preview I tried to run the linker analyzer on the lib but didn't work for me, let's see if now the history will be different

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/BadgeFactory.android.cs
```diff
@@ -0,0 +1,54 @@
...
+	public static IBadgeProvider? GetBadgeProvider()
+	{
+		var launcherType = GetLauncherType();
+		if (launcherType is null)
+		{
+			return null;
+		}
+
+		providers.TryGetValue(launcherType, out var provider);
+		return provider ?? new DefaultBadgeProvider();
```
pictos, line 36: can't we cache this `DefaultBadgeProvider`?

.gitignore
```diff
@@ -263,3 +263,4 @@ paket-files/
 
 # Visual Studio Code
 .vscode
+/samples/workload-install.ps1
```
brminnick, line 266: Is this specific to your local build machine, Vlad? I haven't seen a `workload-install` PowerShell script in our repo before.

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -28,6 +28,9 @@
     -->
     <!--<EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
     <CompilerGeneratedFilesOutputPath>$(BaseIntermediateOutputPath)\GeneratedFiles</CompilerGeneratedFilesOutputPath>-->
+
+    <!-- Uncomment these line if installation fails -->
```
brminnick, line 32: Let's add some additional details to this comment to make it more specific. 

eg 

If installation of what fails? 
Should I uncomment line 32 or just line 33?

samples/CommunityToolkit.Maui.Sample/Pages/Essentials/BadgePage.xaml
```diff
@@ -0,0 +1,43 @@
...
+                Title="Badge">
+    
+    <Grid
+        RowDefinitions="*, *"
+        ColumnDefinitions="*, *"
+        VerticalOptions="Center">
+
+        <Label
+            Grid.ColumnSpan="2"
+            Text="Badge"
```
brminnick, line 18:
```suggestion
            Text="Toggle Home Screen Icon Badge Count"
```

samples/CommunityToolkit.Maui.Sample/Pages/Essentials/BadgePage.xaml
```diff
@@ -0,0 +1,43 @@
...
+                x:Class="CommunityToolkit.Maui.Sample.Pages.Essentials.BadgePage"
+                xmlns:vm="clr-namespace:CommunityToolkit.Maui.Sample.ViewModels.Essentials"
+                x:TypeArguments="vm:BadgeViewModel"
+                x:DataType="vm:BadgeViewModel"
+                Title="Badge">
+    
+    <Grid
+        RowDefinitions="*, *"
+        ColumnDefinitions="*, *"
+        VerticalOptions="Center">
```
brminnick, line 14:
```suggestion
        ColumnSpacing="12"
        RowSpacing="12"
        VerticalOptions="Center">
```

samples/CommunityToolkit.Maui.Sample/Platforms/Android/AndroidManifest.xml
```diff
@@ -4,6 +4,11 @@
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 	<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
 	<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+
+  <!-- Samsung -->
+  <uses-permission android:name="com.sec.android.provider.badge.permission.READ" />
+  <uses-permission android:name="com.sec.android.provider.badge.permission.WRITE" />
```
brminnick, line 10: Just a reminder that we'll need to include this in the docs 😊

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/BadgeFactory.android.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+	static readonly Dictionary<string, IBadgeProvider> providers = new();
+	static readonly DefaultBadgeProvider defaultBadgeProvider = new();
+
+	/// <summary>
+	/// Register provider for launcher type
+	/// </summary>
+	/// <param name="launcherType">Launcher type</param>
+	/// <param name="provider">Provider implementation</param>
+	public static void SetProvider(string launcherType, IBadgeProvider provider)
```
brminnick, line 19: Let's change this to `SetBadgeProvider` to match the other public method name, `GetBadgeProvider`

```suggestion
	public static void SetBadgeProvider(string launcherType, IBadgeProvider provider)
```

samples/CommunityToolkit.Maui.Sample/Platforms/Android/MainApplication.cs
```diff
@@ -9,6 +11,9 @@ public class MainApplication : MauiApplication
 	public MainApplication(IntPtr handle, JniHandleOwnership ownership)
 		: base(handle, ownership)
 	{
+		var samsungProvider = new SamsungBadgeProvider();
+		BadgeFactory.SetProvider("com.sec.android.app.launcher", samsungProvider);
+		BadgeFactory.SetProvider("com.sec.android.app.twlauncher", samsungProvider);
```
brminnick, line 16: Just a reminder that we'll need to include this in the docs 😊

src/CommunityToolkit.Maui.Core/Essentials/Badge/IBadge.shared.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace CommunityToolkit.Maui.ApplicationModel;
+
+/// <summary>
+/// Badge
+/// </summary>
+public interface IBadge
```
brminnick, line 6: Let's also include a `int GetCount()` (or `int Count { get; }`).

Looking at the sample, after we add `Badge.GetCount()`, we can remove the `int count` field:

```cs
	[RelayCommand]
	void Decrement()
	{
		if (count > 0)
		{
			count--;
			badge.SetCount(count);
		}
	}
```

Becomes

```cs
	[RelayCommand]
	void Decrement()
	{
		if (badge.GetCount() > 0)
		{
			badge.SetCount(count);
		}
	}
```

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/DefaultBadgeProvider.android.cs
```diff
@@ -0,0 +1,59 @@
...
+			Application.Context.SendBroadcast(intent);
+		}
+		catch (Exception ex)
+		{
+			Debug.WriteLine("(DEFAULT) unable to set badge: " + ex.Message);
+		}
+	}
+	
+
+	bool IsSupported()
```
brminnick, line 50: Let's make this method name a bit more specific. Eg "Is what supported?".

src/CommunityToolkit.Maui.Core/Essentials/Badge/BadgeImplementation.net.cs
```diff
@@ -0,0 +1,11 @@
+namespace CommunityToolkit.Maui.ApplicationModel;
+
+/// <inheritdoc />
+public class BadgeImplementation : IBadge
+{
+	/// <inheritdoc />
+	public void SetCount(int count)
+	{
+		throw new NotImplementedException();
```
brminnick, line 9: Let's change this to a `NotSupportedException`

```suggestion
		throw new NotSupportedException($"{nameof(IBadge)} requires a platform-specific implementation of .NET");
```

src/CommunityToolkit.Maui.Core/Essentials/Badge/BadgeImplementation.tizen.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace CommunityToolkit.Maui.ApplicationModel;
+
+using Tizen.Applications;
+using Tizen.NUI;
```
brminnick, line 4: Let's put the `namespace` below the `using` statements to remain consistent with existing code formatting

```suggestion
using Tizen.Applications;
using Tizen.NUI;

namespace CommunityToolkit.Maui.ApplicationModel;
```

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/DefaultBadgeProvider.android.cs
```diff
@@ -0,0 +1,59 @@
...
+
+		var component = Application.Context.PackageManager?.GetLaunchIntentForPackage(packageName)?.Component;
+		if (component is null)
+		{
+			return;
+		}
+
+		if (!IsSupported())
+		{
+			return;
```
brminnick, line 32: Let's add a `Debug.WriteLine` here to let the dev know what happened instead of failing silently

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/DefaultBadgeProvider.android.cs
```diff
@@ -0,0 +1,59 @@
...
+
+		var packageName = Application.Context.PackageName;
+		if (packageName is null)
+		{
+			return;
+		}
+
+		var component = Application.Context.PackageManager?.GetLaunchIntentForPackage(packageName)?.Component;
+		if (component is null)
+		{
```
brminnick, line 26: Let's add a `Debug.WriteLine` here to let the dev know what happened instead of failing silently

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/DefaultBadgeProvider.android.cs
```diff
@@ -0,0 +1,59 @@
...
+	public void SetCount(int count)
+	{
+		if (count < 0)
+		{
+			return;
+		}
+
+		var packageName = Application.Context.PackageName;
+		if (packageName is null)
+		{
```
brminnick, line 20: Let's add a `Debug.WriteLine` here to let the dev know what happened instead of failing silently

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/DefaultBadgeProvider.android.cs
```diff
@@ -0,0 +1,59 @@
...
+
+class DefaultBadgeProvider : IBadgeProvider
+{
+	const string intentAction = "android.intent.action.BADGE_COUNT_UPDATE";
+
+	public void SetCount(int count)
+	{
+		if (count < 0)
+		{
+			return;
```
brminnick, line 15: Let's throw an `IndexOutOfRangeException` instead of failing silently:

```suggestion
			throw new IndexOutOfRangeException("Cannot set badge count below zero");
```

brminnick (changes requested):

.gitignore
VladislavAntonyuk: it is a Tizen script. each time I open the solution, this file is created. as this is a gitignore file, I don't see any problem adding this file.

src/CommunityToolkit.Maui.Core/Essentials/Badge/IBadge.shared.cs
VladislavAntonyuk: we don't know the current count (at least on Android). For android we'll return value 0.

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/DefaultBadgeProvider.android.cs
VladislavAntonyuk: other platforms (ios, maccatalyst) ignore negative value, so we shouldn't throw exception to be consistent between different platforms

src/CommunityToolkit.Maui.Core/Essentials/Badge/BadgeImplementation.windows.cs
```diff
@@ -0,0 +1,35 @@
...
+			var badge = new BadgeNotification(badgeXml);
+			badgeUpdater.Update(badge);
+		}
+	}
+
+	/// <inheritdoc />
+	public int GetCount()
+	{
+		var badgeUpdater = BadgeUpdateManager.CreateBadgeUpdaterForApplication();
+		return 0;
```
brminnick, line 33: Windows' `GetCount()` method always returns `0` instead of the actual badge count

brminnick (changes requested):
Hey Vlad! It looks like we have a bug in Windows' `GetCount()` implementation; it always returns `0`

src/CommunityToolkit.Maui.Core/Essentials/Badge/IBadge.shared.cs
```diff
@@ -0,0 +1,18 @@
...
+	/// <summary>
+	/// Set the badge count
+	/// </summary>
+	/// <param name="count">Badge count</param>
+	void SetCount(int count);
+
+	/// <summary>
+	/// Get the badge count
+	/// </summary>
+	int GetCount();
```
pictos, line 17: Any reason to choose a method instead of a property? From implementation on all platforms, this can be a property. I would say using a property is more a C# like

src/CommunityToolkit.Maui.Core/Essentials/Badge/BadgeImplementation.macios.cs
```diff
@@ -11,6 +11,7 @@ public void SetCount(int count)
 	{
 		UNUserNotificationCenter.Current.RequestAuthorization(UNAuthorizationOptions.Badge, (r, e) =>
 		{
+			Debug.WriteLine($"Error Requesting Authorization to Set Badge Count: {e.Description}");
```
pictos: I'm 90% sure this will not be shipped on the release package... Maybe changing it to `Console.WriteLine` or using the `Trace` class that will have symbols for both Debug and Release?

https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debug?view=net-7.0&redirectedfrom=MSDN

src/CommunityToolkit.Maui.Core/Essentials/Badge/IBadge.shared.cs
brminnick: That may be my fault. I recommend `int GetCount()` to Vlad as a corollary to `void SetCount()`. 

But I think you're right - as a dev, I'd rather just access a Property

src/CommunityToolkit.Maui.Core/Essentials/Badge/BadgeImplementation.macios.cs
brminnick: Dang - you're totally right. 

It looks like a bunch of our services use `Debug.WriteLine`. 

Could you open a new Bug Report on this? 

I think the best move is to just merge this PR with the `Debug.WriteLine` included, and then open a new PR to fix all of them, including `AnimationBehavior`, `BaseConverter`, `ColorAnimationExtensions`, `StateContainer`, etc.

src/CommunityToolkit.Maui.Core/Essentials/Badge/BadgeImplementation.windows.cs
VladislavAntonyuk: it is not supported. removed GetCount method

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/BadgeFactory.android.cs
```diff
@@ -0,0 +1,55 @@
...
+			return defaultBadgeProvider;
+		}
+
+		providers.TryGetValue(launcherType, out var provider);
+		return provider ?? defaultBadgeProvider;
+	}
+
+	static string? GetLauncherType()
+	{
+		var intent = new Intent(Intent.ActionMain);
```
pictos, line 42:
```suggestion
		using var intent = new Intent(Intent.ActionMain);
```

src/CommunityToolkit.Maui.Core/Essentials/Badge/Android/DefaultBadgeProvider.android.cs
```diff
@@ -0,0 +1,57 @@
...
+		}
+		catch (Exception ex)
+		{
+			Debug.WriteLine($"{nameof(DefaultBadgeProvider)} unable to set badge count: " + ex.Message);
+		}
+	}
+
+	static bool CanSetBadgeCounter()
+	{
+		var intent = new Intent(intentAction);
```
pictos, line 49:
```suggestion
		using var intent = new Intent(intentAction);
```

src/CommunityToolkit.Maui.Core/Essentials/Badge/BadgeImplementation.tizen.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace CommunityToolkit.Maui.ApplicationModel;
+
+/// <inheritdoc />
+public class BadgeImplementation : IBadge
+{
+	/// <inheritdoc />
+	public void SetCount(uint count)
+	{
+		var appId = Application.Current.ApplicationInfo.ApplicationId;
+		if (count <= 0)
```
pictos, line 13: it can't be less than zero
```suggestion
		if (count is 0)
```

src/CommunityToolkit.Maui.Core/Essentials/Badge/BadgeImplementation.windows.cs
```diff
@@ -0,0 +1,28 @@
...
+using Windows.UI.Notifications;
+
+/// <inheritdoc />
+public class BadgeImplementation : IBadge
+{
+	/// <inheritdoc />
+	public void SetCount(uint count)
+	{
+		var badgeUpdater = BadgeUpdateManager.CreateBadgeUpdaterForApplication();
+		if (count <= 0)
```
pictos, line 13:
```suggestion
		if (count is 0)
```

brminnick (approved):
Nice work, Vlad!!

PR #1165, opened 2023-04-28
PopupService

### LINKED ISSUE #981: [Proposal] PopupService
### Feature name

PopupService

### Link to discussion

https://github.com/CommunityToolkit/Maui/discussions/933

### Progress tracker

- [x] Android Implementation
- [x] iOS Implementation
- [x] MacCatalyst Implementation
- [x] Windows Implementation
- [x] Tizen Implementation
- [x] Unit Tests
- [x] Samples
- [ ] Documentation

### Summary

The following proposal provides the ability to register and show popups with associated view models. It will also allow this to be done from a view model.

### Motivation

To make it easier to show popups from a view model.

## PopupService 

```csharp
using System.ComponentModel;
using CommunityToolkit.Maui.Views;

namespace Popups;

public class PopupService
{
    private readonly IServiceProvider serviceProvider;

    private readonly static IDictionary<Type, Type> viewModelToViewMappings = new Dictionary<Type, Type>();

    public PopupService(IServiceProvider serviceProvider)
    {
        this.serviceProvider = serviceProvider;
    }

    public static void AddTransientPopup<TPopupView, TPopupViewModel>(IServiceCollection services)
        where TPopupView : Popup
        where TPopupViewModel : INotifyPropertyChanged
    {
        viewModelToViewMappings.Add(typeof(TPopupViewModel), typeof(TPopupView));

        services.AddTransient(typeof(TPopupView));
        services.AddTransient(typeof(TPopupViewModel));
    }

    public Task<object> ShowPopupAsync<TViewModel>()
    {
        var viewModelType = typeof(TViewModel);

        var viewModel = (TViewModel)this.serviceProvider.GetService<TViewModel>();

        if (viewModel is null)
        {
            throw new InvalidOperationException($"Unable to resolve type {viewModelType} please make sure that you have called RegisterPopup");
        }

        return ShowPopupAsync<TViewModel>(viewModel);
    }

    public Task<object> ShowPopupAsync<TViewModel>(TViewModel viewModel)
    {
        var viewModelType = typeof(TViewModel);

        var view = (Popup)this.serviceProvider.GetRequiredService(viewModelToViewMappings[viewModelType]);

        view.BindingContext = viewModel;

        // TODO: Do we need to provide an override for determine the main page? Or we at least need to verify there is a main page.
        return Application.Current.MainPage.ShowPopupAsync(view);
    }
}
```

## ServiceCollectionExtension

```csharp
public static class ServiceCollectionExtensions
{
    public static IServiceCollection AddTransientPopup<TPopupView, TPopupViewModel>(this IServiceCollection services)
        where TPopupView : Popup
        where TPopupViewModel : INotifyPropertyChanged
    {
        PopupService. AddTransientPopup<TPopupView, TPopupViewModel>(services);

        return services;
    }
}
```

### Usage Syntax

```cs
// Registration in MauiProgram.cs
builder.Services.AddTransientPopup<ConfirmationPopupView, ConfirmationPopupViewModel>();

// Showing of popup from a VM
popupService.ShowPopupAsync<ConfirmationPopupViewModel>();
```

### Drawbacks

It is heavily MVVM focused

### Alternatives

_No response_

### Unresolved Questions

_No response_

bijington:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #981

 ### PR Checklist ###
 

src/CommunityToolkit.Maui/PopupService.cs
```diff
@@ -0,0 +1,127 @@
...
+		return GetMainPage().ShowPopupAsync(popup);
+	}
+
+	static Page GetMainPage()
+	{
+		var currentApplication = Application.Current;
+
+		ArgumentNullException.ThrowIfNull(currentApplication);
+
+		var mainPage = currentApplication.MainPage;
```
pictos, line 92: This isn't the best way to get the current Page, because it can be the `NavigationPage` instead of the `ContentPage` presented on the screen.

I believe that you want the presented page that calls the popup. So, it's better to use this [method instead](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/PageExtensions.cs#L9)
We used it [here](https://github.com/communityToolkit/maui/blob/main/src/CommunityToolkit.Maui/Views/Popup/PopupExtensions.shared.cs#L88)  to find the page that called the popup and set it as its `Parent`.

Or you can try to find the page using the ViewModel

ne0rrmatrix: Pedro helped me with FullScreen implementation of Page with:
```csharp
﻿namespace CommunityToolkit.Maui.Extensions;

// Since MediaElement can't access .NET MAUI internals we have to copy this code here
// https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/PageExtensions.cs
static class PageExtensions
{
	internal static Page GetCurrentPage(this Page currentPage)
	{
		if (currentPage.NavigationProxy.ModalStack.LastOrDefault() is Page modal)
		{
			return modal;
		}
		if (currentPage is FlyoutPage fp)
		{
			return GetCurrentPage(fp.Detail);
		}
		if (currentPage is Shell shell && shell.CurrentItem?.CurrentItem is IShellSectionController ssc)
		{
			return ssc.PresentedPage;
		}
		if (currentPage is IPageContainer<Page> pc)
		{
			return GetCurrentPage(pc.CurrentPage);
		}

		return currentPage;
	}
}
```
Use this to get Page:
```csharp
protected Page CurrentPage =>
PageExtensions.GetCurrentPage(Application.Current?.MainPage ??
		                            throw new NullReferenceException("MainPage is null."));
```
And to actually get the page use this:
```csharp
Page page = CurrentPage;
```
I had no idea of how to get the current page in a generalized way. That code snippet will do it.

src/CommunityToolkit.Maui/PopupService.cs
```diff
@@ -0,0 +1,127 @@
...
+	static readonly IDictionary<Type, Type> viewModelToViewMappings = new Dictionary<Type, Type>();
+
+	/// <summary>
+	/// Creates a new instance of <see cref="PopupService"/>.
+	/// </summary>
+	/// <param name="serviceProvider">The <see cref="IServiceProvider"/> implementation.</param>
+	public PopupService(IServiceProvider serviceProvider)
+	{
+		this.serviceProvider = serviceProvider;
+	}
```
brminnick, line 20: I recommend using the existing references to `IServiceProvider`. This removes the requirement for the dev to inject `IServiceProvider`:

```suggestion
	static readonly IDictionary<Type, Type> viewModelToViewMappings = new Dictionary<Type, Type>();

	static IServiceProvider serviceProvider =>
#if WINDOWS
			MauiWinUIApplication.Current.Services;
#elif ANDROID
            MauiApplication.Current.Services;
#elif IOS || MACCATALYST
			MauiUIApplicationDelegate.Current.Services;
#else
			throw InvalidOperationException($"{nameof(IServiceProvider)} requires a platform specific implementation");
#endif
```

src/CommunityToolkit.Maui/PopupService.cs
```diff
@@ -0,0 +1,127 @@
...
+	{
+		ArgumentNullException.ThrowIfNull(viewModel);
+
+		var popup = GetPopup(viewModel);
+
+		GetMainPage().ShowPopup(popup);
+	}
+
+	/// <inheritdoc cref="IPopupService.ShowPopup{TViewModel}(IDictionary{string, object})"/>
+	public void ShowPopup<TViewModel>(IDictionary<string, object> query) where TViewModel : IQueryAttributable
```
brminnick, line 47: Let's include `INotifyPropertyChanged` here to match the `TViewModel` constraints on similar methods
```suggestion
	public void ShowPopup<TViewModel>(IDictionary<string, object> query) where TViewModel : IQueryAttributable, INotifyPropertyChanged
```

src/CommunityToolkit.Maui/PopupService.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.ComponentModel;
+
+namespace CommunityToolkit.Maui;
+
+/// <inheritdoc cref="IPopupService"/>
+public class PopupService : IPopupService
+{
+	readonly IServiceProvider serviceProvider;
+
+	static readonly IDictionary<Type, Type> viewModelToViewMappings = new Dictionary<Type, Type>();
```
claybrooks, line 11: Why have this mapping when the IServiceProvider can handle the view-to-viewmodel mapping for you?  Is the assumption that Popup view implementations won't make use of DI? Or is it more about invoking the popup via the viewmodel type rather than the view type in application code?

src/CommunityToolkit.Maui/PopupService.cs
```diff
@@ -0,0 +1,127 @@
...
+
+		ArgumentNullException.ThrowIfNull(mainPage);
+
+		return mainPage;
+	}
+
+	Popup GetPopup<TViewModel>(TViewModel viewModel)
+	{
+		var viewModelType = typeof(TViewModel);
+		var popup = this.serviceProvider.GetService(viewModelToViewMappings[viewModelType]) as Popup;
```
claybrooks, line 102: If my Popup view implementation expects a ViewModel injection, this line will create a second instance of the ViewModel, rather than use the one that is passed in to this function.

pictos: We don't create a new instance of an viewModel. We create an instance of pop-up, what is desired, since pop-up can't be reused.

src/CommunityToolkit.Maui/PopupService.cs
bijington: That is a good question! The initial plan was to make it cleaner to show a Popup (View) from a View Model by only knowing the type of the view model and not the view.

src/CommunityToolkit.Maui/PopupService.cs
bijington: That is a very good point! I wonder what the cleanest option would be here... we could:
- consider scoped lifetimes but that runs the risk of people not registering as scoped or moves the responsibility on us to scope (which may be good or bad depending on how we look at it)
- check if the `BindingContext` of the Popup already has a value once created and if it is of the expected view model type
- any other ideas?

claybrooks: If my Popup expects TViewModel as a constructor parameter, it will be created again. 

CsharpBindingPopup.xaml.cs provided in the samples is an example of a popup that would exhibit this behavior.

src/CommunityToolkit.Maui/PopupService.cs
bijington: Is this better/cleaner? I really can't decide... `IServiceProvider` is registered automatically for us by .NET MAUI but it might be nice to remove the constructor...

src/CommunityToolkit.Maui/PopupService.cs
claybrooks: It is possible to use the IServiceProvider to that benefit.  This way there is no worry about being in sync with other type mappings that may live in IServiceProvider.

`
internal interface ITypeMapper<TInType>
{
    Type MappedType { get; }
}

internal class ViewModelToViewTypeMapper<TInType, TOutType> : ITypeMapper<TInType>
{
     Type MappedType => typeof(TOutType);
}
`
Usage:
`
_services.AddSingleton<ITypeMapper<TViewModel>, ViewModelToViewTypeMapper<TViewModel, TView>>();
...
var viewType = _serviceProvider.GetRequiredService<ITypeMapper<TViewModel>>().MappedType;
`

Either way, small potatoes.

claybrooks: It is possible to use the IServiceProvider to that benefit.  This way there is no worry about being in sync with other type mappings that may live in IServiceProvider.

```
internal interface ITypeMapper<TInType>
{
    Type MappedType { get; }
}

internal class ViewModelToViewTypeMapper<TInType, TOutType> : ITypeMapper<TInType>
{
     Type MappedType => typeof(TOutType);
}
```
usage:
```
_services.AddSingleton<ITypeMapper<TViewModel>, ViewModelToViewTypeMapper<TViewModel, TView>>();
...
var viewType = _serviceProvider.GetRequiredService<ITypeMapper<TViewModel>>().MappedType;
```

src/CommunityToolkit.Maui/PopupService.cs
claybrooks: In my implementation, I rely on the TViewModel type to get the TView instance directly.  Then check the BindingContext as you suggest.

1. If the BindingContext is set, then probably safe to move on.
2. If the BindingContext is not set, create the ViewModel and assign it.

bijington: Yeah I think this should _mostly_ be ok. There could be the odd scenario where the developer has injected and set something different as the `BindingContext` and perhaps we should throw an exception in that scenario to make it clear that something isn't as expected. 

For now I have opted to use this approach. I will try and update the XML documentation to make it as clear as possible what is happening.

src/CommunityToolkit.Maui.Core/IArgumentsReceiver.cs
```diff
@@ -0,0 +1,13 @@
...
+/// Represents an implementation that can receive arguments during specific lifecycle events.
+/// </summary>
+public interface IArgumentsReceiver
+{
+	/// <summary>
+	/// Sets the arguments ready for use.
+	/// </summary>
+	/// <param name="arguments">A set of arguments.</param>
+	void SetArguments(IDictionary<string, object> arguments);
+}
```
bijington, line 13: This might not stick around but the intention is to not rely on `IQueryAttributable` based on an initial discussion with @PureWeen and this open issue/discussion https://github.com/dotnet/maui/issues/10294. I suspect we might opt to move towards something like `IViewAppearing` or similar but I wanted to make a clean break from `IQueryAttributable` for now so we don't misdirect anyone...

src/CommunityToolkit.Maui.UnitTests/PopupServiceTests.cs
```diff
@@ -0,0 +1,133 @@
+﻿using System.ComponentModel;
+using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.UnitTests.Mocks;
+using CommunityToolkit.Maui.Views;
+using Xunit;
+
+namespace CommunityToolkit.Maui.UnitTests;
+
+public class PopupServiceTests : BaseHandlerTest
```
bijington, line 9: We should unit test and prove that only the expected instances are created during Popup creation.

src/CommunityToolkit.Maui.Core/IArgumentsReceiver.cs
pictos: We could use a [`IReadOnlyDictionary`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlydictionary-2?view=net-8.0) here, this will ensure the collection is immutable. This is based on feedback that I got for `IQueryAttributable`.

bijington: Excellent idea!

src/CommunityToolkit.Maui/PopupService.cs
claybrooks: There are still extra instances of TViewModel being created in the ShowPopup/ShowPopupAsync() methods.  These should be removed and GetViewModel<TViewModel>() should only be called when BindingContext is null

MDale-RAC:
Any updates on this?

bijington:
> Any updates on this?

No I haven't been able to determine what the best option is for `IArgumentsReceiver` - whether we want a different name or better abstraction. Any input is welcome :)

bijington:
As an alternative to requiring an interface when presenting the popup for the view model to receive data we could try something like this:

```csharp
public Task<object?> ShowPopupAsync<TViewModel>(Action<TViewModel> onPresenting) where TViewModel : INotifyPropertyChanged
{
    ArgumentNullException.ThrowIfNull(onPresenting);

    var popup = GetPopup(typeof(TViewModel));

    var viewModel = GetViewModel<TViewModel>();

    popup.BindingContext = viewModel;

    onPresenting.Invoke(viewModel);

    return CurrentPage.ShowPopupAsync(popup);
}

```

The `onPresenting` action would be invoked once the view model is created and assigned to the popup BindingContext, just before the popup is presented.

It could then be used as follows:

```csharp
public class MyViewModel : INotifyPropertyChanged
{
    public event PropertyChangedEventHandler? PropertyChanged;

    public void LoadData(int id)
    {
    }
}

void Show()
{
    ShowPopupAsync<MyViewModel>(onPresenting: viewModel => viewModel.LoadData(123));
}
```

What do people think? Is it clean or a bit yucky?

VladislavAntonyuk:
> As an alternative to requiring an interface when presenting the popup for the view model to receive data we could try something like this:
> 
> ```cs
> public Task<object?> ShowPopupAsync<TViewModel>(Action<TViewModel> onPresenting) where TViewModel : INotifyPropertyChanged
> {
>     ArgumentNullException.ThrowIfNull(onPresenting);
> 
>     var popup = GetPopup(typeof(TViewModel));
> 
>     var viewModel = GetViewModel<TViewModel>();
> 
>     popup.BindingContext = viewModel;
> 
>     onPresenting.Invoke(viewModel);
> 
>     return CurrentPage.ShowPopupAsync(popup);
> }
> ```
> 
> The `onPresenting` action would be invoked once the view model is created and assigned to the popup BindingContext, just before the popup is presented.
> 
> It could then be used as follows:
> 
> ```cs
> public class MyViewModel : INotifyPropertyChanged
> {
>     public event PropertyChangedEventHandler? PropertyChanged;
> 
>     public void LoadData(int id)
>     {
>     }
> }
> 
> void Show()
> {
>     ShowPopupAsync<MyViewModel>(onPresenting: viewModel => viewModel.LoadData(123));
> }
> ```
> 
> What do people think? Is it clean or a bit yucky?

I like it, but make it Func<ViewModel, Task>

src/CommunityToolkit.Maui/PopupService.cs
```diff
@@ -0,0 +1,171 @@
...
+using System.ComponentModel;
+
+namespace CommunityToolkit.Maui;
+
+/// <inheritdoc cref="IPopupService"/>
+public class PopupService : IPopupService
+{
+	readonly IServiceProvider serviceProvider;
+
+	static readonly IDictionary<Type, Type> viewModelToViewMappings = new Dictionary<Type, Type>();
```
pictos, line 13: since this is private and not injected, we should use the concrete type, changing in the future, if needed, will not be a breaking change.
```suggestion
	static readonly Dictionary<Type, Type> viewModelToViewMappings = new Dictionary<Type, Type>();
```

src/CommunityToolkit.Maui/PopupService.cs
```diff
@@ -0,0 +1,171 @@
...
+
+		var viewModel = AssignBindingContext<TViewModel>(popup);
+
+		onPresenting.Invoke(viewModel);
+
+		CurrentPage.ShowPopup(popup);
+	}
+
+	/// <inheritdoc cref="IPopupService.ShowPopup{TViewModel}(Func{TViewModel, Task})"/>
+	public async void ShowPopup<TViewModel>(Func<TViewModel, Task> onPresenting) where TViewModel : INotifyPropertyChanged
```
pictos, line 67: I think we don't need this overload, the way _sync  over async_ works, the users can have the same behavior using the method at line `53`, they just to make the `Action` async.

bijington: I guess I wasn't sure if we wanted to allow users to wait until their async calls had finished before showing the popup rather than after. I am all for a smaller API surface so happy to remove the `Func<TViewModel, Task>` option though

pictos: With the `Task ShowPopupAsync (Func<TViewModel, Task> onPresenting)` we will await until the Func completes to show the popup because we await it and after that, we show the popup.

bijington: I have removed the Task options for now

pictos: @bijington 
Just to picture the behavior on my mind... What should be the behavior when I do:

```csharp

var result = await PopupService.ShowPopup( async (vm) => {

var data = await myService.LoadDataAsync();
Items.AddRange(data);
});
```

Here the Popup should be visible during the `await myService.LoadDataAsync();` or it should wait until the `Func` finish his execution to show up?

bijington: I expect the Popup to be visible during the `await myService.LoadDataAsync();` I have added a sample that shows this. It shows the Popup and then simulates the running of updates with the Popup showing the updates. I think this is enough, if we wanted to prevent the Popup from showing then the `Func<Task>` might need to return however I don't know if that would result in a good user experience.

The final piece of the puzzle is a clean way to allow the view model to return the value back from the Popup

src/CommunityToolkit.Maui.Core/IPopupService.cs
```diff
@@ -0,0 +1,54 @@
...
+/// <summary>
+/// Provides a mechanism for displaying <see cref="CommunityToolkit.Maui.Core.IPopup"/>s based on the underlying view model.
+/// </summary>
+public interface IPopupService
+{
+	/// <summary>
+	/// Resolves and displays a <see cref="CommunityToolkit.Maui.Core.IPopup"/> and <typeparamref name="TViewModel"/> pair that was registered with <c>AddTransientPopup</c>.
+	/// </summary>
+	/// <typeparam name="TViewModel">The type of the view model registered with the <see cref="CommunityToolkit.Maui.Core.IPopup"/>.</typeparam>
+	void ShowPopup<TViewModel>() where TViewModel : INotifyPropertyChanged;
```
prostopsih, line 14: Hi, can we use Type viewModelType instead of TViewModel in interface and make an extension with TViewModel?

prostopsih, line 14: Is INotifyPropertyChanged constraint really necessary here? Can we use class instead?

src/CommunityToolkit.Maui.Core/IPopupService.cs
```diff
@@ -0,0 +1,54 @@
...
+
+	/// <summary>
+	/// Resolves and displays a <see cref="CommunityToolkit.Maui.Core.IPopup"/> and <typeparamref name="TViewModel"/> pair that was registered with <c>AddTransientPopup</c>.
+	/// The supplied <paramref name="onPresenting"/> provides a mechanism to invoke any methods on your view model in order to load or pass data to it.
+	/// </summary>
+	/// <typeparam name="TViewModel">The type of the view model registered with the <see cref="CommunityToolkit.Maui.Core.IPopup"/>.</typeparam>
+	/// <param name="onPresenting">An <see cref="Action{TViewModel}"/> that will be performed before the popup is presented.</param>
+	void ShowPopup<TViewModel>(Action<TViewModel> onPresenting) where TViewModel : INotifyPropertyChanged;
```
prostopsih, line 22: I don't really like onPresenting parameter here, in my case, I would like to have a control of Presenting, Presented directly in ViewModel, I've got a custom solution for my needs, I used an interface, with proper naming it might be something like IPresentableViewModel, and inside of a service implementation you could call the method if viewModel implements this interface. There is a [link](https://github.com/prostopsih/Nikklim.App.Navigation) to my custom solution, it has more than just Popups, but anyway, I would like to extend your proposal with my ideas if there is any interest in it.

src/CommunityToolkit.Maui.Core/IPopupService.cs
brminnick: The PopupService is created to facilitate using Popups with the MVVM architecture.

In the MVVM architecture, the way the ViewModel communicates updates to the View is via `INotifyPropertyChanged`. If your class isn't implementing `INotifyPropertyChanged`, then it isn't really a ViewModel and doesn't require this method; you can use the other, existing, methods to display your Popup: https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/views/popup#presenting-a-popup

prostopsih: ViewModel may have only commands and in this case INotifyPropertyChanged is not needed, I just don't like adding unnecessary constraints, it may make a problem in some very niche case

bijington: It is possible but what is your use case for wanting to supply a Type instead?

bijington: We did discuss whether we would need the INPC constraint and I believe we came to the conclusion that forcing this constraint would make people consider why they need it. I believe we also agreed that it wouldn't be a breaking change to remove the constraint rather than add it.

src/CommunityToolkit.Maui.Core/IPopupService.cs
bijington: I'm guessing your repository is private as I cannot see the contents when clicking the link.

The part I like about not using an interface is that it lets the caller use whatever implementation they like and it doesn't tie them into a constrained interface that we have to provide. It'll be good to see your code though to compare

src/CommunityToolkit.Maui.Core/IPopupService.cs
prostopsih: In the codebase that I work with, unfortunately, we have a  viewModel type passing from other parts of the code, so I know the type of viewModel needed only on runtime, it  I believe I'm not the only one with such issue. Also as far as I know, it's a good practice to pass a type as a parameter and then have a generic extension

src/CommunityToolkit.Maui.Core/IPopupService.cs
prostopsih: I apologise for it, it's public now

bijington: I think I am inclined to leave this as-is for now. I worry about defining yet another interface for users to have to implement when it might force them down a route which won't suit their scenario. You could quite easily wrap up this method with an extension to solve your needs though:

```csharp
public void ShowPopup<TViewModel>(this IPopupService popupService, IDictionary<string, object> arguments) where TViewModel : IPresentableViewModel, INotifyPropertyChanged
{
    popupService.ShowPopup<TViewModel>(onPresenting: viewModel => viewModel.OnPresenting(arguments);
}
```

Something like the above albeit I am making a number of assumptions about your potential implementation which is probably wrong :).

src/CommunityToolkit.Maui.Core/IPopupService.cs
bijington: I'd be OK with expanding the popup service to support this but perhaps we can open up a proposal to modify it once this PR gets in?

src/CommunityToolkit.Maui/PopupService.cs
```diff
@@ -0,0 +1,160 @@
...
+			throw new InvalidOperationException(
+				$"Unable to resolve popup type for {viewModelType} please make sure that you have called {nameof(AddTransientPopup)}");
+		}
+
+		return popup;
+	}
+
+	TViewModel GetViewModel<TViewModel>()
+	{
+		var viewModel = this.serviceProvider.GetService<TViewModel>();
```
VladislavAntonyuk, line 150: Here may be a problem if ViewModel is Scoped we need to create a scope and resolve VM from it.

VladislavAntonyuk (approved):

brminnick (approved):
Thanks Shaun!! I'm excited to see PopupService finally become a reality!!

I'm going to merge this right away so that we can squeeze it into this week's release. 

Could you prioritize getting the PopupService docs done before Monday so that PopupService doesn't go too long without proper documentation? 

FYI - I made a small tweak, removing PopupService's functionality that automatically assigned BindingContexts. I renamed the method to `ValidateBindingContext` (below) to ensure that the user has properly assigned the Popup View's BindingContext. However, I think it's out-of-scope for the PopupService to automatically assign it for them. 

For example, .NET MAUI doesn't automatically assign a ContentPage's BindingContext when a user passes it into `IServiceCollection`. 

I'm happy to circle back to this discussion in the future if we get requests from the community. But I agree that PopupService should be opinionated, and it should require the correct BindingContext. And us throwing an `InvalidOperationException` is enough to inform the dev of the requirement, in lieu of doing it for them under the hood (which they may not expect/want). 

```cs
static void ValidateBindingContext<TViewModel>(Popup popup, out TViewModel bindingContext)
{
 if (popup.BindingContext is not TViewModel viewModel)
 {
 	throw new InvalidOperationException($"Unexpected type has been assigned to the BindingContext of {popup.GetType().FullName}. Expected type {typeof(TViewModel).FullName} but was {popup.BindingContext?.GetType().FullName?? "null"}");
 }

 bindingContext = viewModel;
}

src/CommunityToolkit.Maui/PopupService.cs
brminnick: Noted. Let's keep this in mind if/when a community member opens up an Issue / Feature Request.

PR #1179, opened 2023-05-10
Add IconTintColorBehavior for windows

kperdlich:
### Description of Change ###
Adds windows support for the IconTintColorBehavior using the [Windows.UI.Composition API](https://learn.microsoft.com/en-us/windows/uwp/composition/visual-layer). It is based on the original proposal of the XCT implementation:  https://github.com/xamarin/XamarinCommunityToolkit/pull/1934. 

 

 ### Linked Issues ###
 

 - Fixes #

 ### PR Checklist ###
 

kperdlich:
@dotnet-policy-service agree

pictos (commented):
Thanks for the work, here are some comments

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.windows.cs
```diff
@@ -0,0 +1,255 @@
...
+
+		var maskBrush = compositor.CreateMaskBrush();
+		maskBrush.Source = sourceColorBrush;
+		maskBrush.Mask = compositor.CreateSurfaceBrush(loadedSurfaceMask);
+
+		spriteVisual = compositor.CreateSpriteVisual();
+		spriteVisual.Brush = maskBrush;
+		spriteVisual.Size = new Vector2(width, height);
+		spriteVisual.AnchorPoint = Vector2.Zero;
+		spriteVisual.CenterPoint = new Vector3(width * .5f, height * .5f, 0f);
```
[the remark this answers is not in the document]
kperdlich: Its for applying rotations on the center of the image, but since rotations are applied on the parent its doesn't matter. Removed it 👍

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.windows.cs
```diff
@@ -0,0 +1,234 @@
...
+using WButton = Microsoft.UI.Xaml.Controls.Button;
+using WImage = Microsoft.UI.Xaml.Controls.Image;
+
+namespace CommunityToolkit.Maui.Behaviors;
+
+public partial class IconTintColorBehavior
+{
+	SpriteVisual? spriteVisual;
+	Vector2? originalImageSize;
+	bool IsUpdate => originalImageSize is not null;
```
brminnick, line 19: Let's make this variable name more descriptive. 
* Is something being updated or is something updated?
* What is being updated?

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.windows.cs
```diff
@@ -0,0 +1,233 @@
...
+			ApplyTintColor(platformView, bindable, TintColor);
+		}
+	}
+
+	void ApplyTintColor(FrameworkElement platformView, View element, Color? color)
+	{
+		if (color is null)
+		{
+			RemoveTintColor(platformView);
+		}
```
brminnick, line 89: What happens if the user passes in `Colors.Transparent`? 

Should we also call `RemoveTintColor()`?

brminnick (changes requested):
Thanks @kperdlich! This is looking really good! 

I have just a small change request and a question about how we'll handle `Colors.Transparent`.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.windows.cs
kperdlich: Basically it was to differentiate in general between the initial apply of the behavior or a later update by a property change. I don't need to handle this case explicitly anymore since I use the `ImageSource.ImageOpened` to get notified when the image size is ready. So it doesn't matter if its the initial apply of the behavior or coming from a property change.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.windows.cs
kperdlich: I reworked that part so currently its not required to handle this case explicitly.  Technically we could still add a check and skip the composition part what might be beneficial for performance/memory usage. What do you think? @brminnick

brminnick:
Hey @kperdlich! Is this PR ready to graduate from `Draft` to `Ready for review`?

kperdlich:
@brminnick Yeah, I think it should be ready.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.windows.cs
```diff
@@ -0,0 +1,223 @@
...
+	SpriteVisual? spriteVisual;
+	Vector2? originalImageSize;	
+
+	/// <inheritdoc/>
+	protected override void OnAttachedTo(View bindable, FrameworkElement platformView)
+	{
+		ApplyTintColor(platformView, bindable, TintColor);
+
+		bindable.PropertyChanged += OnElementPropertyChanged;
+		this.PropertyChanged += (s, e) =>
```
VladislavAntonyuk, line 27: should we unsubscribe to this event on detach?

brminnick: I had a similar thought, but then noticed that this matches the existing implementation in `IconTintColorBehavior.ios.cs`:

https://github.com/CommunityToolkit/Maui/blob/21cb4eac178627022790cff0af7c376977b2e81a/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.ios.cs#L14-L20

pictos: It's fine to not unsubscribe, since the event is related to the same object, then will have the same life time, so it will be collected.

To illustrate that imagine that you've a Page with a button on that page, then you subscribe to the Click event, if you use xaml, you can't unsubscribe that event and even with that the Page will not leak.

```csharp
class MyPage : ContentPage

Button loginButton = new();

public MyPage()
{
   loginButton.Clicked += (s,e) =>
  {
     // Do something
  }
}
```

Since the life time of Page and Button are the same, the event will be collected correctly.

pictos (dismissed):
@kperdlich running our sample I saw an issue on the Bindings, if you click the last `ImageButton` it should switch the color every time that you click on it, but the change occurs just once.

I verified on Android and it works as expected there, if you want to use it as a reference.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.windows.cs
```diff
@@ -0,0 +1,223 @@
...
+			case WButton button:
+				if (TryGetButtonImage(button, out var image))
+				{
+					RestoreOriginalImageSize(image);
+					ElementCompositionPreview.SetElementChildVisual(image, null);
+				}
+				break;
+
+			default:
+				throw new NotSupportedException($"{nameof(IconTintColorBehavior)} only currently supports {typeof(WImage)} and {typeof(WButton)}.");
```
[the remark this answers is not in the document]
kperdlich: Removed it from `RemoveTintColor()` since the check should be enough in  `ApplyTintColor()`.

kperdlich:
> @kperdlich running our sample I saw an issue on the Bindings, if you click the last `ImageButton` it should switch the color every time that you click on it, but the change occurs just once.
> 
> I verified on Android and it works as expected there, if you want to use it as a reference.

Ah nice catch. Was a regression when I reworked the image loading for Source property changes. Should work now.

brminnick (approved):

brminnick:
Thanks @kperdlich!!

Just FYI - I also realized that we didn't include macOS in the original implementation, so I've added that to this PR. It was just an oversight on our part and the change doesn't require any additional logic.

For the curious, I just renamed `IconTintColorBehavior.ios.cs` -> `IconTintColorBehavior.macios.cs`. This filename tells the compiler to include `IconTintColorBehavior` in both the iOS DLL and the macOS DLL. We don't need to change any logic since iOS + MacCatalyst use the same APIs for this feature.

pictos:
@lothrop I believe we don't work on `Button` control, just `ImageButton` and `Image` controls

Areshosu:
Using it as a regular Image doesn't work well 
image

PR #1232, opened 2023-06-11
Initial work on adding remaining animations

bijington:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #1231

 ### PR Checklist ###
 

src/CommunityToolkit.Maui/Animations/ScaleAnimation.shared.cs
```diff
@@ -0,0 +1,48 @@
...
+	public double Scale
+	{
+		get => (double)GetValue(ScaleProperty);
+		set => SetValue(ScaleProperty, value);
+	}
+
+	/// <summary>
+	/// Initializes a new instance of <see cref="ScaleAnimation"/>.
+	/// </summary>
+	public ScaleAnimation() : base(340)
```
VladislavAntonyuk, line 31: is there a reason why we use different values for each animation?

src/CommunityToolkit.Maui.UnitTests/Animations/BaseAnimationTests.cs
```diff
@@ -0,0 +1,31 @@
...
+		Label label = new();
+
+		label.EnableAnimations();
+
+		var stopwatch = new Stopwatch();
+		stopwatch.Start();
+		await animation.Animate(label);
+		stopwatch.Stop();
+
+		stopwatch.ElapsedMilliseconds.Should().BeCloseTo(animation.Length, 50);
```
VladislavAntonyuk, line 27: can we get this value from some constant in Animation from library code?

src/CommunityToolkit.Maui/Animations/RotateAnimation.shared.cs
```diff
@@ -0,0 +1,62 @@
...
+			typeof(double),
+			typeof(RotateAnimation),
+			180.0,
+			BindingMode.TwoWay,
+			defaultValueCreator: GetDefaultRotationProperty);
+
+	/// <summary>
+	/// Gets or sets the rotation used by the animation.
+	/// </summary>
+	public double Rotation
```
VladislavAntonyuk, line 23: maybe name it Angle?

src/CommunityToolkit.Maui/Animations/ScaleAnimation.shared.cs
```diff
@@ -0,0 +1,48 @@
...
+	{
+
+	}
+
+	/// <inheritdoc />
+	public override async Task Animate(VisualElement view)
+	{
+		ArgumentNullException.ThrowIfNull(view);
```
VladislavAntonyuk, line 39: somewhere we check view for null (RotateAnimation), somewhere we throw an exception.

src/CommunityToolkit.Maui/Animations/RotateAnimation.shared.cs
bijington: This is the property name that was used in XCT and it relates to the property that it is animating (`VisualElement.Rotation`). I would like to investigate more around our use of `RotateTo` vs `RelRotateTo` though.

src/CommunityToolkit.Maui/Animations/ScaleAnimation.shared.cs
bijington: This again comes from XCT, I guess these are the best defaults for a good look and feel of each specific animation.

src/CommunityToolkit.Maui/Animations/RotateAnimation.shared.cs
VladislavAntonyuk: let's keep Rotation: https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.visualelement.rotation?view=net-maui-7.0#microsoft-maui-controls-visualelement-rotation

src/CommunityToolkit.Maui.UnitTests/Animations/BaseAnimationTests.cs
bijington: This might highlight that the code doesn't read well. the `Length` property of the `BaseAnimation` is that value we pass into the constructor. The `50` is my attempt at verifying that the completed duration was within 50 milliseconds of the expected duration. Perhaps this could be a calculated percentage instead. I still need to find a reliable way of verifying this because running the tests locally don't yield positive test results for all yet.

bijington:
Yes it will

bijington:
I believe I have already addressed that issue in this PR I'm just struggling to get the unit test to prove it

src/CommunityToolkit.Maui/Animations/ShakeAnimation.shared.cs
```diff
@@ -0,0 +1,73 @@
...
+	{
+
+	}
+
+	/// <inheritdoc />
+	public override async Task Animate(VisualElement view)
+	{
+		ArgumentNullException.ThrowIfNull(view);
+
+		var duration = (uint)(Length / Math.Ceiling(StartFactor / ReducingAmount)) / 2;
```
bijington, line 62: I must be missing something obvious but I can't work out why this fails when being tested. The current values mean there are 3 shakes crossing the x == 0 point twice. This should result in 6 `await`s however the test results in a failure because the animation completes within roughly 200ms. Given the above I calculate with a `Length == 300`, `duration` should equal 50 and therefore result in a passing test. However it doesn't appear to work

bijington:
As far as I can tell it might be down to our use of `Task.Delay` in the mocking of the animation layer. I have been trying to investigate with a more accurate option for mocking

PR #1258, opened 2023-06-19
Call InvalidateMeasure on DockLayout when properties change affecting layout

### LINKED ISSUE #1250: [BUG] DockLayout doesn't update when DockPosition updated via Hot Reload / code behind

### Current Behavior

DockLayout doesn't fully work with Hot Reload - when updating the `DockPosition` attached property, or the `ShouldExpandLastChild`/`HorizontalSpacing`/`VerticalSpacing` regular properties, the change isn't reflected until something forces it to layout again. Updating via code behind doesn't update immediate either (as is typical of these Hot Reload issues - since setting the property is what Hot Reload does).

### Expected Behavior

Property updates should be reflected immediately.

### Steps To Reproduce

1. Open and run the solution from the repro repository, debugging in order to use Hot Reload. I tested with Windows, though the issue likely happens on all platforms.
2. Open MainPage.xaml in VS
3. For the `RightToBottom` button, change the `DockPosition` from `Right` to `Bottom`.
Result: Hot Reload applies the change, but the button doesn't move, demonstrating that Hot Reload is broken here.
4. Click the `Move RightToLeft Button` button, which calls code behind to move the right button to the bottom.
Result: The UI doesn't update here either, demonstrating the same problem happens with code behind.
5. Add `ShouldExpandLastChild`/`HorizontalSpacing`/`VerticalSpacing` properties to the DockLayout.
Result: Hot Reload applies the change, but the layout doesn't update.

Now resize the window.
Result: Buttons are drawn in their new position, demonstrating that the changes were applied but they just didn't trigger the refresh immediately like they should have.

https://github.com/BretJohnson/DockLayoutHotReloadNotWorking

### Environment

```markdown
- .NET MAUI CommunityToolkit: 5.2.0
- OS: Win 11
- .NET MAUI: 7.0.86
```

### Anything else?

_No response_

BretJohnson:
### Description of Change ###

When properties are updated affecting layout, `InvalidateMeasure` should be called to trigger the layout to happen, normally called via a propertyChanged notification. This PR adds that for `DockLayout`. The `InvalidateMeasure` method added here is similar to what's called for Grid properties in the MAUI repo.

 ### Linked Issues ###
 

 - Fixes #1250

 ### PR Checklist ###
 
Note that I attempted adding unit tests, but none of my attempts were able to catch the behavior difference when `InvalidateMeasure` is called. This may be something normally done by "device tests" in the MAUI repo. So I ended up just testing manually against my test repo, which all looked good.

### DIFF
--- a/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.windows.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.windows.cs
@@ -30,7 +30,7 @@ public sealed partial class SpeechToTextImplementation
 		speechRecognitionEngine = null;
 		speechRecognizer = null;
 	}
-	
+
 	async Task<string> InternalListenAsync(CultureInfo culture,
 		IProgress<string>? recognitionResult,
 		CancellationToken cancellationToken)
--- a/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.windows.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.windows.cs
@@ -30,7 +30,7 @@ public sealed partial class SpeechToTextImplementation
 		speechRecognitionEngine = null;
 		speechRecognizer = null;
 	}
-
+	
 	async Task<string> InternalListenAsync(CultureInfo culture,
 		IProgress<string>? recognitionResult,
 		CancellationToken cancellationToken)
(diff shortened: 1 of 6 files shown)


brminnick:
Thanks Bret! 

Just curious - are there any severe performance impacts to this? In other words, should we wrap InvalidateMeasure in `#if DEBUG` so that it works in Hot Reload and avoids any potential performance problems?

BretJohnson:
> Thanks Bret!
> 
> Just curious - are there any severe performance impacts to this? In other words, should we wrap InvalidateMeasure in `#if DEBUG` so that it works in Hot Reload and avoids any potential performance problems?

No, doing the InvalidateMeasure when updating properties that can affect layout is expected - all the built in MAUI controls do it. Note that it's not only needed for Hot Reload; it's also needed when those properties are updated (beyond what's in the original XAML) by C# code. So it's a potential app runtime problem too, not just a Hot Reload issue.

brminnick (approved):
Thanks Bret!

PR #1260, opened 2023-06-21
Fix coordinate usage on Windows Maps

jfversluis:
By providing the coordinates as a string there would still be an offset to where the coordinates should actually be placed and where they ended up on the map.

This change removes the JavaScript quotes and instead does a `ToString(CurrentCulture.InvariantCulture);` on the C# objects to the decimal separator is not converted to a comma and will not interfere with the JavaScript syntax.

Follow up from #1257


VladislavAntonyuk (approved):
😃

PR #1262, opened 2023-06-21
Remove `VerticalStackLayout` from `StateContainerController`

### LINKED ISSUE #1118: [BUG] StateContainer - Constraints are not working properly

### Current Behavior

I have found multiple bugs but I think they can be caused by the same issue: Layouts ignore set constraints when used inside of state container as a state.

| Layout | Layout not inside of state container (working) | Layout inside of the container (not working) |
| -- | -- | -- |
| Grid + Label | image | image |
| ScrollView | image | image
| ItemView | image | image |

But if set a specific height to that layout it shows up as:
image

These all seem like they originate from how the bounds of the sub-layouts are calculated if they are calculated at all.

### Expected Behavior

Images with expected behavior are in the previous paragraph.

### Steps To Reproduce

I included an application repository where you can try working and not working layouts.

1. We can just run without doing any edits why launches the app which is using a state container.
2. If we want to see working constraints we can uncomment the Working Constraints region.
3. If we want to see a working Scroll View we can uncomment the Working Scroll View region.
4. If we want to see a working refresh view we can uncomment the Working Refresh View region.

https://github.com/zZHorizonZz/MAUI-Bug-Report-App

### Environment

```markdown
- OS: Windows 10 Enterprise (OS build 19044.2604)
- .NET MAUI: 7.0
```

### Anything else?

_No response_

brminnick:
### Description of Change ###

This PR removes the `VerticalStackLayout` from the `StateContainer`.

.NET MAUI has [different behavior than Xamarin.Forms](https://github.com/dotnet/maui/issues/8888) for `StackLayout`. In .NET MAUI, `VerticalStackLayout` expands to the size of its children. When the children require more space than is available on the screen, `VerticalStackLayout` will continue to "display" the additional `View` below the visible area on the screen.

The example below uses a `ScrollView` inside of `StateContainer`. When `StateContainerController` places the `ScrollView` inside of a `VerticalStackLayout`, it does not scroll because the size of the `VerticalStackLayout` extends past the bottom of the screen. 

| `StateContainer` with `VerticalStackLayout` | `StateContainer` without `VerticalStackLayout` (Fixed)| 
| -- | -- |
| Cannot scroll the `ScrollView` because `VerticalStackLayout` has laid out its children past the bottom of the screen  (notice how the Yellow color extends past the `Button` with a `BackgroundColor = Colors.Pink`)| Can scroll because the `ScrollView` is properly sized |
| Simulator Screen Recording - iPhone 14 Pro - 2023-06-21 at 18 32 41 | Simulator Screen Recording - iPhone 14 Pro - 2023-06-21 at 18 31 28 |

 ### Linked Issues ###
 

 - Partially Fixes #1118

 ### PR Checklist ###

I've also updated `StateContainerController` to use `IList<IView> originalContent` instead of `List<View> originalContent` to avoid the need to cast `item` when adding it to the list.

### DIFF
--- a/src/CommunityToolkit.Maui/Layouts/StateContainer/StateContainerController.shared.cs
+++ b/src/CommunityToolkit.Maui/Layouts/StateContainer/StateContainerController.shared.cs
@@ -8,7 +8,7 @@ sealed class StateContainerController
 	readonly WeakReference<Layout> layoutWeakReference;
 
 	string? previousState;
-	IList<IView> originalContent = Enumerable.Empty<IView>().ToList();
+	List<IView> originalContent = Enumerable.Empty<IView>().ToList();
 
 	/// <summary>
 	/// Initialize <see cref="StateContainerController"/> with a <see cref="Layout"/>
--- a/src/CommunityToolkit.Maui/Layouts/StateContainer/StateContainerController.shared.cs
+++ b/src/CommunityToolkit.Maui/Layouts/StateContainer/StateContainerController.shared.cs
@@ -71,6 +71,7 @@ sealed class StateContainerController
 			{
 				VerticalOptions = view.VerticalOptions,
 				HorizontalOptions = view.HorizontalOptions,
+				InputTransparent = true
 			};
 
 			if (grid.RowDefinitions.Count > 0)
--- a/src/CommunityToolkit.Maui/Layouts/StateContainer/StateContainerController.shared.cs
+++ b/src/CommunityToolkit.Maui/Layouts/StateContainer/StateContainerController.shared.cs
@@ -64,36 +64,21 @@ sealed class StateContainerController
 		// Otherwise it would just end up in row 0 : column 0.
 		if (layout is IGridLayout grid)
 		{
-			// We create a VerticalStackLayout spanning the Grid.
-			// It takes VerticalOptions and HorizontalOptions from the
-			// view to allow for more control over how it layouts.
-			var innerLayout = new VerticalStackLayout
-			{
-				VerticalOptions = view.VerticalOptions,
-				HorizontalOptions = view.HorizontalOptions,
-				InputTransparent = true
-			};
-
 			if (grid.RowDefinitions.Count > 0)
 			{
-				Grid.SetRowSpan(innerLayout, grid.RowDefinitions.Count);
+				Grid.SetRowSpan(view, grid.RowDefinitions.Count);
 			}
 
 			if (grid.ColumnDefinitions.Count > 0)
 			{
-				Grid.SetColumnSpan(innerLayout, grid.ColumnDefinitions.Count);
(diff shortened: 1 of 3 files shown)


VladislavAntonyuk (dismissed):

src/CommunityToolkit.Maui/Layouts/StateContainer/StateContainerController.shared.cs
```diff
@@ -8,7 +8,7 @@ sealed class StateContainerController
 	readonly WeakReference<Layout> layoutWeakReference;
 
 	string? previousState;
-	List<View> originalContent = Enumerable.Empty<View>().ToList();
+	IList<IView> originalContent = Enumerable.Empty<IView>().ToList();
```
pictos, line 11: this's a private field and isn't injected anywhere. So, I would say it's fine to use the concrete type here.

brminnick: Hey Pedro! Are you advocating to continue using `View` instead of replacing it with `IView`? 

I made the change to `IView` so that we could remove the casting here: https://github.com/CommunityToolkit/Maui/pull/1262/files#diff-c16a4976748b4af82ec28108ed444db5aa96141fef8d3a90e3b9769e2ea872e9L55-R55

pictos: No, using `List<IView>`, instead of `IList<IView>`

brminnick:
```suggestion
	List<IView> originalContent = Enumerable.Empty<IView>().ToList();
```

pictos (approved):

VladislavAntonyuk (approved):

BobSammers:
Sorry I'm late to question this commit, but is this the correct way to treat this problem? The `VerticalStackLayout` had to go, obviously (I found myself here because of issues it is causing in my layout), but I'm concerned about maintaining the principle that the switched view should fill a containing `Grid`. Why is this desirable? Am I missing something?

What if the consumer _wants_ the `View` to only go in a particular cell? Why force it to fill the containing `Grid`? It seems to me the correct behaviour is to respect `Grid.Row`, `Grid.Column`, `Grid.RowSpan`, etc., either on the `<toolkit:StateContainer.StateViews>`, or on the root of each of the views inside this group (which allows variations upon switching).

If the current behaviour was kept unless these properties were set, that should pretty much eliminate concerns over breaking changes.

Wrapping the `View` selector in a single-celled `Grid` is a partial workaround, but it introduces friction, is not intuitive and doesn't allow different views to occupy different locations in the containing `Grid`'s cell structure.

PR #1264, opened 2023-06-23
Implement AppThemeColor, AppThemeObject, and AppThemeResource

jfversluis:
Still todo:
- [x] Determine if we even want this (I think we do)
- [x] Find the right place/namespace for it
- [x] Add some samples for it
- [x] Add code documentation
- [x] Add conceptual docs: https://github.com/MicrosoftDocs/CommunityToolkit/pull/297

This adds the AppThemeColor and AppThemeResource objects that make it easier to work with colors and other resources in light/dark theme scenarios. This enables you to define colors (and resources) in resource dictionaries, making your XAML less verbose and more reusable.

Original PR on .NET MAUI by Dan Siegel: https://github.com/dotnet/maui/pull/8839


src/CommunityToolkit.Maui/Extensions/AppThemeExtension.shared.cs
```diff
@@ -0,0 +1,91 @@
+﻿namespace CommunityToolkit.Maui.Extensions;
+
+/// <summary>
+/// A XAML markup extension that enables using <see cref="AppThemeColor"/> and <see cref="AppThemeResource"/> from XAML.
+/// </summary>
+[ContentProperty(nameof(Key))]
+public sealed class AppThemeExtension : IMarkupExtension<BindingBase>
```
jfversluis, line 7: We probably want to rename this because now the extension is `AppTheme` which collides with the `AppTheme` in .NET MAUI and Visual Studio suggests the wrong namespace to be imported. Unless you use the full name `AppThemeExtension` but not sure if that's desirable.

src/CommunityToolkit.Maui/AppThemeColor.shared.cs
```diff
@@ -0,0 +1,76 @@
+﻿namespace CommunityToolkit.Maui;
```
jfversluis, line 1: What's going to be the right namespace for this?

src/CommunityToolkit.Maui/AppThemeResource.shared.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace CommunityToolkit.Maui;
```
jfversluis, line 1: What's going to be the right namespace for this?

src/CommunityToolkit.Maui/AppThemeColor.shared.cs
```diff
@@ -0,0 +1,76 @@
...
+
+/// <summary>
+/// Represents a color that is aware of the operating system theme.
+/// </summary>
+public class AppThemeColor
```
jfversluis, line 6: With some effort we could make this an inheritance from `AppThemeResource`? Not sure why that hasn't been done.

brminnick (changes requested):
Hey Gerald! I've only done a partial review, but at first glance it looks like we can simplify `AppThemeColor.cs` a bit. What do you think?

samples/CommunityToolkit.Maui.Sample/Pages/Essentials/AppThemePage.xaml
```diff
@@ -0,0 +1,49 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<pages:BasePage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:pages="clr-namespace:CommunityToolkit.Maui.Sample.Pages"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             x:Class="CommunityToolkit.Maui.Sample.Pages.Essentials.AppThemePage"
+             xmlns:maui="clr-namespace:CommunityToolkit.Maui;assembly=CommunityToolkit.Maui"
```
VladislavAntonyuk, line 6: Let's use toolkit to not confuse people it's a MAUI namepaspace

src/CommunityToolkit.Maui.UnitTests/AppThemeTests.cs
```diff
@@ -0,0 +1,136 @@
+﻿using CommunityToolkit.Maui.Extensions;
+using Xunit;
+
+namespace CommunityToolkit.Maui.UnitTests;
+
+public class AppThemeTests : BaseTest
+{
+	class MockAppInfo : IAppInfo
```
VladislavAntonyuk, line 8: Let's move it to all other mocks

jfversluis:
@brminnick yep thanks! That was my outstanding comment on refactoring it to use the same code. Applied that now. There is one base `AppThemeObject<T>` and 2 convenience objects basically that inherit from it, but don't add anything except for making it easier to work with in both code and XAML: `AppThemeResource` which just uses an `object` and `AppThemeColor` which does the same for colors. What do you think?

I've left `AppThemeObject<T>` public on purpose, maybe people want to extend it for their own things?

src/CommunityToolkit.Maui/Extensions/AppThemeExtension.shared.cs
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		return false;
+	}
+
+	static bool TryGetApplicationLevelResource(string key, out object? resource, out ResourceDictionary? resourceDictionary)
+	{
+		resource = null;
+		resourceDictionary = null;
+		// TODO: Remove reference to Application.Current
```
pictos, line 85: Can we remove this ToDo?

src/CommunityToolkit.Maui/AppThemeColor.shared.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace CommunityToolkit.Maui;
+
+/// <summary>
+/// Represents a color that is aware of the operating system theme.
+/// </summary>
+public class AppThemeColor : AppThemeObject<Color>
```
pictos, line 6: Can we make this `sealed`, can't see any value allowing users to inherit from it

```suggestion
public sealed class AppThemeColor : AppThemeObject<Color>
```

src/CommunityToolkit.Maui/AppThemeResource.shared.cs
```diff
@@ -0,0 +1,8 @@
...
+
+/// <summary>
+/// Represents a resource that is aware of the operating system theme.
+/// </summary>
+public class AppThemeResource : AppThemeObject<object>
```
pictos, line 6: Can we make this `sealed` for the same reason as the `AppThemeColor`?

```suggestion
public sealed class AppThemeResource : AppThemeObject<object>
```

pictos (commented):
This looks good in general, I've feel comments around the `AppThemeObject` class

src/CommunityToolkit.Maui/AppThemeObjectOfT.shared.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace CommunityToolkit.Maui;
+
+/// <summary>
+/// Represents an object that is aware of the operating system theme.
+/// </summary>
+public class AppThemeObject<T>
```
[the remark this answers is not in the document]
jfversluis: Yeah I thought about this later as well, I think this is the right move! Thanks!

src/CommunityToolkit.Maui/AppThemeObjectOfT.shared.cs
```diff
@@ -0,0 +1,45 @@
...
+	/// The <see cref="object"/> that is used when the current theme is unspecified or
+	/// when a value is not provided for <see cref="Light"/> or <see cref="Dark"/>.
+	/// </summary>
+	public T? Default { get; set; }
+
+	internal BindingBase GetBinding()
+	{
+		var binding = new AppThemeBinding();
+
+		if (Light is not null)
```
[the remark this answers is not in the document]
jfversluis: You would think so! However this is needed. If we don't do this it won't work correctly. For instance: when you only specify a color for Default and Dark theme and you set your device to Light theme, instead of Default, it will use Light which is then null.

jfversluis:
OK I think I have all the feedback processed. Then the only question remains if we want to rename `AppThemeExtension` to anything.

If you start typing `<Label TextColor="{AppTheme ...` it will suggest the [`AppTheme`](https://github.com/dotnet/maui/blob/main/src/Essentials/src/AppInfo/AppTheme.shared.cs#L6) type in .NET MAUI. Is that a problem? I think ideally we'd pick a different name to avoid collisions. On the other hand, people will most likely add the xmlns first and then type `toolkit:AppTheme...` and it will suggest the right one?

samples/CommunityToolkit.Maui.Sample/Pages/Essentials/AppThemePage.xaml
```diff
@@ -0,0 +1,48 @@
...
+
+        <Style x:Key="Headline" TargetType="Label">
+            <Setter Property="FontFamily" Value="Segoe UI" />
+            <Setter Property="FontSize" Value="10" />
+            <Setter Property="TextColor" Value="{mct:AppTheme LabelTextColor}" />
+        </Style>
+    </pages:BasePage.Resources>
+
+    <VerticalStackLayout Spacing="10">
+        <Label
```
VladislavAntonyuk, line 23: Can we add a theme toggle button?

src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj
```diff
@@ -69,4 +76,8 @@
     </PackageReference>
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Essentials\" />
```
VladislavAntonyuk, line 80: All changes in this file can be reverted

src/CommunityToolkit.Maui/Essentials/AppTheme/AppThemeObjectOfT.shared.cs
```diff
@@ -0,0 +1,49 @@
...
+
+	/// <summary>
+	/// Gets a bindable object which holds the diffent values for each operating system theme. 
+	/// </summary>
+	/// <returns>A <see cref="AppThemeBinding"/> instance with the respective theme values.</returns>
+	public virtual BindingBase GetBinding()
+	{
+		var binding = new AppThemeBinding();
+
+		if (Light is not null)
```
VladislavAntonyuk, line 32: Do we need a check for null? Is null a valid value? What if null is a value value for T and I want to set it after the previous value?

src/CommunityToolkit.Maui/Extensions/AppThemeExtension.shared.cs
```diff
@@ -0,0 +1,91 @@
...
+	/// </summary>
+	public string? Key { get; set; }
+
+	/// <inheritdoc/>
+	/// <exception cref="ArgumentNullException">Thrown if <paramref name="serviceProvider"/> is <see langword="null"/>.</exception>
+	/// <exception cref="XamlParseException">Thrown if <see cref="Key"/> is <see langword="null"/> or if a resource with <see cref="Key"/> cannot be found.</exception>
+	/// <exception cref="ArgumentException">Thrown if <paramref name="serviceProvider"/> does no implement <see cref="IProvideParentValues"/>.</exception>
+	public BindingBase ProvideValue(IServiceProvider serviceProvider)
+	{
+		if (serviceProvider is null)
```
VladislavAntonyuk, line 20: Use ArgumentException.ThrowIfNull

src/CommunityToolkit.Maui/Extensions/AppThemeExtension.shared.cs
```diff
@@ -0,0 +1,91 @@
...
+		{
+			throw new ArgumentNullException(nameof(serviceProvider));
+		}
+
+		if (Key is null)
+		{
+			throw new XamlParseException("You must specify a key for {AppThemeExtension} that specifies the AppTheme resource to use", serviceProvider);
```
VladislavAntonyuk, line 27: Does sourceProvider a valid parameter?

src/CommunityToolkit.Maui/Extensions/AppThemeExtension.shared.cs
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		if (serviceProvider.GetService(typeof(IProvideValueTarget)) is not IProvideParentValues valueProvider)
```
VladislavAntonyuk, line 30: Can we use generic GetService or GetRequiredService?

src/CommunityToolkit.Maui/Essentials/AppTheme/AppThemeObjectOfT.shared.cs
pictos: Already answered here
https://github.com/CommunityToolkit/Maui/pull/1264#discussion_r1241127495

src/CommunityToolkit.Maui/Extensions/AppThemeExtension.shared.cs
brminnick: Yup, it's a valid parameter for `XamlParseException`

brminnick:
> Then the only question remains if we want to rename AppThemeExtension to anything.

We could change it to `BindingBaseExtension` to align with our existing `IMarkupExtension` classes; their class name references the IMarkupExtension Type (eg `ValueConverterExtension ` implements `IMarkupExtension<ICommunityToolkitValueConverter> `)
- `public class MultiValueConverterExtension : IMarkupExtension<ICommunityToolkitMultiValueConverter>`
- `public abstract class ValueConverterExtension : BindableObject, IMarkupExtension<ICommunityToolkitValueConverter>`


```cs
public sealed class AppThemeExtension : IMarkupExtension<BindingBase>
```
changes to 
```cs
public sealed class BindingBaseExtension : IMarkupExtension<BindingBase>
```

src/CommunityToolkit.Maui/Essentials/AppTheme/AppThemeColor.shared.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace CommunityToolkit.Maui;
```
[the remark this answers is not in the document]
pictos: But this is a `xaml` feature, so because of that it shouldn't go to `Core`

src/CommunityToolkit.Maui/Essentials/AppTheme/AppThemeObjectOfT.shared.cs
VladislavAntonyuk: This may be an issue if null is a valid value. We cannot set it after some "not null" value

src/CommunityToolkit.Maui/Essentials/AppTheme/AppThemeColor.shared.cs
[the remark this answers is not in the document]
jfversluis: I don't think .NET MAUI Core has all the references we need to make this happen. I put it in core first and it couldn't find all the references to types it needed.

src/CommunityToolkit.Maui/Essentials/AppTheme/AppThemeObjectOfT.shared.cs
```diff
@@ -0,0 +1,49 @@
+﻿namespace CommunityToolkit.Maui;
```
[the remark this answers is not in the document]
pictos: But this code has a reference to the Binding system, which is MVVM oriented. The core assembly should be agnostic

IndianaGary:
Re Brandon's chosen rename for `AppThemeExtension`: It occurs to me that `BindingBaseExtension `isn't really appropriate as it doesn't identify it's purpose and is too generic. I suggest changing it to `ThemedColorExtension`, as it concisely specifies its purpose and doesn't clash with other names.

jfversluis:
Thanks for the input @IndianaGary however, this isn't just about colors. It's also resources in general.

So `{ThemedColor MyResource}` might make people think that it's only for colors, which it's not.

In our core maintainers chat I opted for just Theme? `TextColor={Theme MyColorThing}` and `ImageSource={Theme MyImage}`.

However, that maybe makes the names `AppThemeColor` and `AppThemeResource` a bit weird maybe. Should they be `ThemeColor` and `ThemeResource` then? But that will make them different from `AppThemeBinding` (in .NET MAUI) that they're based on.

Or maybe at this point I'm overthinking it 😄

IndianaGary:
That's possible... ii was being guided by the specificity of AppThemeColor and AppThemeResource being distinct markup extensions and the issue of name collisions. ThemedColor IS targeted specifically at color. Similarly, Themed Resource would reference a resource. Trying to come up with a combined extension violates the Single Responsibility principle hence my preference for a targeted name.

jfversluis:
>  Similarly, Themed Resource would reference a resource.

A color is a resource though, even in this code, it literally just inherits from the resource. But this is more about it being clear to the consumer and not the technicalities I guess 😄

src/CommunityToolkit.Maui/Essentials/AppTheme/AppThemeObjectOfT.shared.cs
```diff
@@ -0,0 +1,49 @@
...
+
+/// <summary>
+/// Represents an object that is aware of the operating system theme.
+/// </summary>
+public abstract class AppThemeObject<T>
```
brminnick, line 6: Shall we rename this to `ThemeObject<T>`?

After renaming `ThemeResourceExtension`, it's a bit confusing to be using `<mct:AppThemeColor ...`, `<mct:AppThemeResource ...`, and `{mct:ThemeResource ...` in XAML. 

I assume devs would prefer consistency of each being named `Theme*`

src/CommunityToolkit.Maui/Essentials/AppTheme/AppThemeColor.shared.cs
```diff
@@ -0,0 +1,8 @@
...
+
+/// <summary>
+/// Represents a color that is aware of the operating system theme.
+/// </summary>
+public sealed class AppThemeColor : AppThemeObject<Color>
```
brminnick, line 6: Shall we rename this to `ThemeColor`?

After renaming `ThemeResourceExtension`, it's a bit confusing to be using `<mct:AppThemeColor ...`, `<mct:AppThemeResource ...`, and `{mct:ThemeResource ...` in XAML. 

I assume devs would prefer consistency of each being named `Theme*`

src/CommunityToolkit.Maui/Essentials/AppTheme/AppThemeResource.shared.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace CommunityToolkit.Maui;
+
+/// <summary>
+/// Represents a resource that is aware of the operating system theme.
+/// </summary>
+public sealed class AppThemeResource : AppThemeObject<object>
```
brminnick, line 6: Shall we rename this to `ThemeResource`?

After renaming `ThemeResourceExtension`, it's a bit confusing to be using `<mct:AppThemeColor ...`, `<mct:AppThemeResource ...`, and `{mct:ThemeResource ...` in XAML. 

I assume devs would prefer consistency of each being named `Theme*`

brminnick (commented):
Hey Gang! I think we should rename everything to `Theme...` now that we renamed `ThemeResouceExtension`.

Renaming will give us this consistency in XAML:
- Updated: `<mct:ThemeResource ... />
  - Current  `<mct:AppThemeResource ... />
- Updated: `<mct:ThemeColor ... />
  - Current: `<mct:AppThemeColor ... />
- Current: `{mct:ThemeResource ...}`

jfversluis:
@brminnick So yeah, funny story, now we're back to square 1.

The extension is called: ThemeResource and now we have an object ThemeResource... 😅

brminnick:
Oh no!! 🙈🙃😭

Naming is so hard...

image

IndianaGary:
😊

brminnick (approved):
Thanks Gerald!! I think we finally got the naming:

```cs
public sealed class AppThemeColor : AppThemeObject<Color>;

public sealed class AppThemeObject : AppThemeObject<object>;

public abstract class AppThemeObject<T>;

public sealed class AppThemeResourceExtension : IMarkupExtension<BindingBase>
```

I approve, and I'll defer to you to merge it to give you time to review the updates.

:point_down: Here's how these names look in the Sample app

Screenshot 2023-07-23 at 1 16 38 PM

PR #1269, opened 2023-06-27
Implement MapClicked

### LINKED ISSUE #1268: [BUG] CommunityToolkit/Maui/Maps MapClicked event doesn't trigger on Windows

### Current Behavior

Community toolkit Maps for Windows doesn't trigger the event in code-behind when **MapClicked** is added.

### Expected Behavior

Community toolkit Maps for Windows should trigger the code-behind method set for **MapClicked** in xaml.

### Steps To Reproduce

1. Use https://github.com/jfversluis/MauiToolkitWindowsMapsSample repo.
2. Add **MapClicked** event to the map and assign to a code-behind method.
3. Add a breakpoint in the code-behind method.
4. Run and click the map
Observe: code-behind is not being triggered at all.

https://github.com/jfversluis/MauiToolkitWindowsMapsSample

### Environment

```markdown
- .NET MAUI CommunityToolkit.Maui.Maps 1.0.2
- OS:Windows 10.0.19041.0
- .NET MAUI: 7.0.81/7.0.100
- .NET Runtime: net7.0
```

### Anything else?

_No response_

jfversluis:
### Description of Change ###

Implements `MapClicked` for the Windows map. Also restructures how the events are handled to be more robust and future proof. Can probably be optimized further, but this should be a good starting point.

 ### Linked Issues ###

 - Fixes #1268

 ### PR Checklist ###

 - [x] Changes adhere to [coding standard]

### DIFF
--- a/src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
+++ b/src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
@@ -447,7 +447,7 @@ public partial class MapHandlerWindows : MapHandler
 	{
 		public string Id { get; set; } = string.Empty;
 
-		public string? Payload { get; set; }
+		public object? Payload { get; set; }
 	}
 
 	enum EventIdentifier
--- a/src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
+++ b/src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
@@ -9,6 +9,7 @@ using Microsoft.UI.Xaml.Controls;
 using Microsoft.Web.WebView2.Core;
 using Windows.Devices.Geolocation;
 using IMap = Microsoft.Maui.Maps.IMap;
+using Map = Microsoft.Maui.Controls.Maps.Map;
 
 namespace CommunityToolkit.Maui.Maps.Handlers;
 
@@ -18,9 +19,9 @@ public partial class MapHandlerWindows : MapHandler
 	internal static string? MapsKey;
 	MapSpan? regionToGo;
 
-	JsonSerializerOptions jsonSerializerOptions = new()
+	readonly JsonSerializerOptions jsonSerializerOptions = new()
 	{
-		PropertyNameCaseInsensitive = true
+		PropertyNameCaseInsensitive = true,
 	};
 
 	/// <summary>
@@ -191,7 +192,20 @@ public partial class MapHandlerWindows : MapHandler
 									showTrafficButton: false
 								});
 								loadTrafficModule();
-								Microsoft.Maps.Events.addHandler(map, 'viewrendered', function () { var bounds = map.getBounds(); invokeHandlerAction(bounds); });
+								Microsoft.Maps.Events.addHandler(map, 'viewrendered', function () { var bounds = map.getBounds(); invokeHandlerAction('BoundsChanged', bounds); });
+
+								Microsoft.Maps.Events.addHandler(map, 'click', function (e) {
+									if (!e.isPrimary) {
+										return;
+									}
+
+									var clickedLocation = {
+										latitude: e.location.latitude,
+										longitude: e.location.longitude
+									};
+
+									invokeHandlerAction('MapClicked', clickedLocation);
+								});
 
 								infobox = new Microsoft.Maps.Infobox(map.getCenter(), {
 									visible: false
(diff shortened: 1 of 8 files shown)


src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
```diff
@@ -366,37 +385,80 @@ void WebViewNavigationCompleted(WebView2 sender, CoreWebView2NavigationCompleted
...
+				var clickedLocation = JsonSerializer.Deserialize<Location>(eventMessage.Payload?.ToString() ?? "",
+					jsonSerializerOptions);
+				if (clickedLocation is not null)
+				{
+					VirtualView.Clicked(clickedLocation);
+				}
+				break;
+
+			case EventIdentifier.InfoWindowClicked:
+				var clickedInfoWindowWebView = JsonSerializer.Deserialize<InfoWindow>(eventMessage.Payload?.ToString() ?? "",
```
VladislavAntonyuk, line 420: Does Deserialize return null if string is empty?

src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
```diff
@@ -18,9 +19,9 @@ public partial class MapHandlerWindows : MapHandler
 	internal static string? MapsKey;
 	MapSpan? regionToGo;
 
-	JsonSerializerOptions jsonSerializerOptions = new()
+	readonly JsonSerializerOptions jsonSerializerOptions = new()
 	{
-		PropertyNameCaseInsensitive = true
+		PropertyNameCaseInsensitive = true,
```
VladislavAntonyuk, line 24:
```suggestion
		PropertyNameCaseInsensitive = true
```

src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
```diff
@@ -366,37 +385,80 @@ void WebViewNavigationCompleted(WebView2 sender, CoreWebView2NavigationCompleted
...
+				}
+				break;
 		}
 	}
 
+	class EventMessage
+	{
+		public string Id { get; set; } = string.Empty;
+
+		public object? Payload { get; set; }
```
VladislavAntonyuk, line 450: In which case payload is null? Can we mark it as required?

As we always call ToString, can we change the type to string?

src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
jfversluis: Hm thats a good test! I would expect it to indeed return null, that’s kind of what happened in the previous code as well, right?

src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
jfversluis: We could make it required. It’s not really meant to be instantiated by hand though, it should always come from the Bing Map.

In the first commit it was a string type, I guessed that would work. For some reason that throws an exception 😅

VladislavAntonyuk: Maybe we can make it generic because you pass a parameter that is not actually a string (like InfoWindow).

jfversluis: Everything is serialized as JSON, so it should be a string I guess? But not 100% sure what you mean. Feel free to do any suggestions or make changes if you have an idea around this.

src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
jfversluis: Good call! If you really give it an empty string it will give an exception although it won't surface, but I'll add some code to check for that first.

src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
jfversluis: Do you mean dynamic instead of generic?

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
VladislavAntonyuk: I thought we can put generic type and it can be deserialized automatically. but I see we don't know the target type at this moment.

PR #1309, opened 2023-07-28
Add routing of BindingContext of Popup

### LINKED ISSUE #1166: [BUG] Picker not work correctly on popup

### Current Behavior

Hi. I have a problem with a Picker, but only on the Popup. I have already open a bug in 

https://github.com/dotnet/maui/issues/14832

But I don't know if is a Maui problem or a CommunityToolkit problem.
Can you take a look?
Thanks

### Expected Behavior

The popup should set the SelectedItem correctly

### Steps To Reproduce

Take a look to https://github.com/dotnet/maui/issues/14832, there is a repo

https://github.com/acaliaro/MauiPicker

### Environment

```markdown
- .NET MAUI CommunityToolkit: latest
- OS: iOS 15.7
- .NET MAUI:7
```

### Anything else?

_No response_

cat0363:
This PR solves the problem that the BindingContext of Popup is not routed to ContentPage.

 ### Description of Change ###

I added the following code to pass the Popup's BindingContext to the ContentPage.

[src\CommunityToolkit.Maui\HandlerImplementation\Popup\Popup.macios.cs]
    
    public static void MapOnOpened(PopupHandler handler, IPopup view, object? result)
    {
        handler.PlatformView?.CreateControl(CreatePageHandler, view);
        view.OnOpened();

        static PageHandler CreatePageHandler(IPopup virtualView)
        {
            var mauiContext = virtualView.Handler?.MauiContext ?? throw new NullReferenceException(nameof(IMauiContext));
            var view = (View?)virtualView.Content ?? throw new InvalidOperationException($"{nameof(IPopup.Content)} can't be null here.");
            view.BindingContext = ((Popup)virtualView).BindingContext;    // <= here
            var contentPage = new ContentPage
            {
                Content = view
            };

            contentPage.SetBinding(BindingContextProperty, new Binding { Source = virtualView, Path = BindingContextProperty.PropertyName });

            return (PageHandler)contentPage.ToHandler(mauiContext);
        }
    }

This will take over the BindingContext and not initialize it with null.

 

 ### Linked Issues ###
 

 - Fixes #1166

 ### PR Checklist ###
 
 I simplified the program of Issue #1166 and verified it.
Below are the verification results.

https://github.com/CommunityToolkit/Maui/assets/125236133/a366eff7-0586-41d7-be80-de39aa7314c0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -263,3 +263,4 @@ paket-files/
 
 # Visual Studio Code
 .vscode
+/samples/workload-install.ps1
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,6 +119,9 @@ Here we will have some:
 
 ## Contributing Code - Best Practices
 
+### Debug Logging
+* Always use `Trace.WriteLine()` instead of `Debug.WriteLine` for debug logging because `Debug.WriteLine` is removed by the compiler in Release builds
+
 ### Enums
 * Always use `Unknown` at index 0 for return types that may have a value that is not known
 * Always use `Default` at index 0 for option types that can use the system default option
--- a/samples/CommunityToolkit.Maui.Sample/AppShell.xaml.cs
+++ b/samples/CommunityToolkit.Maui.Sample/AppShell.xaml.cs
@@ -83,6 +83,8 @@ public partial class AppShell : Shell
 		CreateViewModelMapping<VariableMultiValueConverterPage, VariableMultiValueConverterViewModel, ConvertersGalleryPage, ConvertersGalleryViewModel>(),
 
 		// Add Essentials View Models
+		CreateViewModelMapping<AppThemePage, AppThemeViewModel, EssentialsGalleryPage, EssentialsGalleryViewModel>(),
+		CreateViewModelMapping<BadgePage, BadgeViewModel, EssentialsGalleryPage, EssentialsGalleryViewModel>(),
 		CreateViewModelMapping<FileSaverPage, FileSaverViewModel, EssentialsGalleryPage, EssentialsGalleryViewModel>(),
 		CreateViewModelMapping<FolderPickerPage, FolderPickerViewModel, EssentialsGalleryPage, EssentialsGalleryViewModel>(),
 		CreateViewModelMapping<SpeechToTextPage, SpeechToTextViewModel, EssentialsGalleryPage, EssentialsGalleryViewModel>(),
--- a/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
+++ b/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
@@ -22,12 +22,15 @@
     <ApplicationVersion>1</ApplicationVersion>
 
     <!--
-    Uncomment these lines if you need to debug the SG code
+    Uncomment the line below if you need to debug the SG code
     If you see any LongPath issue on Windows, check this doc
     https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd#enable-long-paths-in-windows-10-version-1607-and-later
     -->
     <!--<EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
     <CompilerGeneratedFilesOutputPath>$(BaseIntermediateOutputPath)\GeneratedFiles</CompilerGeneratedFilesOutputPath>-->
+
+    <!-- Uncomment the line below if Android App installation fails -->
+    <!--<EmbedAssembliesIntoApk>true</EmbedAssembliesIntoApk>-->
   </PropertyGroup>
 
   <ItemGroup>
@@ -79,4 +82,5 @@
   <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('windows'))=='false' and $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)'))=='maccatalyst' and $(Configuration) == 'Debug'">
     <RuntimeIdentifiers>maccatalyst-arm64;maccatalyst-x64</RuntimeIdentifiers>
   </PropertyGroup>
+
 </Project>
--- a/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
+++ b/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
@@ -50,7 +50,7 @@
 
     <PackageReference Include="CommunityToolkit.Maui.Markup" Version="3.2.0" />
     <PackageReference Include="CommunityToolkit.Mvvm" Version="8.2.1" />
-    <PackageReference Include="Microsoft.Extensions.Http.Polly" Version="7.0.8" />
+    <PackageReference Include="Microsoft.Extensions.Http.Polly" Version="7.0.9" />
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="7.0.0" />
   </ItemGroup>
 
--- a/samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
+++ b/samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
@@ -1,3 +1,4 @@
+using CommunityToolkit.Maui.ApplicationModel;
 using CommunityToolkit.Maui.Maps;
 using CommunityToolkit.Maui.Markup;
 using CommunityToolkit.Maui.Media;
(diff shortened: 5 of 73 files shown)


VladislavAntonyuk:
Does the issue happen only on Apple devices?

cat0363:
> Does the issue happen only on Apple devices?

Hi, @VladislavAntonyuk 
At least it was working fine on Android.
I can't confirm for Windows. Could someone please confirm?
As for Windows, if it's not urgent, I'll check next week.

pictos:
the `view` is a child of the `ContentPage` and when we set the Binding for that `ContentPage` it should propagate to the `view`, if I didn't miss anything. What's the issue you're seeing?

cat0363:
> view.BindingContext = ((Popup)virtualView).BindingContext;

Thank you for your review.
You're right, that's how it should be. I thought so too until I debugged it.
When the following is executed, the ContentPage's BindingContext is propagated and initialized.

    ContentView = view

The cause is that the content of the Popup is replaced with the content of the newly generated ContentPage.
Binding is set in ContentPage with SetBinding method, but its Source is virtualView, not view.

    contentPage.SetBinding(BindingContextProperty, new Binding { Source = virtualView, Path = BindingContextProperty.PropertyName });

For this, I added the code below.

    view.BindingContext = ((Popup)virtualView).BindingContext;

But that is wrong. Add the following instead of the above

    view.SetBinding(BindingContextProperty, new Binding { Source = (Popup)virtualView, Path = BindingContextProperty.PropertyName });

And I should have changed SetBinding of ContentPage as below.

    contentPage.SetBinding(BindingContextProperty, new Binding { Source = view, Path = BindingContextProperty.PropertyName });

From the above, I think it will be as follows.

    static PageHandler CreatePageHandler(IPopup virtualView)
    {
        var mauiContext = virtualView.Handler?.MauiContext ?? throw new NullReferenceException(nameof(IMauiContext));
        var view = (View?)virtualView.Content ?? throw new InvalidOperationException($"{nameof(IPopup.Content)} can't be null here.");
        view.SetBinding(BindingContextProperty, new Binding { Source = (Popup)virtualView, Path = BindingContextProperty.PropertyName });

        var contentPage = new ContentPage
        {
            Content = view
        };

        contentPage.SetBinding(BindingContextProperty, new Binding { Source = view, Path = BindingContextProperty.PropertyName });

        return (PageHandler)contentPage.ToHandler(mauiContext);
    }

By correcting it as above, it propagates correctly.
I will upload it again next week.

cat0363:
Hi, @VladislavAntonyuk 
Fixed the BindingContext propagation as per the previous post.
Below is the execution result on iOS.

[iOS]

https://github.com/CommunityToolkit/Maui/assets/125236133/b78a5e07-0570-4a80-9d2b-25bc2aa2d112

This issue does not occur on Windows and Android. This issue occurs only on iOS.
Below are the execution results on Android and Windows.

[Android]

https://github.com/CommunityToolkit/Maui/assets/125236133/d700d11a-d2e0-4458-b0ee-625970f41c6e

[Windows]

https://github.com/CommunityToolkit/Maui/assets/125236133/ecdbe193-374c-4b28-ad74-c64925df39ed

cat0363:
Below is the reproduction code I created to confirm that there is a problem with the method of BindingContext propagation.

https://github.com/cat0363/MauiComm-ReproPopupPicker.git

The above partially mimics a portion of the Popup.macios.cs file.

Instead of displaying the Popup on the screen, I set the Content of the ContentView imitating the Popup to the 
Content of cvContent in the layout below.

[MainPage.xaml]

    <Grid RowDefinitions="44,44,*">
        <Button x:Name="btnShow" Grid.Row="0" Text="Show" BackgroundColor="Blue" TextColor="White" Clicked="btnShow_Clicked" />
        <ContentView x:Name="cvContent" Grid.Row="1" />
    </Grid>

[Popup.xaml]

    <Grid>
        <Picker
            Title="Select an Outlet"
            ItemDisplayBinding="{Binding Name}"
            ItemsSource="{Binding Outlets}"
            SelectedItem="{Binding SelectedItemOutlet}" />
    </Grid>

The processing when the Show button is pressed is as follows.

    private void btnShow_Clicked(object sender, EventArgs e)
    {
        var popup = new Popup();
        var view = popup.Content;
        // view.SetBinding(BindingContextProperty, new Binding { Source = popup, Path = BindingContextProperty.PropertyName });
        cvContent.Content = view;
        // cvContent.SetBinding(BindingContextProperty, new Binding { Source = view, Path = BindingContextProperty.PropertyName });
        cvContent.SetBinding(BindingContextProperty, new Binding { Source = popup, Path = BindingContextProperty.PropertyName });
    }

If you run it as above, the same phenomenon will be reproduced.

Therefore, modify the code as follows. This is equivalent to the solution I posted today.

    private void btnShow_Clicked(object sender, EventArgs e)
    {
        var popup = new Popup();
        var view = popup.Content;
        view.SetBinding(BindingContextProperty, new Binding { Source = popup, Path = BindingContextProperty.PropertyName });
        cvContent.Content = view;
        cvContent.SetBinding(BindingContextProperty, new Binding { Source = view, Path = BindingContextProperty.PropertyName });
        // cvContent.SetBinding(BindingContextProperty, new Binding { Source = popup, Path = BindingContextProperty.PropertyName });
    }

You can see that the BindingContext is propagated correctly.

VladislavAntonyuk:
It looks weird why we need it only for macios.  Could it be a .NET MAUI issue as they don't set the binding context?

cat0363:
@VladislavAntonyuk, Thank you for your reply.
This issue occurs even if I run the code that causes the same phenomenon that I created on Android and Windows.
So it's not just iOS. 

https://github.com/cat0363/MauiComm-ReproPopupPicker.git

Below are the results of running the above code on Android and Windows.
iOS, Android, and Windows all have the same results.

[Android]

Screenshot_1690848084

[Windows]

windows

If it is made similarly, this problem will occur on Android and Windows.
As for why this issue is only for macios, it is because Android and Windows are not made like this.

On Android, the LayoutViewGroup obtained by calling the ToPlatform method of the Popup's Content is set to the Dialog's ContentView.

[src\CommunityToolkit.Maui.Core\Views\Popup\MauiPopup.android.cs] : TryCreateContainer

    container = popup.Content.ToPlatform(mauiContext);
    SetContentView(container);

On Windows, the LayoutPanel obtained by calling the View's ToPlatform method is added to the Panel's child elements.
That Panel is set as the Flyout's Content.

[src\CommunityToolkit.Maui\Views\WrapperControl.windows.cs] : Constructor

    FrameworkElement = view.ToPlatform(mauiContext);
    Children.Add(FrameworkElement);

[src\CommunityToolkit.Maui.Core\Views\Popup\MauiPopup.windows.cs] : CreateControl

    Control = createControl(handler);
    Content = Control;

At least, Android and Windows are built differently than iOS.
If you don't mind, can you give me a specific reason why it's a .NET MAUI issue?

Does it mean that the behavior that the SelectedItem of the Picker is initialized when the parent is replaced is not what you intended? I'm sorry if my understanding is wrong.

pictos:
@cat0363 thanks for all explanation and the fix... I think this is a fine change to do. But looks like a bug on Maui side of the BindingContext, so let's say that you have a `ContentPage` with a `Label` on it, something like this:

```xml
<ContentPage>
    <Label Text="{Binding Title}" />
</ContentPage>
```

Then on your code-behind you have this:

```csharp
class MainPage : ContentPage
{
    public MainPage()
    {
        InitializeComponent();
        BindingContext = new MainViewModel();
    }
}
```

So, we have the `Label.TextProperty` bind to the `Title` property on the ViewModel, but at the moment the Label is initialized (on the InitializeComponent method) the BindingContext for the page and Label is `null`, the Binding changed to `MainViewModel`, for both Label and Page, after the initialization and the `Label` will show whatever the `Title` is.

With that, the behavior of the popup and picker should be the same, don't you agree? Because of that, I believe there's a bug or the `Picker` control works in a different way.

cat0363:
@pictos, Thank you for the easy-to-understand explanation.

I agree with your explanation. Certainly it's not that the BindingContext isn't 
propagated. As proof, when you tap the Picker, you can see that the ItemsSource 
is correctly bound.

When the parent is replaced, the ItemsSource is initialized to null, at which time 
the Picker's ItemsSourceChanged method is called and the SelectedItem is initialized 
to null, but the Picker's ItemsSource is set again.

I've found that the initialization that occurs when the parent is swapped causes 
this problem, but I'm not sure if that initialization was intended. Presumably, 
if this BindingContext initialization hadn't happened, this problem wouldn't have occurred.

In the Discussions below, I am asking a similar question to the .NET MAUI side.
So far I haven't received an answer from the .NET MAUI side.

https://github.com/dotnet/maui/discussions/16516

If it's a bug, it seems that BindingContext initialization and Picker behavior are deeply related. . .
By the way, if it's a .NET MAUI bug, is there any way for users to avoid this problem?

One workaround I can think of is something like this:
The following is an implementation example on the side that uses Popup.

    void Button_Clicked(object sender, EventArgs e)
    {
        var popup = new PopupPage();
        popup.Opened += new EventHandler<CommunityToolkit.Maui.Core.PopupOpenedEventArgs>(popup_Opend);
        Shell.Current.ShowPopup(popup);		
    }

    void popup_Opend(object? sender, CommunityToolkit.Maui.Core.PopupOpenedEventArgs e)
    {
        if (sender is PopupPage popup)
        {
            popup.Opened -= new EventHandler<CommunityToolkit.Maui.Core.PopupOpenedEventArgs>(popup_Opend);
            popup.BindingContext = new ViewModelTest();
        }
    }

Instead of setting the BindingContext in the constructor of the class that inherits 
Popup, you can avoid this by setting the BindingContext in the Popup's Opend event.

I will leave the closing of this PR to you. If there is a workaround that can be considered at this time, 
it may be better to present it to the person who posted the issue.

pictos (approved):

PR #1315, opened 2023-07-31
Allow change position after MediaEnded event

### LINKED ISSUE #1043: [BUG] MediaElement StateChanged Event Position doesn't update after media has reached end

### Current Behavior

When you have a handler for the StateChanged event and have enabled ShouldShowPlaybackControls; when the video has played till the end then you use onscreen controls to rewind the video, the MediaPositionChangedEventArgs's Position does not update to the current actual position, it keeps sending the video ends Timespan (=Duration).
Examples include:
-if the video is 2min long (Duration).
-you let it play until the end
-when you use the playback controls to rewind the video
-the e.Position of MediaPositionChangedEventArgs doesn't send the current (rewind) position, put keeps sending the 2mim, which is the MediaElement.Duration.

This is a problem for both Android and iOS****

### Expected Behavior

The MediaPositionChangedEventArgs's Position should update to the current rewind position, MediaElement.Position, and not the MediaElement.Duration

### Steps To Reproduce

1. And a MediaElement to any Page and enable ShouldShowPlaybackControls (=True)
2. Add the Video Url
3. Add an event handler for StateChanged
4. Monitor the MediaPositionChangedEventArgs's (e) Position until the video ends (when ~ (Position = Duration))
5. The e.Position should updates fine until the video reaches the end
6. Using the playback controls, rewind the video to any position < Duration
7. The e.Position doesn't contain the latest rewind TimeSpan, but the video's end one (which is the Duration)

https://github.com/ChiefMonk/MauiMediaElementTesting

### Environment

```markdown
- .NET MAUI CommunityToolkit.MediaElement: 1.0.2
- OS: iOS and Android
- .NET MAUI: 7.0

Packages used:

 <ItemGroup>
   <PackageReference Include="CommunityToolkit.Maui.MediaElement " Version="1.0.2" />
   <PackageReference Include="Microsoft.Extensions.Loggin.Debug " Version="7.0.0" />	  
</ItemGroup>
```

### Anything else?

_No response_

niksedk:
### Description of Change ###
Allow change position after MediaEnded event

The report-status-timer in this PR is:
- started in `OnMediaOpened` (when a new media opened successfully)
- stopped in `OnSourcePropertyChanged` (when opening a new media or setting media to null)

This will keep the status/positioning/progress-bar alive until the media is set to null or a new media is loaded, but dragging the progress-bar-position to the end, will not stop the status/positioning/progress-bar from being alive.

 ### Linked Issues ###
 - Fixes #1043


niksedk:
@dotnet-policy-service agree

niksedk:
@dotnet-policy-service agree

pictos:
I'm pretty sure we have a good reason to keep the timer. So removing it doesn't look like the right solution

niksedk:
> I'm pretty sure we have a good reason to keep the timer. So removing it doesn't look like the right solution

This PR is not removing the timer... it's keeping it active :)

pictos:
> > I'm pretty sure we have a good reason to keep the timer. So removing it doesn't look like the right solution
> 
> This PR is not removing the timer... it's keeping it active :)
> 
> The timer is started in `OnMediaOpened`

And we never stop the timer?

niksedk:
> > > I'm pretty sure we have a good reason to keep the timer. So removing it doesn't look like the right solution
> > 
> > 
> > This PR is not removing the timer... it's keeping it active :)
> > The timer is started in `OnMediaOpened`
> 
> And we never stop the timer?

Ah, good point. OK, the `ClearTimer` is back. MediaElement will now stop the timer when setting the media source to null.

src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
```diff
@@ -255,7 +255,11 @@ public bool ShouldShowPlaybackControls
 	public MediaSource? Source
 	{
 		get => (MediaSource)GetValue(SourceProperty);
-		set => SetValue(SourceProperty, value);
+		set
+		{
+			ClearTimer();
+			SetValue(SourceProperty, value);
```
brminnick, line 261: Just curious - is there a reason we're calling `ClearTimer()` here in the Property's setter instead of calling `ClearTimer()` in `void OnSourcePropertyChanged(MediaSource?)`?

brminnick: I'd prefer to put all of the logic `void OnSourcePropertyChanged(MediaSource?)` for consistency + easier maintainability if it still solves this bug.

niksedk: There was no specific reason for using the property setter instead of `void OnSourcePropertyChanged(MediaSource?)`.

I've moved the `ClearTimer()` to `void OnSourcePropertyChanged(MediaSource?)`... still solves the issues for me.

brminnick: Right on! 🙌

brminnick (approved):

src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
```diff
@@ -372,14 +371,12 @@ public void SeekTo(TimeSpan position)
...
 	public void Stop()
 	{
-		ClearTimer();
 		OnStopRequested();
 		Handler?.Invoke(nameof(StopRequested));
 	}
 
 	internal void OnMediaEnded()
 	{
-		ClearTimer();
```
pictos, line 382: Why not clear the time when the media ended?

niksedk: `OnMediaEnded` will occur when you drag the progress-bar to the end (or play to the end)... very annoying that you need to reload the media (it's actually still loaded) after hitting the progress-bar end, in order to get progress-bar working again. 
Neither libmpv or libvlc works this way.

Starting the timer when loading the media, and stopping it when opening a new media (or setting it to null) makes sense to me.

Developers can use `OnMediaEnded` to unload/null the media file/stream if they really want the media closed and the progress-bar unresponsive.

niksedk: @pictos: Pedro, could you give this a look - it would really help, so I will not need to reload the video after moving the progress bar to the end or playing the last part of the video. It's also how the video players I normally use works (libmpv, libvlc, directshow, mpc). I could really need in order to port this Winforms project to Mai - https://github.com/SubtitleEdit/subtitleedit
TIA :)

pictos:
@jfversluis can you take a look at this PR and make sure that it's fine to change the behavior of the timer? I believe this will be a breaking change anyway.

jfversluis (approved):

Issue #542, opened 2022-08-13
[Proposal] BindableProperty generator


pictos:
# [Bindable Property (BP) generator]

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
  * [ ] iOS Support
  * [ ] Android Support
  * [ ] macOS Support
  * [ ] Windows Support
* [ ] Unit Tests: Not Started
* [ ] Sample: Not Started
* [ ] Documentation: Not Started

## Link to Discussion

- https://github.com/CommunityToolkit/Maui/discussions/458
- https://github.com/CommunityToolkit/Maui/discussions/530

## Summary
[summary]: #summary

This feature will make our lives easier (and in the future the lives of our users) allowing us to create `BindableProperties` without having to type too much (I hope).

> ### Important, this just covers the BindableProperty.Create method!
> ### This does not cover the BindableProperty.CreateAttached

## Motivation
[motivation]: #motivation

We are lazy devs that don't like to write boilerplate code, so the SG is here to save us! Let's use it

## Detailed Design
[design]: #detailed-design

Right now we don't have a final decision on what the API should look like, and it's fine so far, so I'll post here some proposals that we can vote on. Maybe we can have more than one implemented in order to make the class readable (avoid a lot of attributes at the class level for example). It's very easy to implement an attribute that covers all parameters from the [`BindableProperty.Create`](https://docs.microsoft.com/pt-br/dotnet/api/xamarin.forms.bindableproperty.create?view=xamarin-forms) method, so we can do that in our v1.

### 1. Using attributes at the class level

```csharp
 [BindableProperty(PropertyName = "CardTitle", ReturnType = typeof(string), OwnerType = typeof(CardView), DefaultValue = string.Empty, PropertyChangedMethodName = "Invalidate")]
 [BindableProperty(PropertyName = "CardDescription", ReturnType = typeof(string)]
public partial class CardView : ContentView 
{
       public static void Invalidate() { }  // OnPropertyChanged method
}

// SG will generate
public partial class CardView
{
      public static readonly BindableProperty CardTitleProperty = BindableProperty.Create("CardTitle", typeof(string), typeof(CardView), string.Empty, propertyChaged: Invalidate);
    public string CardTitle
    {
         get => (string)GetValue(CardTitleProperty);
         set => SetValue(CardTitleProperty, value);
    }

[...]

PR #1321, opened 2023-08-07
Source Generator for Bindable Property

pictos:
This will be used by the lib only, it shouldn't be part of our public nuget, because of that I deleted the template...

I'll point out the code areas that, I believe, can cause questions. In the future should be easier to port the code to the public SG project and delete this one

Fixes https://github.com/CommunityToolkit/Maui/issues/542


src/CommunityToolkit.Maui.BindablePropertySG/BindablePropertySG.cs
```diff
@@ -0,0 +1,192 @@
...
+
+	public void Initialize(IncrementalGeneratorInitializationContext context)
+	{
+		context.RegisterPostInitializationOutput(ctx => ctx.AddSource("BindablePropertyAttribute.g.cs", SourceText.From(bpAttribute, Encoding.UTF8)));
+
+		var provider = context.SyntaxProvider.ForAttributeWithMetadataName("CommunityToolkit.Maui.BindablePropertySG.BindablePropertyAttribute`1",
+			SyntaxPredicate, SemanticTransform)
+			.Where(static x => x.ClassInformation != default || !x.BPInfos.IsEmpty)
+			.Collect()
+			.SelectMany(static (types, _) => types);
```
pictos, line 53: This was necessary to avoid the provider to be a collection. I prefer to do it here instead of having a `foreach` loop inside the `Execute` method

src/CommunityToolkit.Maui.BindablePropertySG/BindablePropertySG.cs
```diff
@@ -0,0 +1,192 @@
...
+
+
+		context.RegisterSourceOutput(provider, Execute);
+	}
+
+	void Execute(SourceProductionContext context, SemanticValues semanticValues)
+	{
+		var source = GenerateSource(semanticValues);
+		SourceStringService.FormatText(ref source);
```
pictos, line 62: I copied the code as a reference, so there's no code duplication. When this goes to the main project nothing will change

src/CommunityToolkit.Maui.BindablePropertySG/BindablePropertySG.cs
```diff
@@ -0,0 +1,192 @@
...
+		context.AddSource($"{semanticValues.ClassInformation.ClassName}.g.cs", SourceText.From(source, Encoding.UTF8));
+	}
+
+	static string GenerateSource(SemanticValues value)
+	{
+		var sb = new StringBuilder($@"
+// <auto-generated>
+// Test2 : {DateTime.Now}
```
pictos, line 70: I'll leave this here, this will allow us to see if the incremental behavior is working. The value should change only when we edit/add/remove the `BindablePropertyAttribute`, if it changes in any other scenario that could be a bug.

src/CommunityToolkit.Maui.BindablePropertySG/BindablePropertySG.cs
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		sb.AppendLine().Append("}");
+		return sb.ToString();
+
+		static void GenerateBindableProperty(StringBuilder sb, BPInfo info)
+		{
+			/// <summary>
+			/// Backing BindableProperty for the <see cref="PropertyName"/> property.
+			/// </summary>
```
pictos, line 91: For us this will work very well, not sure how to improve this when it goes public.

src/CommunityToolkit.Maui.BindablePropertySG/BindablePropertySG.cs
```diff
@@ -0,0 +1,192 @@
...
+							.Append(");");
+
+			sb.AppendLine().AppendLine();
+		}
+
+		static void GenerateProperty(StringBuilder sb, BPInfo info)
+		{
+
+			/// <inheritdoc />
+			sb.AppendLine("/// <inheritdoc />");
```
pictos, line 118: For us this will work very well, not sure how to improve this when it goes public. This is inherit by the inline docs on the interface

src/CommunityToolkit.Maui.BindablePropertySG/Helpers/EquatableArray{T}.cs
```diff
@@ -0,0 +1,200 @@
+﻿//https://github.com/CommunityToolkit/dotnet/blob/main/src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/EquatableArray%7BT%7D.cs
```
pictos, line 1: We need this because we will work with a collection of BindableProperties... And all the other collections didn't work because they aren't very good with `Equals`, so I grabbed this one from Sergio, and works like a charm.

src/CommunityToolkit.Maui.BindablePropertySG/Helpers/HashCode.cs
```diff
@@ -0,0 +1,188 @@
+﻿// https://github.com/CommunityToolkit/dotnet/blob/main/src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/HashCode.cs
```
pictos, line 1: Again I grabbed some code from Sergio, this is used by the `EquatableArray<T>`, and it's necessary because we are on `.netstandard`. And because of this puppy we have to allow `unsafe` code

Yeah, I know... We are limited by the technology of our era

src/CommunityToolkit.Maui.BindablePropertySG/Models/Records.cs
```diff
@@ -0,0 +1,22 @@
+﻿using CommunityToolkit.Maui.BindablePropertySG.Helpers;
+using Microsoft.CodeAnalysis;
+
+namespace CommunityToolkit.Maui.BindablePropertySG.Models;
+
+record BPInfo
```
pictos, line 6: The `IsExternalInit` class is used on our regular SG to allow the usage of `record`s that are great for keeping the incremental behavior. I didn't added as link because VS complained sometimes

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
```diff
@@ -1,49 +1,24 @@
...
 namespace CommunityToolkit.Maui.Views;
 
 /// <inheritdoc cref="IExpander"/>
 [ContentProperty(nameof(Content))]
+
+[BindableProperty<IView>("Header", PropertyChangedMethodName = nameof(OnHeaderPropertyChanged))]
+[BindableProperty<IView>("Content", PropertyChangedMethodName = nameof(OnContentPropertyChanged))]
+[BindableProperty<bool>("IsExpanded", PropertyChangedMethodName = nameof(OnIsExpandedPropertyChanged))]
+[BindableProperty<object>("CommandParameter")]
+[BindableProperty<ICommand>("Command")]
```
pictos, line 14: Here you've an example of usage

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
```diff
@@ -84,27 +59,6 @@ public Expander()
...
-	}
-
-	/// <inheritdoc />
-	public bool IsExpanded
-	{
-		get => (bool)GetValue(IsExpandedProperty);
-		set => SetValue(IsExpandedProperty, value);
-	}
-
 	/// <inheritdoc />
```
pictos, line 62: Since `Direction` has a custom `set` method I don't use the SG to generate it. Maybe in the future, when we have `partial Property` we can come back here and improve it

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
brminnick: I love those generic attributes! 😍

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
pictos: Isn't a good idea to cast all Enums to `int`, some enums don't have a int value or use the Flag attribute. Here I would say it's best to let devs handle that instead of forcing something

VladislavAntonyuk: sorry, accidentally deleted comment:
We can choose another more generic exception (or even create a custom) or just not cast to int. 
```
if (enum)
{
	public PropertyEnumType PropertyName
	{
		get => (PropertyEnumType)GetValue(PropertyNameProperty);
		set
		{
			if (!Enum.IsDefined(typeof(PropertyEnumType), value))
			{
				throw new InvalidEnumException(nameof(value), value, typeof(PropertyEnumType));
			}

			SetValue(PropertyNameProperty, value);
		}
	}
}
```

Or we can have partial method OnPropertyChanging(), where developer can add additional logic. It will be even more flexible:
public PropertyType PropertyName
	{
		get => (PropertyType)GetValue(PropertyNameProperty);
		set
		{
			OnPropertyChanging(value);
			SetValue(PropertyNameProperty, value);
		}
	}

brminnick:
Thanks Pedro! Just FYI - I renamed the new class and its namespace to `CommunityToolkit.Maui.SourceGenerators.Internal`. 

This opens up the opportunity for us to use this `csproj` for additional internal source generators in the future.

brminnick:
Hey Pedro! This is looking really really cool! 

Could you add the ability to include XML comments in the attribute? Right now `<inheritdoc />` is hardcoded, but many of our already-implemented properties have customized XML docs.

👇 Maybe it'd look something like this?

```cs
[BindableProperty<bool>("IsExpanded", PropertyChangedMethodName = nameof(OnIsExpandedPropertyChanged), XMLComment="<inheritdoc />")]
```

pictos:
@brminnick I didn't have a good approach for XML docs, because they can be small or big, and not sure how that will fit with SG... Maybe let this be solved when C# has `partial properties`?

src/CommunityToolkit.Maui.SourceGenerators.Internal/Models/Records.cs
```diff
@@ -0,0 +1,22 @@
+﻿using CommunityToolkit.Maui.SourceGenerators.Internal.Helpers;
+using Microsoft.CodeAnalysis;
+
+namespace CommunityToolkit.Maui.SourceGenerators.Internal.Models;
+
+record BindablePropertyModel
+{
+	public string PropertyName { get; set; } = string.Empty;
```
brminnick, line 8: ### 1. Add `record XMLTagModel`

I suggest we create a new `record XMLTagModel` and add it to `BindablePropertyModel` as a nullable object. If `BindablePropertyModel.XMLTags` is `null`, then the source generator won't add any XML comments. If it is instantiated, then the source generator will generate the XML documentation for the non-null strings in `XMLTagModel`. 

The nullable string ensures we don't accidentally generate a blank XML comment for `Summary`, for example.

```cs
record XMLTagModel(string? Summary, string? Remarks, string? Returns, string? Param, string? Paramref, string? ExceptionDescription);

record BindablePropertyModel
{
        public XMLTagModel? XMLTags { get; set; }
	public string PropertyName { get; set; } = string.Empty;
// ...
}
```

### 2. Add XML Tag `string?`s to `BindablePropertyModel`

If using an object in the attribute doesn't look/feel right, we can add each XML Tag option to `BindablePropertyModel` as a nullable string.

The nullable string ensures we don't accidentally generate a blank XML comment for `Summary`, for example.

```cs
record BindablePropertyModel
{
	public string? Summary { get; set; }, 
	public string? Remarks { get; set; }
	public string? Returns { get; set; }
	public string? Param { get; set; }
	public string? Paramref { get; set; }
	public string? ExceptionDescription { get; set; }
	public string PropertyName { get; set; } = string.Empty;
// ...
}
```

Then the source generate can add the XML.

If our BindableProperty Source Generator doesn't support XML tags, then we can't use it in the Toolkit because we have so many BindableProperties with unique XML tags.

pictos: For 1: 
The usage of a custom type will not work, because it will not be a valid type for attribute, as you can see [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/attributes#2224-attribute-parameter-types).

For 2:

That will cause a lot of noise on the attribute, making it very big. Don't think the best solution will be using C#, I know that SG can consume xml and json files. The usage of xml files for docs is already used on Essentials and Maui as well, maybe we can mimic that with SG?

VladislavAntonyuk:
@pictos do we need any unresolved issues or open questions? Can we merge this PR as it is?

src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
pictos: With that, we will infer a lot about enums used on SG, in this case I think it's better to leave at it's and measure that on our usage

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui.SourceGenerators.Internal/BindablePropertyAttributeSourceGenerator.cs
```diff
@@ -0,0 +1,192 @@
...
+		var source = GenerateSource(semanticValues);
+		SourceStringService.FormatText(ref source);
+		context.AddSource($"{semanticValues.ClassInformation.ClassName}.g.cs", SourceText.From(source, Encoding.UTF8));
+	}
+
+	static string GenerateSource(SemanticValues value)
+	{
+		var sb = new StringBuilder($@"
+// <auto-generated>
+// Test2 : {DateTime.Now}
```
VladislavAntonyuk, line 70: should we remove it?

src/CommunityToolkit.Maui.SourceGenerators.Internal/Models/Records.cs
```diff
@@ -0,0 +1,22 @@
...
+	public ITypeSymbol? ReturnType { get; set; }
```
VladislavAntonyuk, line 9: do we need the whole ITypeSymbol? can we extract only necessary properties?

src/CommunityToolkit.Maui.SourceGenerators.Internal/BindablePropertyAttributeSourceGenerator.cs
pictos: for now, no... This is to make sure our generator is incremental. We can remove it when this goes public

listepo:
any news?

brminnick (approved):
Thanks Pedro! Thanks Vlad!

PR #1326, opened 2023-08-09
Touch effect maui

christianrr:
I added an first implementation for iOS, ported from XCT. 
Moreover, I fixed some issues with the Android behavior.

For me, native animations now work on both platforms as intended (like in XCT).


christianrr:
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```

@dotnet-policy-service agree company="mutor GmbH"

Issue #992, opened 2023-02-14
[BUG] Maui Community pop up doesn't work in OnAppearing()


BrianJSavage:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [X] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

I used the Maui Community PopUp example to test this issue, after I couldn't get it to work in my new app.

I needed to display a message when a page is first opened, so put the PopUp call in OnAppearing().

It fails there with System.InvalidOperationException: 'Could not locate MauiContext.'

Steps to Reproduce
Create the Popup app as described here: https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/views/popup
Add a call for the pop as below in OnAppearing()
It works with the OnAppearing call commented out, but not if used.

### Expected Behavior

Call pop up from OnAppearing()

### Steps To Reproduce

Download the repo.
Run with and without  commented out DisplayPopup() in OnAppearing().

protected override void OnAppearing()
{
//DisplayPopup();
}

### Link to public reproduction project repository

https://github.com/BrianJSavage/PopUpTest.git

### Environment

```markdown
- .NET MAUI CommunityToolkit: 4.0
- OS: w11
- .NET MAUI: 17.4.4
Latest versions of everything
```

### Anything else?

Nope

bijington:
This is fixed as part of https://github.com/CommunityToolkit/Maui/pull/934

I don't believe we have released this version yet

bijington:
@BrianJSavage I don't see why we wouldn't release this soon (next week or so) but in the meantime you can probably try using the `OnNavigatedTo` method instead of `OnAppearing`

BrianJSavage:
Thanks I'll try that.

dg2k:
> This is fixed as part of #934
> 
> I don't believe we have released this version yet

I am using version 5.1.0 but the issue is still there. Doe that mean the fix is not part of 5.1.0?

pictos:
@dg2k can you provide a small sample that reproduces the issue?

jeff-eats-pubsubs:
Also encountering this issue overriding OnAppearing() to ShowPopupAsync()

PR #934, opened 2023-01-26
Postpone popup show until platform is ready

bijington:
### Description of Change ###

 

Introduces the approach of deferring a `ShowPopup` and `ShowPopupAsync` call to the `OnNavigatedTo` event if the `IsPlatformEnabled` flag is `false`. This matches how .NET MAUI provides similar support for things like `DisplayAlert`, etc.

 ### Linked Issues ###
 

 - Fixes #822
 - Fixes #693

Possibly helps

- #810 **but this is untested**

 ### PR Checklist ###
 
 - [x] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [ ] Has tests (if omitted, state reason in description)
 - [x] Has samples (if omitted, state reason in description)
 - [x] Rebased on top of `main` at time of PR
 - [x] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#contributing-code---best-practices)
 - [ ] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/XYZ 

 ### Additional information ###

PR #1343, opened 2023-08-13
iOS popup fix with Shell or TabbedPage

cucumber-sp:
### Linked Issues ###
#822 
#992 

 - Fixes #
These issues are closed but it still happens if you have Shell or TabbedPage because they're returning true to `page.IsPlatformEnabled` even if child page is not loaded. Instead of checking `page.IsPlatformEnabled` or/and subscribing to it's navigation events, get child page (if it exists) and use it.

 ### PR Checklist ###
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/XYZ 

I tested it on iOS with OnAppearing method inside VM (`Application.Current.MainPage.ShowPopupAsync()`) which returns Shell, so it crashes. It can be avoided by using `Application.Current.MainPage.GetCurrentPage().ShowPopupAsync()` but it might be unclear for some users, so i think it should be checked before trying to display popup for page.


pictos:
You referenced closed/fixed issues... Did you or we missed something?

brminnick (approved):
Thanks @cucumber-sp!

pictos:
Reading the updated description provided by @cucumber-sp , this looks like a bug on .NET MAUI side and his fix works around the issue and doesn't fix it. The popup should use the page provided by the parameter, since that page is the Page that the user will ask the popup to show up, With this change we break it. Now, no matter what, the popup will not be attached to the required Page but by some CurrentPage provided by the system.

cucumber-sp:
@pictos nope, you're wrong here. You could check platform-specific code for popups and you will see that it used the same `.GetCurrentPage()` method while presenting popup. I just moved this thing before `page.IsPlatformEnabled` check so it won't crash if parent page is loaded, but child is not.

pictos:
@cucumber-sp the only platform-specific that has that call is on Windows, and it's because of the way that platform handles it. On others, this should be needed. And the situation that you described should be fixed by #934, since it closes your referenced issues. The correct approach was to create an issue and provide a sample code to investigate the issue.

This PR changes the meaning of the code... So, let's say that you call.

`await MyTabbedPage.ShowPopupAsync(aboutPopup);`

Now with your change, the popup will not be created in the context of the TabbedPage, but will be created in the context of the presented page (a child of the TabPage) and that's not how it was designed to work, it could introduce new bugs and for sure it's a breaking change/

cucumber-sp:
@pictos wrong again, it already did this before (https://github.com/cucumber-sp/CommunityToolkit.Maui/blob/4fbf6a8a64dbed252ab4866f87410b01605cf964/src/CommunityToolkit.Maui/Views/Popup/PopupExtensions.shared.cs#L90). The only thing I did is checking if page is loaded for final page, not the given one.

pictos:
you did what? Your change was just in one method... The line that you pointed out is what I did. You didn't check if the Page is loaded, you grabbed the currentPage. But since I'm wrong, can you educate me how that method works?

PR #1382, opened 2023-09-04
Speech Recognition v2 - StartListening/StopListening

VladislavAntonyuk:
### Description of Change ###

StartListeningAsync is to start listening for speech. The method returns "immediately", it doesn't wait for the recognition result. I would not name it Background, to not confuse users, that it works like a background service and allows to spy for the user.
so flow for users:
StartListeningAsync
Subscribe for events (SpeechCompleted/PartialSpeech)
do some other stuff
Whenever I want, I can stop listening without caring for the result.

In future we can extend the API with Pause/Resume async

trying to replicate this API: https://learn.microsoft.com/en-us/uwp/api/windows.media.speechrecognition.speechcontinuousrecognitionsession.startasync?view=winrt-22621

Added new APIs : 
- StartListeting
- StopListening
- SpeechToTextState
- SpeechRecognitionUpdated event
- SpeechRecognitionCompleted event

Removed:
Offline Speech Recognition on Windows. Maybe that is my pronunciation, but the output result is much worse than with online recognition.

 ### Linked Issues ###
 

 - Inspired by https://github.com/CommunityToolkit/Maui/discussions/1210

 ### PR Checklist ###
 
 - [ ] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/315 

### DIFF
--- a/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/Events/SpeechToTextRecognitionResultUpdatedEventArgs.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/Events/SpeechToTextRecognitionResultUpdatedEventArgs.cs
@@ -1,0 +1,20 @@
+namespace CommunityToolkit.Maui.Media;
+
+/// <summary>
+/// <see cref="EventArgs"/> for <see cref="ISpeechToText.RecognitionResultUpdated"/>
+/// </summary>
+public class SpeechToTextRecognitionResultUpdatedEventArgs : EventArgs
+{
+	/// <summary>
+	/// Speech recognition result
+	/// </summary>
+	public string RecognitionResult { get; }
+
+	/// <summary>
+	/// Initialize a new instance of <see cref="SpeechToTextRecognitionResultUpdatedEventArgs"/>
+	/// </summary>
+	public SpeechToTextRecognitionResultUpdatedEventArgs(string recognitionResult)
+	{
+		RecognitionResult = recognitionResult;
+	}
+}
--- a/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/Events/SpeechToTextRecognitionResultUpdatedEventArgs.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/Events/SpeechToTextRecognitionResultUpdatedEventArgs.cs
@@ -6,15 +6,15 @@ namespace CommunityToolkit.Maui.Media;
 public class SpeechToTextRecognitionResultUpdatedEventArgs : EventArgs
 {
 	/// <summary>
-	/// Speech recognition result
-	/// </summary>
-	public string RecognitionResult { get; }
-
-	/// <summary>
 	/// Initialize a new instance of <see cref="SpeechToTextRecognitionResultUpdatedEventArgs"/>
 	/// </summary>
 	public SpeechToTextRecognitionResultUpdatedEventArgs(string recognitionResult)
 	{
 		RecognitionResult = recognitionResult;
 	}
+
+	/// <summary>
+	/// Speech recognition result
+	/// </summary>
+	public string RecognitionResult { get; }
 }
--- a/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
@@ -54,7 +54,7 @@ public interface ISpeechToText : IAsyncDisposable
 	/// </remarks>
 	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
 	Task StopListenAsync(CancellationToken cancellationToken);
-	
+
 	/// <summary>
 	/// Request permissions for speech to text.
 	/// </summary>
--- a/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
@@ -8,14 +8,19 @@ namespace CommunityToolkit.Maui.Media;
 public interface ISpeechToText : IAsyncDisposable
 {
 	/// <summary>
-	/// Converts speech to text in real time.
+	/// Triggers when SpeechToText has real time updates
 	/// </summary>
-	event EventHandler<OnSpeechToTextRecognitionResultUpdated> RecognitionResultUpdated;
+	event EventHandler<SpeechToTextRecognitionResultUpdatedEventArgs> RecognitionResultUpdated;
 
 	/// <summary>
-	/// Final recognition result.
+	/// Triggers when SpeechToText has completed
 	/// </summary>
-	event EventHandler<OnSpeechToTextRecognitionResultCompleted> RecognitionResultCompleted;
+	event EventHandler<SpeechToTextRecognitionResultCompletedEventArgs> RecognitionResultCompleted;
+
+	/// <summary>
+	/// Triggers when <see cref="CurrentState"/> has changed
+	/// </summary>
+	event EventHandler<SpeechToTextStateChangedEventArgs> StateChanged;
 
 	/// <summary>
 	/// Current listening state
@@ -32,17 +37,23 @@ public interface ISpeechToText : IAsyncDisposable
 	Task<SpeechToTextResult> ListenAsync(CultureInfo culture, IProgress<string>? recognitionResult, CancellationToken cancellationToken);
 
 	/// <summary>
-	/// Converts speech to text in real time.
+	/// Starts the SpeechToText service
 	/// </summary>
+	/// <remarks>
+	/// Real time speech recognition results will be surfaced via <see cref="RecognitionResultUpdated"/> and <see cref="RecognitionResultCompleted"/>
+	/// </remarks>
 	/// <param name="culture">Speak language</param>
 	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
-	Task StartListeningAsync(CultureInfo culture, CancellationToken cancellationToken);
+	Task StartListenAsync(CultureInfo culture, CancellationToken cancellationToken);
 
 	/// <summary>
-	/// Stop listening.
+	/// Stops the SpeechToText service
 	/// </summary>
+	/// <remarks>
+	/// Speech recognition results will be surfaced via <see cref="RecognitionResultCompleted"/>
+	/// </remarks>
 	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
-	Task StopListeningAsync(CancellationToken cancellationToken);
+	Task StopListenAsync(CancellationToken cancellationToken);
 	
 	/// <summary>
 	/// Request permissions for speech to text.
--- a/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
@@ -10,6 +10,21 @@ public interface ISpeechToText : IAsyncDisposable
 	/// <summary>
 	/// Converts speech to text in real time.
 	/// </summary>
+	event EventHandler<OnSpeechToTextRecognitionResultUpdated> RecognitionResultUpdated;
+
+	/// <summary>
+	/// Final recognition result.
+	/// </summary>
+	event EventHandler<OnSpeechToTextRecognitionResultCompleted> RecognitionResultCompleted;
+
+	/// <summary>
+	/// Current listening state
+	/// </summary>
+	SpeechToTextState CurrentState { get; }
+
+	/// <summary>
+	/// Converts speech to text in real time.
+	/// </summary>
 	/// <param name="culture">Speak language</param>
 	/// <param name="recognitionResult">Intermediate recognition result.</param>
 	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
@@ -28,16 +43,6 @@ public interface ISpeechToText : IAsyncDisposable
 	/// </summary>
 	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
 	Task StopListeningAsync(CancellationToken cancellationToken);
-
-	/// <summary>
-	/// Converts speech to text in real time.
-	/// </summary>
-	event EventHandler<OnSpeechToTextRecognitionResultUpdated> RecognitionResultUpdated;
-
-	/// <summary>
-	/// Final recognition result.
-	/// </summary>
-	event EventHandler<OnSpeechToTextRecognitionResultCompleted> RecognitionResultCompleted;
 	
 	/// <summary>
 	/// Request permissions for speech to text.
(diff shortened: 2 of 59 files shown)


src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.ios.cs
```diff
@@ -66,12 +69,20 @@ async Task<string> InternalListenAsync(CultureInfo culture, IProgress<string>? r
...
 			}
 		});
+		return Task.CompletedTask;
+	}
 
+	async Task<string> InternalListenAsync(CultureInfo culture, IProgress<string>? recognitionResult, CancellationToken cancellationToken)
+	{
+		recognitionProgress = recognitionResult;
+		getRecognitionTaskCompletionSource ??= new TaskCompletionSource<string>();
+		await StartListeningAsync(culture, CancellationToken.None);
```
pictos, line 85: Why not use the cancellationToken provided by the user?

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.shared.cs
```diff
@@ -41,4 +85,42 @@ static async Task<bool> IsSpeechPermissionAuthorized()
 		return status is PermissionStatus.Granted;
 	}
 #endif
+}
+
+/// <summary>
+/// Event occurred on SpeechToText Recognition Result Updated
+/// </summary>
+public class OnSpeechToTextRecognitionResultUpdated : EventArgs
```
pictos, line 93: Can those EventArgs be moved to another file?
Also, is a pattern on .NET to finish the name with `Args` to be clear that this model is an `EventArgs`, so consider changing it

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
```diff
@@ -16,10 +16,54 @@ public interface ISpeechToText : IAsyncDisposable
...
+
+	/// <summary>
+	/// Converts speech to text in real time.
+	/// </summary>
+	event EventHandler<OnSpeechToTextRecognitionResultUpdated> RecognitionResultUpdated;
+
+	/// <summary>
+	/// Final recognition result.
+	/// </summary>
+	event EventHandler<OnSpeechToTextRecognitionResultCompleted> RecognitionResultCompleted;
```
pictos, line 40: I'm trying to understand this callback, if I'm using async/await pattern to start/stop listening I would spect that one of those methods (StartListeningAsync or StopListeneningAsync) returns the string result instead of having to subscribe to an eventHandler. Is something that prevents those methods from returning the result?

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.macos.cs
```diff
@@ -80,17 +84,26 @@ async Task<string> InternalListenAsync(CultureInfo culture, IProgress<string>? r
...
 						var s = result.BestTranscription.Segments[i].Substring;
 						currentIndex++;
-						recognitionResult?.Report(s);
+						recognitionProgress?.Report(s);
+						OnRecognitionResultUpdated(s);
 					}
 				}
 			}
 		});
+		return Task.CompletedTask;
```
pictos, line 93: Why not return the `getRecognitionTaskCompletionSource.Task`?

pictos (changes requested):
This looks good in general, just a friendly reminder this will be a breaking change since you're removing a feature on Windows. Also, I could see a lot of places where the `CancellationToken` provided by the user is being ignored, and this isn't ideal.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.ios.cs
VladislavAntonyuk: cancellation token is not used in any of those methods. removed it at all

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
VladislavAntonyuk: we already have a method ListenAsync, that returns result and intermediate result. So the idea of these methods just start and stop listening. we may later add Pause and resume for example

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.macos.cs
VladislavAntonyuk: this method doesn't use getRecognitionTaskCompletionSource. if you call this method, getRecognitionTaskCompletionSource is null.

brminnick (changes requested):
Thanks Vlad! 

Looks like we just need to add unit tests for the new public APIs and write the the docs.

Otherwise, the implementation looks good.

I made a couple tweaks:
- Added `event ISpeechToText.StateChanged` 
- Renamed `ISpeechToText.State` -> `ISpeechToText.CurrentState`
- Renamed `ISpeechToText.StartListeningAsync` -> `ISpeechToText.StartListenAsync`
- Renamed `ISpeechToText.StopListeningAsync` -> `ISpeechToText.StopListenAsync`
- Renamed `OnSpeechToTextRecognitionResultCompleted` -> `SpeechToTextRecognitionResultCompletedEventArgs`
- Renamed `OnSpeechToTextRecognitionResultUpdated` -> `SpeechToTextRecognitionResultUpdatedEventArgs`
- Updated XML Comments for new APIs
- Updated Sample App to be more descriptive 

Screenshot_1694303409

brminnick:
Adding the `do not merge` and `pending documentation` labels until we have the Docs PR opened

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.macos.cs
pictos: Probably I'm not following the flow, so how it's null if it's used by `recognitionTask` at line 69? And also looks like the only place that it gets instantiated is on `InternalListenAsync` method.

VladislavAntonyuk: it is marked as nullable there.
we have 2 methods: IntenalListenAsync and InternalStartListenAsync.
- InternalListenAsync is waiting for the result, so it needs that getRecognitionTaskCompletionSource. (and it initialize it, then calls InternalStartListenAsync, In that case InternalStartListenAsync has defined getRecognitionTaskCompletionSource)
- InternalStartListenAsync is called directly from new method StartListenAsync, so this getRecognitionTaskCompletionSource is not initialized in that case and just ignored

VladislavAntonyuk:
StateChanged event only makes sense on Android, where we manually switch the state, all other platforms have get only property, that returns the real state of speech recognizer.

brminnick (dismissed):
@JoonghyunCho Hey Jay!

Could you incorporate the newly added event, `ISpeechToText.StateChanged` into the Tizen implementation?

We have a new method, `void OnSpeechToTextStateChanged(SpeechToTextState)`, that you can call any time the `SpeechToText` state changes on Tizen.

https://github.com/CommunityToolkit/Maui/blob/7eda8ffc42ce36b259c04472e147092b7e597d00/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.shared.cs#L83-L86 

I'm not sure if Tizen has it's own event, like Windows' `SpeechRecognizer.StateChanged` event, that you can subscribe to, or if you'll need to call `OnSpeechToTextStateChanged()` manually, like we're doing for iOS, MacCatalyst and Windows.

brminnick:
Thanks @vlad! I think we're super close 💯

Once Jay implements `StateChanged` for Tizen, we should be ready to merge!

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/Events/SpeechToTextRecognitionResultUpdatedEventArgs.cs
```diff
@@ -5,16 +5,16 @@ namespace CommunityToolkit.Maui.Media;
...
 	/// <summary>
 	/// Initialize a new instance of <see cref="SpeechToTextRecognitionResultUpdatedEventArgs"/>
 	/// </summary>
 	public SpeechToTextRecognitionResultUpdatedEventArgs(string recognitionResult)
 	{
 		RecognitionResult = recognitionResult;
 	}
+
+	/// <summary>
+	/// Speech recognition result
```
pictos, line 17: This has the same docs as the `SpeechToTextRecognitionResultCompletedEventArgs`, and `SpeechToTextStateChangedEventArgs` can't we merge this as single one? I know, the name is little different, but the arguments and meanings are the same

VladislavAntonyuk: I would keep them separate. The same as in Windows native implementation. They may be extended later with more properties (Choices, recognition accuracy etc)

pictos: we can have a base type with those properties, I believe all Args will have the same properties or most of them will be the same

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.ios.cs
```diff
@@ -58,27 +55,29 @@ async Task<string> InternalListenAsync(CultureInfo culture, IProgress<string>? r
...
-		{
-			StopRecording();
-			getRecognitionTaskCompletionSource.TrySetCanceled();
-		}))
-		{
-			return await getRecognitionTaskCompletionSource.Task;
-		}
+		cancellationToken.ThrowIfCancellationRequested();
+
+		return Task.CompletedTask;
```
pictos, line 81: Completing this method synchrony could could be related to this [issue](https://github.com/CommunityToolkit/Maui/issues/1415)

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.android.cs
```diff
@@ -12,22 +12,68 @@ public sealed partial class SpeechToTextImplementation
...
+	}
+
+	static bool IsSpeechRecognitionAvailable() => SpeechRecognizer.IsRecognitionAvailable(Application.Context);
+
+	[MemberNotNull(nameof(speechRecognizer), nameof(listener), nameof(speechRecognitionListenerTaskCompletionSource))]
+	Task InternalStartListeningAsync(CultureInfo culture, CancellationToken cancellationToken)
 	{
+		speechRecognitionListenerTaskCompletionSource?.TrySetCanceled(cancellationToken);
+		speechRecognitionListenerTaskCompletionSource = new();
+		cancellationToken.Register(() =>
```
pictos, line 70: this should be disposed

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SharedSpeechToTextImplementation.macios.cs
```diff
@@ -47,5 +71,26 @@ void StopRecording()
...
+		return Task.CompletedTask;
+	}
+
+	[MemberNotNull(nameof(getRecognitionTaskCompletionSource))]
+	void ResetGetRecognitionTaskCompletionSource(CancellationToken token)
+	{
+		getRecognitionTaskCompletionSource?.TrySetCanceled(token);
+		getRecognitionTaskCompletionSource = new();
+
+		token.Register(() =>
```
pictos, line 90: This should be disposed

pictos (changes requested):
Just found a couple of areas that we can improve and a question regarding an opened issue.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SharedSpeechToTextImplementation.macios.cs
brminnick: Good catch! Fixed ✅

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.android.cs
brminnick: Good catch! Fixed ✅

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.ios.cs
brminnick: What do you think about merging this PR and addressing https://github.com/CommunityToolkit/Maui/issues/1415 in a future PR? 

As long as this PR isn't introducing that bug (or making it worse), I'd like to merge this because  this PR is currently blocking us from releasing CommunityToolkit.Maui v6.0.0.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/Events/SpeechToTextRecognitionResultUpdatedEventArgs.cs
brminnick: You both make good points. 

I just played around a bit with this in code and I think it's best to keep the current implementation for now.

We can't create `public abstract class SpeechToTextRecognitionEventArgs` and add `public string RecognitionResult { get; }` to it because then we's have a 3rd `EventArgs`, `SpeechToTextStateChangedEventArgs`, that doesn't implement this base class. Same is true for creating `public interface ISpeechToTextRecognitionEventArgs`; we'd always have a 3rd event args that doesn't implement an abstraction which may lead to confusion for devs expecting it.

There's nothing smelly in this code. But if/when we add additional `EventArgs` to `ISpeechToText`, we'd definitely want to rediscuss this.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.ios.cs
pictos: I'm fine with that, just want to point out if completing in a sync way is the cause of that issue. No problem to merge as it's

JoonghyunCho:
Sorry I'm late!!!!! Just added `StateChanged` to Tizen :)

JoonghyunCho:
Not sure why it suddenly cause the vulnerability error with the SkiaSharp. Can you guess anything?

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.macos.cs
brminnick: Any actions here? This is the final open conversation on this PR before we can merge it. 

Given the [security vulnerability that we patched on SkiaSharp today](https://github.com/CommunityToolkit/Maui/issues/1424), I'd like to publish a new release of `CommunityToolkit.Maui` as quickly as possible so that our users are subsequently patched!

pictos: For me it's good to go

VladislavAntonyuk:
Thank you @brminnick , @JoonghyunCho  and @pictos !
@brminnick Can we merge a PR?

brminnick:
For sure! I was just giving it a final look right now 🙌

brminnick (approved):

PR #1080, opened 2023-03-09
Camera view initial work

pictos:
Draft PR around the CameraView implementation

PR #1387, opened 2023-09-05
Camera view initial work (iOS)

Dresel:
As discussed in Discord, this is the draft PR of my contribution for the iOS part for the camera view proposal, mostly as starting point for discussion / feedback.

It is based on the original branch / PR [pj/cameraView-initial-work](https://github.com/CommunityToolkit/Maui/pull/1080). These are the things I have done:

* Added missing AndroidX Camera packages
*  Fixed some compiler warnings regarding IDisposable (CA1001)
* Added the ViewHandler and CameraManager for iOS for previewing and photo triggering
* Updated the sample page (iOS does not seem to be happy without height requests in the vertical stack layout, but I'm no definitely no expert on layouting)

The following things are still missing (might get it done until end of this week):

* Flash & Availability check
* General implementation and testing of the cleanup logic (Disconnect, Dispose, etc.)
* Unify CameraViewHandler - most if not all of the code of the ios / android file could be moved to the shared handler


src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraManager.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	protected virtual partial void PlatformTakePicture()
+	{
+		ArgumentNullException.ThrowIfNull(photoOutput);
+
+		var wrapper = new AVCapturePhotoCaptureDelegateWrapper();
+		var avCapturePhotoSettings = AVCapturePhotoSettings.Create();
+
+		AVCapturePhotoSettings.FromFormat(new NSDictionary<NSString, NSObject>(new[] {AVVideo.CodecKey},
```
pictos, line 59: Since all the values are constants, can we move this `NSDictionary<NSString, NSObject>` to a field? That way we can cache it and avoid extra allocations

src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraManager.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+
+	protected virtual partial void PlatformDisconnect()
+	{
+	}
+
+	protected virtual partial void PlatformTakePicture()
+	{
+		ArgumentNullException.ThrowIfNull(photoOutput);
+
+		var wrapper = new AVCapturePhotoCaptureDelegateWrapper();
```
pictos, line 56: Instead of creating a new delegate each time, how about we create a new `TaskCompletionSource` every time and keep one instance of the delegate?

src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraManager.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+
+	public void Dispose()
+	{
+		Dispose(true);
+		GC.SuppressFinalize(this);
+	}
+
+	protected virtual partial void PlatformConnect()
+	{
+		var device = AVCaptureDevice.GetDefaultDevice(AVMediaTypes.Video) ?? throw new InvalidOperationException();
```
pictos, line 38:
```suggestion
		var device = AVCaptureDevice.GetDefaultDevice(AVMediaTypes.Video) ?? throw new InvalidOperationException("There's no camera available on your device.");
```

src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraManager.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+
+		photoOutput.CapturePhoto(avCapturePhotoSettings, wrapper);
+
+		Task.Run(async () =>
+		{
+			var (output, photo, error) = await wrapper.Task;
+
+			var nsData = photo.FileDataRepresentation;
+
+			if (nsData == null)
```
pictos, line 70: We use pattern matching for null checks
```suggestion
			if (nsData is null)
```

src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraViewHandler.ios.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+	public static Action<byte[]>? Picture { get; set; }
+	
+	CameraManager? cameraManager;
+
+	public static IPropertyMapper<ICameraView, CameraViewHandler> Propertymapper = new PropertyMapper<ICameraView, CameraViewHandler>(ViewMapper)
+	{
+		[nameof(ICameraView.CameraFlashMode)] = MapCameraFlashMode,
+		[nameof(IAvailability.IsAvailable)] = MapIsAvailable
+	};
```
pictos, line 17: This could live in a shared class since it will be the same as Android and other platforms. I put everything on `CameraViewHandler.android.cs` file for simplicity, since I just have one platform implemented.

src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraViewHandler.ios.cs
```diff
@@ -0,0 +1,82 @@
...
+	public static void MapIsAvailable(CameraViewHandler handler, ICameraView view)
+	{
+		var cameraAvailability = (IAvailability)handler.VirtualView;
+		cameraAvailability.UpdateAvailability();
+	}
+
+	public static CommandMapper<ICameraView, CameraViewHandler> Commandmapper = new(ViewCommandMapper)
+	{
+		[nameof(ICameraView.Shutter)] = MapShutter,
+	};
```
pictos, line 28: This can also be in the shared CameraViewHandler file

pictos (changes requested):
@Dresel thanks for your work, for now, here're my recommendations around the code. In general, it looks good and you're in the right direction (;

src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraManager.ios.cs
Dresel: How would you handle concurrent flows / multiple tcs?

Dresel: I could also implement AVCapturePhotoCaptureDelegate on the `CameraManager` if you prefer that.

Dresel:
@pictos I have integrated your feedback and added flash mode for ios. I have also created a gallery page (Views / Camera View).

Do we want to handle permissions? Or do we assume that the user only uses the camera view if permissions are granted?
[Connect](https://github.com/Dresel/CommunityToolkitMaui/blob/pj/cameraView-initial-work/src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraViewHandler.cs#L65) obviously fails if permissions are not granted.

Cleanup / Error handling is still on my to-do but apart from that I think the ios implementation is good to go. Waiting for a colleague so I can test on macOS / MacBook.

Dresel:
@pictos Should I go on with the windows implementation? Should I wait for feedback and further iOS requests / changes? What do you think?

bijington:
> @pictos Should I go on with the windows implementation? Should I wait for feedback and further iOS requests / changes? What do you think?

I'm happy to take a look at the current changes. If you are up for it and available then I would say please go ahead with the windows bits too. Thank you

src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraManager.ios.cs
pictos: > How would you handle concurrent flows / multiple tcs?

This code should run on UIThread, so I don't think we'll have concurrent flows... Or iOS allow the camera to be triggered from a background thread?

> I could also implement AVCapturePhotoCaptureDelegate on the CameraManager if you prefer that.

Need more time to think about it

Dresel:
> > @pictos Should I go on with the windows implementation? Should I wait for feedback and further iOS requests / changes? What do you think?
> 
> I'm happy to take a look at the current changes. If you are up for it and available then I would say please go ahead with the windows bits too. Thank you

Would you consider a (preview) release without a windows implementation or is this mandatory? Asking for a customer who is not in the need of a windows implementation.

pictos:
> Would you consider a (preview) release without a windows implementation or is this mandatory? Asking for a customer who is not in the need of a windows implementation.


@Dresel if there's a Windows control for CamareView (which I believe there isn't) we should implement it. I'll double-check if the winui team released something

pictos:
@Dresel I'm aware of that repo, but we will not go to that path, where we need to create a CameraControl, we want to use whatever the platform gives to us. That said, we (probably) will not ship the CameraView for Windows until the winui team provides a control for it.

Dresel:
I only found the "semi official" camera preview from the [CommunityToolkit Windows repository](https://github.com/CommunityToolkit/Windows) which I guess we won't include?

pictos:
No, we're following this issue https://github.com/microsoft/microsoft-ui-xaml/issues/4710

Dresel:
So I guess the Windows part is off the table? Just let me know if there is anything I can do to advance / finish this PR.

What about [CameraLocation](https://github.com/Dresel/CommunityToolkitMaui/blob/pj/cameraView-initial-work/src/CommunityToolkit.Maui.Core/Handlers/CameraView/CameraManager.shared.cs#L61)? Is this just for testing purpose or something you want to publish (couldn't find it in the initial proposal)?

peruchali:
@pictos @Dresel Is it possible to move this forward without the Windows bits?

pictos:
@peruchali yes, that's the plan

ricardocesarcara:
Hi,
Some release forecast?

tanks

pictos:
@bijington can't remember on the top of my mind, need to take a look into it and see... Will find find time to review this week

Dresel:
@bijington @pictos I'm still eager to help - let me know if there is something for me to do.

StonedHackerman:
> @Dresel I'm aware of that repo, but we will not go to that path, where we need to create a CameraControl, we want to use whatever the platform gives to us. That said, we (probably) will not ship the CameraView for Windows until the winui team provides a control for it.

Why can't we use either [https://github.com/CommunityToolkit/Windows/blob/ce82ac235bded01f8bc2f2b6ff64a5b0574998b9/components/CameraPreview/src/CameraPreview.cs](this) or [https://github.com/hjam40/Camera.MAUI/tree/master/Camera.MAUI/Platforms/Windows](this) for Windows until the WinUI team provides a control for Windows? I've been tinkering with various CameraView implementations on Windows with MAUI for some time, and I'd love to implement a temporary working CameraView for Windows (once you merge the Android and IOS implementations, so that I know the final version of the "interface" I need to provide). 

It can always be swapped later once the WinUI control exists. [https://github.com/microsoft/microsoft-ui-xaml/issues/4710](The issue) you mentioned, which this problem is following, is almost 3 years old, so the chance of an "official" control being implemented soon is like zero. If by some miracle the WinUI team actually implements this control, you can literally click delete on the files I'd commit and use the WinUI thing. 

I don't have the skills and knowledge to implement this for Android or IOS, but I'd love to implement it for Windows if you let me

mroseberry99:
Hello all.  I'm trying to find out if this is still being worked on.  I have a Xamarin.Forms app that utilizes the XTC CameraView.  It's a huge part of our app.  XTC MauiCompat CameraView for iOS does not work and I was hoping that I would find that view here as I am in the process of converting the app to .NET MAUI.  With the looming May 1st end of support deadline for Xamarin.Forms approaching fast, I would like to humbly ask if the CameraView will available in the near future? I read above that @Dresel said the iOS is pretty much ready to go.  I see a lot of the latest dialogue is around the Windows version of the view.  Is it possible to release the iOS and Android versions first without the Windows portion?  Thanks.

PR #1409, opened 2023-09-14
[MacCatalyst] Get correct window associated with the Popup

### LINKED ISSUE #1275: [BUG]

### Current Behavior

ShowPopup() and ShowPopupAsync() results in the popup generating in a wrong window on a MacCatalyst application. This requires that MacCatalyst has multi window enabled.

I originally stumbled upon this issue when I opened up another window and tried to display a popup. It resulted in the popup displaying on a primary screen. Attempting to reproduce works, but sporadically - I could not figure out exactly what was the root cause. I tried opening up windows, focusing, resizing - anything to recreate the issue with 100% success rate (instead of 50%).

After running the tests, I figured out that I can recreate this problem with 100% success rate by opening up multiple windows (not just two). I will provide the steps to reproduce.

### Expected Behavior

ShowPopup and ShowPopupAsync() should display the popup on the window that requested it.

### Steps To Reproduce

1. Clone the project from the GitHub repo.
2. Open the solution in Visual Studio for Mac and click on the "Play" button to run the application in debug mode.
3. You will now see a Mac app running. The main window will have a button that says "Click me to open a new window". Click on it two times to open up two windows. I'll call those "secondary windows".
4. Move those two secondary windows side by side (to make it more easier to see the issue).
5. Click on either one of the buttons on the left window ("Popup" or "PopupAsync" button).
6. If the popup showed up on the right window when it should have showed up on the [same] left window, then you have successfully recreated the bug. If you were unable to replicate this, then try to click on either button on the right window instead. At this point, you should be able to replicate this.

https://github.com/babenkovitaliy/MAUIToolkitPopupBug

### Environment

```markdown
- .NET MAUI CommunityToolkit: 5.2.0
- OS: MacOS Ventura 13.4.1
- .NET MAUI: 7.0.86/7.0.100
- Hardware: MacBook Air M2, 2022
```

### Anything else?

_No response_

babenkovitaliy:
This PR solves an issue with where a MacCatalyst application has multi-windows enabled, but when creating a popup, it opens up to a random window instead of the correct one.

 ### Description of Change ###

There is an issue with this line of code:
`var rootViewController = WindowStateManager.Default.GetCurrentUIViewController() ?? throw new InvalidOperationException($"{nameof(PageHandler.ViewController)} cannot be null.");`

It assumes that the application will be in a single-window state (e.g. Shell app). However, with MacOS, apps [can also be multi-window applications](https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/windows#multi-window-support).

This fix looks through all windows, locates its "View" (e.g. ContentPage), then compares it against the Parent View of the Popup. If it matches, then that is where the popup will display. Previously, it displayed on whatever window this method returned: `WindowStateManager.Default.GetCurrentUIViewController()`

 ### Linked Issues ###
 

 - Fixes #1275 

 ### PR Checklist ###
 
 - [X] Rebased on top of `main` at time of PR (fresh fork, nothing to rebase)

With these code changes, the existing demo app still works without issues. When it comes to creating a demo for this fix, this would require a massive rewrite of the demo app, since the main UI is coded as a Shell app (single window).

### DIFF
--- a/src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
+++ b/src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
@@ -95,10 +95,50 @@ public class MauiPopup : UIViewController
 		_ = View ?? throw new InvalidOperationException($"{nameof(View)} cannot be null.");
 		_ = VirtualView ?? throw new InvalidOperationException($"{nameof(VirtualView)} cannot be null.");
 
-		var rootViewController = WindowStateManager.Default.GetCurrentUIViewController() ?? throw new InvalidOperationException($"{nameof(PageHandler.ViewController)} cannot be null.");
+		UIViewController? rootViewController;
+
+#if MACCATALYST
+		var popupParentElementUIViewController = GetUIViewControllerHostingPopupParentElement(element);
+		rootViewController = popupParentElementUIViewController ?? WindowStateManager.Default.GetCurrentUIViewController() ?? throw new InvalidOperationException($"{nameof(PageHandler.ViewController)} cannot be null.");
+#else
+		rootViewController = WindowStateManager.Default.GetCurrentUIViewController() ?? throw new InvalidOperationException($"{nameof(PageHandler.ViewController)} cannot be null.");
+#endif
+
 		ViewController ??= rootViewController;
 		SetDimmingBackgroundEffect();
 	}
+
+#if MACCATALYST
+	/// <summary>
+	/// The <see cref="IPopup"/> is associated to a view (e.g. ContentPage). Use this method to locate the UIViewController in the native application that is hosting that view.
+	/// </summary>
+	/// <param name="element">An instance of <see cref="IPopup"/>.</param>
+	/// <returns>The UIViewController that is hosting the <i>element.Parent</i> view or <i>null</i> if UIViewController cannot be located.</returns>
+	UIViewController? GetUIViewControllerHostingPopupParentElement(IPopup element)
+	{
+		var scenes = UIApplication.SharedApplication.ConnectedScenes.OfType<UIWindowScene>().ToList();
+
+		foreach (var scene in scenes)
+		{
+			foreach (var window in scene.Windows)
+			{
+				if (window.RootViewController is not Microsoft.Maui.Platform.ContainerViewController rootViewController)
+				{
+					continue;
+				}
+
+				var isViewMatching = rootViewController?.CurrentView == element.Parent;
+
+				if (isViewMatching)
+				{
+					return window.RootViewController;
+				}
+			}
+		}
+
+		return null;
+	}
+#endif
 
 	void SetDimmingBackgroundEffect()
 	{
--- a/src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
+++ b/src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
@@ -116,7 +116,7 @@ public class MauiPopup : UIViewController
 	/// <returns>The UIViewController that is hosting the <i>element.Parent</i> view or <i>null</i> if UIViewController cannot be located.</returns>
 	UIViewController? GetUIViewControllerHostingPopupParentElement(IPopup element)
 	{
-		var scenes = UIApplication.SharedApplication.ConnectedScenes.OfType<UIWindowScene>().ToList();
+		var scenes = UIApplication.SharedApplication.ConnectedScenes.OfType<UIWindowScene>();
 
 		foreach (var scene in scenes)
 		{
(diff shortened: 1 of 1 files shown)


babenkovitaliy:
@microsoft-github-policy-service agree

brminnick:
Love this! @pictos - Does this introduce any breaking changes to Popup?

babenkovitaliy:
> Love this! @pictos - Does this introduce any breaking changes to Popup?

Nope. None so far. No method signature changes (e.g. parameters, return types).

babenkovitaliy:
Oh sorry. Just noticed that you tagged someone else

src/CommunityToolkit.Maui.Core/Views/Popup/MauiPopup.macios.cs
```diff
@@ -95,11 +95,51 @@ public void SetElement(IPopup element)
...
 
+#if MACCATALYST
+	/// <summary>
+	/// The <see cref="IPopup"/> is associated to a view (e.g. ContentPage). Use this method to locate the UIViewController in the native application that is hosting that view.
+	/// </summary>
+	/// <param name="element">An instance of <see cref="IPopup"/>.</param>
+	/// <returns>The UIViewController that is hosting the <i>element.Parent</i> view or <i>null</i> if UIViewController cannot be located.</returns>
+	UIViewController? GetUIViewControllerHostingPopupParentElement(IPopup element)
+	{
+		var scenes = UIApplication.SharedApplication.ConnectedScenes.OfType<UIWindowScene>().ToList();
```
pictos, line 119: Do we need to create the concrete collection? If needed please use `ToArray` instead otherwise just call the method and iterate on it using `foreach`
```suggestion
		var scenes = UIApplication.SharedApplication.ConnectedScenes.OfType<UIWindowScene>();
```

pictos (commented):
Thanks for the PR @babenkovitaliy, I was thinking here... We know what's the Popup.Parent, and it should be a Page, can't we get the information on that Page instead of looking up all available ViewControllers? Since we have a loop inside a loop it could be a O(n²)

pictos:
> Love this! @pictos - Does this introduce any breaking changes to Popup?

From my review it doesn't have any break-change

babenkovitaliy:
> Thanks for the PR @babenkovitaliy, I was thinking here... We know what's the Popup.Parent, and it should be a Page, can't we get the information on that Page instead of looking up all available ViewControllers? Since we have a loop inside a loop it could be a O(n²)

I tried doing that. 

- The Popup.Parent is a "Virtual View" - meaning that it is not native to a specific platform - there are no properties that point it to an associated native UIView. So that makes this task impossible.
- IIRC, there was a method that was able to create a UIViewController from a View, but it created a new instance of it instead of referencing the one that is currently in runtime. So that makes this task impossible.
- That is when I had to go in reverse - I took a Virtual view from Popup.Parent and tried to see if the Native Window had a UIViewController that had a reference to the Virtual View. Thankfully, I found a property under UIWindow.RootViewController called CurrentView. That was the only place that referenced the Virtual View inside a Native View. That is the reason I went with this route.

TLDR: Virtual Windows, Views, ViewControllers etc. know nothing about Native UIWindows, UIViews and UIViewControllers. However, native UIWindows, UIViews and UIViewControllers do reference that Virtual Windows, ViewControllers and Views.

babenkovitaliy:
@dotnet-policy-service agree

pictos:
> The Popup.Parent is a "Virtual View" - meaning that it is not native to a specific platform - there are no properties that point it to an associated native UIView. So that makes this task impossible.

You can do something like 

```cs
var pagehandler = VirtualView.Parent.Handler as PageHandler;
var viewController = pagehandler.ViewController;
``` 

and we're sure it's a `PageHandler` because we force it to be on the `SetElement`  method

babenkovitaliy:
> > The Popup.Parent is a "Virtual View" - meaning that it is not native to a specific platform - there are no properties that point it to an associated native UIView. So that makes this task impossible.
> 
> You can do something like
> 
> ```cs
> var pagehandler = VirtualView.Parent.Handler as PageHandler;
> var viewController = pagehandler.ViewController;
> ```
> 
> and we're sure it's a `PageHandler` because we force it to be on the `SetElement` method

I'll try to look into this and update the PR if this indeed will work. This would be a **much better** solution. The small catch is that I may have to tackle this sometime this weekend.

babenkovitaliy:
Alright. I have made the changes. I have tested out in a separate app I built where I referenced the compiled DLL files. The Popups work nicely now.

Fun, but controversial sidenote: I never realized how much I got spoiled by Visual Studio for Windows IDE. The rebase process is so much more nicer when using the GUI vs the command line (when you are not familiar with the command line). Since I was using Visual Studio for Mac, the GUI rebase process is more painful, so I had to switch over to command line. Ashamed to admit that it took me longer than it should have.

pictos:
@babenkovitaliy if I told you how many times I missed up my branches using rebase, on both GUI and command line, you would not believe it kkkkk.

Thanks for the update and glad that worked as expected (:

pictos (approved):

bijington:
These changes look good. Thanks @babenkovitaliy and @pictos  

Is it good to merge?

pictos:
Yes, it's. Enabled the auto-merge

maonaoda:
However, due to this change, the background color of the popup has become extremely light.
Can I change the background color of the screen where the Popup is located?
is PresentationController.ContainerView.BackgroundColor?

After：
image

Before：
image

PR #1444, opened 2023-10-04
Device test runner

pictos:
This PR adds the Device test runner to our repo. It's WiP but it's nice to have some visibility of you all. It's working and if you want you can start to add tests. I added two tests to be a guide for you. 

I'll need help of someone who knows yaml in order to create a step on our build to run these tests, I know that Maui does that so we can grab for them (?).

Right now these bits are in preview, and uses a private feed, I added a `nuget.config` file adding the source to the needed packages, I would say it's fine to use it since is the one that .NET MAUI uses and it the team is working in making this stable. Also, using it on our repo is a good way to help them provide early feedback.


pictos:
cc: @mattleibow just in case you have something to add (:

src/CommunityToolkit.Maui.DeviceTests/CommunityToolkit.Maui.DeviceTests/CommunityToolkit.Maui.DeviceTests.csproj
```diff
@@ -0,0 +1,73 @@
...
+    <PackageReference Include="xunit.runner.utility" Version="2.5.0-pre.44" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\CommunityToolkit.Maui.Core\CommunityToolkit.Maui.Core.csproj" />
+    <ProjectReference Include="..\..\CommunityToolkit.Maui\CommunityToolkit.Maui.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <MauiXaml Update="Tests\UITests\StatusBarTestPage.xaml">
```
VladislavAntonyuk, line 69: Can be removed

src/CommunityToolkit.Maui.DeviceTests/CommunityToolkit.Maui.DeviceTests/CommunityToolkit.Maui.DeviceTests.csproj
```diff
@@ -0,0 +1,73 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net7.0-android;net7.0-ios;net7.0-maccatalyst</TargetFrameworks>
```
VladislavAntonyuk, line 4: Please use NetVersion instead of net7.0

src/CommunityToolkit.Maui.DeviceTests/CommunityToolkit.Maui.DeviceTests/Tests/BehaviorTest.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using CommunityToolkit.Maui.Behaviors;
+using Xunit;
+
+namespace CommunityToolkit.Maui.DeviceTests.Tests;
+public class BehaviorTest
+{
+	[Fact]
+	public void SuccessfulTest()
```
VladislavAntonyuk, line 13: Can be removed

src/CommunityToolkit.Maui.DeviceTests/CommunityToolkit.Maui.DeviceTests/Tests/BehaviorTest.cs
```diff
@@ -0,0 +1,29 @@
...
+	{
+		Assert.True(true);
+	}
+
+	[Fact]
```
VladislavAntonyuk, line 18: Should it be a UIFact?

pictos: No, UIFact is just when need to test UI-related stuff. For example, to check if the expander is expanded after some action.

src/CommunityToolkit.Maui.DeviceTests/CommunityToolkit.Maui.DeviceTests/Tests/BehaviorTest.cs
pictos: Let's make it run on your CI first

VladislavAntonyuk:
@pictos can we merge this PR as it is (with comments fixes)?

bijington:
It looks like it is forcing us to upgrade to Xcode 16 and iOS 18? Is this supported via Azure Pipelines?

brminnick:
> It looks like it is forcing us to upgrade to Xcode 16 and iOS 18? Is this supported via Azure Pipelines?

You'll need to do two things:
1. Update `CommunityToolkitLibrary_Xcode_Version` to `'16'` https://github.com/CommunityToolkit/Maui/blob/372d87358d0d043a8dc1bea885a978db021eeef5/azure-pipelines.yml#L32
2. Update `TOOLKIT_NET_VERSION` to `8.0.402` https://github.com/CommunityToolkit/Maui/blob/372d87358d0d043a8dc1bea885a978db021eeef5/azure-pipelines.yml#L9C3-L9C22
  - net8.0-ios only support Xcode 16 in .NET v8.0.402+

bijington (commented):
Update to Xcode 16

mattleibow:
I have been iterating on the repo and I have added some features for running on the cli without xharness. It uses tcp connections.

Once I finish the final ios test runner, then I will ship a version and get it to work with skia as a first dogfooder. 

There are some great Android and Apple cli tools from @redth, so I am using that under the hood. 

I also want to make is easier to have a test app and not need all the boilerplate with things like icons and resources. I am hoping to almost emulate a app project with nothing in it except your tests.

This latter idea is maybe for next time I get to it, but hopefully the new cli tool can just replace the xharness way and I can deprecate/remove that dependency.

I see this PR has been here for a while, what has been the main blockers to adding testing support to the toolkit?

pictos:
> I have been iterating on the repo and I have added some features for running on the cli without xharness. It uses tcp connections.
> 
> Once I finish the final ios test runner, then I will ship a version and get it to work with skia as a first dogfooder.
> 
> There are some great Android and Apple cli tools from @Redth, so I am using that under the hood.
> 
> I also want to make is easier to have a test app and not need all the boilerplate with things like icons and resources. I am hoping to almost emulate a app project with nothing in it except your tests.
> 
> This latter idea is maybe for next time I get to it, but hopefully the new cli tool can just replace the xharness way and I can deprecate/remove that dependency.
> 
> I see this PR has been here for a while, what has been the main blockers to adding testing support to the toolkit?

@mattleibow I was working on this one, and one thing that I can remember was, how to configure it to run on CI.

Other than that I believe everything was working as expected

mattleibow:
Ah, nice. OK, so my repo now has workflows (iOS is almost done) and it is pretty simple - especailly with Jon's cool CLI tools to create, start, launch emus and sims.

The workflows are a couple steps long, so you might just be able to copy and make it work.

I really would like this all to be as simple as possible, so if anything is more that 3 chars deep  and 2 lines long, let me know and I can try make it a single stop shop.

mattleibow:
How is this going? Do we still like this? So far, I have been using it in SkiaSharp and the last updates have made it very reliable. I still use xharness as a host, but the TCP runner is working and I am investigating a dotnet run-ability for things like mac and windows.

The dream is still a dotnet test, but that may still be a net11 or later thing. There is hope. But if there is still interest in this I can have a look at all the updates and ensure the device tests can run.

I have workflows that you can copy in the repo if need be. I think now all the nugets are on nuget.org, making things nicer.

I also have a cli tool that does the work, but not required - you just have to capture the logs if you don't want it. I discovered that there is a wasdk nuget that does dotnet run now: https://www.nuget.org/packages/Microsoft.Windows.SDK.BuildTools.WinApp

That is very cool and with dotnet run now working: https://github.com/mattleibow/DeviceRunners/pull/94

TheCodeTraveler:
> The dream is still a dotnet test

That'd be great! To be honest, I never really got this working.

PR #1470, opened 2023-10-23
Add consideration of Loaded event to Popup Show method on Windows

### LINKED ISSUE #1453: [BUG] [Popup][Windows] Calling Pupop in OnAppearing Causes the Program to Crash

### Current Behavior

As shown in the figure,Calling Pupop in OnAppearing Causes the Program to Crash
This issue only occurs onWindows devices, and  Android performs normally.

popup

### Expected Behavior

Pupop can be called normally in OnAppearance

### Steps To Reproduce

1. Open Popup_OnAppearing\ Toolkit.sln
2. Click SHELL to open the second page

https://github.com/BaY1251/MauiDebug.git

### Environment

```markdown
- .NET MAUI CommunityToolkit:6.0.0
- OS:Windows10.0-22H2
- .NET MAUI:7.0.96
```

### Anything else?

_No response_

cat0363:
This PR resolves the issue where Popup crashes when displayed at OnAppearing and OnNavigatedTo timings on Windows.

 ### Description of Change ###

If you display a Popup using Shell.Current.ShowPopupAsync(popup), it will be processed in the AppShell Loaded event instead of the Page Loaded event.

[src\CommunityToolkit.Maui.Core\Handlers\Popup\PopUpHandler.windows.cs]

    public static void MapOnOpened(PopupHandler handler, IPopup view, object? result)
    {
        ArgumentNullException.ThrowIfNull(view.Parent);
        ArgumentNullException.ThrowIfNull(handler.MauiContext);
        var parent = view.Parent.ToPlatform(handler.MauiContext);
        parent.IsHitTestVisible = false;
        handler.PlatformView.XamlRoot = parent.XamlRoot;    // <= parent.XamlRoot is null
        handler.PlatformView.IsHitTestVisible = true;
        handler.PlatformView.IsOpen = true;
        view.OnOpened();
    }

For this reason, parentXaml.Root was null and the following warning was output.

    XamlRoot must be explicitly set for unparented popup.

To avoid a crash when using the Shell.Current.ShowPopupAsync(popup) method to display a Popup at the OnAppearing, OnNavigatedTo timing, I shifted the Popup display to the Loaded event timing. Furthermore, when displaying a Popup using the Shell.Current.ShowPopupAsync(popup) method, the GetCurrentPage method was used to ensure that the display of the Popup is handled by Page's Loaded instead of AppShell's Loaded. 

[src\CommunityToolkit.Maui\Views\Popup\PopupExtensions.shared.cs]

    public static void ShowPopup<TPopup>(this Page page, TPopup popup) where TPopup : Popup
    {
    #if WINDOWS
        void handler(object? sender, EventArgs args)
        {
            page.GetCurrentPage().Loaded -= handler;

            CreateAndShowPopup(page, popup);
        }

        page.GetCurrentPage().Loaded += handler;
    #else
        if (page.IsPlatformEnabled)
        {
            CreateAndShowPopup(page, popup);
        }
        else
        {
            void handler(object? sender, EventArgs args)
            {
                page.Loaded -= handler;

                CreateAndShowPopup(page, popup);
            }

            page.Loaded += handler;
        }
    #endif
    }

    public static Task<object?> ShowPopupAsync<TPopup>(this Page page, TPopup popup) where TPopup : Popup
    {
    #if WINDOWS
        var taskCompletionSource = new TaskCompletionSource<object?>();

        async void handler(object? sender, EventArgs args)
        {
            page.GetCurrentPage().Loaded -= handler;

            try
            {
                var result = await CreateAndShowPopupAsync(page, popup);

                taskCompletionSource.TrySetResult(result);
            }
            catch (Exception ex)
            {
                taskCompletionSource.TrySetException(ex);
            }
        }
        page.GetCurrentPage().Loaded += handler;

        return taskCompletionSource.Task;
    #else
        if (page.IsPlatformEnabled)
        {
            return CreateAndShowPopupAsync(page, popup);
        }
        else
        {
            var taskCompletionSource = new TaskCompletionSource<object?>();

            async void handler(object? sender, EventArgs args)
            {
                page.Loaded -= handler;

                try
                {
                    var result = await CreateAndShowPopupAsync(page, popup);

                    taskCompletionSource.TrySetResult(result);
                }
                catch (Exception ex)
                {
                    taskCompletionSource.TrySetException(ex);
                }
            }
            page.Loaded += handler;

            return taskCompletionSource.Task;
        }
    #endif
    }

 

 ### Linked Issues ###
 

 - Fixes #1453

 ### PR Checklist ###
 
Below is the execution result.

<table>
  <tr>
    <td>[OnAppearing]</td>
    <td>[OnNavigatedTo]</td>
    <td>[OnLoaded]</td>
  </tr>
  <tr>
    <td><video controls width="250" src="https://github.com/CommunityToolkit/Maui/assets/125236133/2f7d8cb0-3359-4c93-ac95-28bb4bd2e31b" /></td>
    <td><video controls width="250" src="https://github.com/CommunityToolkit/Maui/assets/125236133/867d3dbf-6318-423c-a092-a9480419f0e0" /></td>
    <td><video controls width="250" src="https://github.com/CommunityToolkit/Maui/assets/125236133/a39beb28-1178-4211-a06d-40bb6fb4f442" /></td>
  </tr>
</table>

You can see that the Popup is displayed without crashing in any of the execution results.


cat0363:
The code used for verification is below.
https://github.com/cat0363/MauiComm-VerifyPopupShow.git

pictos:
@cat0363 did you have the chance to investigate why the actual approach doesn't work for Windows? This looks like a bug on Maui side, not saying we shouldn't implement this fix here, I just want to make sure where the actual error is, to get it fixed in a better way

cat0363:
@pictos , Thank you for your comment.
I haven't done any research on it being a bug on the .NET MAUI side, so I've only looked into workarounds.
Considering that it works as expected on Android and iOS, it is possible that it is a bug in .NET MAUI, as you say.
I haven't been able to find out why the return value of Popup(Element)'s ToHandler method is null on Windows.

Below is the information I used to resolve this issue.
https://stackoverflow.com/questions/68007005/uwp-how-to-pass-in-a-non-null-xamlroot-of-a-page-to-a-navigationservice-called

Additional Information:
Currently, the only workaround without modifying the code is to use:

    Shell.Current.CurrentPage.ShowPopupAsync(popup);

cat0363:
Correct the PR description. The return value of the ToHandler method is not null. The essence of the problem is that parent.XamlRoot is null.

cat0363:
When displaying a Popup on Windows, it is necessary to set the XamlRoot, but it seems that the Loaded event is 
the timing at which this XamlRoot is set. On Android and iOS, there is no need to set XamlRoot. Because of this, 
the problem is only occurring on Windows where the XamlRoot needs to be specified. I thought it was a problem 
on the .NET MAUI side since iOS and Android work fine, but it seems that's not the case.

The cause of this problem is that I tried to display the Popup at the OnAppearing, OnNavigatedTo timing before the VisualTree was loaded. If the timing is Loaded, it will be after the VisualTree has been loaded, so you can get the expected behavior.

https://stackoverflow.com/questions/68007005/uwp-how-to-pass-in-a-non-null-xamlroot-of-a-page-to-a-navigationservice-called

I'm sorry if my thinking is wrong. 

As an aside, [this correction](https://github.com/CommunityToolkit/Maui/pull/1470#issuecomment-1774400075) is correct because it is based on debugging. Therefore, it is incorrect to say that the return value of ToHandler method is null.

pictos:
@cat0363, @bijington did this PR a while ago fixing the popup errors when users tried to open it on `OnAppearing` method
https://github.com/CommunityToolkit/Maui/pull/934

this code will postpone the shown of the popup when the page is loaded, and it's implemented using the shared layer of .net maui, because of that I believe it should work on Windows out of the box

bijington:
Is the difference that the Page hasn't loaded by the time OnNavigatedTo is called on Windows? If so it might be worth confirming from the MAUI team whether this is expected. I'll make a note to look at the MAUI code again as my original implementation borrowed a lot from their mechanism for handling DisplayAlert, etc

cat0363:
@pictos , Thanks for the PR information.

As described in Issue #1453, there was a difference in the result depending on how ShowPopupAsync was called.

Below, XamlRoot is null.

    Shell.Current.ShowPopupAsync(popup) ;

Below, XamlRoot is not null.

    Shell.Current.CurrentPage.ShowPopupAsync(popup) ;

The former will crash, and the latter will display a popup normally.
The former was AppShell and the latter was MainPage (Current).
The value of the IsPlatformEnabled property was true for the former and false for the latter at the OnAppearing timing.

The page where the Popup is displayed differs between the former and the latter.
If the [PR](https://github.com/CommunityToolkit/Maui/pull/934) previously merged as a solution is the former, it's not working.
The former means AppShell is already loaded and will not be handled by the Loaded event.

The former only executes the following:

    if (page.IsPlatformEnabled)
    {
        return CreateAndShowPopupAsync(page, popup);
    }

The latter routes calls based on the value of the IsPlatformEnabled property.

    if (page.IsPlatformEnabled)
    {
        return CreateAndShowPopupAsync(page, popup);
    }
    else
    {
        var taskCompletionSource = new TaskCompletionSource<object?>();

        async void handler(object? sender, EventArgs args)
        {
            page.Loaded -= handler;

            try
            {
                var result = await CreateAndShowPopupAsync(page, popup);

                taskCompletionSource.TrySetResult(result);
            }
            catch (Exception ex)
            {
                taskCompletionSource.TrySetException(ex);
            }
        }
        page.Loaded += handler;

        return taskCompletionSource.Task;
    }

Is the difference in behavior depending on the caller of Popup intentional? Is it unintentional?
@bijington , At least in the former case, the page is not loading.

pictos:
@PureWeen If I'm not wrong you worked on the initialization lifetime on Maui. Can you confirm if something is off on Windows? It looks like it's since Android, mac and iOS work as expected, but I would like to have your opinion on that

pictos:
Just confirmed with the .Net Maui team, that it's a bug on Windows. Here's the related issue:

- https://github.com/dotnet/maui/issues/12970

@cat0363 can you update your code and add a comment above your fix?

> TODO: This is a workaround for https://github.com/dotnet/maui/issues/12970. Remove when the issue is closed

cat0363:
@pictos , Thank you for confirming whether it is a Windows bug or not.
I added a comment and re-uploaded it.

pictos (dismissed):

brminnick (approved):
Thanks @cat0363!!

PR #692, opened 2022-10-14
Implements tizen platform

myroot:
### Description of Change ###
 * This PR implements handlers for tizen platform
 * I have to update maui versions at least 6.0.540 because on 6.0.540 had a big changes on Tizen handlers. So I update 6.0.312.json to 6.0.541.json
 * I'm not sure main branch is a best place to put this PR. if I should it put another branch, please let me know. 
 * I hope to Tizen handlers are shiped with maui 6.x versions. because many apps build this maui 6.x versions and tizen also need it
 * Tizen platform include on TargetFrameworks when only if current os platform is windows and it run on CI(`TF_BUILD`)

 

 ### Linked Issues ###
 

 - Fixes #704 

 ### PR Checklist ###
 
 - [x] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [x] Has tests (if omitted, state reason in description)
 - [x] Has samples (if omitted, state reason in description)
 - [x] Rebased on top of `main` at time of PR
 - [x] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#contributing-code---best-practices)
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/154 

 ### Additional information ###

 
 
image
image
image

PR #1503, opened 2023-11-07
Add `CancellationToken` Parameter to Methods Returning `Task` and `ValueTask`

brminnick:
### Description of Change ###

This PR updates every method that returns a `Task` or `ValueTask` to include a `CancellationToken` as a parameter.

This ensures that the .NET MAUI Community Toolkit follows [Microsoft's recommended async/await Best Practices](https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming) to always support Task Cancellation.

Requiring a `CancellationToken` also ensures that analyzer `CA2016` ("Forward the 'cancellationToken' parameter methods that take one") correctly alerts developers using our library to propagate their `CancellationToken` to our APIs.

~I've also removed the default values for `CancellationToken` parameters (eg `CancellationToken token = default` -> `CancellationToken token`) . This follows the best-practice recommendation for Libraries to require the developer to make a thoughtful choice as to how to handle cancellation.~

**Update**: Looking at the .NET BCL, their APIs provide default values for `CancellationToken`. Therefore, we too will provide default values for our `CancellationToken` parameters (eg `Cancellation token = default)`).

#### Example: `IDrawingView.GetImageStream`

For example, the `IDrawingView.GetImageStream` method changes from

```cs
ValueTask<Stream> GetImageStream(double imageSizeWidth, double imageSizeHeight, Paint background);
```

becomes

```cs
ValueTask<Stream> GetImageStream(double imageSizeWidth, double imageSizeHeight, Paint background, CancellationToken token = default);
```

 ### Linked Issues ###
 

 - Fixes https://github.com/CommunityToolkit/Maui/discussions/225

 ### PR Checklist ###
 
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/335

In the `.editorconfig`, I've [promoted `CA1068` to `error`](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068) to ensure that we follow best-practices when adding a `CancellationToken` as a parameter to methods.

I've added the https://github.com/CommunityToolkit/Maui/labels/breaking%20change label because this modifies existing APIs. I recommend merging this before we publish `CommunityToolkit v7.0.0` which contains the existing breaking change targeting .NET 8.

This PR also adds a `Timeout` to every Unit Test that returns a `Task` to ensure our Unit Test timeout when a deadlock is encountered.

~I've also added https://github.com/CommunityToolkit/Maui/labels/do%20not%20merge label until Unit Tests are added and Documentation is updated.~


src/CommunityToolkit.Maui.Core/Views/DrawingView/Service/DrawingViewService.macios.cs
```diff
@@ -15,14 +15,19 @@ public static class DrawingViewService
 	/// <param name="lines">Drawing lines</param>
 	/// <param name="imageSize">Maximum image size. The image will be resized proportionally.</param>
 	/// <param name="background">Image background</param>
+	/// <param name="token"><see cref="CancellationToken"/></param>
 	/// <returns>Image stream</returns>
-	public static ValueTask<Stream> GetImageStream(in IList<IDrawingLine> lines, in Size imageSize, in Paint? background)
+	public static ValueTask<Stream> GetImageStream(IList<IDrawingLine> lines, Size imageSize, Paint? background, CancellationToken token)
```
pictos, line 20: One thing that I would do is to make a default value for all the `token` parameters. 

```suggestion
	public static ValueTask<Stream> GetImageStream(IList<IDrawingLine> lines, Size imageSize, Paint? background, CancellationToken token = default)
```

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
```diff
@@ -132,8 +132,10 @@ public void Dispose()
 		Player?.Pause();
 	}
 
-	protected virtual partial ValueTask PlatformSeek(TimeSpan position)
+	protected virtual partial ValueTask PlatformSeek(TimeSpan position, CancellationToken token)
 	{
+		token.ThrowIfCancellationRequested();
```
pictos, line 137: I can remember from some discussion that we didn't throw when the code is related to UI code, since there's no way to `try/catch` inside xaml... And I believe it's the case here (inside this manager), the user can't `try/catch` this, so the app could crash.

pictos (commented):
This is a great idea! I've two comments around the current implementation

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
brminnick: That's fair. 

In regards to this code snipper, I don't think `Seek` is a UI thing, though. Looking at the docs, the `Position` property is read-only, and the only way to change it is in code via the `SeekTo()` method.

image

pictos: This isn't affected when the user moves that _video timeline_? I thought it can be changed by user interaction somehow

brminnick: I’ll dig a bit deeper and run some tests to confirm 👍 

I’ll plan on updating the sample app to include a “Skip Ahead 5 Seconds” button, and I’ll debug down into the code to ensure that we cannot trigger this via the UI.

src/CommunityToolkit.Maui.Core/Views/DrawingView/Service/DrawingViewService.macios.cs
brminnick: Personally, I think it's best that Libraries do **not** provide a default value to `CancellationToken` for two reasons:
1. [Analyzer CA2016 ("Forward the 'cancellationToken' parameter methods that take one")](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2016) correctly alerts developers using our library to propagate their CancellationToken to our APIs
2. It forces the developer to make conscious decision about whether they want to pass in a cancellable Token  or not. Sometimes using `CancellationToken.None` is fine

That being said, I'm happy to disagree and commit here. I understand that most MAUI developers won't use `CancellationToken` and this extra parameter is just forcing them to type more code.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
brminnick: Good news! I confirmed that this code cannot be triggered by the UI, so we're all good!

src/CommunityToolkit.Maui.Core/Views/DrawingView/Service/DrawingViewService.macios.cs
pictos: All the BCL APIs provide a default value for CancellationToken, not sure about the reason behind it

brminnick: Ah - good find! 

Ok - I'll add default value to `CancellationToken` 👍

brminnick:
@pictos Ok! This PR is ready for Review (again!).

✅ Docs are complete https://github.com/MicrosoftDocs/CommunityToolkit/pull/335
✅ Cancellation Unit Tests written
✅ Public APIs Include a Default Value for `CancellationToken` (eg `CancellationToken token = default)
✅ `Contributing.md` Updated to Include `CancellationToken` guidelines

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.windows.cs
```diff
@@ -103,7 +100,7 @@ async Task StopRecording(CancellationToken cancellationToken)
 				speechRecognizer.ContinuousRecognitionSession.Completed -= OnCompleted;
 
 				cancellationToken.ThrowIfCancellationRequested();
-				await speechRecognizer.ContinuousRecognitionSession.StopAsync();
+				await speechRecognizer.ContinuousRecognitionSession.StopAsync().AsTask(cancellationToken);
 			}
```
VladislavAntonyuk, line 104: What is the benefit of it?

src/CommunityToolkit.Maui.Core/Views/DrawingView/Service/DrawingViewService.android.cs
```diff
@@ -19,13 +19,14 @@ public static class DrawingViewService
 	/// <param name="lines">Drawing lines</param>
 	/// <param name="imageSize">Maximum image size. The image will be resized proportionally.</param>
 	/// <param name="background">Image background</param>
+	/// <param name="token"><see cref="CancellationToken"/></param>
 	/// <returns>Image stream</returns>
-	public static ValueTask<Stream> GetImageStream(in IList<IDrawingLine> lines,
-												   in Size imageSize,
-												   in Paint? background)
+	public static ValueTask<Stream> GetImageStream(IList<IDrawingLine> lines, Size imageSize, Paint? background, CancellationToken token = default)
```
VladislavAntonyuk, line 24: Maybe we should just return stream instead of ValueTask.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/SpeechToTextImplementation.windows.cs
brminnick: [`StopAsync()` returns an `IAsyncAction`](https://learn.microsoft.com/en-us/uwp/api/windows.media.speechrecognition.speechcontinuousrecognitionsession.stopasync?view=winrt-22621) (not `Task`) and does not accept a `CancellationToken` as a parameter. 

The only way for us to associate our `CancellationToken` with this long-running asynchronous method, is to first convert it to a `Task` using `.AsTask(CancellationToken)`. 

If it were a `Task`, I could just use `.WaitAsync(cancellationToken)`

brminnick: WinUI has a few of these weird APIs, like [`IAsyncAction`](https://learn.microsoft.com/en-us/uwp/api/windows.media.speechrecognition.speechcontinuousrecognitionsession.stopasync?view=winrt-22621), that are awaitable, but are not derived from `Task`

src/CommunityToolkit.Maui.Core/Views/DrawingView/Service/DrawingViewService.android.cs
brminnick: So any of the underlying methods require async/await? I don’t have my computer in front of me right now, but I think the Tizen implementation uses async/await.

brminnick: Yup - confirmed that `DrawingVewService.tizen.cs` uses async/await:

https://github.com/CommunityToolkit/Maui/blob/ece95a00bc2e959fbfd952877be32890800277fd/src/CommunityToolkit.Maui.Core/Views/DrawingView/Service/DrawingViewService.tizen.cs#L21-L46

Albiet, it is a self-induced; i.e. we are purposefully pushing the logic to a background thread. 

I hesitate to remove the async/await logic until we loop in Jay to do some performance testing. Perhaps we could do that in a future Proposal / PR? I'd really like to get this PR merged before our .NET 8 release since contains many breaking changes to our API surface.

brminnick: image

I [followed the link in the code comment](https://github.com/CommunityToolkit/Maui/pull/692#pullrequestreview-1150202727), and Seungkeun Lee (software engineer at Tizen) recommended deferring to a thread-pool for the Tizen implementation. 

https://github.com/CommunityToolkit/Maui/pull/692/files/c431204d443256918516d508905883cd60c2f146..947056d463d90f2620ece5ed7fe79cf6531f9674#diff-a26ec24614bb6020677735a034f76ab28d2279a4afd7678f7c22357888ce3c92

src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
```diff
@@ -76,31 +75,42 @@ protected override void OnDetachingFrom(VisualElement bindable)
...
-			return Task.CompletedTask;
+			return;
 		}
 
 		View.CancelAnimations();
 
 		try
 		{
-			return AnimationType.Animate(View);
+			await AnimationType.Animate(View, token);
```
brminnick, line 109:
```suggestion
			// We must `await` `AnimationType.Animate()` here in order to properly implement `Options.ShouldSuppressExceptionsInAnimations`
			// Returning the `Task` would cause the `OnAnimate()` method to return immediately, before `AnimationType.Animate()` has completed. Returning immediately exits our try/catch block and thus negates our opportunity to handle any Exceptions which breaks `Options.ShouldSuppressExceptionsInAnimations`.
			await AnimationType.Animate(View, token);
```

src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
```diff
@@ -19,14 +18,14 @@ public class AnimationBehavior : EventToCommandBehavior
...
-		BindableProperty.CreateReadOnly(nameof(AnimateCommand), typeof(ICommand), typeof(AnimationBehavior), default, BindingMode.OneWayToSource, defaultValueCreator: CreateAnimateCommand).BindableProperty;
+		BindableProperty.CreateReadOnly(nameof(AnimateCommand), typeof(Command<CancellationToken>), typeof(AnimationBehavior), default, BindingMode.OneWayToSource, propertyChanging: OnAnimateCommandChanging, defaultValueCreator: CreateAnimateCommand).BindableProperty;
 
 	TapGestureRecognizer? tapGestureRecognizer;
 
 	/// <summary>
 	/// Gets the Command that allows the triggering of the animation.
 	/// </summary>
-	public ICommand AnimateCommand => (ICommand)GetValue(AnimateCommandProperty);
+	public Command<CancellationToken> AnimateCommand => (Command<CancellationToken>)GetValue(AnimateCommandProperty);
```
brminnick, line 28:
```suggestion
	/// <remarks>
 	/// <see cref="AnimateCommand"/> has a <see cref="Type"/> of Command&lt;CancellationToken&gt; which requires a <see cref="CancellationToken"/> as a CommandParameter. See <see cref="Command{CancellationToken}"/> and <see cref="System.Windows.Input.ICommand.Execute(object)"/> for more information on passing a <see cref="CancellationToken"/> into <see cref="Command{T}"/> as a CommandParameter"
 	/// </remarks>
	public Command<CancellationToken> AnimateCommand => (Command<CancellationToken>)GetValue(AnimateCommandProperty);
```

VladislavAntonyuk (approved):

Issue #509, opened 2022-07-22
[Bug] On iOS + MacCatalyst, `AvatarView.BackgroundColor` Does Not Work


brminnick:
### Description

This is a known-issue on iOS + MacCatalyst, [noted in the AvatarView Pull Request](https://github.com/CommunityToolkit/Maui/pull/481#issuecomment-1192126478).

It is a purposefully implemented workaround because there is a bug in .NET MAUI where it paints the BackgroundColor on top of the Image on iOS. 

https://github.com/CommunityToolkit/Maui/blob/68b82412a7e31d961b7ec0da2909f38e2a07ff9a/src/CommunityToolkit.Maui/Views/AvatarView.shared.cs#L324-L341

Here's an example of what it looks like without the workaround when setting `BackgroundColor = Colors.Black`:

<img src="https://user-images.githubusercontent.com/13558917/180352566-088180f4-e99c-4ff2-9d00-c00a1010dedb.png" width=250 />

### Link to Reproduction Sample

https://github.com/CommunityToolkit/Maui/tree/main/samples

### Steps to Reproduce

1. In Visual Studio, launch `CommunityToolkit.Maui.Sample.sln`
2. In Visual Studio, set iOS as the target platform
3. In Visual Studio, build/deploy an iOS app to a simulator or device
4. In the iOS device, navigate to `Views / AvatarView Images Page`
5. In the iOS device, confirm that the Image background color is `Transparent`

### Expected Behavior

The background color of the AvatarView should reflect the `BackgroundColor` property set by the developer.

### Actual Behavior

AvatarView on iOS is hardcoded to set `BackgroundColor` to `Colors.Transparent` to work-around a .NET MAUI bug causing the BackgroundColor to be painted over top of the Image.

https://github.com/CommunityToolkit/Maui/blob/68b82412a7e31d961b7ec0da2909f38e2a07ff9a/src/CommunityToolkit.Maui/Views/AvatarView.shared.cs#L324-L341

### Workaround

Unknown

### Reproduction Imagery

In this example, `BackgroundColor = Colors.White`:

|  iOS / MacCatalyst  | Android |
| -- | -- |
| ![](https://user-images.githubusercontent.com/13558917/180354857-f880e8c9-c9dc-4461-a239-0f3435ca9144.png) | ![Screenshot_1658459618](https://user-images.githubusercontent.com/13558917/180354919-51d1bbdc-bb50-40cf-9b15-bab8bf8b15e6.png) |

PR #1539, opened 2023-11-20
Fix IOS/MacOS AvataView background color with image bug

TRybina132:
- Bug fix
 https://github.com/CommunityToolkit/Maui/issues/509

 ### Description of Change ###

This bug occurs if we use any instance of class derived from ``` GeometryGroup ```  as default clip for image in AvatarView on MacOs or IOS. In AvatarView ``` RoundRectangleGeometry ``` was used as default clip for image. This caused background color to overlay image.
image

In this pull request I decided not to set clip object if AvatarView is used on IOS/MacOS with this approach background does not overlay image on IOS or MacOS and  ``` RoundRectangleGeometry ``` still be used as default image clip for other platforms.

How it looks in MacOS 
image

In IOS
image


TRybina132:
@dotnet-policy-service agree

PR #1582, opened 2023-11-30
Use raw string literal

symbiogenesis:
This will make editing the map HTML text much easier, and isn't indented as much.  I have more changes I would like to make, and this will be the basis for that.

I just used roslyn refactorings to perform this.

Tested the sample app and it works.

### DIFF
--- a/src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
+++ b/src/CommunityToolkit.Maui.Maps/Handler/Map/MapHandler.Windows.cs
@@ -164,212 +164,211 @@ public partial class MapHandlerWindows : MapHandler
 		}
 	}
 
-	static string GetMapHtmlPage(string key)
-	{
-		var str = @$"<!DOCTYPE html>
-				<html>
-					<head>
-						<title></title>
-						<meta http-equiv=""Content-Security-Policy"" content=""default-src 'self' data: gap: https://ssl.gstatic.com 'unsafe-eval' 'unsafe-inline' https://*.bing.com https://*.virtualearth.net; style-src 'self' 'unsafe-inline' https://*.bing.com https://*.virtualearth.net; media-src *"">
-						<meta name=""viewport"" content=""user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width"">
-						<script type='text/javascript' src='https://www.bing.com/api/maps/mapcontrol?key={key}'></script>";
-		str += @"<script type='text/javascript'>
-			                var map;
-							var locationPin;
-							var trafficManager;
-							var infobox;
+	static string GetMapHtmlPage(string key) =>
+		$$"""
+		<!DOCTYPE html>
+		<html>
+			<head>
+				<title></title>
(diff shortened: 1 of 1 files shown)


symbiogenesis:
@dotnet-policy-service agree

symbiogenesis:
@dotnet-policy-service agree

symbiogenesis:
@jfversluis @BretJohnson @rmarinho @VladislavAntonyuk

brminnick:
Thanks @symbiogenesis! I'm a bit hesitant to merge this PR because it doesn't change / improve any behavior and doesn't add any benefit to the end user.

> This will make editing the map HTML text much easier

Can you help me understand this benefit? 

>  I have more changes I would like to make, and this will be the basis for that.

Could you explain the changes you would like to make? 

---

It'd be best for you to submit a [New Feature Proposal](https://github.com/CommunityToolkit/Maui/issues/new?assignees=&labels=new%2Cproposal&projects=&template=open-a-new-feature-proposal.yml&title=%5BProposal%5D+) first to help us understand your end goal behind this PR.

symbiogenesis:
> > This will make editing the map HTML text much easier
> 
> Can you help me understand this benefit?

The previous code was concatenating two strings, one of which was interpolated, and one of which wasn't. And one was verbatim and one wasn't.

Javascript uses curly braces, which would otherwise interfere with the interpolated strings, so that is why they were concatenated, so that the Javascript code wouldn't interfere with interpolation.

But the new-ish feature of [raw string literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string) is awesome, and makes it easy to just have one big multi-line block of code that reads just like a normal file, and allows interpolation using an arbitrary number of '$' symbols to designate how the interpolation should happen (double curly braces, or triple, etc)

This could also be even nicer in the future if Microsoft adds more StringSyntaxAttribute options in the future, like HTML support. ([JetBrains already has](https://www.jetbrains.com/help/rider/Language_Injections.html)).

> Could you explain the changes you would like to make?

I was thinking of pulling in a number of improvements I made to my own codebase. One of which uses the Layer feature of Bing V8 maps to make performance much better when there are a large number of pins. 

Particularly that would enable better performance with future potential features like SelectedPin, or dragging pins, or whatever. I already have all that on Windows in my code.

I would likely need to make upstream changes to fully incorporate some of this.

symbiogenesis:
This would also potentially address one of the [original nits](https://github.com/CommunityToolkit/Maui/pull/604/files#r989449235) when this feature was added originally, which is whether to move the HTML code to its own file for clarity. Or perhaps to allow syntax highlighting. But I think that wasn't specifically to make interpolation easy. This achieves more clarity, without losing interpolation.

jfversluis:
Agreed, I think this provides minimum value for something that we want to deprecate as soon as we have the chance anyway.

@symbiogenesis thank you so much for doing this and spending some time on our project, but unfortunately I don't think we'll be merging this at this time. I'm really sorry about that. I guess this is a good example of something where you want to check with us first to see if it's something we'd consider taking in.

Either way, I do appreciate you caring for our product and I hope to see more of your work!

PR #1635, opened 2024-01-04
Fix Hot Reload with Popup

### LINKED ISSUE #620: [BUG] Popup doesn't work with hot reload

### Current Behavior

`Popup` xaml changes aren't reflected with hot reload.  I'm not sure if this was part of the original intended functionality, but it can be quite tedious to iterate on `Popup` design changes.

### Expected Behavior

`Popup` xaml changes should ideally be reflected during runtime, just like other Views in .NET MAUI.

### Steps To Reproduce

1. Open and run the solution in the project repository
2. Click the button to show a popup
3. Make changes in the popup's xaml during runtime
4. Observe that your changes have not been reflected

PopupHotReloadNotWorking

https://github.com/mavispuford/PopupHotReloadNotWorking

### Environment

```markdown
- .NET MAUI CommunityToolkit: 1.2.0
- OS:
Edition	Windows 10 Enterprise
Version	20H2
OS build	19042.1889
Experience	Windows Feature Experience Pack 120.2212.4180.0
- .NET MAUI: 6.0.486
```

### Anything else?

I've also reproduced this bug in Android.  I haven't tested platforms outside of Windows/Android.

BretJohnson:
### Description of Change ###

Call `AddLogicalChild`/`RemoveLogicalChild` when the popup is shown/closed so that popups are included in the logical (and visual) tree, so that Hot Reload works for them. The Live Visual Tree VS UI now shows popup content as well.

As Shane explained to me, the rule as of MAUI in .NET8 is that whenever `Parent` is set on an `Element` currently, `AddLogicalChild` should be called on the new parent instead. `AddLogicalChild` takes care of both updating the children and parent, whereas updating the `Parent` only updates the parent not children (and will likely be deprecated). Likewise, `RemoveLogicalChild` should be called when the Parent is set to null currently.

 

 ### Linked Issues ###
 Fixes #620 

 ### PR Checklist ###
 
### DIFF
--- a/src/CommunityToolkit.Maui/HandlerImplementation/Popup/Popup.macios.cs
+++ b/src/CommunityToolkit.Maui/HandlerImplementation/Popup/Popup.macios.cs
@@ -25,12 +25,11 @@ public partial class Popup
 			var mauiContext = virtualView.Handler?.MauiContext ?? throw new NullReferenceException(nameof(IMauiContext));
 			var view = (View?)virtualView.Content ?? throw new InvalidOperationException($"{nameof(IPopup.Content)} can't be null here.");
 			view.SetBinding(BindingContextProperty, new Binding { Source = virtualView, Path = BindingContextProperty.PropertyName });
-			var parent = virtualView.Parent as Element;
 			var contentPage = new ContentPage
 			{
-				Content = view,
-				Parent = parent
+				Content = view
 			};
+			var parent = virtualView.Parent as Element;
 			parent?.AddLogicalChild(contentPage);
 
 			return (PageHandler)contentPage.ToHandler(mauiContext);
--- a/src/CommunityToolkit.Maui/HandlerImplementation/Popup/Popup.macios.cs
+++ b/src/CommunityToolkit.Maui/HandlerImplementation/Popup/Popup.macios.cs
@@ -25,11 +25,13 @@ public partial class Popup
 			var mauiContext = virtualView.Handler?.MauiContext ?? throw new NullReferenceException(nameof(IMauiContext));
 			var view = (View?)virtualView.Content ?? throw new InvalidOperationException($"{nameof(IPopup.Content)} can't be null here.");
 			view.SetBinding(BindingContextProperty, new Binding { Source = virtualView, Path = BindingContextProperty.PropertyName });
+			var parent = virtualView.Parent as Element;
 			var contentPage = new ContentPage
 			{
 				Content = view,
-				Parent = virtualView.Parent as Element
+				Parent = parent
 			};
+			parent?.AddLogicalChild(contentPage);
 
 			return (PageHandler)contentPage.ToHandler(mauiContext);
 		}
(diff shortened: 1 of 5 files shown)


pictos (dismissed):
LGTM

src/CommunityToolkit.Maui/HandlerImplementation/Popup/Popup.macios.cs
```diff
@@ -27,9 +27,10 @@ static PageHandler CreatePageHandler(IPopup virtualView)
 			view.SetBinding(BindingContextProperty, new Binding { Source = virtualView, Path = BindingContextProperty.PropertyName });
 			var contentPage = new ContentPage
 			{
-				Content = view,
-				Parent = virtualView.Parent as Element
+				Content = view
 			};
+			var parent = virtualView.Parent as Element;
+			parent?.AddLogicalChild(contentPage);
```
PureWeen, line 33: @pictos can this just be changed to 

```C#
			var parent = (Element)virtualView.Parent;
			parent.AddLogicalChild(contentPage);
```

To be consistent with other platforms? Or are there scenarios where this might actually be null?

BretJohnson: I just updated the other platforms to also be null safe. At first I thought that the `GetCurrentPage` extension method could never return null, which is why I didn't do that originally, but looking closer at its implementation it seems like there are some scenarios where it might - it's at least hard to guarantee that it doesn't. So now it's consistent, with the null safety check.

pictos: If the MainPage is null, what will be the state of a Maui application? I mean, is that a correct behavior?

BretJohnson: If the page is null, the behavior will be the same as it was before my change - the popup likely won't show at all, which arguably seems preferable to introducing an NRE in a scenario that didn't previously NRE. In practice, I don't know if it will ever be null - unlikely, but it's hard to say for certain. @PureWeen may know better. Here's the code for `GetCurrentPage` for instance (which does a `#nullable disable` at the top, unfortunately): https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/PageExtensions.cs#L9. If the shell `CurrentItem` is a `ShellSectionController` and `ShellSectionController.PresentedPage` is null, then `GetCurrentPage` will be null, but I'm not sure if this happens in practice. In any case, I just didn't want to introduce an NRE when there wasn't one there before.

pictos: @BretJohnson, it's fine to not introduce a NRE... From my point of view, Maui should throw it if this state is invalid. It's fine just not show a popup. Thanks for the references (:

VladislavAntonyuk (approved):

src/CommunityToolkit.Maui/HandlerImplementation/Popup/Popup.macios.cs
PureWeen: I think popups can only exist on pages right? so if `GetCurrentPage` is null it seems like something has gone really wrong :-)

PR #1651, opened 2024-01-18
Rebase touch effect

Axemasta:
Rebase of `touch-effect` onto `main`. There we alot of merge conflicts, I have generally kept what is in main with the exception of csproj changes


Axemasta:
Abandoning this PR for now, there was an revert that caused some breaks in this branch and I'll do a better job of avoiding them in my next pr!

PR #1658, opened 2024-01-24
Windows Snackbar rewrite

### LINKED ISSUE #353: [Bug] Snackbar opens a new app process on Windows
### Description

Snackbar opens a new window as shown here when you click the action button on Windows.

### Steps to Reproduce

1. Place a snackbar inside the default template.
2. Click on the toast action button
3. See how a new window appears.

### Expected Behavior
Action executes inside the current open app without opening a new app.

### Actual Behavior

Action executes and a new app process is opened.

### Basic Information

- Version with issue: prev 8 (MAUI prev 14)
- Last known good version: unknown
- IDE: VS

### Workaround
None without modifying the snackbar files, since the snackbar registers itself to the Activated event instead of using the solution below:

I just wanted to share a little what I've discovered after quite a bit of experimentation. I thought it might be useful for your nuget.

On Windows there's two kind of activations for UWP apps: foreground and background. The former will use the current app but you need to modify the package.appxmanifext file if you want to use `ToastNotificationManagerCompat.OnActivated`. You must both modify the appxmanifest as indicated [here](https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=uwp#step-3-handling-activation) in order to make this work without opening a new window. The background activation uses background tasks instead but they require the background tasks to be either registered as an In-Process background task or Out-of-Process background task as shown [here](https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=uwp#handling-background-activation).

For my use case the background task sounded ideal, but I was unable to make it work with MAUI (at least for the In-Process Background tasks). I kept getting argument out of range exception.

I might try the out-of-process background tasks later.

In any case, one way I found to solve this issue using the foreground activation (the one currently used [here](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.windows.cs)), you need to make this change to the `package.appxmanifest`:

in the Package element this needs to be added:
```xml
xmlns:desktop="http://schemas.microsoft.com/appx/manifest/desktop/windows10" 
xmlns:com="http://schemas.microsoft.com/appx/manifest/com/windows10"
IgnorableNamespaces="uap rescap com desktop"
```

Inside of the applications:
```xml
<Applications>
    <Application Id="App" Executable="$targetnametoken$.exe" EntryPoint="$targetentrypoint$">
      <uap:VisualElements />
     <Extensions>

        <!-- Specify which CLSID to activate when toast clicked -->
        <desktop:Extension Category="windows.toastNotificationActivation">
          <desktop:ToastNotificationActivation ToastActivatorCLSID="guid-used-in-maui-csproj" /> 
        </desktop:Extension>

        <!--Register COM CLSID LocalServer32 registry key-->
        <com:Extension Category="windows.comServer">
          <com:ComServer>
            <com:ExeServer Executable="filename-used-for-the-exe-usually-the-assembly-name.exe" Arguments="-ToastActivated" DisplayName="Toast activator">
              <com:Class Id="guid-used-in-maui-csproj" DisplayName="Toast activator"/>
            </com:ExeServer>
          </com:ComServer>
        </com:Extension>

      </Extensions>
    </Application>
  </Applications>
```
Using `$targetnametoken$.exe` for the executable in com:ExeServer didn't work. I had to manually enter the assembly name.

Finally I register in the MauiProgram.cs the event:

```csharp
using Microsoft.Toolkit.Uwp.Notifications;
builder.
... (every other build action)
.ConfigureLifecycleEvents(configure =>
            {
				configure.AddWindows(c =>
				{
					c.OnLaunched((window, args) =>
					{
						ToastNotificationManagerCompat.OnActivated += WindowsNotificationService.OnActivated;
					});
                });
            });
```

I made this minimal repo that works:
https://github.com/emorell96/MauiWithWindowsToasts

I guess to make the snackbar work you could do:

```csharp
	/// <summary>
	/// Show Snackbar
	/// </summary>
	private partial async Task ShowNative(CancellationToken token)
	{
		await DismissNative(token);
		var toastContentBuilder = new ToastContentBuilder()
			.AddText(Text)
			.AddButton(
				new ToastButton { ActivationType = ToastActivationType.Foreground }.SetContent(ActionButtonText));

		var toastContent = toastContentBuilder.GetToastContent();
		toastContent.ActivationType = ToastActivationType.Background;

		dismissedTCS = new();

		var xmlDocument = new XmlDocument();
		xmlDocument.LoadXml(toastContent.GetContent());

		NativeSnackbar = new ToastNotification(xmlDocument);
		//NativeSnackbar.Activated += OnActivated; //this is not needed anymore
		NativeSnackbar.Dismissed += OnDismissed;
		NativeSnackbar.ExpirationTime = DateTimeOffset.Now.Add(Duration);

		ToastNotificationManager.CreateToastNotifier().Show(nativeSnackbar);

		OnShown();
	}

public static void OnActivated(ToastNotification sender, object args)
	{
		if (NativeSnackbar is not null && Action is not null)
		{
			MainThread.BeginInvokeOnMainThread(Action);
		}
	}

```

And you would register OnActivated in an extension method that the user would have to register if he wants to use snackbar on windows? 

I'd be happy to contribute to it.

Hope it helps.

### Reproduction imagery

gif

### LINKED ISSUE #1649: [BUG] Application crashes under Windows when running in desktop mode

### Current Behavior

Open the CommunityToolkit.Samples solution and add this line to the csproj to run in Windows desktop mode (not in store mode):

```
<WindowsPackageType Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">None</WindowsPackageType>
```
also change the `commandName` in `launchsettings.json` to `"Project"`. Then run the project and try out the "Snackbar" and "ToastBar" examples. 

The application crashes with some low-level error.

### Expected Behavior

A notification window is shown

### Steps To Reproduce

See above

https://github.com/CommunityToolkit/Maui/

### Environment

```markdown
- .NET MAUI CommunityToolkit: Latest
- OS: Windows 10
- .NET MAUI: 8.0
```

### Anything else?

When running as store app, everything works as expected.

VladislavAntonyuk:
Replaces ToastNotification with AppNotificationManager. Fixes Snackbar and Toast execute new instance of the app

 ### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #1649 
 - Fixes #353

 ### PR Checklist ###
 
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/374


samples/CommunityToolkit.Maui.Sample/Platforms/Windows/Package.appxmanifest
```diff
@@ -3,9 +3,11 @@
     xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
     xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
     xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
-    IgnorableNamespaces="uap rescap">
+    xmlns:com="http://schemas.microsoft.com/appx/manifest/com/windows10"
+    xmlns:desktop="http://schemas.microsoft.com/appx/manifest/desktop/windows10"
+    IgnorableNamespaces="uap rescap com desktop">
 
-    <Identity Name="maui-package-name-placeholder" Publisher="CN=Microsoft" Version="0.0.0.0" />
+    <Identity Name="6e919706-2634-4d97-a93c-2213b2acc334" Publisher="CN=Microsoft" Version="0.0.0.0" />
```
pictos, line 10: Not sure if we should replace this. That placeholder is used on Resizetizer to update the `Package.appxmanifest`, looks like that placeholder will be replaced by the value on the `ApplicationId`  especified in the csproj.

Can't we use that value instead?

VladislavAntonyuk: reverted. But Resizetizer shouldn't use this, because this value is overridden when publishing the app to the Microsoft Store.

pictos: it uses it to override with the correct value, which will be used by MS store

VladislavAntonyuk: Should I set this value somewhere in csproj? Is it an ApplicationId? The default value "com.companyname.appname" is not correct for Windows. Microsoft store generates its own value after creating Publishing Profile.

pictos: I'm not sure... I would say that's better for you to see what's generated and see if there's another property we can use. For that you just need to generate a new blank maui app and build, at least, the Windows target, then go to:

`\obj\Debug\net8.0-windows10.0.19041.0\win10-x64\resizetizer\m` and compare the appxmanifest generated with one added on sln by the template.

BTW, you will see it generates an ID, but for mobile, so I'm not sure if that's used by MS store
`<mp:PhoneIdentity PhoneProductId="6FB897E8-7984-4CAE-B3BD-E801155582AB" PhonePublisherId="00000000-0000-0000-0000-000000000000" />`

VladislavAntonyuk: if we hardcode the value, it won't be replaced.
I believe there is another build in .NET  MAUI related to resizetizer. Output images/fonts are not included in the final build. I get error message about it during uploading the build file to the market. the workaround looks like this (after build copy resizetizer outputs to the /bin/release folder): https://github.com/VladislavAntonyuk/WorldExplorer/blob/main/.github/workflows/main_world-explorer_client.yml#L80

pictos: > if we hardcode the value, it won't be replaced.

I know that, my point here is that your change is needed in or der to make this feature work. Because that is against the default implementation of .net maui and I believe most apps shouldn't need to change that. And, if there's the case where we should change that value, we take care to document it very well and investigate if will not break other areas or find a way to make it work without changing the defaults of Resizetizer.

What you found looks like a bug on Resizetizer

VladislavAntonyuk: I have reverted this change so that we can keep placeholder value here.
It is probably a bug of the msbuild task that doesn't copy the files to the output folder

pictos: Do you mind to open an issue on .NET MAUI repo about the issue you found?

src/CommunityToolkit.Maui/AppBuilderExtensions.shared.cs
```diff
@@ -36,4 +55,12 @@ public static MauiAppBuilder UseMauiCommunityToolkit(this MauiAppBuilder builder
 
 		return builder;
 	}
+#if WINDOWS
+	static void OnSnackbarNotificationInvoked(
```
pictos, line 59: update docs

src/CommunityToolkit.Maui/AppBuilderExtensions.shared.cs
```diff
@@ -20,6 +22,23 @@ public static MauiAppBuilder UseMauiCommunityToolkit(this MauiAppBuilder builder
 		// Pass `null` because `options?.Invoke()` will set options on both `CommunityToolkit.Maui` and `CommunityToolkit.Maui.Core`
 		builder.UseMauiCommunityToolkitCore(null);
 
+#if WINDOWS
+		builder.ConfigureLifecycleEvents(events =>
```
pictos, line 26: updatedocs

pictos (approved):
LGTM!

Issue #103, opened 2021-09-28
[Proposal] Android-Specific Navigation Bar Effect


brminnick:
# NavigationBarEffect

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
  * [ ] iOS Support
  * [ ] Android Support
  * [ ] macOS Support
  * [ ] Windows Support
* [ ] Unit Tests: Not Started
* [ ] Sample: Not Started
* [ ] Documentation: Not Started

## Summary
[summary]: #summary

Android-specific implementation allowing the user to customize the Navigation Bar

## Detailed Design
[design]: #detailed-design

### NavigationBarEffect.cs

```cs
namespace Xamarin.CommunityToolkit.PlatformConfiguration.AndroidSpecific
{
	[Preserve(Conditional = true)]
	public static class NavigationBarEffect
	{
		static NavigationBarEffect()
		{
			#region Required work-around to prevent linker from removing the platform-specific implementation
#if __ANDROID__
			if (System.DateTime.Now.Ticks < 0)
				_ = new PlatformNavigationBarEffect();
#endif
			#endregion
		}

		public static readonly BindableProperty ColorProperty = BindableProperty.CreateAttached(
			"Color", typeof(Color), typeof(NavigationBarEffect), Color.Default, propertyChanged: TryGenerateEffect);

		public static readonly BindableProperty StyleProperty = BindableProperty.CreateAttached(
			"Style", typeof(NavigationBarStyle), typeof(NavigationBarEffect), NavigationBarStyle.Default, propertyChanged: TryGenerateEffect);

		public static Color GetColor(BindableObject bindable) =>
			(Color)bindable.GetValue(ColorProperty);

		public static void SetColor(BindableObject bindable, Color value) =>
			bindable.SetValue(ColorProperty, value);

		public static NavigationBarStyle GetStyle(BindableObject bindable) =>
			(NavigationBarStyle)bindable.GetValue(StyleProperty);

		public static void SetStyle(BindableObject bindable, NavigationBarStyle value) =>
			bindable.SetValue(StyleProperty, value);

		public static IPlatformElementConfiguration<XFP.Android, FormsElement> SetNavigationBarColor(this IPlatformElementConfiguration<XFP.Android, FormsElement> config, Color color)
		{
			SetColor(config.Element, color);
			return config;
		}

		public static Color GetNavigationBarColor(this IPlatformElementConfiguration<XFP.Android, FormsElement> config) =>
[...]

PR #1665, opened 2024-01-26
Add SystemNavigationBar behavior for android

tranb3r:
### Description of Change ###

Implement the behavior for system navigation bar (android-specific).

 ### Linked Issues ###

#103

 ### PR Checklist ###

- This feature is not available on iOS. Should we hide the page link in the sample app on iOS? It does not seem to be the case for other features that are platform specific.
- StatusBar is only available for android>23, although according to the doc [here](https://developer.android.com/reference/android/view/Window#setStatusBarColor(int)), the min api is 21. I've also set min version 23 for system navigation bar.
- I've copied the StatusBarStyle enum to SystemNavigationBarStyle. It would be better to have a single SystemBarStyle enum, and share it for both system bars, but I don't think it's worth a breaking change for now.

### DIFF
--- a/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SystemNavigationBar/SystemNavigationBarBehavior.shared.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SystemNavigationBar/SystemNavigationBarBehavior.shared.cs
@@ -1,0 +1,87 @@
+
+using System.Runtime.CompilerServices;
+using System.Runtime.Versioning;
+using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.Core.Platform;
+using CommunityToolkit.Maui.Extensions;
+
+namespace CommunityToolkit.Maui.Behaviors;
+
+/// <summary>
+/// <see cref="PlatformBehavior{TView,TPlatformView}"/> that controls the System navigation bar color
+/// </summary>
+[UnsupportedOSPlatform("Windows"), UnsupportedOSPlatform("MacCatalyst"), UnsupportedOSPlatform("MacOS"), UnsupportedOSPlatform("Tizen"), UnsupportedOSPlatform("iOS")]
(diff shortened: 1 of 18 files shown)


src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SystemNavigationBar/SystemNavigationBarBehavior.shared.cs
```diff
@@ -0,0 +1,87 @@
...
+using CommunityToolkit.Maui.Core.Platform;
+using CommunityToolkit.Maui.Extensions;
+
+namespace CommunityToolkit.Maui.Behaviors;
+
+/// <summary>
+/// <see cref="PlatformBehavior{TView,TPlatformView}"/> that controls the System navigation bar color
+/// </summary>
+[UnsupportedOSPlatform("Windows"), UnsupportedOSPlatform("MacCatalyst"), UnsupportedOSPlatform("MacOS"), UnsupportedOSPlatform("Tizen"), UnsupportedOSPlatform("iOS")]
+public class SystemNavigationBarBehavior : PlatformBehavior<Page>
```
pictos, line 14: Since this is a platform-specific just for Android, I don't think this is the best way to do it. You should use the `OnPlatform` [API](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.onplatform-1?view=net-maui-8.0), we don't have it implemented here but you can adapt the implementation we did for XCT.

[Here's a sample on how to use that API](https://learn.microsoft.com/en-us/dotnet/maui/ios/platform-specifics/page-safe-area-layout?view=net-maui-8.0)

pictos (changes requested):
Thanks for your PR. I would that your implement isn't the best solution for this, since we're talking about a feature that will be available just for one platform, doesn't make sense to expose it in a multi-target way and throw exceptions on other platforms. I added some references on how you can approach this.

if you have more questions about the API implementation, let's use the issue to discuss it.

pictos:
closing this in favor of #1678

PR #1673, opened 2024-02-01
Add TouchBehavior

Axemasta:
This is a rebase of the `touch-effect` branch, back ontop of main.

The rebase wasn't pretty and there were lots of conflicts, I sided with main in every conflicted file but there were still some issues that had to be resolved by hand. Where any file was broken by the rebase, I have copied the code from `main` in to replace the contents.


bijington:
@Axemasta thanks for updating this PR. What action do you need now? Do you need a maintainer to review the current changes? Looking on my phone I see changes to popup related bits which might not be intended?

Axemasta:
The outstanding work as I am aware is:

- Fix input transparency on iOS (Code is complete and I have a branch with the fix on my fork)
- Fix unit tests (a couple are failing)
- Update sample app & make the touch behavior sample inline with other MCT sample pages

The next steps are testing and ensuring it works and is ready to merge!

I'm not sure what to do about this branch since its a rebase branch, do we abandon the old touch effect branch and merge this one (when ready)? If we decide to use the old branch, I can't rebase since i dont have push permissions and i'm unsure of how I can rebase it without doing a force push.

bijington:
What's the least amount of effort for you? I can rebase and force push if you need?

brminnick:
You don't need to fret about rebasing, @Axemasta, we have branch protection rules in place that will take care of that automatically for you. 

To keep this PR scoped to the new TouchBevahior control, please remove the all of edits to files that are unrelated to the new TouchBehavior:
- samples/CommunityToolkit.Maui.Sample/App.xaml.cs
- src/CommunityToolkit.Maui.Core/Views/Alert/AlertView.macios.cs
- src/CommunityToolkit.Maui.Core/Views/DrawingView/PlatformView/MauiDrawingView.android.cs
- src/CommunityToolkit.Maui.Core/Views/DrawingView/PlatformView/MauiDrawingView.windows.cs
- src/CommunityToolkit.Maui.Core/Views/DrawingView/PlatformView/MauiDrawingView.windows.cs
- src/CommunityToolkit.Maui.MediaElement/CommunityToolkit.Maui.MediaElement.csproj
- src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/SafeFireAndForgotExtensions.cs
- src/CommunityToolkit.Maui/CommunityToolkit.Maui.csproj

Axemasta:
@brminnick I have addressed all but two of these files in my latest commit, the files not address are:

- `samples/CommunityToolkit.Maui.Sample/App.xaml.cs`, this will be reverted when i add the demo page to the sample
- `src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/SafeFireAndForgotExtensions.cs`, this is part of the touch behavior feature

Axemasta:
I mentioned this in the discord yesterday, after updating the sample app to use a viewmodel I discovered the `TouchBehavior` will not update its binding context & the commands (or any bindings) will not work. This is due to the behavior being a `PlatformBehavior` which does not set a binding context.

I discovered this from the following sources:

- [Binding does not work for Platform dependent Behaviors](https://stackoverflow.com/questions/74607699/binding-does-not-work-for-platform-dependent-behaviors)
- [Binding does not work for Platform dependent Behaviors Maui - #11729](https://github.com/dotnet/maui/issues/11729)
- [Binding does not work for Platform dependent Behaviors MCt - #795 ](https://github.com/CommunityToolkit/Maui/issues/795)

For the time being I have opted to set the binding context of the platform behavior so that the sample functions in the intended way. My question here is will this cause issues for consumers of the behavior? Is there any special handling we will need to do to ensure there aren't issues or would we be better reverting to the old api design, which uses attached properties to achieve the behavior?

@pictos what do you think about this issue?

pictos:
@Axemasta Yeah, as you can see on #795 it's this way by design (as I mentioned in the discussion). 

> For the time being I have opted to set the binding context of the platform behavior so that the sample functions in the intended way

The user can set the BindingContext itself, and use the normal Binding expression. And I believe that we should align our implementation with the general recommendation, that's to set BindingContext on behaviors.

bijington:
> And I believe that we should align our implementation with the general recommendation, that's to set BindingContext on behaviors.

Does this mean the toolkit should set the BindingContext in the behaviors? Or we apply it to our samples?

Axemasta:
> I'm not criticizing anything here, just stating it feels a bit dangerous to allow it

Previously you couldn't do this because it was using an attached effect. The docs are going to have to be quite clear here that here is a specific way you setup binding & you definitely shouldn't apply to a style.

People can still abuse it at their own risk, but the documented way should work fine!

brminnick:
> The docs are going to have to be quite clear

The .NET Maui docs on Behaviors are very explicit that behaviors do not inherit the BindingContext of their parent: https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/behaviors?view=net-maui-8.0#create-a-net-maui-behavior

image

bijington:
Oh good spot! We should link to that on all behavior pages

bijington:
@Axemasta do you think it's fairly safe for me to get started on the docs PR for this?

Axemasta:
> @Axemasta do you think it's fairly safe for me to get started on the docs PR for this?

Yes I think so, the hover feature requires implementing and I'm starting to look at that. Api wise everything is there and aside from hover it behaves how it did in forms

Axemasta:
Current status of the pr, I have fixed hover which is now working on both platforms. Tests immediately all went green as a bonus (minus 1 test that had the wrong default plugged into it).

All that remains dev wise is to fix the issue i'm seeing on android devices where a child view becomes invisible if the parent has the effect applied. I've narrowed it down to `UseNativeAnimation` being the culprit, when we don't apply a ripple view the components render fine and are visible, its only when applying the ripple that the child view gets made invisible. I need to debug further to find the cause.

Axemasta:
> Perhaps it helps you.

Thanks its good to know it wasn't working properly in the XCT. I checked my port and the iOS native animations work but the android ones don't show so I've clearly ported the bug 😂. Currently the ripple animation is actually playing on Android, the issue is that the ripple view swallows any child views therefore causing the ui not to render correctly.

I've actually found the root cause and I'm working on shipping a fix to position the ripple view behind any child views so the order of views it correct.

The XamEffect library looks really cool, I've used the syncfusion ripple in the past and love how slick it makes your controls look. Its a small addition that makes a big difference to the look & feel of the app 😌

Axemasta:
Ok i've been able to fix the issue i was seeing on android with nested view, I've simply removed the code that brings the ripple view to the front, it was causing the problems.

I can verify expected appearances work:
nested android fix

And more complicated nested layouts also work properly:
mullti nest demonstration

This also works for a single view that has no nested views :)

I believe this PR is feature complete now and is ready for a proper review!

bijington:
I hate to suggest big changes but is it a silly idea to suggest a rename to `InteractionBehavior` given this does a lot more than just touch based interactions?

bijington:
Wow! There is a LOT to this TouchBehavior API! I know this came from XCT and @Axemasta you have put in a lot of effort to bring it to the toolkit! I am just trying to comprehend how I can get some docs written. I think I will take the samples page and build on from there.

These question(s) are purely to open a conversation:

- do we know why we have properties like `NormalBackgroundImageSource` that only apply when the behavior is attached to an `Image`? I've been trying to think on how it might be possible (and failing) to only expose these properties when the behavior is attached to an `Image` element.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,789 @@
...
+
+	internal async Task ChangeStateAsync(TouchBehavior sender, bool animated)
+	{
+		var status = sender.Status;
+		var state = sender.State;
+		var hoverState = sender.HoverState;
+
+		AbortAnimations(sender);
+		animationTokenSource = new CancellationTokenSource();
+		var token = animationTokenSource.Token;
```
pictos, line 132: you create this local variable but you use `animationTokenSource.Token` when passing the token to other methods, like at line `153`. Can you refactor to pass the local variable instead?

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,789 @@
...
+			return;
+		}
+
+		var pulseCount = sender.PulseCount;
+
+		if (pulseCount == 0 || (state == TouchState.Normal && !isToggled.HasValue))
+		{
+			if (isToggled.HasValue)
+			{
+				Console.WriteLine($"Touch state: {status}");
```
pictos, line 163: Please, change all `Console.WriteLine` to `Trace.WriteLine` the reason is `Console.*` is trimmed on Release builds, so users will not have a nice log.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,789 @@
...
+
+	internal void Reset()
+	{
+		SetCustomAnimationTask(null);
+		defaultBackgroundColor = default;
+	}
+
+	internal void OnTapped(TouchBehavior sender)
+	{
+		if (!sender.CanExecute || (sender.LongPressCommand != null && sender.InteractionStatus == TouchInteractionStatus.Completed))
```
pictos, line 255:
```suggestion
		if (!sender.CanExecute || (sender.LongPressCommand is not null && sender.InteractionStatus == TouchInteractionStatus.Completed))
```

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,789 @@
...
+		}
+
+		sender.RaiseCompleted();
+	}
+
+	void HandleCollectionViewSelection(TouchBehavior sender)
+	{
+		CollectionView? parent = null;
+		VisualElement? result = null;
+		if (!sender.Element?.TryFindParentElementWithParentOfType(out result, out parent) ?? true)
```
pictos, line 277: Note for the future, on `TryFindParentElementWithParentOfType` add attributes to guide roslyn that `out` can't be null if the return is `true`

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,789 @@
...
+
+		var item = result?.BindingContext ?? result ?? throw new NullReferenceException();
+
+		switch (collectionView.SelectionMode)
+		{
+			case SelectionMode.Single:
+				collectionView.SelectedItem = item;
+				break;
+			case SelectionMode.Multiple:
+				var selectedItems = collectionView.SelectedItems?.ToList() ?? new List<object>();
```
pictos, line 292: `SelectedItems` is already a `IList` so no need to call the `.ToList()` method.
```suggestion
				var selectedItems = collectionView.SelectedItems ?? [];
```

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,789 @@
...
+		}
+	}
+
+	internal void AbortAnimations(TouchBehavior sender)
+	{
+		animationTokenSource?.Cancel();
+		animationTokenSource?.Dispose();
+		animationTokenSource = null;
+		var element = sender.Element;
+		if (element == null)
```
pictos, line 314: Please check other places for `== null` and `!= null` and change to use pattern matching.
```suggestion
		if (element is null)
```

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,789 @@
...
+
+					animationProgress = (double)progress / duration;
+				}
+				animationProgress = 1;
+			}));
+	}
+
+	public void Dispose()
+	{
+		throw new NotImplementedException();
```
pictos, line 787: Can you please implement the Dispose logic?

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/SafeFireAndForgotExtensions.cs
```diff
@@ -0,0 +1,114 @@
+namespace CommunityToolkit.Maui.Behaviors;
+
+internal static class SafeFireAndForgetExtensions
+{
+    #region Private Methods
+
+    private static async void HandleSafeFireAndForget<TException>(ValueTask valueTask, bool continueOnCapturedContext, Action<TException>? onException) where TException : Exception
```
pictos, line 7: Fix for `IDE0040`
```suggestion
    static async void HandleSafeFireAndForget<TException>(ValueTask valueTask, bool continueOnCapturedContext, Action<TException>? onException) where TException : Exception
```

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/SafeFireAndForgotExtensions.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            await valueTask.ConfigureAwait(continueOnCapturedContext);
+        }
+        catch (TException ex) when (onException != null)
+        {
+            onException(ex);
+        }
+    }
+
+    private static async void HandleSafeFireAndForget<TException>(Task task, bool continueOnCapturedContext, Action<TException>? onException) where TException : Exception
```
pictos, line 19:
```suggestion
    static async void HandleSafeFireAndForget<TException>(Task task, bool continueOnCapturedContext, Action<TException>? onException) where TException : Exception
```

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.Methods.shared.cs
```diff
@@ -0,0 +1,168 @@
+﻿namespace CommunityToolkit.Maui.Behaviors;
+
+public partial class TouchBehavior : IDisposable
+{
+	readonly NullReferenceException nre = new(nameof(Element));
```
pictos, line 5: I believe this can be `static`, that way we have just one instance for all TouchBehavior used by devs.
```suggestion
	static readonly NullReferenceException nre = new(nameof(Element));
```

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs
```diff
@@ -0,0 +1,557 @@
...
+using MColor = Microsoft.Maui.Graphics.Color;
+using MView = Microsoft.Maui.Controls.View;
+using PlatformView = Android.Views.View;
+using ParentView = Android.Views.IViewParent;
+
+namespace CommunityToolkit.Maui.Behaviors;
+
+public partial class TouchBehavior
+{
+	private static readonly MColor defaultNativeAnimationColor = MColor.FromRgba(128, 128, 128, 64);
```
pictos, line 21: Please remove all `private` keyword

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs
```diff
@@ -0,0 +1,557 @@
...
+	private ViewGroup? viewGroup;
+
+	private AccessibilityManager? accessibilityManager;
+	private AccessibilityListener? accessibilityListener;
+
+	private bool IsAccessibilityMode => accessibilityManager is not null
+	                                    && accessibilityManager.IsEnabled
+	                                    && accessibilityManager.IsTouchExplorationEnabled;
+
+	private readonly bool isAtLeastM = IsAndroidVersionAtLeast((int) BuildVersionCodes.M);
```
pictos, line 39: this value will be the same for all instances, and we with that we call android APIs just once.
```suggestion
	static readonly bool isAtLeastM = IsAndroidVersionAtLeast((int) BuildVersionCodes.M);
```

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs
```diff
@@ -0,0 +1,557 @@
...
+			{
+				rippleView.Pressed = false;
+				viewGroup?.RemoveView(rippleView);
+				rippleView.Dispose();
+				rippleView = null;
+			}
+		}
+		catch (ObjectDisposedException)
+		{
+			// Suppress exception
```
pictos, line 137: can we create a log for it? It may be useful to alert devs and us that dispose is called more than once.
```suggestion
			Trace.WriteLine("TouchBehavior is already disposed.");
```

pictos (changes requested):
Wow, that's a massive PR. Thanks for moving this work forward. I've some questions and suggestions. Also,

1. Review the code and find places where isn't in sync with our pattern, like not use the `private` keyword for members and `internal` for classes;
2. run `dotnet format` to indent the files correclty.

Thank you again for this amazing work.

Axemasta:
> I hate to suggest big changes but is it a silly idea to suggest a rename to `InteractionBehavior` given this does a lot more than just touch based interactions?

I personally think the name `TouchBehavior` is suitable since all the features are concerned with responding to different touches on the given view 😄

> * do we know why we have properties like `NormalBackgroundImageSource` that only apply when the behavior is attached to an `Image`? I've been trying to think on how it might be possible (and failing) to only expose these properties when the behavior is attached to an `Image` element.

AFAIK this is from the old design when you could change the image source based on touch, it needed to know what the "normal" was. We could see how much of the `Normal` api can be removed now & if everything still functions the same, I know in my port I had issues with views not returning to normal unless that was explicitly set.

bijington:
> I personally think the name `TouchBehavior` is suitable since all the features are concerned with responding to different touches on the given view 😄

It was the `Hover` properties that made me question the name 🙂

> AFAIK this is from the old design when you could change the image source based on touch, it needed to know what the "normal" was. We could see how much of the `Normal` api can be removed now & if everything still functions the same, I know in my port I had issues with views not returning to normal unless that was explicitly set.

Yeah I get that. I'd love to only expose the properties when they can have an effect. We might be limited by what we can achieve there

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
Axemasta: Updated

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
Axemasta: Updated to use pattern matching within scope of this pr

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
Axemasta: Updated

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/SafeFireAndForgotExtensions.cs
Axemasta: Updated

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/SafeFireAndForgotExtensions.cs
Axemasta: Updated

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.Methods.shared.cs
Axemasta: Updated

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs
Axemasta: Done

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs
Axemasta: Done

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs
Axemasta: Added

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
Axemasta: Do you have any ideas of what needs disposing in this class. The previous implementation did not implement `IDisposable`, I assume it was added for a reason?

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
Axemasta: Done

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
pictos: No idea hehe, if there isn't anything that need to be disposable, just remove the interface and the method, please

brminnick: Both `CancellationTokenSource animationTokenSource` and `CancellationTokenSource longPressTokenSource` need to be disposed

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
Axemasta: This has been updated to not use the local variable

pictos:
@Axemasta looks like the unit test are breaking the build, can you take a look on those?

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,828 @@
...
+			this.durationMultiplier = null;
+
+			try
+			{
+				await RunAnimationTask(sender, state, hoverState, animationTokenSource.Token, durationMultiplier.GetValueOrDefault()).ConfigureAwait(false);
+			}
+			catch (TaskCanceledException ex)
+			{
+				Trace.WriteLine(ex);				
+			}
```
pictos, line 179: So this one will need some investigation... The `RunAnimationTask` [method](https://github.com/Axemasta/CommunityToolkit.Maui/blob/rebase-touch-effect/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs#L747) returns a `Task<bool>`, but we never look the bool value for it, so I would say that we can just return a `Task`. With that's possible, then we can change this code to be

```suggestion
			try
			{
				await RunAnimationTask(sender, state, hoverState, animationTokenSource.Token, durationMultiplier.GetValueOrDefault()).ConfigureAwait(ConfigureAwaitOptions.SuppressThrowing);
			}
			catch (TaskCanceledException ex)
			{
				Trace.WriteLine(ex);				
			}
```

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,828 @@
...
+			{
+				aspect = sender.NormalBackgroundImageAspect;
+			}
+		}
+
+		try
+		{
+			if (sender.ShouldSetImageOnAnimationEnd && duration > TimeSpan.Zero)
+			{
+				await Task.Delay(duration, token).ConfigureAwait(ConfigureAwaitOptions.ForceYielding);
```
pictos, line 397: @brminnick I believe you added the `ForceYielding` here, what would be the case? I tried to find the reason to yield it but I failed

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.macios.cs
```diff
@@ -0,0 +1,327 @@
...
+
+			if (behavior.IsDisabled)
+			{
+				return;
+			}
+
+			isCanceled = false;
+			startPoint = GetTouchPoint(touches);
+
+			await HandleTouch(TouchStatus.Started, CancellationToken.None, TouchInteractionStatus.Started);
```
pictos, line 142: Since this is the last line of the method, we don't care where the continuation will run.
```suggestion
			await HandleTouch(TouchStatus.Started, CancellationToken.None, TouchInteractionStatus.Started).ConfigureAwait(false);
```

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs
```diff
@@ -0,0 +1,581 @@
...
+
+		isHoverSupported = false;
+	}
+
+	static ColorStateList GetColorStateList(MColor? color)
+	{
+		var animationColor = color;
+		animationColor ??= defaultNativeAnimationColor;
+
+		return new ColorStateList([Array.Empty<int>()], [animationColor.ToAndroid()]);
```
pictos, line 147: let's type less xP
```suggestion
		return new ColorStateList([[]], [animationColor.ToAndroid()]);
```

pictos (commented):
I believe we're moving into the right direction, I've some more code suggestions

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
```diff
@@ -0,0 +1,828 @@
...
+		longPressTokenSource = new CancellationTokenSource();
+
+		try
+		{
+			await Task.Delay(sender.LongPressDuration, longPressTokenSource.Token).WaitAsync(token);
+		}
+		catch (TaskCanceledException)
+		{
+			return;
+		}
```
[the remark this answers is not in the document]
brminnick: No, we don't want to suppress all exceptions. 

We only want to catch the TaskCanceledException.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs
Axemasta: This isn't my code but I'd like to defend the practise of Array.Empty, its good code hygiene because only 1 empty array will be created in the app domain. Does `[]` work the same way?, if so ignore me... However as a rule I think more code should be using the `Array.Empty`/`Enumerable.Empty` paradigm!

pictos: `[]` makes sure that compiler will use the best way to initialize that array, so under the hood it will not allocate, just like Array.Empty, as you can see [here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATABgLABQ2AjIYRsQJwAUARIDwbgIPu0CUA3GQYQG4CGUAAhgBbAA4AXAJ4BBKIIC8A2VF6SAdAFExUgDwBLAHbiAfNXadD4gNoBdIdplyAwgMW22QA=)

jfversluis:
Looks like there is a build error for the sample app because now we have a duplicate `Splash.svg` file. Was that added intentionally? Should we remove the one that is added in this PR?

brminnick (changes requested):
Thanks @Axemasta! This PR is getting really really close! We're ~90% of the way there!

I found a couple bugs in the sample app that need to be fixed. And we need to add [missing unit tests for TouchBehavior](https://dev.azure.com/dotnet/CommunityToolkit/_build/results?buildId=104051&view=codecoverage-tab) before we can merge this PR:

### iOS To-Do
- `Native | Long Press | Hover`
  -  [ ] A normal click (not a long press) increases the `Long press count`
- `Image | Native` 
  - [ ] No image is displayed

### MacCatalyst To-Do
- `Native | Long Press | Hover`
  -  [ ] A normal click (not a long press) increases the `Long press count`
- `Image | Native` 
  - [ ] No image is displayed
- `Input Transparency for nested layouts` 
  - [ ] Gray overlay appears over this section

### Android To-Do
- `Native | Long Press | Hover`
  -  [ ] A normal click (not a long press) increases the `Long press count`
- `Image | Native` 
  - [ ] No image is displayed

### Windows To-Do
- `Image | Touch` 
  - [ ] Tapping does not toggle image
- `Scale | Fade | Animated` 
  - [ ] Tapping does not scale, fade or animate layout
- `Native | Long Press | Hover`
  -  [x] A normal click (not a long press) increases the `Long press count`
- `Color | Rotation | Pulse | Animated` 
  - [ ] Tapping does animate, rotate or change color
- `Image | Native` 
  - [ ] No image is displayed
- `Nested effect` 
  - [ ]  Tapping the nested child triggers both the Child event and the Parent event
- `Input Transparency for nested layouts` 
  - [ ] Gray overlay appears over this section

### Unit Tests To-Do
Here's the link to the code coverage results for this PR where you can see where we're missing Unit Tests: https://dev.azure.com/dotnet/CommunityToolkit/_build/results?buildId=104051&view=codecoverage-tab

We don't need to reach 100% code coverage, but we should at least shoot for +80%.

- [ ] Add Unit Tests for TouchBehavior (currently at 41% coverage)
- [ ] Add Unit Tests for GestureManager (currently at 65% coverage)

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.cs
brminnick: It was a part of troubleshooting why animations weren't triggering. I now defer calling `ConfigureAwaitOptions.ForceYielding` until `RunAnimationTask()`;

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.macios.cs
brminnick: You're absolutely correct, but let's avoid this since a future PR could add more code below this line at the end of the method and cause bigger headaches than we earn from this optimization.

brminnick (changes requested):
Ok! This PR is officially 99% of the way there!

There are just a few small bugs that I need help fixing:

### Android To-Do
- `Image | Native` 
  - [ ] NativeAnimation appears away from the Image
- `Nested effect` 
  - [x]  Tapping the nested child triggers only Parent Command

### Windows To-Do
- `Nested effect` 
  - [x]  Tapping the nested child triggers both the Child Command and the Parent Command

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/ImageTouch/ImageTouchBehavior.cs
bijington, line 1: Should this be named `ImageTouchBehavior.shared.cs`?

brminnick: Yup! Good catch 💯

brminnick (approved):
Ok gang! I think we're finally at a point where we're ready to approve this PR a begin writing the documentation!

Thank you to @bijington for graciously volunteering to write the docs and @CliffAgius for offering to help! 

And thank you @Axemasta + @kphillpotts for getting this started and doing the bulk of the work!!

bijington:
The docs PR is at: https://github.com/MicrosoftDocs/CommunityToolkit/pull/392

src/CommunityToolkit.Maui.UnitTests/Behaviors/ImageTouchBehaviorTests.cs
```diff
@@ -0,0 +1,258 @@
...
+		Assert.Equal(Aspect.AspectFit, view.Aspect);
+
+		imageTouchBehavior.HandleTouch(TouchStatus.Started);
+		Assert.Equal(Aspect.AspectFill, view.Aspect);
+
+		imageTouchBehavior.HandleTouch(TouchStatus.Canceled);
+		Assert.Equal(Aspect.AspectFit, view.Aspect);
+	}
+
+	[Fact]
```
brminnick, line 120:
```suggestion
	[Fact(Timeout = (int)TestDuration.Short)]
```

src/CommunityToolkit.Maui.UnitTests/Behaviors/ImageTouchBehaviorTests.cs
```diff
@@ -0,0 +1,258 @@
...
+		imageTouchBehavior.HandleHover(HoverStatus.Entered);
+		Assert.Same(hoveredImageSource, view.Source);
+		Assert.Equal(hoveredImageSource, view.Source);
+
+		imageTouchBehavior.HandleHover(HoverStatus.Exited);
+		Assert.Same(normalImageSource, view.Source);
+		Assert.Equal(normalImageSource, view.Source);
+	}
+
+	[Fact]
```
brminnick, line 145:
```suggestion
	[Fact(Timeout = (int)TestDuration.Short)]
```

brminnick (commented):
Update Unit Tests

brminnick (dismissed):

brminnick (approved):

MitchBomcanhao:
fyi I've just tried the 99.0.0-preview1678 nightly build (the first where TouchBehavior seems to appear in), and it seems to make my windows app crash on launch with an unspecified error and on android nothing really happens when adding a touchbehaviour to a control.

MitchBomcanhao:
@pictos to get the windows app failing on launch, Just get a blank new app, add that nightly build to the project, and launch the app. it'll fail with the error shown below. It'll take me longer to figure out how to make a sample project added to github (which is a requirement to open an issue) than it will for you to try this.

```
Unspecified error

   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|39_0(Int32 hr)
   at WinRT.ExceptionHelpers.ThrowExceptionForHR(Int32 hr)
   at ABI.Microsoft.Windows.AppNotifications.IAppNotificationManagerMethods.Register(IObjectReference _obj)
   at Microsoft.Windows.AppNotifications.AppNotificationManager.Register()
   at CommunityToolkit.Maui.AppBuilderExtensions.<>c.<UseMauiCommunityToolkit>b__0_3(Application _, LaunchActivatedEventArgs _)
   at Microsoft.Maui.MauiWinUIApplication.<>c__DisplayClass1_0.<OnLaunched>b__3(OnLaunched del)
   at Microsoft.Maui.LifecycleEvents.LifecycleEventServiceExtensions.InvokeLifecycleEvents[TDelegate](IServiceProvider services, Action`1 action)
   at Microsoft.Maui.MauiWinUIApplication.OnLaunched(LaunchActivatedEventArgs args)
   at Microsoft.UI.Xaml.Application.Microsoft.UI.Xaml.IApplicationOverrides.OnLaunched(LaunchActivatedEventArgs args)
   at ABI.Microsoft.UI.Xaml.IApplicationOverrides.Do_Abi_OnLaunched_0(IntPtr thisPtr, IntPtr args)
```

brminnick:
I confirmed that TouchBehavior works as expected in CommunityToolkit.Maui.Sample on both Android + Windows

| Android | Windows | 
| -- | -- |
|  Screenshot_1711644268 |  | 

@MitchBomcanhao Could you open an Issue and provide a reproduction sample today? We are planning to release TouchBehavior in `CommunityToolkit.Maui v8.0.0` tomorrow (Friday 28 March).

MitchBomcanhao:
@brminnick I will not be able to get that done today as I'm about at the end of the working day... I came across this issue on two different projects (one a production app and another a brand new blank app where there's no other code modified other than updating maui, adding the toolkit and initialising it in mauiprogram.cs. perhaps there's some OS level issue where you might be running different windows versions? fyi I'm on windows 10 build 19045.4170 and VS 17.9.4 - that's pretty much all I got for you right now :/

ryan-autonomousutilities:
Are there any examples of using a CommandParameter for one of these touch events (I'm trying to use the new LongTap)?  I can't find one in the samples, and no matter what I try, null gets passed to the command.  I just need an example to see the proper syntax.

brminnick:
You'll need to manually set `TouchBehavior.BindingConext`. 

Here's more information: https://stackoverflow.com/questions/78245724/both-touchbehavior-commandparameter-and-touchbehavior-longpresscommandparamet

PR #1692, opened 2024-02-13
Media Element Full Screen Support for Android and Windows

vhugogarcia:
### Description of Change ###

@ne0rrmatrix and myself have been working very closely to have the full screen feature for Android and Windows implemented on the media element. It has been an **amazing team work** with collaborative meetings, etc. Implements probably one of the most wanted features from developers on which apps a media element is required. 

 ### Linked Issues/Discussions ###

 - The initial discussion can be found here: https://github.com/CommunityToolkit/Maui/discussions/929

 ### PR Checklist ###
 
 - [ ] Has tests (no need of unit tests)
 - [ ] Documentation created or updated.

MacOS and iOS already integrate the full screen feature, so no additional work is needed for those platforms. Please find below a couple of videos demoing the full screen on Android and Windows.

### Windows Sample
https://github.com/CommunityToolkit/Maui/assets/1047398/c7876902-2961-4f85-b003-a73d0fa13a7e

### Android Sample

https://github.com/CommunityToolkit/Maui/assets/1047398/ffe83af7-c0f8-4296-8f75-3d8124d504cf


src/CommunityToolkit.Maui.MediaElement/CommunityToolkit.Maui.MediaElement.csproj
```diff
@@ -48,6 +48,12 @@
     <GenerateDocumentationFile>True</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net8.0-ios|AnyCPU'">
+    <CreatePackage>false</CreatePackage>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Remove="Extensions\" />
+  </ItemGroup>
```
pictos, line 56: this can be deleted

```suggestion
```

src/CommunityToolkit.Maui.MediaElement/CommunityToolkit.Maui.MediaElement.csproj
```diff
@@ -63,6 +69,12 @@
...
   <ItemGroup>
-    <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiPackageVersion)"/>
+    <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiPackageVersion)" />
+  </ItemGroup>
+  <ItemGroup>
+    <MauiImage Include="Resources\whitefs.png" BaseSize="36,36" Resize="true" pack="true" PackagePath="." />
+  </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Extensions\" />
   </ItemGroup>
```
pictos, line 79: this can be removed 
```suggestion
```

samples/CommunityToolkit.Maui.Sample/Platforms/Android/MainActivity.cs
```diff
@@ -1,9 +1,21 @@
...
 public class MainActivity : MauiAppCompatActivity
 {
+	protected override void OnPostCreate(Bundle? savedInstanceState)
+	{
+		try
+		{
+			base.OnPostCreate(savedInstanceState);
+		}
+		catch(Exception)
+		{
```
pictos, line 17: wondering if we should log at least the exception message... Also, friendly reminder to add this on our docs, since this should be on app project.

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
```diff
@@ -22,18 +27,124 @@ public class MauiMediaElement : CoordinatorLayout
 	public MauiMediaElement(Context context, StyledPlayerView playerView) : base(context)
 	{
 		this.playerView = playerView;
+		this.context = context;
 
-		// Create a RelativeLayout for sizing the video
-		RelativeLayout relativeLayout = new(context)
+		playerView.FullscreenButtonClick += PlayerView_FullscreenButtonClick;
```
pictos, line 32: let's rename this handler to match the C# pattern

```suggestion
		playerView.FullscreenButtonClick += OnFullscreenButtonClick;
```

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
```diff
@@ -22,18 +27,124 @@ public class MauiMediaElement : CoordinatorLayout
...
+	{
+		// Ensure there is a player view
+		if (this.playerView is null)
+		{
+			return;
+		}
+
+		// Ensure current activity exists
+		var currentActivity = Platform.CurrentActivity;
+		if (currentActivity is null)
```
pictos, line 53: we could reduce the number of lines on those checks... Something like

```suggestion
		if (Platform.CurrentActivity is not Activity currentActivity
		    || currentActivity.Window is not Window currentWindow)
		    {
		        return;
		    }
```

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
```diff
@@ -20,9 +43,158 @@ public class MauiMediaElement : Grid, IDisposable
...
+		};
+		btn.Click += ButtonSetFullScreenStatus;
+
+		Image image = new()
+		{
+			Width = 40,
+			Height = 40,
+			HorizontalAlignment = Microsoft.UI.Xaml.HorizontalAlignment.Right,
+			VerticalAlignment = Microsoft.UI.Xaml.VerticalAlignment.Top,
+			Source = new BitmapImage(new Uri("ms-appx:///whitefs.png")),
```
pictos, line 62: If we cache this `BitmapImage` and reuse, it will work as expected?

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
```diff
@@ -20,9 +43,158 @@ public class MauiMediaElement : Grid, IDisposable
...
+			Width = 40,
+			Height = 40
+		};
+		ButtonContainer.Children.Add(image);
+		ButtonContainer.Children.Add(btn);
+
 		Children.Add(this.mediaPlayerElement);
+		Children.Add(ButtonContainer);
+
+		mediaPlayerElement.PointerMoved += MediaPlayerElement_PointerMoved;
```
pictos, line 78: rename to align on C# parttern.
```suggestion
		mediaPlayerElement.PointerMoved += OnMediaPlayerElementPointerMoved;
```

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
```diff
@@ -20,9 +43,158 @@ public class MauiMediaElement : Grid, IDisposable
...
+
+	void InitializeTimer()
+	{
+		if (timer is not null)
+		{
+			return;
+		}
+
+		timer = new();
+		timer.Tick += OnTimer_Tick;
```
pictos, line 97:
```suggestion
		timer.Tick += OnTimerTick;
```

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
```diff
@@ -20,9 +43,158 @@ public class MauiMediaElement : Grid, IDisposable
...
+
+	static AppWindow GetAppWindowForCurrentWindow()
+	{
+		// let's cache the CurrentPage here, since the user can navigate or background the app
+		// while this method is running
+		var currentPage = CurrentPage;
+
+		if (currentPage?.GetParentWindow().Handler.PlatformView is not MauiWinUIWindow window)
+		{
+			throw new InvalidOperationException();
```
pictos, line 185: can we add helpful message for the user?

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
pictos: Also, if I understand it correctly, this Ticker jazz is to hide the button after 5 seconds, right? Wouldn't be easier to make `MediaPlayerElement_PointerMoved` async and add a Task.Delay(TimeSpan.FromSeconds(5))?

```cs
async void MediaPlayerElement_PointerMoved(object sender, PointerRoutedEventArgs e)
	{
		e.Handled = true;
		mediaPlayerElement.PointerMoved -= MediaPlayerElement_PointerMoved;
		ButtonContainer.Visibility = Microsoft.UI.Xaml.Visibility.Visible;
		await Task.Delay(TimeSpan.FromSeconds(5));
		ButtonContainer.Visibility = Microsoft.UI.Xaml.Visibility.Collapsed;
		mediaPlayerElement.PointerMoved += MediaPlayerElement_PointerMoved;
	}

**Review by pictos — COMMENTED** [review]
Great work guys! This looks great and the community will be very happy when this get merged. I've some suggestions for you

File: src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs (hunk 9 above)
Comment (ne0rrmatrix) [reply src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs:185]: ```throw new InvalidOperationException($"{nameof(Application.Current.MainPage)} cannot be null.");```

File: src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs (hunk 6 above)
Comment (ne0rrmatrix) [reply src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs:62]: During testing there were no issues. I set full screen and restored previous state. Back and forth a few times with zero issues. I was using sample app.

File: src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs (hunk 8 above)
Comment (ne0rrmatrix) [reply src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs:97]: Yes I think that is a much better approach.

File: src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs (hunk 6 above)
Comment (pictos) [reply src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs:62]: great, so let's cache the image instead of creating a new one everytime

File: src/CommunityToolkit.Maui.MediaElement/CommunityToolkit.Maui.MediaElement.csproj (hunk 2 above)
Comment (vhugogarcia) [reply src/CommunityToolkit.Maui.MediaElement/CommunityToolkit.Maui.MediaElement.csproj]: Thanks @pictos for the recommendation. It has been applied. Marked as resolved.

File: samples/CommunityToolkit.Maui.Sample/Platforms/Android/MainActivity.cs (hunk 3 above)
Comment (vhugogarcia) [reply samples/CommunityToolkit.Maui.Sample/Platforms/Android/MainActivity.cs:17]: Thanks @pictos for the recommendation, we added a Trace to show the error message. It has been applied. For the documentation, let me touch base with @bijington to give me some insights how to add this to the documentation.

File: src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs (hunk 5 above)
Comment (vhugogarcia) [reply src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs]: Very nice way to reduce the lines. I learned something new today! Thanks @pictos for the recommendation. It has been applied. Marked as resolved.

File: src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs [hunk 10]
```diff
@@ -9,9 +25,17 @@ namespace CommunityToolkit.Maui.Core.Views;
...
 public class MauiMediaElement : Grid, IDisposable
 {
+	bool navbarExist;
+	bool isDisposed;
+
+	static readonly AppWindow? appWindow = GetAppWindowForCurrentWindow();
 	readonly MediaPlayerElement mediaPlayerElement;
 
-	bool isDisposed;
+	ImageSource source { get; set; } = new BitmapImage(new Uri("ms-appx:///whitefs.png"));
```
pictos, line 34: I think this one can be `static` and used across all MediaElement instances

```suggestion
	static readonly ImageSource source = new BitmapImage(new Uri("ms-appx:///whitefs.png"));
```

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
```diff
@@ -9,9 +25,17 @@ namespace CommunityToolkit.Maui.Core.Views;
...
+	Button btn { get; set; }
+	Grid grid { get; set; } = new();
+	Grid ButtonContainer { get; set; }
+	Popup popup { get; set; } = new();
```
pictos, line 38: those can be fields, since they are private

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
vhugogarcia: The recommendation has been applied @pictos . Thanks for the suggestion. Marked as resolved.

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
vhugogarcia: The properties changed to fields as recommended. Thanks @pictos . Marked as resolved.

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
```diff
@@ -22,18 +28,106 @@ public class MauiMediaElement : CoordinatorLayout
...
+		}
+		else
 		{
-			LayoutParameters = new CoordinatorLayout.LayoutParams(LayoutParams.MatchParent, LayoutParams.MatchParent)
+			currentActivity.RequestedOrientation = Android.Content.PM.ScreenOrientation.Portrait;
+			
+			// Show again the SystemBars and Status bar
+			if (OperatingSystem.IsAndroidVersionAtLeast(30))
+			{
+				currentWindow.SetDecorFitsSystemWindows(true);
```
bijington, line 107: Should we be caching this value before we go into `IsFullScreen == true` so that we only reset it back if it needs to be?

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
```diff
@@ -22,18 +28,106 @@ public class MauiMediaElement : CoordinatorLayout
...
+			if (currentWindow.DecorView is FrameLayout layout)
+			{
+				RemoveView(playerView);
+				layout.AddView(playerView, new FrameLayout.LayoutParams(ViewGroup.LayoutParams.MatchParent, ViewGroup.LayoutParams.MatchParent));
+			}
+		}
+		else
 		{
-			LayoutParameters = new CoordinatorLayout.LayoutParams(LayoutParams.MatchParent, LayoutParams.MatchParent)
+			currentActivity.RequestedOrientation = Android.Content.PM.ScreenOrientation.Portrait;
```
bijington, line 102: Should we be caching this value before we go into `IsFullScreen == true` so that we only reset it back if it needs to be?

src/CommunityToolkit.Maui.MediaElement/CommunityToolkit.Maui.MediaElement.csproj
```diff
@@ -66,4 +66,7 @@
     <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiPackageVersion)"/>
     <PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiPackageVersion)"/>
   </ItemGroup>
+  <ItemGroup>
+    <MauiImage Include="Resources\whitefs.png" BaseSize="36,36" Resize="true" pack="true" PackagePath="." />
```
bijington, line 70: If we are resizing this would it be better as an SVG instead?

src/CommunityToolkit.Maui.MediaElement/Extensions/PageExtensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace CommunityToolkit.Maui.Extensions;
+
+// Since MediaElement can't access .NET MAUI internals we have to copy this code here
+// https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Platform/PageExtensions.cs
+static class PageExtensions
+{
+	internal static bool isFullScreen { get; set; } = false;
```
bijington, line 7: I don't think this is actually used anywhere so we could remove it? Or if we need to keep it we might want to Pascal case it

bijington (commented):
Just some potentially minor comments that you may just be able to ignore but I seem to recall we had some issues around Popups where it was messing with the system bar settings on Android

src/CommunityToolkit.Maui.MediaElement/Extensions/PageExtensions.cs
ne0rrmatrix: Yes I forgot to remove that

bijington:
@pictos and @ne0rrmatrix I couldn't get this into the Discord channel for some reason but this is my initial attempt at docs page for the background Android application exception. Let me know what you think

Screenshot 2024-02-15 at 21 19 33

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
vhugogarcia: Thanks @bijington for your feedback. This is something up to bring to discussion in later phases, because the expectation of the behavior of the Full screen on mobile for example is that if you are in portrait mode, the video goes to full screen in landscape mode and if they cancel the full screen then it goes back to portrait mode. 

However, if the app is on a tablet and you have it oriented in landscape mode, when you return back from full screen it will maintain the state, that is what happened in our results. This is how the YouTube app works for mobile and table.

I believe we can do this way right now, and once we start getting feedback we can learn and improve over the time if needed.

What do you think?

bijington: If the correct orientation is maintained even if we are requesting portrait then that is good. Strange but good :)

Out of interest have you tested the full screen mode in an application that doesn't support landscape orientation? Does that behave?

pictos (commented):
# Testing on Android:

## Screen doesn't follow the screen rotation:

During my tests, I rotated the device, but the video didn't adjust correctly. See image below:
image

src/CommunityToolkit.Maui.MediaElement/CommunityToolkit.Maui.MediaElement.csproj
vhugogarcia: Thanks @bijington for the feedback. The image has been replaced by the SVG version. Marked as resolved.

src/CommunityToolkit.Maui.MediaElement/Extensions/PageExtensions.cs
vhugogarcia: This has been removed. Marked as resolved.

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
vhugogarcia: Yes, it maintains the portrait mode. If you block your app on an orientation it will respect the orientation. We tested that particular case on an Android tablet.

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
```diff
@@ -9,9 +25,17 @@ namespace CommunityToolkit.Maui.Core.Views;
...
 public class MauiMediaElement : Grid, IDisposable
 {
+	bool navbarExist;
+	bool isDisposed;
+
+	static readonly AppWindow? appWindow = GetAppWindowForCurrentWindow();
 	readonly MediaPlayerElement mediaPlayerElement;
 
-	bool isDisposed;
+	static ImageSource source { get; set; } = new BitmapImage(new Uri("ms-appx:///fullscreen.png"));
```
pictos, line 34:
```suggestion
	static ImageSource source = new BitmapImage(new Uri("ms-appx:///fullscreen.png"));
```

pictos (approved):
For me, after the last request, it looks good to go!

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.windows.cs
vhugogarcia: Thanks @pictos for the recommendation. It has been applied. Marked as resolved.

vhugogarcia:
@pictos the last commit improves the rotation issue you were facing on Android. James and I ran some tests to confirm. Thanks again for testing.

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
ne0rrmatrix: With this code ``` urrentActivity.RequestedOrientation = Android.Content.PM.ScreenOrientation.Landscape;  ```
or ``` urrentActivity.RequestedOrientation = Android.Content.PM.ScreenOrientation.Portrait; ```
it appears once set it becomes locked. This is unexpected behavior and possibly a bug? Why we don't know. But auto rotation becomes disabled if either is set anywhere in code. I talked to @vhugogarcia and we decided to remove it to restore default rotation behavior back. 

I tested changes with removal of above and auto rotation now works as expected.

brminnick (changes requested):
Hey Gang! This is looking really good and I know many many devs are super excited for this 💯



On Android, the full-screen buttons overlap the built-in software nav buttons at the bottom. The Android bottom navigation button bar blocks me from tapping on the MediaElement full-screen buttons.

In this screen capture, I am trying to click the button on the bottom-right to close the full screen, but the Android navigation buttons on the bottom are blocking me from clicking it.

ne0rrmatrix:
> Hey Gang! This is looking really good and I know many many devs are super excited for this 💯
> 
> 
> On Android, the full-screen buttons overlap the built-in software nav buttons at the bottom. The Android bottom navigation button bar blocks me from tapping on the MediaElement full-screen buttons.
> 
> In this screen capture, I am trying to click the button on the bottom-right to close the full screen, but the Android navigation buttons on the bottom are blocking me from clicking it.
> 
> 👆 This screen capture is from a Android Pixel 7 emulator running Android API 29

Which API? 33, 34, or 30 and below? Or all?

vhugogarcia:
Thanks @brminnick for the feedback. We just committed a fix for the bottom navigation bar overlapping the player controls.

brminnick:
Thanks @vhugogarcia! I see you've updated the logic for Android 30+, but the [problem I pointed out in my comment above](https://github.com/CommunityToolkit/Maui/pull/1692#pullrequestreview-1886809515) is happening on Android API 29. 

Could you also update the logic for supporting Android API 29 (and below)?

## Reproduction

### Emulator

Android Pixel 7 emulator running Android API 29

### Steps
1. Launch `CommunityToolkit.Maui.Samples`
2. Navigate to Views -> MediaElement
3. In the MediaElement page, wait for video playback to bgin
4. Once the video has started, tap the Full Screen button on the video
5. In full screen mode, swipe up from the bottom of the emulator to reveal the software navigation bar
6. Confirm that the software navigation bar now blocks input to both the settings gear icon and the fullscreen button

### Screen Capture

Here's a screen capture demonstrating the bug. In this, I show the swipe-up gesture from the bottom that reveals the software navigation bar. The software navigation bar subsequently blocks the Settings icon and the Full Screen button.

ScreenFlow

vhugogarcia:
oh! sure thing @brminnick , I will work on it.

vhugogarcia:
@brminnick I just noticed that even with the fix applied, your particular scenario where you are in full screen, and then return back from background, the navigation bar sticks on the screen and it does not hide anymore, that is the reason of it being overlapping on that particular scenario.

I personally have no idea how to solve or workaround this particular scenario. we will review and see what we can do 👀 😃

brminnick:
Yup - I've experienced this error before. 

As the error indicates, you need to add an overloaded constructor to the Android implementation of CommunityToolkit.Maui.Core.Views.MauiMediaElement. 

> No constructor found for CommunityToolkit.Maui.Core.Views.MauiMediaElement::.ctor(System.IntPtr, Android.Runtime.JniHandleOwnership)"	

** MauiMediaElement.android.cs **

```cs
#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
#pragma warning disable IDE0060 // Remove unused parameter
	public MauiMediaElement(nint ptr, JniHandleOwnership jni) : base(Platform.AppContext)
	{
		//Fixes no constructor found exception
	}
#pragma warning restore CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
#pragma warning restore IDE0060 // Remove unused parameter
```

For example, here's how I implemented in my GitTrends app: 

https://github.com/brminnick/GitTrends/blob/e0162223f7402ac018bf4e0c69182c589433ce6f/GitTrends.Android/CustomRenderers/VideoPlayerViewCustomRenderer.cs#L37C1-L46C59

vhugogarcia:
Thanks @brminnick for sharing your solution! We just implemented it for the media element and it worked as expected, the app is not crashing anymore in Android. However, I noticed that once we are back from background the media element is not visible at all, this is because the dispose method was invoked. 

Any idea or thoughts or ideas on how to fix the issue when the app is returning from background? cc: @jfversluis @ne0rrmatrix 

image

vhugogarcia:
Thank you so much @ne0rrmatrix , that fixed the issue smoothly. 🚀 .

@brminnick the solution provided by James worked as expected, I just did a commit and push of the change, and it is ready for the final round of tests, let us know how it goes 👍🏻 .

Thanks all for the support and help 😃

bijington:
Great work guys! For the documentation, do we have any behavioral changes? If not then we might not need to add anything.

ne0rrmatrix:
> Great work guys! For the documentation, do we have any behavioral changes? If not then we might not need to add anything.

Yes the ```LaunchMode.SingleInstance``` should be referenced in docs because developers need to know that. The app might behave incorrectly unless that is set. We will get bug reports if users implement full screen and do not use that. It will crash the app otherwise.

bijington:
> > Great work guys! For the documentation, do we have any behavioral changes? If not then we might not need to add anything.
> 
> 
> 
> Yes the ```LaunchMode.SingleInstance``` should be referenced in docs because developers need to know that. The app might behave incorrectly unless that is set. We will get bug reports if users implement full screen and do not use that. It will crash the app otherwise.

Is that in addition to this change https://github.com/MicrosoftDocs/CommunityToolkit/pull/370 or is the crash you are referring to the one mentioned in that PR?

ne0rrmatrix:
> > > Great work guys! For the documentation, do we have any behavioral changes? If not then we might not need to add anything.
> > 
> > 
> > Yes the `LaunchMode.SingleInstance` should be referenced in docs because developers need to know that. The app might behave incorrectly unless that is set. We will get bug reports if users implement full screen and do not use that. It will crash the app otherwise.
> 
> Is that in addition to this change [MicrosoftDocs/CommunityToolkit#370](https://github.com/MicrosoftDocs/CommunityToolkit/pull/370) or is the crash you are referring to the one mentioned in that PR?

Yes it should be added. It is separate and is required. If you do not put that in it will not work when returning from background. It will result in bug reports.

bijington:
@ne0rrmatrix and @vhugogarcia I can add some detail to the docs PR that we have open. Do we have anything more specific than "The app might behave incorrectly unless that is set." is there a common way in which the application won't behave? Or is this statement sufficient?

ne0rrmatrix:
> @ne0rrmatrix and @vhugogarcia I can add some detail to the docs PR that we have open. Do we have anything more specific than "The app might behave incorrectly unless that is set." is there a common way in which the application won't behave? Or is this statement sufficient?

The app will crash if returning from background process if this is not set. I would phrase it differently but I lack the proper language to describe it.

vhugogarcia:
> @ne0rrmatrix and @vhugogarcia I can add some detail to the docs PR that we have open. Do we have anything more specific than "The app might behave incorrectly unless that is set." is there a common way in which the application won't behave? Or is this statement sufficient?

It can be also something like this wording:

> **Warning:** The app might behave incorrectly unless `LaunchMode.SingleInstance` is set in the main activity. This ensures that only one instance of the app is running at a time, and prevents the Media Element from being reinitialized or duplicated. To set the launch mode, add the following attribute to the `MainActivity` class in the `Platforms/Android`  folder.

ne0rrmatrix:
That sounds perfect. I feel that is a great explanation and resolves all issues I have. Thank you.

bijington:
@ne0rrmatrix and @vhugogarcia how does this look: https://github.com/MicrosoftDocs/CommunityToolkit/pull/370/commits/44bc8592862708ca7f9045e7f94d0e7cb739bc0a

vhugogarcia:
> @ne0rrmatrix and @vhugogarcia how does this look: [MicrosoftDocs/CommunityToolkit@44bc859](https://github.com/MicrosoftDocs/CommunityToolkit/commit/44bc8592862708ca7f9045e7f94d0e7cb739bc0a)

Looks great @bijington thanks for updating the docs 😃

ne0rrmatrix:
> @ne0rrmatrix and @vhugogarcia how does this look: [MicrosoftDocs/CommunityToolkit@44bc859](https://github.com/MicrosoftDocs/CommunityToolkit/commit/44bc8592862708ca7f9045e7f94d0e7cb739bc0a)

Looks great. I think that is written very well and provide the needed info to developers. Thank you for updating the docs :)

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
vhugogarcia: Thanks @bijington for the suggestion. It is now being cached.

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
```diff
@@ -13,6 +17,18 @@ namespace CommunityToolkit.Maui.Core.Views;
 public class MauiMediaElement : CoordinatorLayout
 {
 	readonly StyledPlayerView playerView;
+	int defaultSystemUiVisibility;
+	bool isSystemBarVisible;
+	bool isFullScreen;
+
+#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
```
bijington, line 24: Sorry if this has come up before but is this safe? We are introducing this constructor to avoid an exception being thrown but do we know why that constructor is being called on Android? I just want to make sure that we really want to ignore the `CS8618` warning here?

bijington: In fact line 82 checks against `null`, should we just make `playerView` nullable? e.g. `readonly StyledPlayerView? playerView;`

bijington (commented):
Thanks @vhugogarcia and @ne0rrmatrix. I just added a question around the nullable warning that we have suppressed but the rest of the changes look good to me

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
brminnick: @bijington Here's the background behind the additional constructor: https://github.com/CommunityToolkit/Maui/pull/1692#issuecomment-1955099758

I've experienced the same crash myself in my GitTrends app, and this workaround fixed it without any consequences. That said, I don't know 100% of the consequences that this could introduce. After we launch this update, we can keep an eye out for stack traces that include this constructor.

brminnick (approved):
Great work guys!! 

I'm good merging this, pending [@bijington's review comment on the additional constructor](https://github.com/CommunityToolkit/Maui/pull/1692#discussion_r1513396120).

src/CommunityToolkit.Maui.MediaElement/Views/MauiMediaElement.android.cs
pictos: I believe it's fine to ignore the nullable warnings here. This ctor will be called by `Xamarin.Android` only, so if it throws NRE, it will be a bug on their side

brminnick: Agreed. I think it's best to keep `playerView` non-nullable.

We can always circle back if we start seeing many NullReferenceExceptions being thrown. Those stack traces will also help us understand how `net8.0-android` is using this constructor.

ne0rrmatrix:
> Yup - I've experienced this error before.
> 
> As the error indicates, you need to add an overloaded constructor to the Android implementation of CommunityToolkit.Maui.Core.Views.MauiMediaElement.
> 
> > No constructor found for CommunityToolkit.Maui.Core.Views.MauiMediaElement::.ctor(System.IntPtr, Android.Runtime.JniHandleOwnership)"
> 
> **MauiMediaElement.android.cs**
> 
> ```cs
> #pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
> #pragma warning disable IDE0060 // Remove unused parameter
> 	public MauiMediaElement(nint ptr, JniHandleOwnership jni) : base(Platform.AppContext)
> 	{
> 		//Fixes no constructor found exception
> 	}
> #pragma warning restore CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
> #pragma warning restore IDE0060 // Remove unused parameter
> ```
> 
> For example, here's how I implemented in my GitTrends app:
> 
> https://github.com/brminnick/GitTrends/blob/e0162223f7402ac018bf4e0c69182c589433ce6f/GitTrends.Android/CustomRenderers/VideoPlayerViewCustomRenderer.cs#L37C1-L46C59

I can see why an empty constructor can happen here. But I though setting Single Instance prevented this? I believe constructor gets called a second time when returning from background which crashes the app. But with single instance set it will not call constructor. I use that specifically to solve issues like this. If it does not get called a second time which it should not be when returning from background this is is unnecessary? Or am I wrong?

brminnick:
> But I though setting Single Instance prevented this? I believe constructor gets called a second time when returning from background which crashes the app. But with single instance set it will not call constructor. I use that specifically to solve issues like this. If it does not get called a second time which it should not be when returning from background this is is unnecessary? Or am I wrong?

I wish I knew the answer to this. 

My vote is to keep the constructor; the code includes a link to the discussion we've had around it. We can always remove it in a future PR.

bijington (approved):

bijington:
Thanks all for the detail and effort. This looks great!

Issue #1693, opened 2024-02-13
[BUG] Visual studio Build Error on 17.8.7


ne0rrmatrix:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [X] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

Attempting to build Maui Community toolkit fails and multiple projects do not build.

### Expected Behavior

Building the project should work from visual studio

### Steps To Reproduce

1. Open the Sample Solution from Maui CommunityToolkit
2. Build the Solution.

### Link to public reproduction project repository

https://github.com/CommunityToolkit/Maui

### Environment

```markdown
- .NET MAUI CommunityToolkit: 7.0.1
- .Net Maui CommunityToolkit MediaElement: 3.0.1
- OS: Windows
- .NET MAUI: 8.0.6
```

### Anything else?

Visual studio version 17.8.7 shipped today and Maui was updated to 8.0.6 which requires updating the project files with updated nugets for Microsoft.Maui.Controls. Currently ```Version="$(MauiPackageVersion)"``` is not working and the workaround is to manually set ```Version="8.0.6"``` for Controls.

PR #1694, opened 2024-02-13
🔥 FIX - fix unit tests

pictos:
Match versions of `Microsoft.Maui.Controls` and `Microsoft.Maui.Controls.Compatibility`

Fixes https://github.com/CommunityToolkit/Maui/issues/1693

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,6 +21,7 @@ variables:
   PathToCommunityToolkitAnalyzersCodeFixCsproj: 'src/CommunityToolkit.Maui.Analyzers.CodeFixes/CommunityToolkit.Maui.Analyzers.CodeFixes.csproj'
   PathToCommunityToolkitMediaElementAnalyzersCodeFixCsproj: 'src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes.csproj'
   PathToCommunityToolkitAnalyzersUnitTestCsproj: 'src/CommunityToolkit.Maui.Analyzers.UnitTests/CommunityToolkit.Maui.Analyzers.UnitTests.csproj'
+  DotNetMauiRollbackFile: 'https://maui.blob.core.windows.net/metadata/rollbacks/8.0.6.json'
   CommunityToolkitSampleApp_Xcode_Version: '15.1.0'
   CommunityToolkitLibrary_Xcode_Version: '15.0.1'
 
@@ -74,7 +75,7 @@ jobs:
       - task: CmdLine@2
         displayName: 'Install Latest .NET MAUI Workload'
         inputs:
-          script: dotnet workload install maui --skip-sign-check --source https://api.nuget.org/v3/index.json 
+          script: dotnet workload install maui --from-rollback-file $(DotNetMauiRollbackFile) --source https://api.nuget.org/v3/index.json
 
       - pwsh: |
           Invoke-WebRequest 'https://raw.githubusercontent.com/Samsung/Tizen.NET/main/workload/scripts/workload-install.ps1' -OutFile 'workload-install.ps1'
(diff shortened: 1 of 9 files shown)


brminnick:
@JoonghyunCho Could you help us bump Tizen to support the latest version of .NET / MAUI?

pictos:
@JoonghyunCho from our logs

> Id: Samsung.NET.Sdk.Tizen.Manifest-8.0.200
An exception was caught: Response status code does not indicate success: 404 (The specified blob does not exist.).


Hope that helps to track the error

JoonghyunCho:
Didn't expect to use .NET 8.0.200 for MAUI `8.0.6`. Installing Tizen Workload step is passed properly.

brminnick (dismissed):

brminnick (dismissed):

brminnick (dismissed):

brminnick:
The problem is not in our code; it builds fine locally on my machine. 

The problem is that the pipeline is installing an old version of .NET MAUI. 
Screenshot 2024-02-15 at 12 38 25 AM

I cannot think of a reason why this is happening, other than there's a problem with the workload manifests being advertised for .NET 8.0.200. 

That being said, on my local machine, when I run `dotnet workload install maui`, the correct latest version of .NET MAUI is installed. I'm unable to reproduce the problem in the build pipeline locally.

StevilKnevil:
I've checked out and built this locally. I was able to reproduce the issue. I modified `Directory.Build.props` to have `<MauiPackageVersion>8.0.6</MauiPackageVersion>` and that seemed to resolve the issue for me

VladislavAntonyuk:
Could you please also update System.Speech nuget package to the stable version in CommunityToolkit.Maui.Core

azure-pipelines.yml
```diff
@@ -21,6 +21,7 @@ variables:
   PathToCommunityToolkitAnalyzersCodeFixCsproj: 'src/CommunityToolkit.Maui.Analyzers.CodeFixes/CommunityToolkit.Maui.Analyzers.CodeFixes.csproj'
   PathToCommunityToolkitMediaElementAnalyzersCodeFixCsproj: 'src/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes.csproj'
   PathToCommunityToolkitAnalyzersUnitTestCsproj: 'src/CommunityToolkit.Maui.Analyzers.UnitTests/CommunityToolkit.Maui.Analyzers.UnitTests.csproj'
+  DotNetMauiRollbackFile: 'https://maui.blob.core.windows.net/metadata/rollbacks/8.0.6.json'
```
brminnick, line 24: I don't love this workaround, but it's the only way I can for the hosted agent to install the latest version of .NET MAUI.

We'll hopefully be able to remove using the rollback file in a future release (or whenever this bug is fixed)

VladislavAntonyuk (approved):

brminnick:
> The problem is not in our code; it builds fine locally on my machine.
> 
> The problem is that the pipeline is installing an old version of .NET MAUI:
> 
> `Manifest Version: 8.0.0-rc.1.9171/8.0.100-rc.1`
> 
> Screenshot 2024-02-15 at 12 38 25 AM
> I cannot think of a reason why this is happening, other than there's a problem with the workload manifests being advertised for .NET 8.0.200.
> 
> That being said, on my local machine, when I run `dotnet workload install maui`, the correct latest version of .NET MAUI is installed. I'm unable to reproduce the problem in the build pipeline locally.

Ok -I figured out a work-around: We can first install maui using the roll-back file; this forces the workload manifest to update. Then, when we call `dotnet workload install maui`, it installs the latest version of .NET MAUI. 

```bash
dotnet workload install maui --from-rollback-file $(DotNetMauiRollbackFile) --source https://api.nuget.org/v3/index.json
dotnet workload install maui
```

PR #1699, opened 2024-02-17
Increase `Microsoft.Maui.*` Dependencies to v8.0.7

brminnick:
This PR increases the dependencies for the following NuGet Package to v8.0.7:
-  `Microsoft.Maui.Core`
- `Microsoft.Maui.Controls`
- `Microsoft.Maui.Controls.Compatibility`
- `Microsoft.Maui.Controls.Maps`
- `Microsoft.Maui.Essentials`

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
 <Project>
   <PropertyGroup>
     <NetVersion>net8.0</NetVersion>
-    <MauiPackageVersion>8.0.6</MauiPackageVersion>
+    <MauiPackageVersion>8.0.7</MauiPackageVersion>
     <LangVersion>latest</LangVersion>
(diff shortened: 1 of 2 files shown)


VladislavAntonyuk (dismissed):

VladislavAntonyuk:
Should we update azure-pipeline as well?

brminnick:
> Should we update azure-pipeline as well?

Which part of the pipeline? Our $(TOOLKIT_NET_VERSION)? 

I've been going back-and-forth about always targeting the latest version of .NET MAUI instead. Maybe we do that when we add .NET 9 support?

VladislavAntonyuk:
Sorry, I reviewed from phone and didn’t share the line. I meant MauiRollbackFile points to 8.0.6 instead of 8.0.7

PR #1758, opened 2024-03-17
Add CameraView

### LINKED ISSUE #259: [Proposal] CameraView minimal port
# [CameraView minimal port]

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
  * [ ] iOS Support
  * [ ] Android Support
  * [ ] macOS Support
  * [ ] Windows Support
* [ ] Unit Tests: Not Started
* [ ] Sample: Not Started

## Link to Discussion

- #106 
- #206

## Summary
[summary]: #summary

Today on XCT we have a very complete `CameraView` implementation but with a lot of bugs and a complex codebase to maintain. In the table below you can see all features implemented by each platform.

||Android|iOS|Windows|
|---|:---:|:---:|:---:|
|Preview|✅|✅|✅|
|Preview Aspect|✅|🚫|🚫|
|Mirror preview|✅|🚫|🚫|
|Zoom|✅|✅|✅|
|Take photo|✅|✅|✅|
|Flash|✅|✅|✅|
|Take video|✅|✅|✅|
|Record audio|✅|✅|✅|
|Video stabilization|✅|✅|✅|

My plan, described in discussion #206, is to drop some features for the v1.0 of this lib. and use `CameraX` implementation for Android. You can see in the table below what features we will have for v1.0.

||Android|iOS|Windows|
|---|:---:|:---:|:---:|
|Preview|✅|✅|✅|
|Take photo|✅|✅|✅|
|Flash|✅|✅|✅|

After this first implementation, we can add the missing features to all platforms.

## Motivation
[motivation]: #motivation

The motivation for this plan was the fact that `CameraView` lacks in quality right now and this's our chance to fix it and provide a more reliable Camera control for our community.

## Detailed Design
[design]: #detailed-design

### CameraView.shared.cs

```cs
public class CameraView : View
{
  public static readonly BindableProperty IsCameraViewBusyProperty;
  public static readonly BindableProperty IsAvailableProperty;
  public static readonly BindableProperty FlashModeProperty;
  
  // Use WeakEventManager 
  public event EventHandler<bool>? OnAvailable;
  public event EventHandler<MediaCapturedEventArgs>? MediaCaptured; 
  public event EventHandler<string>? MediaCaptureFailed;
  
  public bool IsCameraViewBusy { get; }
  public bool IsAvailable { get; }
  public CameraOptions CameraOptions { get; set; } 
  public void Shutter();
}
```

## Usage Syntax
[usage]: #usage-syntax

### XAML Usage

```xml
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             x:Class="MyLittleApp.MainPage">

     <StackLayout>
        <xct:CameraView
                x:Name="cameraView"
                FlashMode="On"
                MediaCaptured="CameraView_MediaCaptured"
                HorizontalOptions="FillAndExpand"
                VerticalOptions="FillAndExpand" />
    </StackLayout>
</ContentPage>
```

### C# Usage

```cs
class MyPage : ContentPage
{
  public MyPage()
  {
    Content = new CameraView
    {
      FlashMode = CameraFlashMode.Off,
    }.FillAndExpand()
     .Invoke(cameraView => cameraView.MediaCaptured += HandleMediaCaptured);
  }

  void HandleMediaCaptured(object? sender, MediaCapturedEventArgs e)
  {
    //...
  }
}
```

## Drawbacks
[drawbacks]: #drawbacks

We will (probably*) break our users on v1.0 since the API will be missing a lot of features.

* there's a chance that we can implement all features for v1.0 if we have a lot of community help.

## Alternatives
[alternatives]: #alternatives

- Follow the #106 proposal and just port the XCT code to here, but with that, we will bring all issues (if we don't solve them during the port, what I believe it's very hard to do)

## Unresolved Questions
[unresolved]: #unresolved-questions

zhitaop:
### Description of Change ###
As described in https://github.com/CommunityToolkit/Maui/issues/259#issuecomment-1984316965, this PR adds the initial  implementation of the CameraView proposal on MacCatalyst, Windows, Android and iOS, and a corresponding demo page in the sample app . It's based on the fantastic work from the following PR/ branch https://github.com/CommunityToolkit/Maui/pull/1080 and https://github.com/CommunityToolkit/Maui/pull/1387.
Below is the summary of the proposed API design, it has been updated with feedback from  https://github.com/CommunityToolkit/Maui/issues/259#issuecomment-1984316965
```c#
public class CameraView
{
    public static readonly BindableProperty SelectedCameraProperty; // To be discussed: ActiveCameraProperty?
    public static readonly BindableProperty ZoomFactorProperty;
    public static readonly BindableProperty ResolutionProperty;
    public static readonly BindableProperty IsCameraBusyProperty;
    public static readonly BindableProperty IsAvailableProperty;
    public static readonly BindableProperty FlashModeProperty;
    public static readonly BindableProperty IsTorchOnProperty;

    public event EventHandler<MediaCapturedEventArgs> MediaCaptured;
    public event EventHandler<MediaCaptureFailedEventArgs> MediaCaptureFailed;

    public CameraInfo? SelectedCamera { get; set; } // To be discussed: ActiveCamera?
    public float ZoomFactor { get; set; }
    public Size CaptureResolution { get; set; } //  renamed from `Resolution`
    public bool IsCameraBusy { get; set; }
    public bool IsAvailable { get; set; }
    public bool IsTorchOn { get; set; }
    public CameraFlashMode FlashMode { get; set; }

    public void Shutter()
    public void Start()
    public void Stop()
}

public enum CameraFlashMode
{
    Off,
    On,
    Auto
}

public class CameraInfo
{
    public string Name { get; }
    public string DeviceId { get; }
    public CameraPosition Position { get; }
    public bool IsFlashSupported { get; } //To be discussed:  adding a list of capabilities like Auto Focus, etc?
    public float MinimumZoomFactor { get; }
    public float MaximumZoomFactor { get; }
    public ObservableCollection<Size> SupportedResolutions { get; } // renamed from `AvailableResolutions`. 
    //Observable makes it easier for binding as the resolution would only be updated after the camera view is initialized on Windows 
}

public enum CameraPosition // To be discussed: adding External value
{
    Unknown,
    Rear,
    Front
}

public class CameraProvider
{
    public ObservableCollection<CameraInfo> AvailableCameras { get; }
    public void RefreshAvailableCameras();
}
```

 ### Linked Issues ###
 

 - Fixes #259
 - Closes #1080 
 - Closes #1387


src/CommunityToolkit.Maui.CameraView/CommunityToolkit.Maui.CameraView.csproj
```diff
@@ -0,0 +1,77 @@
...
+    <!--Adding the above package reference requires Workaround for https://github.com/xamarin/AndroidX/issues/800#issuecomment-1774639627 -->
+    <!-- Tracking issue https://github.com/xamarin/AndroidX/issues/764 -->
+    <PackageReference Include="Xamarin.AndroidX.Collection" Version="1.4.0.1" />
+    <PackageReference Include="Xamarin.AndroidX.Collection.Ktx" Version="1.4.0.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiPackageVersion)"/>
+    <PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiPackageVersion)"/>
+    <PackageReference Include="CommunityToolkit.Mvvm" Version="8.2.2" />
```
bijington, line 74: I don't think we should be bringing in a dependency on the MVVM toolkit here.

src/CommunityToolkit.Maui.CameraView/CameraInfo.cs
```diff
@@ -0,0 +1,80 @@
...
+#elif ANDROID
+using AndroidX.Camera.Core;
+#elif WINDOWS
+using Windows.Media.Capture.Frames;
+using Windows.Media.MediaProperties;
+#endif
+
+namespace CommunityToolkit.Maui.Core;
+
+public partial class CameraInfo : ObservableObject
```
bijington, line 16: As mentioned in the project file, I don't think we should be bringing in a dependency on the MVVM Toolkit.

Out of interest though, does this info really need to change? I expect that this could be immutable as it represents a physical camera?

src/CommunityToolkit.Maui.CameraView/CameraView.shared.cs
```diff
@@ -0,0 +1,146 @@
...
+
+    public CameraFlashMode CameraFlashMode
+    {
+        get => (CameraFlashMode)GetValue(CameraFlashModeProperty);
+        set => SetValue(CameraFlashModeProperty, value);
+    }
+
+    public CameraInfo? SelectedCamera
+    {
+        get { return (CameraInfo?)GetValue(SelectedCameraProperty); }
```
bijington, line 44: This can be an expression body to match the other properties.

```suggestion
        get => (CameraInfo?)GetValue(SelectedCameraProperty);
```

src/CommunityToolkit.Maui.CameraView/CameraView.shared.cs
```diff
@@ -0,0 +1,146 @@
...
+        set { SetValue(SelectedCameraProperty, value); }
```
bijington, line 45: This can be an expression body to match the other properties.

```suggestion
        set => SetValue(SelectedCameraProperty, value);
```

src/CommunityToolkit.Maui.CameraView/CameraView.shared.cs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    public float ZoomFactor
+    {
+        get { return (float)GetValue(ZoomFactorProperty); }
+        set { SetValue(ZoomFactorProperty, value); }
```
bijington, line 51: This can be an expression body to match the other properties.

```suggestion
        get => (float)GetValue(ZoomFactorProperty);
        set => SetValue(ZoomFactorProperty, value);
```

src/CommunityToolkit.Maui.CameraView/CameraView.shared.cs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    public Size CaptureResolution
+    {
+        get { return (Size)GetValue(CaptureResolutionProperty); }
+        set { SetValue(CaptureResolutionProperty, value); }
```
bijington, line 57: This can be an expression body to match the other properties.

```suggestion
        get => (Size)GetValue(CaptureResolutionProperty);
        set => SetValue(CaptureResolutionProperty, value);
```

src/CommunityToolkit.Maui.CameraView/CameraView.shared.cs
```diff
@@ -0,0 +1,146 @@
...
+        remove => weakEventManager.RemoveEventHandler(value);
+    }
+
+    public event EventHandler<MediaCapturedEventArgs> MediaCaptured
+    {
+        add => weakEventManager.AddEventHandler(value);
+        remove => weakEventManager.RemoveEventHandler(value);
+    }
+
+    public void OnAvailable()
```
bijington, line 117: Do we need this empty method?

src/CommunityToolkit.Maui.CameraView/CameraView.shared.cs
```diff
@@ -0,0 +1,146 @@
+﻿using System.ComponentModel;
+using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.Core.Primitives;
+
+namespace CommunityToolkit.Maui.Views;
+
+public partial class CameraView : View, ICameraView
+{
+    static readonly WeakEventManager weakEventManager = new();
```
bijington, line 9: Should this really be `static`?

src/CommunityToolkit.Maui.CameraView/Handlers/CameraViewHandler.cs
```diff
@@ -0,0 +1,127 @@
...
+
+using Microsoft.Maui.Handlers;
+using CommunityToolkit.Maui.Core.Views;
+using CommunityToolkit.Maui.Extensions;
+
+namespace CommunityToolkit.Maui.Core.Handlers;
+
+public partial class CameraViewHandler : ViewHandler<ICameraView, NativePlatformCameraPreviewView>, IDisposable
+{
+    public static Action<byte[]>? Picture { get; set; }
```
bijington, line 11: Is this used?

src/CommunityToolkit.Maui.CameraView/Handlers/CameraViewHandler.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            MapCameraFlashMode(this, view);
+            view.ZoomFactor = 1.0f;
+        }
+    }
+
+    protected override async void ConnectHandler(NativePlatformCameraPreviewView platformView)
+    {
+        base.ConnectHandler(platformView);
+        await cameraManager!.CheckPermissions();
```
bijington, line 98: As a general rule we want to avoid use of the `!` operator.

src/CommunityToolkit.Maui.CameraView/Interfaces/ICameraView.cs
```diff
@@ -0,0 +1,20 @@
...
+    Size CaptureResolution { get; set; }
+    bool IsTorchOn { get; }
+    void Shutter();
+    void OnAvailable();
+    void OnMediaCaptured(Stream imageData);
+    void Start();
+    void Stop();
+
+    // TODO: Create a custom Exception type and pass as parameter providing more info about the error
+    void OnMediaCapturedFailed();
```
bijington, line 19: How does this tie in with the `MediaCaptureFailedEvent`? Should it?

src/CommunityToolkit.Maui.CameraView/Primitives/CameraFlashMode.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace CommunityToolkit.Maui.Core.Primitives;
+
+public enum CameraFlashMode
+{
+	Off,
+	On,
+	Auto,
+	// Create a BP boolean for torch
+	//Torch
```
bijington, line 9: I think this can be removed?

src/CommunityToolkit.Maui.CameraView/Providers/CameraProvider.android.cs
```diff
@@ -0,0 +1,92 @@
+﻿using System.Diagnostics;
+using Java.Lang;
+using Android.Content;
+using Android.Graphics;
+using Android.Hardware.Camera2;
+using Android.Hardware.Camera2.Params;
+using AndroidX.Core.Content;
+using AndroidX.Camera.Core;
+using AndroidX.Camera.Camera2.InterOp;
```
bijington, line 9: Is it right that we are referring to Camera2 here? I think we wanted to migrate to CameraX where possible

src/CommunityToolkit.Maui.CameraView/CameraInfo.cs
```diff
@@ -0,0 +1,80 @@
...
+{
+    public string? Name { get; internal set; }
+    public string? DeviceId { get; internal set; }
+    public CameraPosition? Position { get; internal set; }
```
bijington, line 20: Can these values ever be null? If we are setting them perhaps we can use something other than null and remove the `?`s?

src/CommunityToolkit.Maui.CameraView/Views/CameraManager.ios.cs
```diff
@@ -0,0 +1,273 @@
...
+namespace CommunityToolkit.Maui.Core.Views;
+
+public partial class CameraManager
+{
+    AVCaptureSession? captureSession;
+    AVCapturePhotoOutput? photoOutput;
+    AVCaptureInput? captureInput;
+    AVCaptureDevice? captureDevice;
+
+    // TODO: Check if we really need this
```
bijington, line 19: Do we really need this? :)

src/CommunityToolkit.Maui.CameraView/Views/CameraManager.ios.cs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    public partial void UpdateZoom(float zoomLevel)
+    {
+        if (!Initialized || captureDevice is null)
+        {
+            return;
+        }
+
+        if (zoomLevel < (float)captureDevice.MinAvailableVideoZoomFactor || zoomLevel > (float)captureDevice.MaxAvailableVideoZoomFactor)
```
bijington, line 145: I think this could be rewritten as:

```suggestion
        if (zoomLevel is < (float)captureDevice.MinAvailableVideoZoomFactor or  > (float)captureDevice.MaxAvailableVideoZoomFactor)
```

which might make it a tiny bit easier to read.

src/CommunityToolkit.Maui.CameraView/Views/CameraManager.shared.cs
```diff
@@ -0,0 +1,93 @@
...
+    internal bool Initialized { get; private set; } = false;
+
+    internal Action? Loaded { get; set; }
+
+    public CameraManager(IMauiContext mauiContext, ICameraView cameraView)
+    {
+        this.mauiContext = mauiContext;
+        this.cameraView = cameraView;
+        this.cameraProvider = IPlatformApplication.Current?.Services.GetService<CameraProvider>() ?? throw new NullReferenceException();
+		this.currentCamera = cameraView.SelectedCamera ??= cameraProvider.AvailableCameras.First();
```
bijington, line 22: Is this safe? Could there be an occasion where there are not any `AvailableCameras`?

bijington (changes requested):
@zhitaop thank you so much for this PR! I have added my initial comments and most I don't think require much effort if they do need to be changed. 

I plan to try and test some bits later this week. I would also love to look at possible macOS support which I am happy to contribute, I wouldn't let this hold up the PR completing though, worst case I can always open a follow up PR to add that in.

src/CommunityToolkit.Maui.CameraView/CommunityToolkit.Maui.CameraView.csproj
jfversluis: Agreed, if we actually need the property values to change lets do it manually, lets not take an extra dependency here

jfversluis:
I added the Analyzers for the initialization code for you!

src/CommunityToolkit.Maui.CameraView/Views/CameraManager.shared.cs
zhitaop: Added exception handling

src/CommunityToolkit.Maui.CameraView/CameraInfo.cs
zhitaop: Removed nullable

src/CommunityToolkit.Maui.CameraView/Providers/CameraProvider.android.cs
zhitaop: Some camera characteristics are only supported in Camera2, such as supported resolutions.

src/CommunityToolkit.Maui.CameraView/Views/CameraManager.ios.cs
zhitaop: It seems like the [right-hand part of a relational pattern must be a constant expression](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#relational-patterns), so this syntax is not supported here.

src/CommunityToolkit.Maui.CameraView/CameraInfo.cs
zhitaop: The reason for making it observable is on Windows some camera properties such as zoom factors can only be queried after initializing a `MediaCapture` object, which is done in the `CameraManager` after starting the view. Doing so in `CameraProvider` where the `CameraInfo` is initialized is not desirable as this class could be used without starting the CameraView.

bijington:
@zhitaop Thanks for the responses. Sorry things have been chaotic here so I haven't responded or reviewed yet. I'll try and take a look this week

bijington (dismissed):
@zhitaop Thanks for your responses, I have added some further comments. I think the majority looks great and we could be close to getting this ready to merge. Let me know your thoughts and whether you need any assistance

samples/CommunityToolkit.Maui.Sample/Platforms/Android/AndroidManifest.xml
```diff
@@ -10,6 +10,7 @@
   <uses-permission android:name="com.sec.android.provider.badge.permission.WRITE" />
 
 	<uses-permission android:name="android.permission.RECORD_AUDIO" />
+	<uses-permission android:name="android.permission.CAMERA" />
```
brminnick, line 13: Reminder that we'll need to add this to the `Platform specific initialization` section of the `CameraView` docs

samples/CommunityToolkit.Maui.Sample/Platforms/MacCatalyst/Info.plist
```diff
@@ -30,5 +30,7 @@
 	<string>ISpeechToText Uses Speech Recognition</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>ISpeechToText Uses Microphone</string>
+	<key>NSCameraUsageDescription</key>
+	<string>ICameraView Uses Camera</string>
```
brminnick, line 34: Reminder that we'll need to add this to the `Platform specific initialization` section of the `CameraView` docs

samples/CommunityToolkit.Maui.Sample/Platforms/iOS/Info.plist
```diff
@@ -34,5 +34,7 @@
 	<string>ISpeechToText Uses Speech Recognition</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>ISpeechToText Uses Microphone</string>
+	<key>NSCameraUsageDescription</key>
+	<string>ICameraView Uses Camera</string>
```
brminnick, line 38: Reminder that we'll need to add this to the `Platform specific initialization` section of the `CameraView` docs

brminnick:
We're getting closer! I was able to get `CameraView` working on macOS. I've updated the title + description of this PR to include MacCatalyst 👍

image

The one thing I can't confirm is CameraView on Windows; I'm running Windows on a VM on my MacBook, and each time I run the app I get a `CameraViewException` telling me that it cannot find a camera. I'm unsure if it's my setup that is causing the problem, or if a refactor edit I made accidentally injected the bug. But, once we get Windows working and the Docs written, we're ready to ship!

src/CommunityToolkit.Maui.CameraView/CommunityToolkit.Maui.CameraView.csproj
```diff
@@ -0,0 +1,82 @@
...
+    <SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">21.0</SupportedOSPlatformVersion>
+    <SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.17763.0</SupportedOSPlatformVersion>
+    <TargetPlatformMinVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.17763.0</TargetPlatformMinVersion>
+    <RootNamespace>CommunityToolkit.Maui</RootNamespace>
+    <AndroidGenerateResourceDesigner>false</AndroidGenerateResourceDesigner>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <GitInfoReportImportance>high</GitInfoReportImportance>
+    <PackageId>CommunityToolkit.Maui.CameraView</PackageId>
```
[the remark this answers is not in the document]
brminnick: I vote for `CommunityToolkit.Maui.Camera` since it gives us options to add more features to this NuGet Package in the future.

That being said, I'm also cool with `CommunityToolkit.Maui.CameraView`.

zhitaop:
@bijington @brminnick Thanks a lot for the review and help with the support on macOS! Sorry I haven't got the chance to look at the comments earlier, but I guess most of them are addressed in the latest changes?

@brminnick Regarding the exception on Windows, recently we found a bug where the CameraView doesn't work on certain Windows machine. It could be fixed by changing the line https://github.com/CommunityToolkit/Maui/blob/45978391ff250dcd7055aa6891cf0c6d332ff6da/src/CommunityToolkit.Maui.CameraView/CameraManager.windows.cs#L129 to `PhotoCaptureSource = PhotoCaptureSource.Auto`.

I'm not sure if that's the same bug that you encountered, but please feel free to give a try.

bijington:
Yes the CameraView will ship in a new nuget package that doesn't exist yet. @jfversluis i assume this means we can ship a previous nuget for it right?

If not then you can test it out in the sample application for now

jfversluis:
If with previous you mean preview then yes 😄 I will have a look at getting this across the finish line in the upcoming week!

bijington:
Haha doh!

bijington:
> @bijington @brminnick Thanks a lot for the review and help with the support on macOS! Sorry I haven't got the chance to look at the comments earlier, but I guess most of them are addressed in the latest changes?
> 
> @brminnick Regarding the exception on Windows, recently we found a bug where the CameraView doesn't work on certain Windows machine. It could be fixed by changing the line
> 
> https://github.com/CommunityToolkit/Maui/blob/45978391ff250dcd7055aa6891cf0c6d332ff6da/src/CommunityToolkit.Maui.CameraView/CameraManager.windows.cs#L129
> 
> to `PhotoCaptureSource = PhotoCaptureSource.Auto`.
> I'm not sure if that's the same bug that you encountered, but please feel free to give a try.

Thanks for the update. I wonder if there is value in just swapping to using `Auto` or allowing the developing to choose this through an options class when registering the CameraView plugin. Something like:

```csharp
    builder
    .UseMauiApp<App>()
    // Initialize the .NET MAUI Community Toolkit CameraView by adding the below line of code
    .UseMauiCommunityToolkitCameraView(options =>
    {
#if WINDOWS
        options.PhotoCaptureSource = PhotoCaptureSource.Auto;
#elif
    });
```

Thoughts?

bijington:
From reading this https://learn.microsoft.com/en-us/uwp/api/windows.media.capture.photocapturesource?view=winrt-18362 I think we should just go with `Auto`

samples/CommunityToolkit.Maui.Sample/ViewModels/Views/CameraView/CameraViewModel.cs
```diff
@@ -0,0 +1,96 @@
...
+
+	partial void OnSelectedResolutionChanged(Size value)
+	{
+		UpdateResolutionText();
+	}
+
+	[RelayCommand]
+	async Task CaptureImage(CameraView cameraView, CancellationToken token)
+	{
+		await cameraView.CaptureImage(token);
```
bijington, line 47: This currently breaks MVVM, is this fine for a v1 or do we want to consider exposing `Command`s on the `CameraView` to support the MVVM architecture more cleanly?

brminnick: > breaks MVVM

Because we calling the View in the ViewModel? Agreed. 

Would the solution be to add a `public ICommand CaptureImageCommand { get; set; }`?

jfversluis:
@MFinkBK there is now a CameraView NuGet package in the artifacts to try out! Please let us know how it goes! Do keep in mind that things are not finalized until they are merged so expect (breaking) changes to happen.

@bijington I disabled the Tizen target for now on CameraView because it was giving build errors. I'm fine with adding a dummy implementation (or @JoonghyunCho a real implementation...?) but we'll need to make sure it builds! :) uncomment the `TargetFrameworks` node for Tizen in the CameraView csproj to enable it again.

JoonghyunCho:
not sure if Tizen can support CameraView APIs. Let me check in the coming week.

JoonghyunCho:
I think we should put dummy implementation or set as not supported for Tizen... I don't see any way to test or verify the implementation... 😭

samples/CommunityToolkit.Maui.Sample/ViewModels/Views/CameraView/CameraViewModel.cs
bijington: Yes along those lines but it wants to be readonly. I will add those now

bijington: I've added them in now. @brminnick would you be willing to review that change?

JoonghyunCho:
omg 😮  I will add Tizen fix

JoonghyunCho:
Looks like you made it builds.

bijington:
Docs PR is at: https://github.com/MicrosoftDocs/CommunityToolkit/pull/428

src/CommunityToolkit.Maui.Camera/CameraManager.macios.cs
```diff
@@ -0,0 +1,299 @@
...
+		previewView.Session = captureSession;
+
+		return previewView;
+	}
+
+	/// <inheritdoc/>
+	public void Dispose()
+	{
+		Dispose(true);
+		GC.SuppressFinalize(this);
```
pictos, line 43: No need for this call
```suggestion
```

src/CommunityToolkit.Maui.Camera/CameraManager.macios.cs
```diff
@@ -0,0 +1,299 @@
...
+		if (zoomLevel < (float)captureDevice.MinAvailableVideoZoomFactor || zoomLevel > (float)captureDevice.MaxAvailableVideoZoomFactor)
+		{
+			return;
+		}
+
+		captureDevice.LockForConfiguration(out NSError error);
+		if (error is not null)
+		{
+			Console.WriteLine(error);
+			Debug.WriteLine(error);
```
pictos, line 67: Just `Trace` is presented in `Release` mode
```suggestion
			System.Diagnostics.Trace.WriteLine(error);
```

src/CommunityToolkit.Maui.Camera/CameraManager.macios.cs
```diff
@@ -0,0 +1,299 @@
...
+		}
+
+		if (cameraView.SelectedCamera is null)
+		{
+			await cameraProvider.RefreshAvailableCameras(token);
+			cameraView.SelectedCamera = cameraProvider.AvailableCameras?.FirstOrDefault() ?? throw new CameraViewException("No camera available on device");
+		}
+
+		captureDevice = cameraView.SelectedCamera.CaptureDevice ?? throw new CameraViewException($"No Camera found");
+		captureInput = new AVCaptureDeviceInput(captureDevice, out var err);
```
pictos, line 153: we declare the `err` but never use it, is that intentional?

src/CommunityToolkit.Maui.Camera/CameraManager.macios.cs
```diff
@@ -0,0 +1,299 @@
...
+		var data = result.Photo.FileDataRepresentation;
+
+		if (data is null)
+		{
+			// TODO: Pass NSError information
+			cameraView.OnMediaCapturedFailed();
+			return;
+		}
+
+		var dataBytes = new byte[data.Length];
```
pictos, line 210: Good usage of `ArrayPool`

src/CommunityToolkit.Maui.Camera/CameraManager.windows.cs
```diff
@@ -0,0 +1,209 @@
...
+	public MediaPlayerElement CreatePlatformView()
+	{
+		mediaElement = new MediaPlayerElement();
+		return mediaElement;
+	}
+
+	public void Dispose()
+	{
+		Dispose(true);
+		GC.SuppressFinalize(this);
```
pictos, line 29: No need for this call, as mentioned before
```suggestion
```

src/CommunityToolkit.Maui.Camera/Providers/CameraProvider.macios.cs
```diff
@@ -0,0 +1,93 @@
...
+using CoreVideo;
+using UIKit;
+
+namespace CommunityToolkit.Maui.Core;
+
+partial class CameraProvider
+{
+	public partial ValueTask RefreshAvailableCameras(CancellationToken token)
+	{
+		AVCaptureDeviceType[] deviceTypes =
```
pictos, line 13: this can be a static readonly collection, to avoid extra allocations

brminnick (dismissed):
Ok, gang! I think this PR is ready to merge! 🚀

I've added the `needs discussion` label to ensure we review it in our monthly standup this week 👍

pictos (commented):
This is a solid work! I've some code comments and some general points:

1. Some places `ValueTask` is used and in other places `Task` are used, even if there's a  chance to that method completes synchronously, is that intentional?

2. Methods that have a `CancellationToken` as parameter and are visible outside the lib (protected, public, etc)  should be optional or have an overload without them, to follow the BCL pattern.

3. I see a lot of exceptions that could be thrown on Manager and other types, how devs can treat them without breaking the application? For example, an exception is thrown if there’s no camera available on device, should we break the app (since there’s no way to easily try/catch those exceptions) if there’s no camera? Also, in most cases the dev can’t predict which device the app will run.

MFinkBK:
@jfversluis Thanks, the NuGet package is now there. What I will definitely need (but can be a follow-up issue and PR) is CameraPosition.External, e.g. for USB webcams on Windows.

src/CommunityToolkit.Maui.Camera/CameraManager.android.cs
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	protected virtual partial void PlatformDisconnect()
+	{
+
+	}
+
+	protected virtual partial ValueTask PlatformTakePicture(CancellationToken token)
+	{
+		imageCapture?.TakePicture(cameraExecutor!, imageCallback!);
```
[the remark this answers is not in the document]
brminnick: Ah- good catch! We'll certainly remove that 💯

Is there an analyzer or `.editorconfig` we can add to avoid this?

src/CommunityToolkit.Maui.Camera/CameraManager.android.cs
```diff
@@ -0,0 +1,353 @@
...
+
+			var buffer = GetFirstPlane(img.GetPlanes())?.Buffer;
+
+			if (buffer is null)
+			{
+				image.Close();
+				return;
+			}
+
+			var imgData = new byte[buffer.Capacity()];
```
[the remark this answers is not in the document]
brminnick: Good call! 

I haven't used `ArrayPool` myself. How do we implement it here?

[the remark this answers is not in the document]
pictos: Not sure who is working on it, so is it fine if I push the changes?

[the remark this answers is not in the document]
brminnick: > s it fine if I push the changes?

Of course - go for it! Thanks man!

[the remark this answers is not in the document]
pictos: Done

src/CommunityToolkit.Maui.Camera/CameraManager.android.cs
```diff
@@ -0,0 +1,353 @@
...
+	ICamera? camera;
+	ICameraControl? cameraControl;
+	Preview? cameraPreview;
+	ResolutionSelector? resolutionSelector;
+	ResolutionFilter? resolutionFilter;
+
+	public void Dispose()
+	{
+		Dispose(true);
+		GC.SuppressFinalize(this);
```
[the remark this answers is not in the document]
brminnick: It's cool to keep it. 

It'll ensure we don't accidentally create a bug in a future PR that implements a finalizer.

src/CommunityToolkit.Maui.Camera/CameraManager.windows.cs
brminnick: It's cool to keep it. It'll save us a bug in the future if a PR adds a finalizer and/or a dev subclasses this and adds a finalizer.

src/CommunityToolkit.Maui.Camera/Providers/CameraProvider.macios.cs
brminnick: Done ✅

src/CommunityToolkit.Maui.Camera/CameraManager.android.cs
```diff
@@ -0,0 +1,353 @@
...
+
+			resolutionSelector?.Dispose();
+			resolutionSelector = null;
+
+			resolutionFilter?.Dispose();
+			resolutionFilter = null;
+		}
+	}
+
+	protected virtual async partial Task PlatformConnectCamera(CancellationToken token)
```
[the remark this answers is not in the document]
brminnick: Let's punt this to a future PR because we have many methods that return `ValueTask` and `Task` that also accept a `CancellationToken` without a default parameter. 

I agree with you. But let's keep the codebase consistent for this PR, and in a future PR we can add the default parameter everywhere.

src/CommunityToolkit.Maui.Camera/CameraManager.android.cs
```diff
@@ -0,0 +1,353 @@
...
+	{
+		var cameraProviderFuture = ProcessCameraProvider.GetInstance(context);
+		if (previewView is null)
+		{
+			return;
+		}
+
+		var cameraProviderTCS = new TaskCompletionSource();
```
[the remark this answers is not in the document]
brminnick: Good catch! I've implemented this timeout, but in a slightly different way: 
1. I've gone through the codebase, and everywhere where we call this method, I ensure we don't pass in `CancellationToken.None`
2. I've updated the return statement to also await the passed-in `CancellationToken`: 
   ```cs
    await cameraProviderTCS.Task.WaitAsync(token);
    ```

This method is currently only called by the CameraViewHandler. Here's how I updated the handler:

```cs
	protected override async void ConnectHandler(NativePlatformCameraPreviewView platformView)
	{
		base.ConnectHandler(platformView);

		var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(3));

		await (cameraManager?.ArePermissionsGranted() ?? Task.CompletedTask);
		await (cameraManager?.ConnectCamera(cancellationTokenSource.Token) ?? Task.CompletedTask);
		await cameraProvider.RefreshAvailableCameras(cancellationTokenSource.Token);
	}
```

src/CommunityToolkit.Maui.Camera/CameraManager.macios.cs
brminnick: Changed to `_` ✅

src/CommunityToolkit.Maui.Camera/CameraManager.android.cs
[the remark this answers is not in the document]
pictos: @brminnick but this code can be called by users, I know isn't productive to protect devs from themself, but in this case having this inside the method implementation is safer, even for future changes on this lib.

Also, this method is visible for external assemblies and forces the usage of `CancellationToken` which isn't the default pattern from BCL. Is it intentional?

brminnick:
Thanks @pictos - great review! I've incorporated your comments here: https://github.com/CommunityToolkit/Maui/pull/1758/commits/0f85de2244c95678f6477b6a60a81781aa1f9cee

> I see a lot of exceptions that could be thrown on Manage
I went through every branch where we `throw new CameraException()`, and the only time we do that now when no cameras are available on the device  is when the user does something that requires the camera. For example, when the dev calls `StartCameraPreview()` or `UpdateCaptureResolution()` (methods that require a camera), we'll throw `CameraException`. 

This should ensure devs are able to catch the `CameraException` now.

src/CommunityToolkit.Maui.Camera/CameraManager.android.cs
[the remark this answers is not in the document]
brminnick: > forces the usage of CancellationToken.... Is it intentional?

Yup - all of our public APIs that return `Task` and `ValueTask` in this library require a CancellationToken parameter. We did this initiative a few months ago: https://github.com/CommunityToolkit/Maui/pull/1503

I'm totally cool with you opening a new PR that adds the default parameter of `CancellationToken.None`. We'd need to add it on every existing public API that returns `Task` + `ValueTask`, and I'd rather not muddle up this 3,000-line PR any more than it needs.

> but this code can be called by users, I know isn't productive to protect devs from themself, but in this case having this inside the method implementation is safer, even for future changes on this lib

You're correct. And we may see an Issue opened where we have to tell the dev to pass in a `CancellationToken`. But for now, let's wrap up this PR and let devs start using CameraView v1.0. If the community starts asking for an automatic timeout, we can always implement it later. It's one of those things that is easier to implement later than remove after implementing it now.

brminnick (approved):
Ok gang! I think this is ready to merge. 🚀

That being said, Pedro brought up some good comments in his review that we should discuss first! Let's chat about it in tomorrow's standup to get a group consensus before we click `merge` 🙌

zhitaop:
@brminnick @bijington  Thanks for all the great work! There has been definitely great progress and the code looks much more organized and consistent to the rest of the toolkit now. 
After some testing on Windows using the sample app, I did notice a few issues that might worth looking into:
1. Due to the access level of `CameraProvider` , there is no way to access the `AvaiableCameras` other than `CameraView.GetAvailableCameras`. My original intent is for `CameraProvider` to be a public singleton class decoupled from `CameraView` so that it can be used to access the available `CameraInfo` without a CameraView instance. Potential use case includes selecting the initial camera to open before initializing the CameraView. Both Android and iOS use a similar pattern in their native camera APIs. 
2. The camera switcher and some of the original debug text (camera name, zoom level, etc) are missing on the sample camera view page
3.  `UpdateCameraInfo` in `CameraManage.Windows.cs` is not needed anymore, since now `CameraInfo` is fully initialized in `CameraProvider.Windows.cs`  and does not need to be updated.

I haven't got the chance to look into these issues myself, so just leave them here as discussion points before the merge!

brminnick (approved):
I think we're good to go! Thanks @zhitaop!!!

FYI - We discussed your latest review in our [standup on Thursday](https://www.youtube.com/watch?v=ImFHkInnCpg). Here's a quick rundown:
> 1. Due to the access level of CameraProvider , there is no way to access the AvaiableCameras other than CameraView.GetAvailableCameras. 

The interface `ICameraProvider` is public and we add it to the `IServiceCollection`; it can be accessed by developers using Dependency Injection or `IPlatformApplication.Current.Services.GetRequiredService<ICameraProvider>()`. I've added an example of this to the Sample App in `CameraViewViewModel`.

> The camera switcher and some of the original debug text (camera name, zoom level, etc) are missing on the sample camera view page

Yea, that was me being a bit lazy. I removed it because the `AvailableCameras` requires async code and after refactoring its async implementation, the value was null when the page first displayed. You're totally welcome to re-add it in a future PR, though! I was just being a bit lazy removing it and not re-implementing the UI with the async implementation, wanting to not delay this PR any further. 

> UpdateCameraInfo in CameraManage.Windows.cs is not needed anymore, since now CameraInfo is fully initialized in CameraProvider.Windows.cs and does not need to be updated.

Removed ✅

PR #1771, opened 2024-03-22
Not use UseMauiMaps

jfversluis:
As part of enabling the .NET MAUI Maps for Windows we call `UseMauiMaps` in the .NET MAUI SDK.

While this worked before, it was unnecessary to do so. Due to a recent change in .NET MAUI, which now throws a exception with a clear error message, this now became problematic. Because of this change in .NET MAUI, now an exception is always thrown, also through our library when you're actually trying to enable Windows.

This change removes the call to .NET MAUI Maps `UseMauiMaps`, users will have to call this themselves, which would then look something like:

```csharp
  builder
	  .UseMauiApp<App>()
	  .ConfigureFonts(fonts =>
	  {
		  fonts.AddFont("OpenSans-Regular.ttf", "OpenSansRegular");
		  fonts.AddFont("OpenSans-Semibold.ttf", "OpenSansSemibold");
	  });

    #if IOS || ANDROID
    builder.UseMauiMaps();
    #endif
    
    #if WINDOWS
    // Initialize the .NET MAUI Community Toolkit Maps by adding the below line of code
    builder.UseMauiCommunityToolkitMaps("key")
    #endif

   return builder.Build();
```

Fixes https://github.com/dotnet/maui/issues/21360

### DIFF
--- a/src/CommunityToolkit.Maui.Maps/AppHostBuilderExtensions.shared.cs
+++ b/src/CommunityToolkit.Maui.Maps/AppHostBuilderExtensions.shared.cs
@@ -13,12 +13,13 @@ public static class AppHostBuilderExtensions
 	/// <returns><see cref="MauiAppBuilder"/></returns>
 	public static MauiAppBuilder UseMauiCommunityToolkitMaps(this MauiAppBuilder builder, string key)
 	{
-		builder.UseMauiMaps();
 		builder.ConfigureMauiHandlers(handlers =>
 		{
 #if WINDOWS
 			CommunityToolkit.Maui.Maps.Handlers.MapHandlerWindows.MapsKey = key;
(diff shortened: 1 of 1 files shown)


jfversluis:
Validated this works as expected

brminnick:
It's a bit late for my review now, but [I would've recommended increasing our `MauiPackageVersion` to `v8.0.10`](https://github.com/CommunityToolkit/Maui/blob/ee03326ab43f35f90d27d92a37b9df06d6663f27/Directory.Build.props#L5) since this PR specifically targets a bug in that version of .NET MAUI. 

We'll likely increase it to the latest version of .NET MAUI, `v8.0.14`, pending any unforeseen circumstances prior to our next release (once TouchBehavior is merged).

PR #1780, opened 2024-03-28
Add Platform-Specific Option to Enable Snackbar on Windows

brminnick:
Snackbar now requires additional code to be added on Windows to the Package.appxmanifest file: https://github.com/MicrosoftDocs/CommunityToolkit/pull/374

Currently, without adding this additional code to `Package.appxmanifest`, `CommunityToolkit.Maui` causes all Windows apps to crash on launch. 

This PR adds `Options.ShouldEnableSnackbarOnWidows` to allow users to opt-in to enabling Snackbar on Windows, ensuring that they have added the additional required code to `Package.appxmanifest`.

```cs
var builder = MauiApp.CreateBuilder()
  .UseMauiCommunityToolkit(options =>
  {
    options.SetShouldEnableSnackbarOnWindows(true);
  })
```

### DIFF
--- a/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
+++ b/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
@@ -18,7 +18,7 @@ public partial class Snackbar : ISnackbar
 	{
 		if (OperatingSystem.IsWindows() && !Options.ShouldEnableSnackbarOnWindows)
 		{
-			throw new InvalidOperationException($"Additional setup is required in the Package.appxmanifest file to enable {nameof(Snackbar)} on Windows. Additonally, `{nameof(AppBuilderExtensions.UseMauiCommunityToolkit)}(options => options.{nameof(Options.SetShouldEnableSnackbarOnWindows)}({true.ToString().ToLower()});` must be called to enable Snackbar on Windows. Additonally, Platform Specific Initialization section of the {nameof(Snackbar)} documentaion for more information: https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar")
+			throw new InvalidOperationException($"Additional setup is required in the Package.appxmanifest file to enable {nameof(Snackbar)} on Windows. Additonally, `{nameof(AppBuilderExtensions.UseMauiCommunityToolkit)}(options => options.{nameof(Options.SetShouldEnableSnackbarOnWindows)}({true.ToString().ToLower()});` must be called to enable Snackbar on Windows. See the Platform Specific Initialization section of the {nameof(Snackbar)} documentaion for more information: https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar")
 			{
 				HelpLink = "https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar"
 			};
--- a/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
+++ b/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
@@ -18,7 +18,7 @@ public partial class Snackbar : ISnackbar
 	{
 		if (OperatingSystem.IsWindows() && !Options.ShouldEnableSnackbarOnWindows)
 		{
-			throw new InvalidOperationException($"Additional setup is required in the Package.appxmanifest file to enable {nameof(Snackbar)} on Windows. Additonally, `{nameof(AppBuilderExtensions.UseMauiCommunityToolkit)}(options => options.{nameof(Options.SetShouldEnableSnackbarOnWindows)}({true.ToString().ToLower()});` must be called to enable Snackbar on Windows. See the Platform Specific Initialization section of the {nameof(Snackbar)} documentaion for more information: https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar")
+			throw new InvalidOperationException($"Additional setup is required in the Package.appxmanifest file to enable {nameof(Snackbar)} on Windows. Additonally, `{nameof(AppBuilderExtensions.UseMauiCommunityToolkit)}(options => options.{nameof(Options.SetShouldEnableSnackbarOnWindows)}({bool.TrueString});` must be called to enable Snackbar on Windows. See the Platform Specific Initialization section of the {nameof(Snackbar)} documentaion for more information: https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar")
 			{
 				HelpLink = "https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar"
 			};
--- a/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
+++ b/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
@@ -16,13 +16,15 @@ public partial class Snackbar : ISnackbar
 	/// </summary>
 	public Snackbar()
 	{
-		if (OperatingSystem.IsWindows() && !Options.ShouldEnableSnackbarOnWindows)
+#if WINDOWS
+		if (!Options.ShouldEnableSnackbarOnWindows)
 		{
 			throw new InvalidOperationException($"Additional setup is required in the Package.appxmanifest file to enable {nameof(Snackbar)} on Windows. Additonally, `{nameof(AppBuilderExtensions.UseMauiCommunityToolkit)}(options => options.{nameof(Options.SetShouldEnableSnackbarOnWindows)}({bool.TrueString.ToLower()});` must be called to enable Snackbar on Windows. See the Platform Specific Initialization section of the {nameof(Snackbar)} documentaion for more information: https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar")
 			{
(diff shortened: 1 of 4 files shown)


src/CommunityToolkit.Maui/Options.cs
```diff
@@ -1,32 +1,83 @@
...
+
+		static void OnSnackbarNotificationInvoked(Microsoft.Windows.AppNotifications.AppNotificationManager sender,
+													Microsoft.Windows.AppNotifications.AppNotificationActivatedEventArgs args)
+		{
+			Alerts.Snackbar.HandleSnackbarAction(args);
+		}
+#endif
+
+
+		ShouldEnableSnackbarOnWindows = true;
```
bijington, line 81: Should this be:

```suggestion
		ShouldEnableSnackbarOnWindows = value;
```

And furthermore should we only perform the above registrations if value is `true`?

bijington: Or can we rename the option to something like `EnableSnackbarOnWindows`

bijington (changes requested):
Good work @brminnick I think we should either use the `value` parameter or remove it though.

src/CommunityToolkit.Maui/Options.cs
brminnick: Whoops - good catch! 

Let's keep the variable / method name similar to the existing ones, eg `ShoudSuppressExceptions`

brminnick:
Thanks Shaun! Good catch. Updated ✅

src/CommunityToolkit.Maui/Options.cs
```diff
@@ -1,32 +1,83 @@
...
+	/// <remarks>
+	/// Default value is false.
+	/// </remarks>
 	public void SetShouldSuppressExceptionsInBehaviors(bool value) => ShouldSuppressExceptionsInBehaviors = value;
+
+	/// <summary>
+	/// Enables <see cref="Alerts.Snackbar"/> for Windows
+	/// </summary>
+	/// <remarks>
+	/// Additional setup is required in the Package.appxmanifest file to enable Snackbar on Windows. See the <see cref="Alerts.Snackbar"/> <a href="https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar">Platform Specific Initialization Documentation</a> for more information. Default value is false.
```
brminnick, line 52:
```suggestion
	/// Additional setup is required in the Package.appxmanifest file to enable <see cref="Alerts.Snackbar"/> on Windows. See the <a href="https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar">Snackbar Platform Specific Initialization Documentation</a> for more information. Default value is false.
```

src/CommunityToolkit.Maui/Options.cs
```diff
@@ -1,32 +1,83 @@
...
+	/// </remarks>
+	public void SetShouldEnableSnackbarOnWindows(bool value)
+	{
```
VladislavAntonyuk, line 55: If value is false return from the method

bijington: Or just change to `public void EnableSnackbarOnWindows()`?

brminnick: I think it'd still be smart for us to set `Options.ShouldEnableSnackbarOnWindows = value` to match the existing implementations like `ShouldSuppressExceptionsInBehaviors`. We may want to use it for something in the future (or not, but it's there if we need it!).

I just added a bit of control flow to check if `value is true && builder is not null`, and then throw an `InvalidOperationException` if for some reason a user tries to set this value outside of `.UseMauiCommunityToolkit()`.



```cs
#if WINDOWS
		if (value is true && builder is null)
		{
			throw new InvalidOperationException($"{nameof(SetShouldEnableSnackbarOnWindows)} must be called using the {nameof(AppBuilderExtensions.UseMauiCommunityToolkit)} extension method. See the Platform Specific Initialization section of the {nameof(Alerts.Snackbar)} documentaion for more inforamtion: https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar)");
		}
		else if (value is true && builder is not null)
		{
			builder.ConfigureLifecycleEvents(events =>
			{
				events.AddWindows(windows => windows
					.OnLaunched((_, _) =>
					{
						Microsoft.Windows.AppNotifications.AppNotificationManager.Default.NotificationInvoked += OnSnackbarNotificationInvoked;
						Microsoft.Windows.AppNotifications.AppNotificationManager.Default.Register();
					})
					.OnClosed((_, _) =>
					{
						Microsoft.Windows.AppNotifications.AppNotificationManager.Default.NotificationInvoked -= OnSnackbarNotificationInvoked;
						Microsoft.Windows.AppNotifications.AppNotificationManager.Default.Unregister();
					}));
			});

			static void OnSnackbarNotificationInvoked(Microsoft.Windows.AppNotifications.AppNotificationManager sender,
														Microsoft.Windows.AppNotifications.AppNotificationActivatedEventArgs args)
			{
				Alerts.Snackbar.HandleSnackbarAction(args);
			}
		}
#endif
```

brminnick: Ah - I figured out the perfect use-case for `Options.ShouldEnableSnackbarOnWindows`! 

We can use it verify the constructor of 

```cs
public Snackbar()
	{
		if (OperatingSystem.IsWindows() && !Options.ShouldEnableSnackbarOnWindows)
		{
			throw new InvalidOperationException($"Additional setup is required in the Package.appxmanifest file to enable {nameof(Snackbar)} on Windows. Additonally, `{nameof(AppBuilderExtensions.UseMauiCommunityToolkit)}(options => options.{nameof(Options.SetShouldEnableSnackbarOnWindows)}({true.ToString().ToLower()});` must be called to enable Snackbar on Windows. Additonally, Platform Specific Initialization section of the {nameof(Snackbar)} documentaion for more information: https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar")
			{
				HelpLink = "https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar"
			};
		}
		// ...
	}
```

brminnick: ^This'll also help every dev who inevitably doesn't read the docs (like me 😀) to be alerted to the new breaking change requirement on Windows for Snackbar.

brminnick (commented):
Thanks Shaun + Vlad! I think I'm happy with the implementation now. 

We've added enough fail-safes to the code to assist future devs who haven't yet seen the platform-specific implementation docs for `Snackbar` on Windows. Hopefully throwing these exceptions at runtime will help folks find the docs instead of opening a new Issue on the repo. 

And if they do open an issue for one of these InvalidOperationExceptions, we can just copy/paste the link from the stack trace, point them to the docs and close the issue.

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
```diff
@@ -16,6 +16,14 @@ public partial class Snackbar : ISnackbar
 	/// </summary>
 	public Snackbar()
 	{
+		if (OperatingSystem.IsWindows() && !Options.ShouldEnableSnackbarOnWindows)
+		{
+			throw new InvalidOperationException($"Additional setup is required in the Package.appxmanifest file to enable {nameof(Snackbar)} on Windows. Additonally, `{nameof(AppBuilderExtensions.UseMauiCommunityToolkit)}(options => options.{nameof(Options.SetShouldEnableSnackbarOnWindows)}({true.ToString().ToLower()});` must be called to enable Snackbar on Windows. See the Platform Specific Initialization section of the {nameof(Snackbar)} documentaion for more information: https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar")
```
VladislavAntonyuk, line 21: Please use Boolean TrueString instead of bool.ToString

src/CommunityToolkit.Maui/Options.cs
```diff
@@ -1,32 +1,88 @@
...
+#if WINDOWS
+		if (value is true && builder is null)
+		{
+			throw new InvalidOperationException($"{nameof(SetShouldEnableSnackbarOnWindows)} must be called using the {nameof(AppBuilderExtensions.UseMauiCommunityToolkit)} extension method. See the Platform Specific Initialization section of the {nameof(Alerts.Snackbar)} documentaion for more inforamtion: https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar)");
+		}
+		else if (value is true && builder is not null)
+		{
```
VladislavAntonyuk, line 65: Do we need to write “is true” or is it only for readability?

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
brminnick: Cool! I hadn't heard of `bool.TrueString` before. 

The problem is that `bool.TrueString` returns `True` with an uppercase `T`, so I kept `.ToLower()`. 

We could also just hardcode the text `true` instead. I didn't mean to make this more complicated than it is 😆

src/CommunityToolkit.Maui/Options.cs
brminnick: It's purely aesthetic, just for readability. 

Since `value` doesn't follow our typical pattern of naming a boolean using a question, it felt strange to read the `if` statement in English: "if value and builder is not null" compared to "if value is not null and builder is not null".

pictos (approved):

PR #1791, opened 2024-04-03
For All `CommunityToolkit.Maui.Behavior`s, Ensure `BindingContext` Is Automatically Set to the Attached View's `BindingContext`

brminnick:
### Description of Change ###

The goal of this Pull Request is to ensure that every `Behavior` in `CommunityToolkit.Maui.Behavior` automatically set's the its `BindingContext` to match the `BindingContext` of the attached `View`.

This PR creates/updates the following classes/interfaces:
- `ICommunityToolkitBehavior` - an interface meant for internal consumption, used to align the functionality of `BaseBehavior` and `BasePlatformBehavior`
  ```cs
  protected TView? View { get; set; }
  
  internal bool TrySetBindingContextToAttachedViewBindingContext();
  internal bool TryRemoveBindingContext();
  internal void AssignViewAndBingingContext(TView bindable);
  internal void UnassignViewAndBingingContext(TView bindable);
  protected void OnViewPropertyChanged(object? sender, PropertyChangedEventArgs e)
  protected void OnViewPropertyChanged(TView sender, PropertyChangedEventArgs e);
  ```
- `BaseBehavior : ICommunityToolkitBehavior`
  - `BaseBehavior` now implements `ICommunityToolkitBehavior`
- `BasePlatformBehavior<TView> : BasePlatformBehavior<TView, PlatformView> where TView : Element`
  - Uses the default `PlatformView` for each operating system
- `BasePlatformBehavior<TView, TPlatformView> : PlatformBehavior<TView, TPlatformView>, ICommunityToolkitBehavior<TView> where TView : Element where TPlatformView : class`

 ### Linked Issues ###
 

 - Fixes https://github.com/CommunityToolkit/Maui/discussions/1785

 ### PR Checklist ###
 
The unit tests for `BasePlatformBehavior` can't test `OnAttachedTo()` because `OnAttachedTo()` is only called in platform-specific code for `Microsoft.Maui.Controls.PlatformBehavior`.

I also updated the Unit Tests to use Collection Expressions for a small performance bump.


src/CommunityToolkit.Maui/Behaviors/BaseBehavior.shared.cs
```diff
@@ -93,29 +52,17 @@ protected override void OnDetachingFrom(TView bindable)
 	/// <param name="property"></param>
 	/// <param name="defaultBinding"></param>
 	/// <returns></returns>
-	[MemberNotNullWhen(true, nameof(bindingField), nameof(getContextMethod))]
+	[Obsolete($"{nameof(IsBound)} is no longer used by {nameof(CommunityToolkit)}.{nameof(CommunityToolkit.Maui)} and will be removed in a future release")]
```
brminnick, line 55: I was able to remove the need for this method which removes the need for Reflection.

However, I've kept it here and marked it `[Obsolete]` because removing it would be a breaking change to the API. We can remove this method for the next major version bump.

src/CommunityToolkit.Maui/Behaviors/ICommunityToolkitBehavior.shared.cs
```diff
@@ -0,0 +1,96 @@
+using System.ComponentModel;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace CommunityToolkit.Maui.Behaviors;
+
+/// <summary>
+/// A common interface to be used across <see cref="BaseBehavior{TView}"/> and <see cref="BasePlatformBehavior{TView,TPlatformView}"/>
+/// </summary>
+public interface ICommunityToolkitBehavior<TView> where TView : Element
```
brminnick, line 10: This Interface primarily contains default methods in order to align the functionality of `BaseBehavior` and `BasePlatformBehavior`.

We had to use an interface with default methods because we cannot use the same `abstract class` for both `BaseBehavor : Behavior` and `BasePlatformBehavior : PlatformBehavior` because these base behavior classes inherit from different .NET MAUI classes.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/BasePlatformBehavior.shared.cs
```diff
@@ -0,0 +1,75 @@
...
+public abstract partial class BasePlatformBehavior<TView, TPlatformView> : PlatformBehavior<TView, TPlatformView>, ICommunityToolkitBehavior<TView>
+	where TView : Element
+	where TPlatformView : class
+{
+	/// <summary>
+	/// View used by the Behavior
+	/// </summary>
+	protected TView? View { get; set; }
+
+	TView? ICommunityToolkitBehavior<TView>.View
```
pictos, line 42: `ICommunityToolkitBehavior` interface is `public`, this will allows any external member to set the `View` value when cast it to the `interface`, is it intentional?

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/BasePlatformBehavior.shared.cs
```diff
@@ -0,0 +1,75 @@
...
+	/// <param name="e"></param>
+	protected virtual void OnViewPropertyChanged(TView sender, PropertyChangedEventArgs e)
+	{
+
+	}
+
+	/// <inheritdoc/>
+	protected override void OnAttachedTo(TView bindable, TPlatformView platformView)
+	{
+		base.OnAttachedTo(bindable, platformView);
```
pictos, line 61: we don't need to call the base method here, [since the implementation is empty](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Interactivity/PlatformBehavior.cs#L128).

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/BasePlatformBehavior.shared.cs
```diff
@@ -0,0 +1,75 @@
...
+
+		((ICommunityToolkitBehavior<TView>)this).AssignViewAndBingingContext(bindable);
+	}
+
+	/// <inheritdoc/>
+	protected override void OnDetachedFrom(TView bindable, TPlatformView platformView)
+	{
+		base.OnDetachedFrom(bindable, platformView);
+
```
pictos, line 70: we don't need to call the base method here, [since the implementation is empty](https://github.com/dotnet/maui/blob/main/src/Controls/src/Core/Interactivity/PlatformBehavior.cs#L135).

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.macios.cs
```diff
@@ -119,4 +106,20 @@ static void SetUIImageViewTintColor(UIImageView imageView, View element, Color c
 		imageView.TintColor = color.ToPlatform();
 	}
 
+	void OnElementPropertyChanged(object? sender, PropertyChangedEventArgs e)
```
pictos, line 109: we already subscribe to the `bindable.PropertyChanged` on base, does make sense to use the override method instead?

src/CommunityToolkit.Maui/Behaviors/Validators/MultiValidationBehavior.shared.cs
```diff
@@ -65,7 +65,7 @@ protected override async ValueTask<bool> ValidateAsync(object? value, Cancellati
 			return c.ValidateNestedAsync(token).AsTask();
 		})).ConfigureAwait(false);
 
-		var errors = children.Where(c => c.IsNotValid).Select(GetError);
+		var errors = children.Where(c => c.IsNotValid).Select(GetError).ToList();
```
pictos, line 68: It could be a performance regression, so I would ask, do we need to make it a concrete collection here?

From the diff, It's needed when inside the `if`, at line 78. So, it did allocate the collection in a specific case.

src/CommunityToolkit.Maui/Behaviors/Validators/MultiValidationBehavior.shared.cs
```diff
@@ -75,31 +75,27 @@ protected override async ValueTask<bool> ValidateAsync(object? value, Cancellati
...
+		if (e.NewItems is not null)
 		{
 			foreach (var child in e.NewItems.OfType<ValidationBehavior>())
 			{
-				child.TrySetBindingContext(new Binding
-				{
-					Path = BindingContextProperty.PropertyName,
-					Source = this
-				});
+				((ICommunityToolkitBehavior<VisualElement>)child).TrySetBindingContextToAttachedViewBindingContext();
```
pictos, line 90: if it fails do we want to log some message or throw an error?

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/BasePlatformBehavior.shared.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.ComponentModel;
+
+namespace CommunityToolkit.Maui.Behaviors;
+
+
+/// <summary>
+/// Abstract class for our behaviors to inherit.
+/// </summary>
+/// <typeparam name="TView">The <see cref="VisualElement"/> that the behavior can be applied to</typeparam>
+public abstract partial class BasePlatformBehavior<TView> : BasePlatformBehavior<TView, PlatformView>
```
pictos, line 22: We need this member to be `public` in order to use on our `public` Behaviors, but do we want external assemblies to create it? As you mentioned on discussion we would like to break the .NET MAUI default behavior outside our lib, and I agree with that. We're setting the `BindingContext` here, but we _know_ what we're doing, could be dangerous to expose it to the world, so my suggestion is to add an `internal ctor` for those `abstract` classes.

pictos (commented):
it looks a lot of work, great that you could fit it soo fast! I've some points there

src/CommunityToolkit.Maui/Behaviors/ICommunityToolkitBehavior.shared.cs
pictos: Should the interface be `internal`?

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/BasePlatformBehavior.shared.cs
brminnick: I prefer to keep the call to `base.OnAttached()` as a best practice. 

If .NET MAUI ever updates their implementation, we'll already be prepared. 

Plus, I literally lost over an hour today trying to figure out why `BasePlatformBehavior.OnAttached()` wasn't getting called. It turns out none of our PlatformBehaviors were calling `base.OnAttached()`. 

Had we always called `base.OnAttached()`, despite it not being immediately necessary, I would've saved a ton of time on this PR.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/BasePlatformBehavior.shared.cs
brminnick: I prefer to keep the call to `base.OnAttached()` as a best practice. 

If .NET MAUI ever updates their implementation, we'll already be prepared. 

Plus, I literally lost over an hour today trying to figure out why `BasePlatformBehavior.OnAttached()` wasn't getting called. It turns out none of our PlatformBehaviors were calling `base.OnAttached()`. 

Had we always called `base.OnAttached()`, despite it not being immediately necessary, I would've saved a ton of time on this PR.

src/CommunityToolkit.Maui/Behaviors/Validators/MultiValidationBehavior.shared.cs
brminnick: It's actually a performance bump recommended by reshapes since every path in the code was making multiple calls to the same IEnumerable.

Axemasta:
Does the BaseBehavior need to account for being attached to multiple `VisualElement`s? Currently with the `TouchBehavior` it will get wierd... all views attached will receive inputs but only the last view the behavior was added to will be animated.

Since specifically the `TouchBehavior` needs to be applied 1:1 (behavior : view) should the `BaseBehavior` have a way of preventing attaching to multiple views?

brminnick:
> Does the BaseBehavior need to account for being attached to multiple VisualElements? Currently with the TouchBehavior it will get weird... all views attached will receive inputs but only the last view the behavior was added to will be animated.
> 
> Since specifically the TouchBehavior needs to be applied 1:1 (behavior : view) should the BaseBehavior have a way of preventing attaching to multiple views?

Good question - yes! I followed the existing pattern I found in `BaseBehavior` and extended it to `BasePlatformBehavior` via `ICommunityToolkitBehavior`. It'll automatically set the `Behavior.BindingContext` to match the `BindingContext` of the first `VisualElement` it is attached to. If the same `Behavior` is then attached to another `VisualElement`, it will keep the `BindingContext` from the first `VisualElement`.

src/CommunityToolkit.Maui/Behaviors/ICommunityToolkitBehavior.shared.cs
brminnick: I wish. 

It has to be `public` because all of our public Behaviors are implementing it. 

I did make all of its methods + properties internal or protected so that we don't need to support it as a stand-alone external interface.

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/BasePlatformBehavior.shared.cs
brminnick: It's a `protected` interface property! 

Since it's marked `protected`, devs won't be able to manually set its value for any of our behaviors from a different class. Here's a screenshot of the compiler error I get when trying to change `ICommunityToolkitBehavior.View` in the IDE:

image

brminnick: And here's another screenshot from the same code showing that the variable `behavior` is cast to `ICommunityToolkitBehavior`:
image

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/IconTintColor/IconTintColorBehavior.macios.cs
brminnick: Great catch! Done  ✅

src/CommunityToolkit.Maui/Behaviors/Validators/MultiValidationBehavior.shared.cs
brminnick: Agreed! I'll only use `Trace.WriteLine` when `TrySetBindingContextToAttachedViewBindingContext` fails; I won't use us when `TryRemoveBindingContext` fails since that also returns `false` when there is no `BindingContext` to remove which wouldn't necessarily be an error.

Done ✅

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/BasePlatformBehavior.shared.cs
brminnick: Done ✅

VladislavAntonyuk (dismissed):

brminnick:
Ok gang! I think this is ready to go. 

I refactored a couple things just now: 
- In `ICommunityToolkitBehavior`, changed `internal OnViewPropertyChanged(object? sender, PropertyChangedEventArgs e)` -> `protected OnViewPropertyChanged(object? sender, PropertyChangedEventArgs e)`
- Implemented CollectionExpressions for Unit Tests

pictos (approved):

tranb3r:
@brminnick @pictos 
I think there is a regression in 8.0.1.
In a CollectionView, the BindingContext is not set correctly when using TouchBehavior. When I tap on an item, the command from another item is executed.
I did not have this problem with 8.0.0 and setting the BindingContext manually.
Also, with 8.0.1, even if I set the BindingContext manually, it does not fix the issue.
Could you please double-check? I do not have the time to build a repro right now, but it's easy to reproduce.

PR #1815, opened 2024-04-14
Fix TouchBehavior for CollectionView

### LINKED ISSUE #1783: [BUG] TouchBehavior crash

### Current Behavior

I have a CollectionView like this

```
<CollectionView
    IsVisible="{Binding isNotBusy}"
    ItemsSource="{Binding CoursesAssigned}">
    <CollectionView.ItemTemplate>
        <DataTemplate x:DataType="model:Course">
            <Border
                Margin="50"
                Stroke="Purple"
                StrokeThickness="1">
                <Border.StrokeShape>
                    <RoundRectangle CornerRadius="8,8,8,8" />
                </Border.StrokeShape>
                <Border.Behaviors>
                    <mct:TouchBehavior Command="{Binding Path=GoToDetailsCommand, Source={x:RelativeSource AncestorType={x:Type vm:MyCoursesPageViewModel}}}"
                        CommandParameter="{Binding .}"
                        DefaultBackgroundColor="Transparent"
                        HoveredBackgroundColor="CornflowerBlue"
                        PressedBackgroundColor="CornflowerBlue" />
                </Border.Behaviors>`
```

But when I run it, I get the error

`Operation is not valid due to the current state of the object.`

### Expected Behavior

My command is declared in my viewModel

```
 [RelayCommand]
 async Task GoToDetails(Course course) {

     await appService.NavigateTo($"{nameof(MyCoursesDetailPage)}",
         true, new Dictionary<string, object> {
         {"Course", course }
     });
 }
```

So I am expecting my command to go the MyCoursesDetailPage when I click the border

### Steps To Reproduce

1. Create a new Maui project
2. Install the mct (Maui community toolkit)
3. Create some sort of class to represent an object, like fruits. 
4. create a collectionView
5. Create a border
6. attach the Touch behavior to a border

https://github.com/eduardoagr/TouchBehaviorExample

### Environment

```markdown
- .NET MAUI CommunityToolkit:
- OS: Windows, Android
- .NET MAUI:
```

### Anything else?

_No response_

### LINKED ISSUE #1804: [BUG][Regression][8.0.1] TouchBevavior binding not working in a CollectionView

### Current Behavior

TouchBehavior binding is not working in a CollectionView when using 8.0.1.
It seems like the wrong item is binded after the CollectionView is scrolled a bit (recycling issue?).
This is a regression in 8.0.1. No issue in 8.0.0.

I mentionned this issue in the PR but so far nobody has reacted.
https://github.com/CommunityToolkit/Maui/pull/1791#issuecomment-2039825987

### Expected Behavior

I expect it to work the same as in 8.0.0.

### Steps To Reproduce

1. Open and run solution from repro.
2. Tap on any item. Scroll the CollectionView. Tap again any item.
3. At some point (probably after recycling of views), the alert shows that the TouchBehavior binds to the wrong item.

[untitled.webm](https://github.com/CommunityToolkit/Maui/assets/12516566/0f8168b4-9c79-4b2a-abcb-2d24d3c1cdba)

https://github.com/tranb3r/Issues/tree/main/MauiAppMct801TouchBehavior

### Environment

```markdown
- .NET MAUI CommunityToolkit:8.0.1
- OS:Android
- .NET MAUI:8.0.14
```

pictos:
### Description of Change ###

Remove the `IsSet` check on `ICommunityToolkitBehavior`, this check will fail for any control that recycle views, since the behavior on that particular view will be already set, causing it to not update. Removing this check will make sure we can update the `BindingContext`.

 ### Linked Issues ###
 

 - Fixes #1804
 - Fixes #1783

 ### PR Checklist ###
 

src/CommunityToolkit.Maui/Behaviors/ICommunityToolkitBehavior.shared.cs
```diff
@@ -21,7 +21,12 @@ internal bool TrySetBindingContextToAttachedViewBindingContext()
 			throw new InvalidOperationException($"{nameof(ICommunityToolkitBehavior<TView>)} can only be used for a {nameof(Behavior)}");
 		}
 
-		if (behavior.IsSet(BindableObject.BindingContextProperty) || View is null)
```
bijington, line 24: What will happen in scenarios where a developer explicitly sets their own BindingContext for the behavior? I think this could would overwrite it with the BindingContext from the View?

bijington: Perhaps the safer change is to remove the binding altogether and subscribe to the BindingContext changed event on the View and assign to the Behavior?

pictos: Not being able to re-assing the `BindingContext` will make all Behaviors un-usable on recycling/virtualization scenarios.

What do you think the correct approach would be?

bijington: I have changed the `OnAttachedTo` method in `BaseBehavior` to look as follows:

```csharp
protected override void OnAttachedTo(TView bindable)
{
	base.OnAttachedTo(bindable);

	bindable.PropertyChanging += (sender, args) =>
	{
		var currentViewBindingContext = ((TView)sender).BindingContext;
			
		if (args.PropertyName == nameof(BindingContext) && 
		    ReferenceEquals(this.BindingContext, currentViewBindingContext))
		{
			this.BindingContext = null;
		}
	};
		
	bindable.PropertyChanged += (sender, args) =>
	{
		if (args.PropertyName == nameof(BindingContext) &&
		    this.BindingContext is null)
		{
			this.BindingContext = ((TView?)sender)?.BindingContext;				
		}
    };
}
```

It might not be the prettiest but I believe it should work. What does everyone think? Any ideas on how to make it nicer?

pictos: @bijington what's the difference between using event handlers and the interface? At the end we will update the BindingContext value

bijington: Unless I'm missing something my code should only update the BindingContext if the behavior previously used the value from the View. This means we wouldn't override the value of a user supplied a BindingContext.

pictos: Oh I see now, yeah you're right about the behavior on your code... But from my memory there isn't anything on .NET MAUI that prevents the BindingContext to be changed... it shouldn't be a `readonly` or `immutable`  value. So I'm curious, why this implementation was used

bijington: I believe the reasoning is that we only inherit the BindingContext from the attached control if the developer hasn't explicitly set the BindingContext themselves. I have to be honest and say I haven't checked the order things happen then and when the BindingContext becomes not null

bijington: I have reverted the change and added in two examples from @tranb3r example repo. Perhaps @tranb3r you can test this sample and let us know how it compares to your expectation. The code works as I expect on iOS/macOS in our sample app

pictos: @bijington will take a look asap. Did you add the failing case to the sample app? Asking to make sure  Ican use it during development

bijington: Yes there are now 2 CollectionViews in the sample app on the TouchBehaviorPage

tranb3r: > I have reverted the change and added in two examples from @tranb3r example repo. Perhaps @tranb3r you can test this sample and let us know how it compares to your expectation. The code works as I expect on iOS/macOS in our sample app

These are indeed the same cases as in my repro. Not working on android, unless using 8.0.0.

bijington: @tranb3r thank you for confirming.

@tranb3r and @pictos the changes that I have just pushed appear to solve the issue on Android based on my testing.

src/CommunityToolkit.Maui/Behaviors/ICommunityToolkitBehavior.shared.cs
```diff
@@ -21,11 +21,40 @@ internal bool TrySetBindingContextToAttachedViewBindingContext()
...
+		if (View.BindingContext?.GetType() == behavior.BindingContext?.GetType())
+		{
+			for (var i = 0; i < 10; i++)
+			{
+				if (parent is null)
+				{
+					break;
+				}
+
+				if (parent is CollectionView)
```
pictos, line 43: I believe for testing this is enough for now... But collectionView isn't the only control that may have virtualization

bijington: Good point!

pictos: still not enough, because devs can apply virtualization to any kind of control or layout... I'm still didn't get the reason to be so conservative on allowing changing the BindingContext. In other words, we shouldn't prevent the BindingContext to be mutable. You may have cases where the dev reuses the control across the page and wants the control's BindingContext to change when that happens

bijington: I see that we want to automatically inherit the `BindingContext` from the `View` it is attached to **only if the `BindingContext`** is not already set. It appears there are issues around checking whether the `BindingContext` is already set especially when the behavior is used inside something like the `CollectionView`  and the behavior is attached to recycled views.

pictos: Yeah, but the thing is that after the first set, that value will always be `true`, and that's causing the issue. Because of that I removed the `IsSet` check, as mentioned on the PR description.

I think this happens because when setting the new value, they don't use the [RemoveBinding](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.bindableobject.removebinding?view=net-maui-8.0) method, just `BindinContext = someValue`, causing the `IsSet` to never be false again

bijington: I didn't confirm the instances but I did discover that OnAttachedTo was being called before OnDetachedFrom when the recycling was happening. Which may explain why IsSet is always true after the first time.

src/CommunityToolkit.Maui/Behaviors/ICommunityToolkitBehavior.shared.cs
tranb3r: Sorry to ask again, but I'm still wondering if this BindingContext auto-set feature (introduced in 8.0.1) is really required?
It seems to me that 8.0.0 was fine. If the goal was only to simplify xaml code, I'm not sure it's worth it, considering the difficulty you have to figure out how to fix this regression.
I'm probably missing something, but could you please explain which bug did you try to fix in 8.0.1? Or at least post a code sample for the original issue?

bijington: I recall there was a lot of developers reporting an issue that their bindings didn't work. @brminnick Is this correct?

tranb3r: I suppose their bindings didn't work because they were not setting the BindingContext manually.

BusterIT:
Just simply push this solution. It is more performant than always set the binding context new:

	internal bool TrySetBindingContextToAttachedViewBindingContext()
	{
		if (this is not Behavior behavior)
		{
			throw new InvalidOperationException($"{nameof(ICommunityToolkitBehavior<TView>)} can only be used for a {nameof(Behavior)}");
		}

		if (View is null)
		{
			return false;
		}

		if (!behavior.IsSet(BindableObject.BindingContextProperty))
		{
			behavior.SetBinding(BindableObject.BindingContextProperty, new Binding
			{
				Source = View,
				Path = BindableObject.BindingContextProperty.PropertyName
			});
			return true;
		}

		if (!Equals(behavior.BindingContext, View.BindingContext))
		{
			behavior.BindingContext = View.BindingContext;
			return true;
		}

		return false;
	}

sstobinski:
Hello, what about the review? Isn't the merge already possible?

DavidV1603:
Hello, this is really important for my app. Would it be possible to merge this or is there a problem left?

bijington:
@BusterIT I don't think that solutions works for all use cases. It won't work if you have a behavior attached to a view inside a CollectionView and it's bound to something outside of that view. It will work on the first pass but not if a view is recycled. Unless I am missing something

bijington:
For the others asking, I don't recall the state of this PR. I can try to take a look at it soon but have any of you tested that this solves your problem?

BusterIT:
> @BusterIT I don't think that solutions works for all use cases. It won't work if you have a behavior attached to a view inside a CollectionView and it's bound to something outside of that view. It will work on the first pass but not if a view is recycled. Unless I am missing something

While the concerns about handling behaviors in recycled views within a CollectionView are valid, the modifications I introduced aim to ensure the Behavior's BindingContext is initially synchronized with the View's BindingContext more reliably than before. The original code did not address changes to the BindingContext post-initialization, which is a limitation that remains consistent even after my changes.

To address the concerns about recycling:
1. The check `if (!Equals(behavior.BindingContext, View.BindingContext))` ensures that if the View's BindingContext changes, which can happen in a recycled view, it updates the Behavior's BindingContext correspondingly. This means my solution should handle the initial cases more robustly.
2. The scenario where the binding might need to refresh dynamically due to external bindings outside of the view's direct context still needs consideration. To robustly handle the recycling scenarios, further enhancements might be required, such as hooking into more lifecycle events of the View or using additional observables to monitor changes in the BindingContext actively.

In summary, the modifications do not introduce new limitations but rather provide a foundation on which more robust handling of dynamic contexts and view recycling can be built.

brminnick:
This PR is resolved in #2215

PR #1820, opened 2024-04-16
Add AOT Compatibility

### LINKED ISSUE #1509: [Proposal] Add Support for NativeAOT
### Feature name

Add Support for NativeAOT

### Link to discussion

https://github.com/CommunityToolkit/Maui/issues/1460#issuecomment-1804383350

### Progress tracker

- [ ] Android Implementation
- [ ] iOS Implementation
- [ ] Windows Implementation
- [ ] Tizen Implementation
- [ ] Unit Tests
- [ ] Samples
- [ ] Documentation

### Summary

This Proposal seeks to make all .NET MAUI CommunityToolkit libraries safe to use with NativeAOT.

This requires the library to be safe to use with the [.NET Trimmer](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-options): 

  > The `IsTrimmable` property defaults to true when configuring a project as AOT-compatible with <IsAotCompatible>true</IsAotCompatible>

https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0#enable-project-specific-trimming

### Motivation

In .NET 9, .NET MAUI will add [NativeAOT support](https://learn.microsoft.com/dotnet/core/deploying/trimming/trimming-options): https://github.com/dotnet/maui/issues/18658.

This give us approximately one year until .NET 9 releases in November 2024 to accomplish this Proposal.

#### Directory.Build.props

We will make two changes to `Directory.Build.proprs`:
1. Replace the `<IsTrimmable>false` property with `<IsAotCompatible>true`
  > [The IsTrimmable property defaults to true when configuring a project as AOT-compatible with <IsAotCompatible>true</IsAotCompatible>](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0#enable-project-specific-trimming)
2. Add `<EnableTrimAnalyzer>true</EnableTrimAnalyzer>`

```xml
<IsAotCompatible>true</IsAotCompatible>
<EnableTrimAnalyzer>true</EnableTrimAnalyzer>
```

#### Resolve Trim Warnings

Once `<EnableTrimAnalyzer>` has been enabled, we will then need to resolve every Trimmer warning. We can use the Microsoft docs to help us accomplish it: https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0#resolve-trim-warnings

### Usage Syntax

To use .NET Trimming, the developer will add `<TrimMode>full</TrimMode>` to their CSProj. The .NET MAUI team will be adding `<TrimMode>full</TrimMode>` to the .NET MAUI Templates in .NET 9: https://github.com/dotnet/maui/issues/18658

### Drawbacks

We may not fix all of the bugs introduced by the Trimmer in our first PR that adds NativeAOT Compatibility, and that's ok.

As new Trimmer-related bugs are reported, we will prioritize them and provide hot fixes.

### Alternatives

There are no alternatives.

Now that .NET MAUI will be adding support for NativeAOT in .NET 9, every .NET MAUI developer is required to ensure their library is Trimmer safe.

### Unresolved Questions

_No response_

### Additional Information

How to make libraries compatible with native AOT: https://devblogs.microsoft.com/dotnet/creating-aot-compatible-libraries/

simonrozsival:
### Description of Change ###

This PR marks all product assemblies as AOT-compatible and fixes all warnings produced by trim and AOT analyzers. The list of warnings was surprisingly short:
- `BaseBehavior.IsBound` is inherently trim-unsafe, but since it's already `[Obsolete]`, I think it is OK to just mark it as `[RequiresUnreferencedCode]` (any app or library which would call this method will get a trimming warning and will have to deal with it)
- `ServiceCollectionExtensions` methods need `[DynamicallyAccessedMembers]` attributes matching `IServiceCollection`
- `PopupService.AddTransientPopup` needs `[DynamicallyAccessedMembers]` attributes matching `IServiceCollection`
- `ValueConverterExtension.IsValidTargetType` is the most complicated case I think. It is not possible to fix this case with `[DAM]` attributes because as I propagated the attribute to all callsites, I reached a point where I would need to add the attribute to `Microsoft.Maui.Controls.IValueConverter` which wasn't possible. I tried to simplify the method, because it seemed too complicated. I'll leave some comments in the code to explain my reasoning.

#### TODO (incomplete list)
- [ ] Fix Windows specific warnings (PopupExtensions.SetColor, JSON deserialization in MapHandler
- [ ] Remove all uses of `new Binding(...)` in favor of expression-based bindings (not available yet, still WIP in MAUI)
- [ ] Update type converters of types which define implicit operators (FileMediaSource, MediaSource, ResourceMediaSource, UriMediaSource)

 ### Linked Issues ###

 - Fixes #1509

 ### PR Checklist ###
 

src/CommunityToolkit.Maui/Extensions/ValueConverterExtension.shared.cs
```diff
@@ -13,18 +13,14 @@ public ICommunityToolkitValueConverter ProvideValue(IServiceProvider serviceProv
 
 	private protected static bool IsValidTargetType<T>(in Type targetType)
 	{
-		if (IsConvertingToString(targetType) && CanBeConvertedToString())
```
simonrozsival, line 16: The `CanBeConvertedToString` looks for `ToString` method which should be available on any .NET type. The only reason I can think of why the method would be missing is due to trimming. Validating if a converter can be applied to a type this way doesn't make much sense to me though. Maybe I'm missing something here?

src/CommunityToolkit.Maui/Extensions/ValueConverterExtension.shared.cs
```diff
@@ -13,18 +13,19 @@ public ICommunityToolkitValueConverter ProvideValue(IServiceProvider serviceProv
...
+		if (IsConvertingToString(targetType))
 		{
 			return true;
 		}
 
-		try
+		if (typeof(T).IsAssignableFrom(targetType))
 		{
-			var instanceOfT = default(T);
-			instanceOfT ??= (T?)Activator.CreateInstance(targetType);
```
simonrozsival, line 24: The use of `Activator.CreateInstance` is unclear to me:
- When `T` is non-nullable value type, it is not called
- When `T` is a nullable value type, it returns `null` and doesn't change the outcome
- When `T` is a reference type, it would try to call the public parameterless constructor of `targetType` and it would try to cast it to `T?`
    - note that this cast won't call any custom explicit or implicit operators because we're casting to a generic type
    - the cast will throw `InvalidCastException` when `targetType` is not a subclass of `T` and the method will return `false`
    - otherwise, we will try to convert the instance of `typeType` to `targetType` which will always succeed and the method returns `true`
- It seem to me that the only relevant usecase for the `CreateInstance` is when `T` is a nullable value type and `targetType` is a reference type such that `Convert.ChangeType(new T(), targetType)` can convert `T` to `targetType`. Isn't that just in the cases when `T` is `object` though and in that case anything can be converted to `object`? The other type that works with `ChangeType` is `string` but you can't create an instance of string using `CreateInstance`.

I definitely need to revisit this later, I'm not 100% sure the simplified code is actually equivalent to the original one.

brminnick (approved):
Thanks for getting this started @simonrozsival!

I confirmed our Sample App runs on iOS when `<PublishAOT>enabled</PublishAOT>` with no problems.

simonrozsival:
@brminnick thanks a lot for finishing the PR!

PR #1858, opened 2024-05-04
Fix IOS crash on invalid or missing file extension on Resource Media Source

### LINKED ISSUE #1856: [BUG] MediaElement throws when setting source from Resource if file extension omitted (on iOS)

### Current Behavior

I accidentally omitted the ".wav" when specifying a sound file present in the Resources/Raw of my Maui project.  (That is, I called MediaSource.FromResource("whistling") instead of ("whistling.wav").)

The result was a remarkably long (for me) trace stack which led back to line 233 of MediaManager.macios.cs, where it calls
<code>string extension = Path.GetExtension(path)[1..];</code>
... but GetExtension returns an empty string.  The error is provoked by the attempt to start at index 1... of a zero-length string.  Which provokes a fatal crash of the application.

### Expected Behavior

If I had supplied (say) an empty string, or a bad file name, to MediaSource.FromResource, I'd merely have gotten a null <code>asset</code> and thus a null <code>PlayerItem</code>.  No sound, but also no app crash.  But a string with no extension end up causing a full-blown app crash instead.  Presumably this is not the desired behaviour.

### Steps To Reproduce

(No repro project provided - sorry, I have special needs kids calling.  This one seems straightforward enough not to need it.)

N/A (see above)

### Environment

```markdown
- .NET MAUI CommunityToolkit: Core is 8.0.1, MediaElement is 3.1.0
- OS: iOS 17.4.1
- .NET MAUI: 8.0 (I think?)
```

### Anything else?

Anyway, having traced the call stack back to there, it's pretty clear that this simply wants something along the lines of the following tweak to line 229 of MediaManager.macios.cs:
<code>if (!string.IsNullOrWhiteSpace(path) && !string.IsNullOrWhiteSpace(Path.GetExtension(path)))</code>
Or perhaps turning line 233 into:
<code>string extension = Path.GetExtension(path);
 extension = (!string.IsNullOrWhiteSpace(extension)) ? extension[1..] : string.Empty;</code>
(Such a shame that there's no equivalent for the ?[ operator for an _empty_ - but not null - string, or the change to line 233 would be trivial!)

ne0rrmatrix:
- Bug fix

 ### Description of Change ###
The primary change made to the code involves enhancing the validation of the `path` variable. Previously, the code checked if the `path` was neither null nor consisted solely of whitespace. The update extends this validation by also verifying that the `path` includes an extension. This adjustment is crucial for ensuring that operations such as retrieving the directory name, the filename without its extension, and the file extension itself are only executed on valid file paths. This modification aims to mitigate potential errors or unexpected outcomes when the code encounters paths that do not correspond to actual files, which would characteristically include extensions.

### Summary of Changes:
- Enhanced validation of the `path` variable to include a check for the presence of a file extension, in addition to the existing checks for null or whitespace content. This ensures that operations reliant on the path being a file (with an extension) are only performed on suitable paths, thereby improving the robustness of the code against invalid path inputs.

_Reference to Code Changes:_
- The condition for checking the `path` variable has been updated to include a check for a file extension.
 
### Linked Issues ###

 - Fixes #1856

 ### PR Checklist ###
 
Fixes issue with invalid or missing file extension for video source that caused BSOD on IOS when trying to set source without a file extension.

### DIFF
--- a/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
@@ -225,8 +225,8 @@ public partial class MediaManager : IDisposable
 		else if (MediaElement.Source is ResourceMediaSource resourceMediaSource)
 		{
 			var path = resourceMediaSource.Path;
-
-			if (!string.IsNullOrWhiteSpace(path))
+		
+			if (!string.IsNullOrWhiteSpace(path) && Path.HasExtension(path))
 			{
 				string directory = Path.GetDirectoryName(path) ?? "";
 				string filename = Path.GetFileNameWithoutExtension(path);
--- a/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
@@ -225,7 +225,7 @@ public partial class MediaManager : IDisposable
 		else if (MediaElement.Source is ResourceMediaSource resourceMediaSource)
 		{
 			var path = resourceMediaSource.Path;
-		
+
 			if (!string.IsNullOrWhiteSpace(path) && Path.HasExtension(path))
 			{
 				string directory = Path.GetDirectoryName(path) ?? "";
@@ -235,6 +235,10 @@ public partial class MediaManager : IDisposable
 					extension, directory);
 
 				asset = AVAsset.FromUrl(url);
+			}
+			else
+			{
+				Logger.LogWarning("Invalid file path for ResourceMediaSource.");
 			}
 		}
 


vhugogarcia (commented):
Thanks @ne0rrmatrix for adding this validation step. I think it won't hurt and it is pretty straightforward. 

cc: @jfversluis @brminnick

bijington:
How do the other platforms behave? Is the result of this PR that the setting of the resource silently fails? If so is that the desired behavior? If we are receiving an invalid source I think an appropriate exception or trace log would be a good option

ne0rrmatrix:
> How do the other platforms behave? Is the result of this PR that the setting of the resource silently fails? If so is that the desired behavior? If we are receiving an invalid source I think an appropriate exception or trace log would be a good option

This fix only applies to IOS and should have zero impact on any other platform. The code is only in the macios branch and should not be included in any other platform. As to how other platform behave I have not tested. I'm sure someone can file a bug report if there is an issue. With this PR I am focused on the issue that was reported and anything else is outside of scope.

bijington:
> This fix only applies to IOS and should have zero impact on any other platform. The code is only in the macios branch and should not be included in any other platform. As to how other platform behave I have not tested. I'm sure someone can file a bug report if there is an issue. With this PR I am focused on the issue that was reported and anything else is outside of scope.
> 
> 
> 
> Edit: I added a log message on error about invalid file paths if it fails.

Thanks, my question on how the other platforms behave was not to suggest we fix those also but to make sure that we are being consistent across all platforms. If Android throws an exception then I don't think iOS should silently fail.

I haven't been able to test on Android as my emulators/SDK is messed up so I might be sending us down the wrong path. I just want to make sure that we are considering the implication of the change.

I also don't see an issue with throwing an exception. Now it should be a specific exception providing a suitable message to help developers.

ne0rrmatrix:
> > This fix only applies to IOS and should have zero impact on any other platform. The code is only in the macios branch and should not be included in any other platform. As to how other platform behave I have not tested. I'm sure someone can file a bug report if there is an issue. With this PR I am focused on the issue that was reported and anything else is outside of scope.
> > Edit: I added a log message on error about invalid file paths if it fails.
> 
> Thanks, my question on how the other platforms behave was not to suggest we fix those also but to make sure that we are being consistent across all platforms. If Android throws an exception then I don't think iOS should silently fail.
> 
> I haven't been able to test on Android as my emulators/SDK is messed up so I might be sending us down the wrong path. I just want to make sure that we are considering the implication of the change.
> 
> I also don't see an issue with throwing an exception. Now it should be a specific exception providing a suitable message to help developers.

On both windows and android the OS itself outputs an error message saying something about an invalid file. It does nothing else. So with the changes to IOS it will be the same behavior across all platforms. Except tizen which I don't have the ability to test.

bijington:
Ok great then ignore my previous comments. Thanks for the effort in checking that

bijington (approved):
Thank you @ne0rrmatrix

jfversluis (approved):

PR #1870, opened 2024-05-09
fixed sample application crashed with MediaElement for Windows

olekssokol:
Sample application crashed in MediaElement page when the "Tab" key switches the active state to the "Change Aspect" button, and then the "Space" key executes it.
In this case, the problem is due to the DisplayActionSheet call from the Microsoft.Maui.Controls package, it should display the window and return the pressed button, but instead of the window, it immediately returns null, which was not handled in the Sample project.


olekssokol:
@dotnet-policy-service agree

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -51,8 +51,8 @@
     <!-- Custom Fonts -->
     <MauiFont Include="Resources\Fonts\*" />
 
-    <PackageReference Include="Microsoft.Maui.Controls" Version="[$(MauiPackageVersion),)"/>
-    <PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="[$(MauiPackageVersion),)"/>
+    <PackageReference Include="Microsoft.Maui.Controls" Version="8.0.21" />
```
brminnick, line 54: Is the root cause a MAUI bug that is fixed in 8.0.21? If so, we'll want to bump the `MauiPackageVersion` variable in `Directory.Build.props`.

bijington: But we want to be wary of when `8.0.21` will ship in Visual Studio based on the last community standup

olekssokol: @brminnick, yes, thanks, you're right. I just sent a commit with changes to Directory.Build.props

brminnick: Thanks @olekssokol!

@bijington - yup! I'll add the `blocked` tag so that we don't accidentally merge this PR until 8.0.24 is included in the Visual Studio (stable) release.

brminnick (dismissed):
Confirmed that the templates for the current release of Visual Studio (stable) are using `Microsoft.Maui.Controls v8.0.21`

brminnick:
Dang - in `Microsoft.Maui.Controls v8.0.21` there is a breaking change in how `Application` is loaded which is breaking our Unit Tests. There's an internal field in `Application` that needs to be initialized before our tests can run: `readonly Lazy<IResourceDictionary> _systemResources;`. Currently it is throwing a `NullReferenceException`.

brminnick (approved):

brminnick:
Ok - there's one more bug we need to fix with the AppTheme test now.

I just spent the last few hours digging into it and couldn't figure it out. @olekssokol any chance you could take a look?

olekssokol:
Yes, of course, I'll take a look at it

brminnick (approved):
Thanks @olekssokol! I finally figured out how to get the Unit Tests fixed ✅

PR #1782, opened 2024-03-31
Lock Screen Controls and Metadata Support for Windows, Android, IOS and Mac Catalyst

ne0rrmatrix:
- ## Feature/Proposal ##
## Description of Change ##
Add Lock Screen Media Controls to Windows, Android, iOS, and Mac Catalyst. Metadata like Title, Artist, and Album art will be displayed for playing media. 

 ## Linked Issues ##

 - Fixes #1682 
 - Fixes #1667
 - Fixes #1765

 ## PR Checklist ##
 
 - [x] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [ ] Has tests (if omitted, state reason in description)
 - [x] Has samples (if omitted, state reason in description)
 - [x] Rebased on top of `main` at time of PR
 - [x] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#contributing-code---best-practices)
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/414

 ## Additional information ##
Add support for Lock Screen Controls to Windows, Mac, IOS, and Android. Support for adding Metadata in XAML or code behind.  </br>

<p align="center">
Example usage in XAML:
</p>

```csharp
 <toolkit:MediaElement
     x:Name="MediaElement"
     WidthRequest="{OnIdiom Tablet=410, Default=380}"
     ShouldAutoPlay="True"
     Source="https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"
     MetaDataArtworkUrl="https://michellevella.com/cdn/shop/products/BeatlesLonelyHearts4UP_1728x.jpg?v=1660582971"
     MetaDataTitle="Big Buck Bunny"
     MetaDataArtist="Blender Foundation"
     SourceType="Audio"
   MediaEnded="OnMediaEnded"
     MediaFailed="OnMediaFailed"
     MediaOpened="OnMediaOpened"
     PositionChanged="OnPositionChanged"
     StateChanged="OnStateChanged"
     SeekCompleted="OnSeekCompleted"/>
```
<p align="center">
Example in code behind:
</p>

```csharp
async void ChangeSourceClicked(Object sender, EventArgs e)
{
	var result = await DisplayActionSheet("Choose a source", "Cancel", null,
		loadOnlineMp4, loadHls, loadLocalResource, resetSource);

	switch (result)
	{
[...]

PR #1905, opened 2024-05-28
Add initial media options class and extension

bijington:
### Description of Change ###

 

The aim of this PR is to highlight how we can initially make `MediaElement` more configurable but we should consider how this could enable developers to configure any of our internal defaults classes.

 ### Linked Issues ###
 

 - Fixes #

 ### PR Checklist ###
 

bijington:
@ne0rrmatrix this was my initial thinking, I have noticed that we don't currently set these values when controlling playback within our codebase but I believe your PR #1782 does.

@VladislavAntonyuk what are your thoughts on expanding this out of a `MediaOptions` class to enable support for also controlling the SpeechToText `AVAudioSession` values?

@everyone what do you think to expanding our options classes to allow for configuring all defaults within the toolkit? We already have classes that represent the defaults e.g. `TouchBehaviorDefaults` if we make this configurable it could make the usage of the controls simpler.

src/CommunityToolkit.Maui.MediaElement/Options/MediaOptions.macios.cs
```diff
@@ -0,0 +1,26 @@
+using AVFoundation;
+
+namespace CommunityToolkit.Maui;
+
+partial class MediaOptions
```
pictos, line 5: I would say those `Options` should be immutable. So we can refactor it to be a record and remove the `setter`s

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.shared.cs
```diff
@@ -19,6 +19,8 @@ namespace CommunityToolkit.Maui.Core.Views;
 /// </summary>
 public partial class MediaManager
 {
+	internal static MediaOptions DefaultMediaOptions { get; set; } = new();
```
pictos, line 22: Here's is fine to be mutable, since users may want to change how options based on the `MediaSource`. But, it does make sense on the current implementation? The `MediaOptions` is only configurable on `AppBuilder`, so it shouldn't change after app startup, right?

pictos (commented):
I really like the idea of extending our features using these `Options` objects. I believe it will help devs to customize the behavior of the feature easily.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.shared.cs
brminnick: Agreed. Let's make these `{ get; private set;}` for now to follow the same architecture in `CommunityToolkit.Maui.Options`: https://github.com/CommunityToolkit/Maui/blob/d6d6e187b4f6136558bc3be2d2ecd8cca06fc96f/src/CommunityToolkit.Maui/Options.cs#L19-L22

We can always make it a `public set;` in the future without breaking changes.

ne0rrmatrix: Do we want to add bindable properties so we can use this at runtime for changes with `MediaSource` ?

ne0rrmatrix:
I was just thinking a good first option would be add the media element service:
```
internal static bool ShouldEnableServiceOnAndroid { get; private set; }

/// <summary>
/// Sets the value to enable the MediaElement service on Android.
/// </summary>
/// <remarks>
/// Default value is false.
/// </remarks>
public void SetShouldEnableServiceOnAndroid(bool value)
{
	if (value is true && builder is null)
	{
		throw new InvalidOperationException("The MauiAppBuilder must be provided to enable Service on Android.");
	}
	ShouldEnableServiceOnAndroid = value;
}
```

ne0rrmatrix:
I am also thinking we should follow the same pattern we have for Core and Main community toolkit

bijington:
I completely forgot this Pr existed. Thanks for the reminder. I'm happy to keep things consistent

PR #1932, opened 2024-06-13
Set unloaded event to make sure popup is shown on Windows.

### LINKED ISSUE #1931: [BUG] [Windows] The second popup will fail to show after MAUI 8.0.60

### Current Behavior

This is a very strange issue. Since MAUI version 8.0.60 the second popup will fail. Only the second time you want to show a popup will fail. The popup will not show up and if you are using CommunityToolkit.MVVM [RelayCommand] the button (triggering the popup) will be disabled, because the Task will not return. See video of the problem.

https://github.com/CommunityToolkit/Maui/assets/1301586/0cc33ecd-a70e-40fc-a295-5f2255d4e126

### Expected Behavior

The popup should always be shown also the second popup.

### Steps To Reproduce

1. Download latest version of CommunityToolkit
2. Open Directory.Build.props in root folder of project
3. Change MauiPackageVersion property to 8.0.60
4. Open Visual Studio and run the sample application
5. Go to Views / Multiple Popups Page
6. Click on "Simple Popup"
7. Close the popup by clicking outside the popup
8. Click on "C# Binding Popup"
9. The button will be disabled and the popup will not appear

https://github.com/CommunityToolkit/Maui

### Environment

```markdown
- .NET MAUI CommunityToolkit: 9.0.1
- OS: Windows 11 (net8.0-windows10.0.19041.0)
- .NET MAUI: 8.0.60
```

### Anything else?

To prevent the button from disabling while using CommunityToolkit.MVVM is to use [RelayCommand(AllowConcurrentExecutions = true)] instead of [RelayCommand]

GuidoNeele:
### Description of Change ###

 
This fixes an issue with MAUI 8.0.60 and later where the `Page.Loaded` event isn't triggering resulting in a hanging Task.

 ### Linked Issues ###
 

 - Fixes #1931 

 ### PR Checklist ###
 
The question is should this be added to the CommunityToolkit or be fixed in MAUI itself. At the moment Popups don't work as expected when used with 8.0.60 and higher on Windows. This fixes that issue. The code is already a workaround so adding the Unloaded events is probably not that exciting and could possibly be removed if issue https://github.com/dotnet/maui/issues/12970 is fixed.

### DIFF
--- a/src/CommunityToolkit.Maui/Views/Popup/PopupExtensions.shared.cs
+++ b/src/CommunityToolkit.Maui/Views/Popup/PopupExtensions.shared.cs
@@ -63,6 +63,7 @@ public static partial class PopupExtensions
 
 		async void handler(object? sender, EventArgs args)
 		{
+			page.GetCurrentPage().Unloaded -= (_, _) => { };
 			page.GetCurrentPage().Loaded -= handler;
 
 			try
@@ -76,6 +77,8 @@ public static partial class PopupExtensions
 				taskCompletionSource.TrySetException(ex);
 			}
 		}
+
+		page.GetCurrentPage().Unloaded += (_, _) => { };
 		page.GetCurrentPage().Loaded += handler;
 
 		return taskCompletionSource.Task.WaitAsync(token);


brminnick:
Thanks @GuidoNeele! What is the root cause of the bug?

I'm not sure I understand how subscribing to an empty function helps fix a bug.

bijington:
It seems strange for .NET MAUI to expect a level of symmetry in listening for both `Loaded` and `Unloaded` events.

I wonder if any of the team can comment on this

src/CommunityToolkit.Maui/Views/Popup/PopupExtensions.shared.cs
```diff
@@ -63,6 +63,7 @@ void handler(object? sender, EventArgs args)
 
 		async void handler(object? sender, EventArgs args)
 		{
+			page.GetCurrentPage().Unloaded -= (_, _) => { };
```
pictos, line 66: this will not work as expected, the compiler will create a new delegate to represent this one and it will be different from the delegate subscribed at line `81`. In order words, it will never unsubscribe the various `Unloaded` registrations causing a memory leak.
The fix is to create a local variable that will represent this delegate and use it

pictos, line 66: this will not work as expected, the compiler will create a new delegate to represent this one and it will be different from the delegate subscribed at line `81`. In order words, it will never unsubscribe the various `Unloaded` registrations causing a memory leak.
The fix is to create a local variable that will represent this delegate and use it

pictos (changes requested):
Not sure if we'll move forward with this workaround, but if we do this need to be fixed

src/CommunityToolkit.Maui/Views/Popup/PopupExtensions.shared.cs
GuidoNeele: Yep I was in doubt about that. Will fix it.

brminnick (approved):
This is certainly a strange bug, but I confirm the bug exists and that the workaround does fix it. 

We'll continue to keep an eye on https://github.com/dotnet/maui/issues/12970, which will allow us to remove the entire workaround if/when it is fixed.

PR #1950, opened 2024-06-18
Reinstate MediaElement MediaWidth & MediaHeight functionality

### LINKED ISSUE #1708: [BUG] MediaElement.MediaWidth and MediaElement.MediaHeight are always 0

### Current Behavior

The values of MediaWidth and MediaHeight are always 0, or at least start as 0 if there is some non-obvious way to get them to get set to the dimensions of the media. The MediaOpened and SizeChanging events get triggered, but the properties are always 0 in those as well.

Similarly, the MediaElement itself is not visible (Height=.36...) unless the HeightRequest is explicitly set or maybe some other non-obvious conditions have to be met.

### Expected Behavior

The MediaWidth and MediaHeight properties should reflect the dimensions of the media the source points to.

Also, in my opinion, if there is no HeightRequest set then the MediaElement should size itself to the same aspect ratio as the media based on however it gets its width, at least assuming the Aspect property is AspectFit or AspectFill. At the very least it should probably just make itself square. And that could work even if the Aspect property was set to Fill.

In other words, I think that by default the MediaElement should display the media set in the Source property. But even if that isn't possible, it would help for the two properties above to reflect the dimensions so that I can calculate it myself (this is what I had to do in Xamarin).

### Steps To Reproduce

1. Open and solution from reproduction repository.
2. Run the MediaElement project.
3. Enter the URL for a video file in the Entry below the race car. The video used in a lot of MediaElement examples can be used: https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4
4. The State should display Paused, since the video isn't playing (and you can't play it because it isn't visible).
5. MediaWidth and MediaHeight both display 0.

https://github.com/bzd3y/ReproductionProject

### Environment

```markdown
- .NET MAUI CommunityToolkit: 3.0.1
- OS: Android Api 34. I haven't tried on iOS.
- .NET MAUI: 8.0.3
```

### Anything else?

_No response_

jfversluis:
### Description of Change ###

It seems like we never implemented the functionality to populate `MediaWidth` and `MediaHeight`. This PR fixes that.

While I did change some APIs I don't think it should be breaking changes.

* Populate `MediaWidth` and `MediaHeight` for iOS, Android, macOS, Windows and Tizen
  * Might not always show for live streamed content on iOS/macOS
* Moved the `Position` and `Duration` labels in the sample app to under the slider that also shows the current playback position
* Added a field right under the video to show the dimensions
* Switched around some properties in `MediaElement.shared.cs` so that we can set the values from our `MediaManager` but they are read-only properties to the outside work
* Updated comment on the above properties to reflect that sometimes width/height are not reported for iOS/macOS

 ### Linked Issues ###

 - Fixes #1708

 ### PR Checklist ###

 - [x] Has tests
 - [x] Has samples
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/430

### DIFF
--- a/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
@@ -3,6 +3,7 @@ using AVKit;
 using CommunityToolkit.Maui.Core.Primitives;
 using CommunityToolkit.Maui.Views;
 using CoreFoundation;
+using CoreGraphics;
 using CoreMedia;
 using Foundation;
 using MediaPlayer;
@@ -292,10 +293,9 @@ public partial class MediaManager : IDisposable
 		{
 			MediaElement.MediaOpened();
 
-#if IOS
-			MediaElement.MediaWidth = (int?)Player.CurrentItem?.Asset.NaturalSize.Width ?? 0;
-			MediaElement.MediaHeight = (int?)Player.CurrentItem?.Asset.NaturalSize.Height ?? 0;
-#endif
+			var mediaDimensions = GetVideoDimensions(PlayerItem);
+			MediaElement.MediaWidth = mediaDimensions.Width;
+			MediaElement.MediaHeight = mediaDimensions.Height;
 
 			if (MediaElement.ShouldAutoPlay)
 			{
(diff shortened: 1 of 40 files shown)


vhugogarcia:
Thanks @jfversluis for adding the functionality to get the media width and height. 

I ran a few tests and it worked very good on Android, iOS and MacCatalyst. I'm unable to test it on Windows, maybe somebody else can give us a hand with the Windows testing.

jfversluis:
@ne0rrmatrix would you mind maybe testing this for a bit? Especially Windows but if you have some time poke at the rest too just so we have an extra set of eyes that looked at this :)

ne0rrmatrix:
I'll take a look at it now

ne0rrmatrix:
Works exactly as expected on Android and Windows. On iOS it behaves as expected on everything except HLS where it shows 0x0 and does not get width or height. I'm Going to test on a physical device to double check this is not a bug in emulator.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
```diff
@@ -624,4 +631,38 @@ void RateChanged(object? sender, NSNotificationEventArgs args)
 			}
 		}
 	}
+
+	(int Width, int Height) GetVideoDimensions(AVPlayerItem avPlayerItem)
```
ne0rrmatrix, line 635: Do we want to use a tuple? Or do we want to use a `Record Struct`

jfversluis:
> except HLS where it shows 0x0 and does not get width or height

This is (unfortunately) expected

Issue #2005, opened 2024-07-08
[BUG] Microsoft.Maui.Controls.Compatibility dependency


filipnavara:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [X] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

CommunityToolkit.Maui has a transitive dependency to `Microsoft.Maui.Controls.Compatibility` NuGet package. It means that consumers of `CommunityToolkit.Maui` now get the reference whether they want it or not.

Cursory test seems to show that there's very little code depending on the compatibility package (`IVisualNativeElementRenderer` usage in `TouchBehavior.macios.cs`) and `Color.ToAndroid()` extension method. The `ToAndroid` method usage can be replaced with `ToPlatform` extension method in the base MAUI package. Usage of `IVisualNativeElementRenderer` may be a bit more difficult to get rid of but probably solvable by getting rid of that compatibility kludge, using reflection, or rewriting it to detect the situation from the view hierarchy. 

### Expected Behavior

No dependency on `Microsoft.Maui.Controls.Compatibility`.

### Steps To Reproduce

1. Add `CommunityToolkit.Maui` NuGet dependency.
2. Observe that `Microsoft.Maui.Controls.Compatibility` gets transitively included.

### Link to public reproduction project repository

-

### Environment

```markdown
- .NET MAUI CommunityToolkit: 9.0.0
- OS: Android, iOS
- .NET MAUI: 8.0.61
```

BaY1251:
Great

GeorgeLeithead:
@vhugogarcia @brminnick I believe that this has already been merged into main and released, and as such i think this can be closed?

brminnick:
Thanks George! Yup - this is completed via https://github.com/CommunityToolkit/Maui/pull/2006

PR #2006, opened 2024-07-08
Remove `Microsoft.Maui.Controls.Compatibility` Dependency 

filipnavara:
### Description of Change ###

Replace one extension with another one to reduce dependencies on compatibility package.

 ### Linked Issues ###

 - Contributes to #2005

 ### PR Checklist ###

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,7 +27,7 @@ variables:
   PathToCommunityToolkitAnalyzersUnitTestCsproj: 'src/CommunityToolkit.Maui.Analyzers.UnitTests/CommunityToolkit.Maui.Analyzers.UnitTests.csproj'
   DotNetMauiRollbackFile: 'https://maui.blob.core.windows.net/metadata/rollbacks/8.0.6.json'
   CommunityToolkitSampleApp_Xcode_Version: '15.2.0'
-  CommunityToolkitLibrary_Xcode_Version: '15.2.0'
+  CommunityToolkitLibrary_Xcode_Version: '15.3.0'
 
 trigger:
   branches:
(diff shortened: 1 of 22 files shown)


brminnick:
Hey @JoonghyunCho! Could you help me fix this error on the Tizen implementation of MediaElement? 

We're looking to remove our dependency on `Microsoft.Maui.Controls.Compatibility`, and it looks like `ResourcePath.GetPath()`  is no longer available in .NET MAUI without the compatibility NuGet Package. 

Could you see about retrieving the path to the `Resources` folder without using the `ResourcePath` class?

filipnavara:
Note the the `<UseMaui>true</UseMaui>` property still brings in the dependency. That needs to be removed and we need to fix the iOS build of TouchBehavior.

brminnick:
> we need to fix the iOS build of TouchBehavior

What's the problem we need to fix for this?

filipnavara:
> What's the problem we need to fix for this?

https://github.com/CommunityToolkit/Maui/blob/cc1fb686815587076bbad45a1903f3ec3039315d/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.macios.cs#L28-L32

`IVisualNativeElementRenderer` comes from the compatibility package. Now I would argue that it has no business being there and it was just copied from the Xamarin.CommunityToolkit, but it may be more complex. If we want to support it on compatibility controls that use `UIButton` then it may need some additional changes/tests.

brminnick (dismissed):
Thanks @filipnavara!!

brminnick (approved):

filipnavara:
Thanks for pushing it through and fixing the last bits! Much appreciated!

Issue #2050, opened 2024-07-22
[DOC] Update Sample Application to demonstrate correct way to initialize Maps in MauiProgram.cs


mikelor:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [X] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

The current code [in MauiProgram.cs](https://github.com/mikelor/CommunityToolkit-Maui/blob/5dd3ec5bbc6cdcc1d5e06f9a9050e017815c94ce/samples/CommunityToolkit.Maui.Sample/MauiProgram.cs#L70) does not enclose the UseCommunityToolkitMaps("key") in a #ifdef WINDOWS directive.

The [2.0.1 Release](https://github.com/CommunityToolkit/Maui/releases/tag/2.0.1-maps) of the Community Toolkit Maps component introduced a change that requires the #ifdef Windows directive when compiling a multiple platform application with Maps.

This can cause confusion for people trying to incorporate Maps into the application due to the fact that since [Maui 8.0.10](https://github.com/dotnet/maui/releases/tag/8.0.10) now throws an exception on Windows when using the standard Maps component.

This can cause hours of wasted effort tracking down the cause of the exception.

### Expected Behavior

The current Sample code would include the #ifdef compiler directive for WINDOWS around the `.UseCommunityToolkitMaps("key")` method.

### Further Discussion
Unsure whether to add just the `#ifdef WINDOWS` directive or also include the `#IFDEF IOS || ANDROID` as well. If doing the latter, we would also want to update the appropriate platforms to include setting API Keys.

### Steps To Reproduce

1. Add Maps support for Android by calling `.UseMauiMaps()`.
2. Get an exception when running on Windows that Maps is not supported on Windows.

### Link to public reproduction project repository

https://github.com/mikelor/CommunityToolkit-Maui/

### Environment

```markdown
- .NET MAUI CommunityToolkit: v9.0.2
- OS: Windows 11
- .NET MAUI: 8.0.70
```

jfversluis:
Fix was merged for the docs a while ago but seems this wasn't closed.

PR #2051, opened 2024-07-22
Fix for #2050 - Add explicit compiler directive to make Maps sample guidance clearer.

mikelor:
### Description of Change ###

 
Fix for #2050 by adding a compiler directive for Windows platform around `.UseCommunityToolkitMaps()` method call.

 ### Linked Issues ###
 
#2050 

 - Fixes #

 ### PR Checklist ###
 
 - [X] Has samples (if omitted, state reason in description)


samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
```diff
@@ -67,7 +67,11 @@ public static MauiApp CreateMauiApp()
 								.UseMauiCommunityToolkitMarkup()
 								.UseMauiCommunityToolkitCamera()
 								.UseMauiCommunityToolkitMediaElement()
+
+#if WINDOWS
```
VladislavAntonyuk, line 71: it looks not correct. we need to call UseMauiMaps() if not Windows

bijington: Also what will happen if a user doesn't supply a valid key? If it results in an exception I don't think think should be the default behavior, the number of users interested in using maps will be a subset and it would be annoying if it throws an exception

VladislavAntonyuk: it won't throw an exception but displays an error message in the middle of the page, but Maps is usable. the exception is thrown only if the key is null or empty

VladislavAntonyuk (changes requested):
please fix the sample for all platforms

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
mikelor: I thought about adding the directives for Android and IOS.

This would also require modifying the Manifest file for Android as well as the IOS implementation.

Lastly I would change the title of the menu items and pages to remove the word "Windows"

I decided against it since the The Community Toolkit only provides the Windows implementation.

It felt like I was changing the purpose of the sample.

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
```diff
@@ -68,9 +68,13 @@ public static MauiApp CreateMauiApp()
 								.UseMauiCommunityToolkitCamera()
 								.UseMauiCommunityToolkitMediaElement()
```
VladislavAntonyuk, line 70: MacCatalyst?

mikelor: I'm unable to validate that. 
In addition, the 2.0.1 release notes to CommunityToolkit.Maui.Maps only referenced WINDOWS, IOS, and ANDROID.

[CommunityToolkit.Maui.Maps 2
0.1 Release Notes](https://github.com/CommunityToolkit/Maui/releases/tag/2.0.1-maps)

VladislavAntonyuk: Please add MacCatalyst. It works.

VladislavAntonyuk (commented):
Left 1 comment. All other looks good.

mikelor:
Thanks for the feedback. Pull request has been updated.

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
mikelor: Thanks for the feedback. Pull Request amended.

bijington:
Did this get closed for a reason? Sorry I think we were inactive, I am happy to proceed with this if you wanted to reopen it?

mikelor:
I needed to do some other work and ended up killing this. I'll submit another one tomorrow (Friday)

PR #2113, opened 2024-08-11
Fix memory leak on touchBehavior android

### LINKED ISSUE #2082: [BUG] TouchBehavior Memory Leak once the behavior has been included in a DataTemplate

### Current Behavior

I have noticed the following problem. As soon as I want to use the touch behavior from the CommunityTool in a DataTemplate, the GarbageCollector no longer manages to clear this and all parent elements from memory.
This problem does not exist with the GestureRecognizer.

I provide a small example application for this.

### Expected Behavior

The expected behavior would be that the GarbageCollector manages to delete the TouchBehavior from memory.

### Steps To Reproduce

In this sample application you can navigate from the MainPage to a page with the implementation of the GestureRecognizer and to a page with the implementation of the Touch behavior. As soon as you navigate back from the respective page to the MainPage, the current memory is displayed in a list. Here you can see the exponential increase in memory when you navigate back and forth between the MainPage and the page with the Touch behavior. If you navigate between the MainPage and the page with the implementation of the GestureRecognizer, the memory remains constant.

I have also implemented the MemoryToolKit by Adam Essenmacher to observe in the debug output which objects are successfully thrown out of memory and which are not.

https://github.com/sstobinski/TouchBehavior_MemoryLeak

### Environment

```markdown
- .NET MAUI: 8.0.61/8.0.100         SDK 8.0.300
- Microsoft.Maui.Controls: 8.0.61
- Microsoft.Maui.Controls.Compatibility: 8.0.61
- .NET MAUI CommunityToolkit: 9.0.2
- AdamE.MemoryToolkit.Maui: 1.0.0
```

### Anything else?

_No response_

pictos:
### Description of Change ###

This PR fixes the memory leak on Android's implementation of TouchBehavior, and adds a new type `WeakWrapper`, that is to help maintainer and contributors to create and use `WeakReference` types.

From my investigations the issue was because of the strong `TouchBehavior` reference inside the `AccessibilityListener`, changing to be a `WeakReference` fixes the leak, as you can see in the video below.

https://github.com/user-attachments/assets/5b19a40e-3626-4ae2-a461-33a7e7238557

also, from the video, you can see it will say there's a leak, but in fact the object will take a while to be free, probably it implements a `Finalizer` which will make it live longer. Just saying, because some user can complain about it.

 ### Linked Issues ###
 

 - Fixes #2082

 ### PR Checklist ###
 

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs
```diff
@@ -118,7 +119,7 @@ static ColorStateList GetColorStateList(MColor? color)
 
 	void UpdateClickHandler()
 	{
-		if (view is null || !view.IsAlive())
+		if (view is null)
```
pictos, line 122:
```suggestion
		if (view is null || !view.IsAlive())
```

VladislavAntonyuk (dismissed):

brminnick (dismissed):
Thanks Pedro!! 

Just FYI - I removed `WeakWrapper` since we were only using it once in `TouchBehavior.android.cs`; we can always add it later if we end up relying on `WeakReference<T>` more. 

I also updated `Dispose(bool)` to properly dispose of the `WeakReference.Target` be setting it to `null`.

pictos:
@brminnick the Dispose method isn't being called, because of that I used the `WeakReference`, with that there's no need to clean anything. The reason the cleanup isn't being executed is because the `Element` is set to null before the `OnDetaching` method being called. I strongly recommend to revert the change to previous commits

pictos:
> Is this a bug in CommunityToolkit.Maui? If so, we should fix it.

I'm not sure if it's a bug or the way the BaseBehavior works, I assumed isn't a bug. The `Element` [here](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/TouchBehavior.android.cs#L78) is `null` because [this piece](
https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui/Behaviors/ICommunityToolkitBehavior.shared.cs#L70) is executed before, leading the clean-up to leave early.

So, there's no bug in the .net-android

> It's still a best practice to set WeakReference.Target to null when we are disposing the object; doing this does not introduce any harm

not sure about this one... But in this case, it doesn't make any difference just add more code that will not be executed

brminnick (approved):

rafalka:
Hi, 
Good job but there's another memory leak on Android...
Take look into `OnDetachedFrom()`
```CSharp
	protected override void OnDetachedFrom(VisualElement bindable, AView platformView)
	{
		base.OnDetachedFrom(bindable, platformView);

		view = platformView;

		if (Element is null)
		{
			return;
		}
...
```
We exiting when `Element` is null.
The thing is that  `Element`  is always null because it returns a `View`  `base.OnDetachedFrom()` calls `UnassignViewAndBingingContext()` which sets it to null. 

I think that this null check is spare.

Issue #2054, opened 2024-07-23
[♻️ Housekepping] Until .NET 9 we should avoid Application.MainPage


pictos:
[The .NET MAUI team is deprecating](https://github.com/dotnet/maui/pull/23692) the usage of `Application.MainPage` because it, normally, isn't what devs/we expect to be. In a MultiWindow scenario it can return the wrong Page.

One way to workaround this is to change `Application.Current.MainPage` to `Application.Current.Windows[0].Page`.

At least for me, isn't 100% clear what `Windows[0]` means: 
 - it's the first window at collection
 - it's the topmost window
 - it's the focused window

For join the discussion at .NET MAUI repo, please go [here](https://github.com/dotnet/maui/issues/23518).

Maybe, today we already have a bug for multi-window apps, but since no user has reported any issue we don't need to rush on this.

VladislavAntonyuk:
Seems like the issue is solved, we don't use MainPage in solution

PR #2128, opened 2024-08-18
Make iOS StatusBarBehavior more robust and not throw exceptions

jfversluis:
### Description of Change ###

While working on a Blazor Hybrid project I implemented the StatusBarBehavior. I also added the new way of setting the root page for a .NET MAUI app by not using MainPage (also see #2054)

I ran into an issue that when setting the status bar color from XAML, I would get an exception. Probably due to the fact that the window isn't created yet, but the page is being created that wants to set the status bar.

To be ahead of the curve for this new change in .NET MAUI, I adapted this code a little so that it doesn't thrown an exception but handles it more gracefully. The actual functionality still seems to work fine.

### DIFF
--- a/samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/StatusBarBehaviorPage.xaml
+++ b/samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/StatusBarBehaviorPage.xaml
@@ -86,7 +86,7 @@
             <Label Text="Red" />
             <Slider
                 Margin="20,0"
-                Maximum="255"
+                Maximum="1"
                 MaximumTrackColor="Red"
                 Minimum="0"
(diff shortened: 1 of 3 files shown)


pictos (approved):

brminnick (approved):
Thanks Gerald! 

Just FYI - I fixed a bug on the StatusBarBehavior page. The min/max values were incorrect, causing `StatusBarColor` to always return `[Color: Red=1, Green=1, Blue=1, Alpha=1]` instead of the actual color requested by the sliders. 

I also renamed `UpdateStatusBarAppearance ` to `TryUpdateStatusBarAppearance` since it is no longer guaranteed that the method will succeed. And I have it now print to `Trace.WriteLine` when it fails.

PR #2140, opened 2024-08-21
Remove log messages that were confusing and not required.

ne0rrmatrix:
- Bug fix

 ### Description of Change ###

Remove confusing log Message in `MediaElement` that led to questions about app behavior.

 ### PR Checklist ###

There was a question asked on StackOverflow about the error message and its possible implications in an app crash from a developer. The log message displayed an error about a class being deprecated as of Android 33. This log message has no impact on behavior and has confused developers about the apps behavior. This PR removes the message as it is not needed and is confusing people

### DIFF
--- a/src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
@@ -230,7 +230,6 @@ class MediaControlsService : Service
 	{
 		if (Build.VERSION.SdkInt >= BuildVersionCodes.Tiramisu)
 		{
-			System.Diagnostics.Trace.WriteLine($"{LocalBroadcastManager.GetInstance} not supported on Android 13 and above.");
 			return;
 		}
 		var intent = new Intent(receiver);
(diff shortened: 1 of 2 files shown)


src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
```diff
@@ -703,7 +703,6 @@ void BroadcastUpdate(string action)
 	{
 		if (Build.VERSION.SdkInt >= BuildVersionCodes.Tiramisu)
 		{
-			Logger.LogError("{LocalBroadcastManager} not supported on Android 13 and above.", typeof(LocalBroadcastManager));
```
bijington, line 706: Is it worth a comment here for future developers to indicate why we don't need to do anything on later versions?

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
```diff
@@ -230,7 +230,6 @@ static void BroadcastUpdate(string receiver, string action)
 	{
 		if (Build.VERSION.SdkInt >= BuildVersionCodes.Tiramisu)
 		{
-			System.Diagnostics.Trace.WriteLine($"{LocalBroadcastManager.GetInstance} not supported on Android 13 and above.");
```
bijington, line 233: Is it worth a comment here for future developers to indicate why we don't need to do anything on later versions?

ne0rrmatrix: You linked a comment on Stack Overflow in Discord where someone was confused about these messages. I figured since they were not required as the code is never run in those scenarios putting the trace message on screen served no purpose.

bijington: I meant add something like the following:

`// on versions above 33 this is no longer required, this functionality is provided by XYZ`

brminnick (approved):

PR #2174, opened 2024-09-04
Add missing setter for AnimateCommand

BretJohnson:
### Description of Change ###

Fix https://developercommunity.visualstudio.com/t/Hot-reload-stops-working-when-using-Comm/10707756

Control properties should have both a getter and setter so that Hot Reload (and C# code) can update them.
AnimateCommand was missing the setter - likely just an oversight. Hot Reload now works with this change.

 

 ### Linked Issues ###
 - Fixes # https://developercommunity.visualstudio.com/t/Hot-reload-stops-working-when-using-Comm/10707756

 ### PR Checklist ###
 
### DIFF
--- a/src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
@@ -28,7 +28,10 @@ public class AnimationBehavior : EventToCommandBehavior
 	/// <remarks>
 	/// <see cref="AnimateCommand"/> has a <see cref="Type"/> of Command&lt;CancellationToken&gt; which requires a <see cref="CancellationToken"/> as a CommandParameter. See <see cref="Command{CancellationToken}"/> and <see cref="System.Windows.Input.ICommand.Execute(object)"/> for more information on passing a <see cref="CancellationToken"/> into <see cref="Command{T}"/> as a CommandParameter"
 	/// </remarks>
-	public Command<CancellationToken> AnimateCommand => (Command<CancellationToken>)GetValue(AnimateCommandProperty);
+	public Command<CancellationToken> AnimateCommand {
+		get => (Command<CancellationToken>)GetValue(AnimateCommandProperty);
+		set => SetValue(AnimateCommandProperty, value);
+	}
 
 	/// <summary>
 	/// The type of animation to perform.


brminnick (approved):

bijington:
@BretJohnson thanks for this. We had the property as read-only because we never wanted it set by developers, I know that we cover this by https://github.com/CommunityToolkit/Maui/blob/19cfdc65d446cb34cc3a193fcbc140207f9ce805/src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs#L21 but does this have any potential side effects?

BretJohnson:
@bijington - What's the reason you didn't want it set by developers? It's of course set by developers in the XAML, like here: https://github.com/CommunityToolkit/Maui/blob/19cfdc65d446cb34cc3a193fcbc140207f9ce805/samples/CommunityToolkit.Maui.Sample/Pages/Behaviors/AnimationBehaviorPage.xaml#L83

And developers that create the effect in C# code probably want a way to set it.

bijington:
Oh sorry that may have been a bad choice of words. I meant that we want developers to be able to bind to the property but always receive our instance of the command, so we are binding `OneWayToSource`. I understand the setter is something separate but we do not want developers to write something like:

```csharp
var animationBehavior = new AnimationBehavior();
animationBehavior.AnimateCommand = new Command(() => blah);
```

The purpose of this command is to allow the triggering of the animation from the view model and not the typical direction of view interaction into the view model.

I am actually intrigued how this ever compiled let alone worked if we didn't provide a setter?

BretJohnson:
cc @etvorun, who may have some thoughts on the best way to handle this

bijington:
Perhaps we are using Commands in a direction that they were initially intended but the use of a boolean flag to trigger an animation like this feels yuckier. I am happy for guidance on a good alternatives to this approach.

bijington:
Can Hot Reload work with an `internal` setter rather than a `public` one?

BretJohnson:
FYI - WPF disallows OneWayToSource bindings with ReadOnly properties. See https://stackoverflow.com/questions/658170/onewaytosource-binding-from-readonly-property-in-xaml, for instance. However, MAUI appears to allow them, as that's what's getting used here. @evtorun and I are discussing now how to best handle this. This might be best fixed in the Hot Reload code, removing the check for an accessible setter in the case of ReadOnly OneWayToSource bound properties. We're investigating doing that.

@brminnick  - can you revert this PR for now; thanks
@bijington - thanks for quick and helpful input here, describing exactly what you're doing

BretJohnson:
> Can Hot Reload work with an `internal` setter rather than a `public` one?

Hot Reload requires the setter to be public or internal with InternalsVisibileTo set in the assembly. internal with InternalsVisibileTo may be a fallback if we can't easily fix on the Hot Reload side. I'll create a separate PR in the future in the unlikely case we want to go that route.

brminnick:
I'll update the setter to `internal`, then 👍

BretJohnson:
> I'll update the setter to `internal`, then 👍
@brminnick - For now, can you just revert the PR.
`internal` won't work unless `InternalsVisibleTo` is also set. But we won't need any of that if we fix on the Hot Reload side.
I can create a new PR later if we want to make any changes here, but I think we'll be able to fix it in the Hot Reload code.

brminnick:
Reverted ✅

bijington:
@BretJohnson thank you for this discussion too!

BretJohnson:
I created a new PR for this fix/workaround, with some updates. See https://github.com/CommunityToolkit/Maui/pull/2288 for the latest.

PR #2185, opened 2024-09-06
Add UseMauiCommunityToolkitNavigationBar method

tranb3r:
This is a simple code change proposal following discussion #1631.
Adding this method would allow people to use the NavigationBar feature and get rid of all the unnecessary code in MCT when using dotnet trimmer.
If this is approved, then we can add similar methods for other features (either new or existing) later.


brminnick:
Hi @tranb3r, I sincerely appreciate you contributing this code and tuning into our monthly standups. 

As we discussed in our September standup, we will not be moving forward with modularizing `UseMauiCommunityToolkit()` because the benefit does not outweigh the scope and maintenance costs.

Issue #2089, opened 2024-08-03
Offline SpeechToText


VladislavAntonyuk:
### Feature name

SpeechToText

### Progress tracker

- [ ] Android Implementation
- [ ] iOS Implementation
- [ ] MacCatalyst Implementation
- [ ] Windows Implementation
- [ ] Tizen Implementation
- [ ] Unit Tests
- [ ] Samples
- [ ] Documentation

### Summary

Free Speech recognition functionality that is available offline.

More details with samples at https://vladislavantonyuk.azurewebsites.net/articles/Speech-recognition-with-.NET-MAUI

### Motivation

Allows offline SpeechToText to allow without internet connection. 

### Detailed Design

https://developer.apple.com/documentation/speech
https://developer.android.com/reference/android/speech/SpeechRecognizer
https://learn.microsoft.com/en-us/uwp/api/windows.media.speechrecognition?view=winrt-22621

The same API as online Speech Recognition but with Offline suffix 

### Usage Syntax

```markdown
var isAuthorized = await speechToText.RequestPermissions();
		if (isAuthorized)
		{
			try
			{
				RecognitionText = await speechToText.StartListeningOffline(CultureInfo.GetCultureInfo(Locale?.Language ?? "en-us"), new Progress<string>(partialText =>
					{
						RecognitionText += partialText + " ";
					}), cancellationToken);
			}
			catch (Exception ex)
			{
				await Toast.Make(ex.Message).Show(cancellationToken);
			}
		}
		else
		{
			await Toast.Make("Permission denied").Show(cancellationToken);
		}
```

Issue #1779, opened 2024-03-27
[BUG] SpeechToText throws Objective-C exception on iOS


tibuprophen:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [X] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

The speech-to-text service does not start recording/listening. It throws an exception instead - depending on the tts mode it is handled or not.
![Simulator Screenshot - iPhone 15 - 2024-03-27 at 21 57 57](https://github.com/CommunityToolkit/Maui/assets/952777/b54140b6-bb8e-456a-b9a2-4840be871ab6)

### Expected Behavior

The speech-to-text service should start listening. 

### Steps To Reproduce

1. Open the repository
2. Run the sample application on iOS Simulator
3. Either the exception is handled and a toast is shown or unhandled and crashes with the exception. 

### Link to public reproduction project repository

https://github.com/CommunityToolkit/Maui

### Environment

```markdown
- .NET MAUI CommunityToolkit: 7.0.1
- OS: macOs 14.3.1 (23D60)
- .NET MAUI: Manifest: 8.0.7/8.0.100, Source: 8.0.200, Controls: 8.0.14
- Xcode: 15.2
- Simulator: iPhone 15 iOS 17.2
```

### Anything else?

Seems to work partially on a physical device. I've re-implemented the tts feature after months in my app, but there appears to be a bug that words/sentences are cut. When I started debugging this bug, I experienced the obj-c crash.

PR #2242, opened 2024-10-01
Offline Speech Recognition #2089

VladislavAntonyuk:
### Description of Change ###

Added 2 new methods for Offline Speech Recognition, Removed ListenAsync method, as it is impossible (with current implementation) to correctly Stop Listening the recognition.  I also added a new State, allowing us to see if Listening is Active, but Silience.

 ### Linked Issues ###
#2089 , #1779 

 - Fixes #

 ### PR Checklist ###
 
### DIFF
--- a/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
@@ -28,15 +28,6 @@ public interface ISpeechToText : IAsyncDisposable
 	SpeechToTextState CurrentState { get; }
 
 	/// <summary>
-	/// Converts speech to text in real time.
-	/// </summary>
-	/// <param name="culture">Speak language</param>
-	/// <param name="recognitionResult">Intermediate recognition result.</param>
-	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
-	/// <returns>Final recognition result</returns>
-	Task<SpeechToTextResult> ListenAsync(CultureInfo culture, IProgress<string>? recognitionResult, CancellationToken cancellationToken = default);
-
-	/// <summary>
 	/// Starts the SpeechToText service
 	/// </summary>
 	/// <remarks>
@@ -54,6 +45,25 @@ public interface ISpeechToText : IAsyncDisposable
 	/// </remarks>
 	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
 	Task StopListenAsync(CancellationToken cancellationToken = default);
+	
+	/// <summary>
+	/// Starts the Offline SpeechToText service
+	/// </summary>
+	/// <remarks>
+	/// Real time speech recognition results will be surfaced via <see cref="RecognitionResultUpdated"/> and <see cref="RecognitionResultCompleted"/>
+	/// </remarks>
+	/// <param name="culture">Speak language</param>
+	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
+	Task StartOfflineListenAsync(CultureInfo culture, CancellationToken cancellationToken = default);
+
+	/// <summary>
+	/// Stops the OfflineSpeechToText service
+	/// </summary>
+	/// <remarks>
+	/// Speech recognition results will be surfaced via <see cref="RecognitionResultCompleted"/>
+	/// </remarks>
+	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
+	Task StopOfflineListenAsync(CancellationToken cancellationToken = default);
 
 	/// <summary>
 	/// Request permissions for speech to text.
--- a/samples/CommunityToolkit.Maui.Sample/Pages/Essentials/SpeechToTextPage.xaml
+++ b/samples/CommunityToolkit.Maui.Sample/Pages/Essentials/SpeechToTextPage.xaml
@@ -32,6 +32,17 @@
                 ItemDisplayBinding="{Binding ., Converter={StaticResource PickerLocaleDisplayConverter}}"/>
 
             <Label
+                Text="State"
+                FontAttributes="Bold"/>
+
+            <Label
+                Text="{Binding State}"
+                FontSize="18"
+                HorizontalOptions="Center"
+                HorizontalTextAlignment="Center"
+                MinimumHeightRequest="100" />
+
+            <Label
                 Text="Language Output"
                 FontAttributes="Bold"/>
 
@@ -46,44 +57,6 @@
                 Text="Play"
                 Command="{Binding PlayCommand}"
                 HorizontalOptions="Center" />
-
-            <Border
-                StrokeThickness="2"
-                Stroke="#808080"
-                StrokeShape="RoundRectangle 8,8,8,8"
-                Padding="12">
-                <Border.Content>
-
-                    <Grid RowDefinitions="*,60"
-                          ColumnDefinitions="*,*"
-                          RowSpacing="12"
-                          ColumnSpacing="12">
-
-                        <Button
-                            Grid.Row="0"
-                            Grid.Column="0"
-                            Text="ListenAsync"
-                            Command="{Binding ListenCommand}"
-                            HorizontalOptions="End" />
-
-                        <Button
-                            Grid.Row="0"
-                            Grid.Column="1"
-                            Text="Cancel Token"
-                            Command="{Binding ListenCancelCommand}"
-                            HorizontalOptions="Start" />
-
-                        <Label
-                            Grid.Row="1"
-                            Grid.ColumnSpan="2"
-                            Text="The `ListenAsync` API allows you to await the final speech recognition results using async/await. `ListenAsync` is cancelled via CancellationToken."
-                            HorizontalOptions="Center"
-                            HorizontalTextAlignment="Center"
-                            FontSize="12"/>
-
-                    </Grid>
-                </Border.Content>
-            </Border>
 
             <Border
                 StrokeThickness="2"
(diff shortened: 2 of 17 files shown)


src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
```diff
@@ -28,32 +28,42 @@ public interface ISpeechToText : IAsyncDisposable
...
+	/// <summary>
+	/// Starts the Offline SpeechToText service
 	/// </summary>
 	/// <remarks>
 	/// Real time speech recognition results will be surfaced via <see cref="RecognitionResultUpdated"/> and <see cref="RecognitionResultCompleted"/>
 	/// </remarks>
 	/// <param name="culture">Speak language</param>
 	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
-	Task StartListenAsync(CultureInfo culture, CancellationToken cancellationToken = default);
+	Task StartOfflineListenAsync(CultureInfo culture, CancellationToken cancellationToken = default);
```
bijington, line 57: Do you think there is a way that we could expose this via a different interface? Then we can keep the related methods together and separate from unrelated methods

VladislavAntonyuk: It's a great idea, I think we should keep only 2 methods here (Start/Stop), but implement the interface with Offline and Online Speech to Text services. We can even provide a default strategy to choose Online or Offline implementation based on internet connectivity.
Let's bring this discussion to our stand up.

bijington: Yes perfect! I think that should provide a nice API surface for developers then

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
bijington, line 1: Is it correct that `SpeechToText` lives in the `CommunityToolkit.Maui.Media` namespace?

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/OfflineSpeechToText.shared.cs
```diff
@@ -0,0 +1,29 @@
+using System.Globalization;
+
+namespace CommunityToolkit.Maui.Media;
+
+/// <inheritdoc cref="ISpeechToText"/>
+public static class OfflineSpeechToText
```
bijington, line 6: I wonder if we should add additional `<remarks>` to make it clear to the user that this is providing offline support. I know it's in the name but adding it to intellisense may also help.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/OfflineSpeechToTextImplementation.shared.cs
```diff
@@ -0,0 +1,80 @@
+using System.Globalization;
+using Microsoft.Maui.ApplicationModel;
+
+namespace CommunityToolkit.Maui.Media;
+
+/// <inheritdoc cref="ISpeechToText"/>
+public sealed partial class OfflineSpeechToTextImplementation : ISpeechToText
```
bijington, line 7: This might be by design but I don't think it is currently easy for a developer to register both online and offline speech to text implementations via DI? They would have to register/resolves them as multiples and somehow key them to know which is which. This might be fine. If we have an interface for offline speech to text then we could make that task simpler. Thoughts?

bijington (commented):
@VladislavAntonyuk thanks for this. I think it looks great! I only had a few comments to get your thoughts on some things.

Finally, this might be outside of the scope of this PR but I wanted to raise it because I think it should be in a follow-up PR; I would love the ability to chain the 2 implementations together when registering with DI, something like:

```csharp
builder.Services.AddSpeechToText(SpeechToText.Default).WithFallback(OfflineSpeechToText.Default);
```

And in theory developers could chain the other way round:

```csharp
builder.Services.AddSpeechToText(OfflineSpeechToText.Default).WithFallback(SpeechToText.Default);
```

What do you think to the above? We might have to wrap this in another class rather than complicating the flow of the 2 current implementations so it probably is best in a follow-up PR.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
VladislavAntonyuk: yes, it is the same namespace as Text To Speech in MAUI ([Microsoft.Maui.Media](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.media?view=net-maui-8.0))

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/OfflineSpeechToTextImplementation.shared.cs
VladislavAntonyuk: They can use keyed service. Updated Sample app

VladislavAntonyuk:
Thank you @bijington.
if user registers SpeechToText like this `builder.Services.AddSpeechToText(OfflineSpeechToText.Default).WithFallback(SpeechToText.Default);`
he still needs to somehow resolve the implementation.

Also, we need to provide the lifetime of the service. And what is under the hood of `AddSpeechToText`? We'll still have the same `AddSingleton(OfflineSpeechToText.Default)` call in that method.

bijington:
> Thank you @bijington. if user registers SpeechToText like this `builder.Services.AddSpeechToText(OfflineSpeechToText.Default).WithFallback(SpeechToText.Default);` he still needs to somehow resolve the implementation.
> 
> Also, we need to provide the lifetime of the service. And what is under the hood of `AddSpeechToText`? We'll still have the same `AddSingleton(OfflineSpeechToText.Default)` call in that method.

Yes I agree the developer will need to define the lifetime of the service which increases the complexity.

Perhaps we could move the `WithFallback` method onto the ISpeechToText interface instead, then the developer could write something like:

```csharp
builder.Services.AddSingleton(OfflineSpeechToText.Default.WithFallback(SpeechToText.Default));
```

Then `WithFallback` could look something like:

```csharp
public static void ISpeechToTextExtensions
{
    public ISpeechToText WithFallback(this ISpeechToText primary, ISpeechToText secondary)
    {
        return new PriorityBasedSpeechToText(primary, secondary);
    }
}

internal class PriorityBasedSpeechToText : ISpeechToText
{
    readonly ISpeechToText primaryService;
    readonly ISpeechToText secondaryService;

    public PriorityBasedSpeechToText(this ISpeechToText primary, ISpeechToText secondary)
    {
        primaryService = primary;
        secondaryService = secondary;
    }

    public Task<SpeechToTextRecognitionResult> StartListenAsync()
    {
        // attempt primary, if fails fallback to secondary...
    }
}
```

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
```diff
@@ -27,24 +27,15 @@ public interface ISpeechToText : IAsyncDisposable
...
-
 	/// <summary>
 	/// Starts the SpeechToText service
 	/// </summary>
 	/// <remarks>
 	/// Real time speech recognition results will be surfaced via <see cref="RecognitionResultUpdated"/> and <see cref="RecognitionResultCompleted"/>
 	/// </remarks>
-	/// <param name="culture">Speak language</param>
+	/// <param name="options"><see cref="SpeechToTextOptions"/></param>
 	/// <param name="cancellationToken"><see cref="CancellationToken"/></param>
```
bijington, line 37: What will the `cancellationToken` parameter do if it is canceled? Will it stop listening and result in a non-successful recognition result? If so I think we should add this detail to the xml docs.

bijington (changes requested):
@VladislavAntonyuk thanks for this! I have a few comments and I think they are mostly pretty minor

samples/CommunityToolkit.Maui.Sample/ViewModels/Essentials/OfflineSpeechToTextViewModel.cs
```diff
@@ -0,0 +1,150 @@
...
+	[ObservableProperty, NotifyCanExecuteChangedFor(nameof(StartListenCommand))]
+	bool canStartListenExecute = true;
+
+	[ObservableProperty, NotifyCanExecuteChangedFor(nameof(StopListenCommand))]
+	bool canStopListenExecute = false;
+
+	public OfflineSpeechToTextViewModel(ITextToSpeech textToSpeech)
+	{
+		this.textToSpeech = textToSpeech;
+		this.speechToText = OfflineSpeechToText.Default;
```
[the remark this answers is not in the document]
VladislavAntonyuk: I just left it for demo purposes. In online speech to text I resolve it from DI.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
VladislavAntonyuk: No, Cancellation Token will abort StartListening operation. If listening is started you can only cancel it using StopListening()

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
```diff
@@ -27,24 +27,15 @@ public interface ISpeechToText : IAsyncDisposable
...
-	Task<SpeechToTextResult> ListenAsync(CultureInfo culture, IProgress<string>? recognitionResult, CancellationToken cancellationToken = default);
-
 	/// <summary>
 	/// Starts the SpeechToText service
 	/// </summary>
 	/// <remarks>
 	/// Real time speech recognition results will be surfaced via <see cref="RecognitionResultUpdated"/> and <see cref="RecognitionResultCompleted"/>
 	/// </remarks>
-	/// <param name="culture">Speak language</param>
+	/// <param name="options"><see cref="SpeechToTextOptions"/></param>
```
bijington, line 36:
```suggestion
	/// <param name="options">The <see cref="SpeechToTextOptions"/> used to customize how the speech will be detected and converted to text.</param>
```

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/ISpeechToText.shared.cs
bijington, line 37:
```suggestion
	/// <param name="cancellationToken">A <see cref="CancellationToken"/> that provides the ability to cancel the starting process. Note that listening has started then you can only stop listening by calling <see cref="StopListening"/>.</param>
```

bijington: Thanks I have added some xml comment improvements

samples/CommunityToolkit.Maui.Sample/ViewModels/Essentials/OfflineSpeechToTextViewModel.cs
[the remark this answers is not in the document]
bijington: Perhaps we can add a comment to highlight this? Just to make it clear that we are showing both options

bijington (commented):
Thanks @VladislavAntonyuk I have added some xml docs improvements but the rest looks good to me

samples/CommunityToolkit.Maui.Sample/ViewModels/Essentials/OfflineSpeechToTextViewModel.cs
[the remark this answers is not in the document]
VladislavAntonyuk:
```suggestion
		// For demo purposes. You can resolve dependency from the DI container,
		this.speechToText = OfflineSpeechToText.Default;
```

bijington (approved):
LGTM! Do we just need to write the docs? Is this something you are happy to do? If not I'm happy to write something

VladislavAntonyuk:
> LGTM! Do we just need to write the docs? Is this something you are happy to do? If not I'm happy to write something

Thank you Shaun! I will create docs PR this weekend.

pictos (changes requested):
The sample app is crashing for me. Here's the log that I could get.
image

VladislavAntonyuk:
> The sample app is crashing for me. Here's the log that I could get.
> 
> I ran on iphone 14 pro. And allowed all permissions.
> 
> image

It’s text to speech. I will remove it from the offline sample

pictos:
will try again during this week, asap

bijington (approved):
LGTM! Thanks for this @VladislavAntonyuk

Issue #1374, opened 2023-08-28
[BUG] CommunityToolkit.Maui.Core.Platform.StatusBar causes app to fail on iOS Testflight


needledrag:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [X] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

I've tested my .net Maui app in Android and iOS, debug, release and on device, Android now in Closed Testing on Google Play.
Just uploaded to App Connect and testing on a device using TestFlight.

On both devices I get the error in appcenter as the app loads the splash screen and goes blank: 
```
  "exception": {
    "type": "System.InvalidOperationException",
    "message": "RootViewController cannot be null",
[...]

PR #2243, opened 2024-10-01
Make use of a specific iOS + MacCatalyst TFM

### LINKED ISSUE #1075: [BUG] Could not set up parent class, due to: Invalid generic instantiation

### Current Behavior

Attempting to debug on a physical iOS device from crashes on startup with System.TypeLoadException Could not set up parent class, due to: Invalid generic instantiation.

### Expected Behavior

App does not crash on start up.

### Steps To Reproduce

1. Debug run sample app on physical device

https://github.com/KieranMaclagan/MCTAppBuilderExtensionsTest

### Environment

```markdown
- OS: macOS Ventura 13.2.1
- .NET MAUI: 7.0.49
- VS Mac: 17.5.1
- Device: iPhone 8, iOS 16.3.1
```

### Anything else?

CommunityToolkit.Maui 4.0.0 does experience this

CommunityToolkit.Maui 3.1.0 does not experience this

Stack Trace:

>    at CommunityToolkit.Maui.AppBuilderExtensions.<>c.<UseMauiCommunityToolkit>b__0_0(IMauiHandlersCollection h) in /_/src/CommunityToolkit.Maui/AppBuilderExtensions.shared.cs:line 28
   at Microsoft.Maui.Hosting.HandlerMauiAppBuilderExtensions.HandlerRegistration.AddRegistration(IMauiHandlersCollection builder)
   at Microsoft.Maui.Hosting.Internal.MauiHandlersFactory.CreateHandlerCollection(IEnumerable`1 registrationActions)
   at Microsoft.Maui.Hosting.Internal.MauiHandlersFactory..ctor(IEnumerable`1 registrationActions)
   at Microsoft.Maui.Hosting.HandlerMauiAppBuilderExtensions.<>c.<ConfigureMauiHandlers>b__1_0(IServiceProvider sp)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitFactory(FactoryCallSite factoryCallSite, RuntimeResolverContext context)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2[[Microsoft.Extensions.DependencyInjection.ServiceLookup.RuntimeResolverContext, Microsoft.Extensions.DependencyInjection, Version=7.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60],[System.Object, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].VisitCallSiteMain(ServiceCallSite callSite, RuntimeResolverContext argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2[[Microsoft.Extensions.DependencyInjection.ServiceLookup.RuntimeResolverContext, Microsoft.Extensions.DependencyInjection, Version=7.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60],[System.Object, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].VisitCallSite(ServiceCallSite callSite, RuntimeResolverContext argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.Resolve(ServiceCallSite callSite, ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.CreateServiceAccessor(Type serviceType)
   at System.Collections.Concurrent.ConcurrentDictionary`2[[System.Type, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Func`2[[Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope, Microsoft.Extensions.DependencyInjection, Version=7.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60],[System.Object, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].GetOrAdd(Type key, Func`2 valueFactory)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType)
   at Microsoft.Maui.MauiContext.WrappedServiceProvider.GetService(Type serviceType)
   at Microsoft.Maui.MauiContext.WrappedServiceProvider.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[IMauiHandlersFactory](IServiceProvider provider)
   at Microsoft.Maui.MauiContext.<.ctor>b__2_0()
   at System.Lazy`1[[Microsoft.Maui.IMauiHandlersFactory, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1[[Microsoft.Maui.IMauiHandlersFactory, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1[[Microsoft.Maui.IMauiHandlersFactory, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].CreateValue()
   at System.Lazy`1[[Microsoft.Maui.IMauiHandlersFactory, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].get_Value()
   at Microsoft.Maui.MauiContext.get_Handlers()
   at Microsoft.Maui.Platform.ElementExtensions.SetHandler(INativeObject nativeElement, IElement element, IMauiContext context)
   at Microsoft.Maui.Platform.ElementExtensions.SetApplicationHandler(IUIApplicationDelegate platformApplication, IApplication application, IMauiContext context)
   at Microsoft.Maui.MauiUIApplicationDelegate.FinishedLaunching(UIApplication application, NSDictionary launchOptions)
   at UIKit.UIApplication.UIApplicationMain(Int32 argc, String[] argv, IntPtr principalClassName, IntPtr delegateClassName)
   at UIKit.UIApplication.Main(String[] args, Type principalClass, Type delegateClass)
   at MCTAppBuilderExtensionsTest.Program.Main(String[] args) in /Users/kmaclagan/Projects/MCTAppBuilderExtensionsTest/MCTAppBuilderExtensionsTest/Platforms/iOS/Program.cs:line 13

bijington:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #1075 

 ### PR Checklist ###
 

brminnick:
Hey Shaun! Two quick thoughts:

1. If we make this change for iOS, let's also do it for MacCatalyst too
2. Is it possible to create unit test that validates this fixes the NativeAOT/Trimming bugs? 
    - We may need to first merge the [Device Tests PR](https://github.com/CommunityToolkit/Maui/pull/1444)
    - If you can figure out how to get CI / CD pipeline working for the [Device Tests PR](https://github.com/CommunityToolkit/Maui/pull/1444), we can merge it now to help you complete this bug fix 
      - i.e. We don't need to write a complete DeviceTest suite in order to first merge the [Device Tests PR](https://github.com/CommunityToolkit/Maui/pull/1444)

bijington:
I am actually starting to wonder whether this might also fix other issues such as #1374 based on this documentation page https://learn.microsoft.com/en-us/dotnet/standard/frameworks#precedence I read it as apps could be built targeting `net8.0` support as this may be closer to what the developer is targeting. It's either going to help or make it worse...

filipnavara (approved):

bijington:
@brminnick I'm not convinced we can create a unit test for this. I do wonder if we could add something to the pipeline to make sure that the packaged information meets what we expect.

brminnick:
Let's chat about it in tomorrow's standup! 

I'm fairly confident that as long as we set up our CI / CD pipeline correctly to match our targeted frameworks and run the Device Tests in RELEASE configuration with NativeAOT enabled, `<PublishAot>true</PublishAot>`, it should replicate any errors our users are experiencing.

filipnavara:
We can replicate the error in a CI pipelines but it’s not straightforward. The latest breakage required .NET 9 RC SDK installed side-by-side to trigger it. Alternatively you can deliberately install old (fixed) workload set instead of the latest one. The trouble with that is that you either need to use custom manifests or start with 8.0.4xx SDK which introduced the workload set support.

bijington:
I was thinking of just inspecting the nuget package generated and verifying it targets the expected TFMs for Apple platforms

bijington:
I guess a nice test would be to try consuming the nightly build and see if that still exhibits the issue. I've not be able to reproduce the issue @filipnavara is this something that you could easily test/verify?

filipnavara:
> I guess a nice test would be to try consuming the nightly build and see if that still exhibits the issue. I've not be able to reproduce the issue @filipnavara is this something that you could easily test/verify?

Nightly build of what? CommunityToolkit.Maui? If you have a link I can test the scenarios easily and report back.

bijington:
There might be a neater way but I have just knocked up a little console app that could be published as a dotnet tool to perform this verification. The code is:

```csharp
using System.IO.Compression;
using CommandLine;

namespace PackageVerifier;

class Program
{
    static void Main(string[] args)
    {
        Parser.Default.ParseArguments<Options>(args)
            .WithParsed<Options>(o =>
            {
                Console.WriteLine($"Verifying package '{o.PackagePath}'");

                var extractDirectory = Path.Join(Environment.CurrentDirectory, "abc");
                Console.WriteLine($@"Extracting package to directory '{extractDirectory}'");
                
                ZipFile.ExtractToDirectory(o.PackagePath, extractDirectory);

                Console.WriteLine($"Checking for TFMs: '{string.Join(',', o.TargetFrameworkMonikers)}'");
                
                var libDirectory = Path.Join(extractDirectory, "lib");
                var directories = new DirectoryInfo(libDirectory).GetDirectories().Select(x => x.Name).ToList();

                foreach (var unexpectedDirectory in directories.Except(o.TargetFrameworkMonikers))
                {
                    Console.WriteLine($"WARNING: TFM '{unexpectedDirectory}' was not expected");
                }
                
                foreach (var unexpectedDirectory in o.TargetFrameworkMonikers.Except(directories))
                {
                    Console.WriteLine($"ERROR: TFM '{unexpectedDirectory}' is missing");
                }
                
                foreach (var unexpectedDirectory in o.TargetFrameworkMonikers.Intersect(directories))
                {
                    Console.WriteLine($"SUCCESS: Expected TFM '{unexpectedDirectory}' exists");
                }
            });
    }
}
```

Usage is something like:

```bash
dotnet packageverifier --tfms "net9.0-ios17.5" --package-path "/Users/shaunlawrence/Documents/work/projects/toolkits/mct/src/Maui/src/CommunityToolkit.Maui/bin/Release/CommunityToolkit.Maui.1.0.0-pre1.nupkg.zip"
```

And the output currently looks something like:

```log
Verifying package '/Users/shaunlawrence/Documents/work/projects/toolkits/mct/src/Maui/src/CommunityToolkit.Maui/bin/Release/CommunityToolkit.Maui.1.0.0-pre1.nupkg.zip'
Extracting package to directory '/Users/shaunlawrence/Documents/work/projects/investigations/PackageVerifier/PackageVerifier/bin/Debug/net8.0/abc'
Checking for TFMs: 'net9.0-ios17.5'
WARNING: TFM 'net9.0-maccatalyst17.5' was not expected
WARNING: TFM 'net9.0' was not expected
WARNING: TFM 'net9.0-android35.0' was not expected
SUCCESS: Expected TFM 'net9.0-ios17.5' exists
```

With a bit more effort it could be modified to be more user friendly and allow it to cause a pipeline build to fail. What do you think?

bijington:
> > I guess a nice test would be to try consuming the nightly build and see if that still exhibits the issue. I've not be able to reproduce the issue @filipnavara is this something that you could easily test/verify?
> 
> Nightly build of what? CommunityToolkit.Maui? If you have a link I can test the scenarios easily and report back.
> 
> I assume I just need to use this feed:
> 
> ```
>     <add key="CommunityToolkit-PullRequests" value="https://pkgs.dev.azure.com/dotnet/CommunityToolkit/_packaging/CommunityToolkit-PullRequests/nuget/v3/index.json" />
> ```
> 
> ...and then reference the build number from this PR, right?

Yes that looks like it is the correct feed and approach for the build number as per here https://github.com/CommunityToolkit/Maui/wiki/Preview-Packages#pull-requests-

Let me find the build number

filipnavara:
> Let me find the build number

`99.0.0-build-2243.110238+e95e119` .. no need, it's in the Checks.

filipnavara:
Yes, I can confirm that with .NET 9 RC 1 SDK I now get the correct iOS assets. With 9.1.0 on NuGet.org I get the generic `net8.0` assets.

bijington:
> Yes, I can confirm that with .NET 9 RC 1 SDK I now get the correct iOS assets. With 9.1.0 on NuGet.org I get the generic `net8.0` assets.

That's fantastic news! Thank you so much for testing this and all of the input to get us to this stage!

filipnavara:
> Thank you so much for testing this and all of the input to get us to this stage!

Thanks for addressing it so quickly. :-) Let me know if there's anything else you need to test.

bijington:
I have had some fun playing around with this so I have stuck it in a GitHub repo: https://github.com/bijington/package-verifier

VladislavAntonyuk (dismissed):

brminnick:
Hey Shaun! I just tested this NuGet Package on a local environment that only has .NET SDK v8.0.303 installed (which does not include `net-ios18.0`, and IDE still allowed me to add the NuGet Package and compile my code with no warnings or build errors. I assume that my test is using the code in the `net8.0` version of `CommunityToolkit.Maui` because the app crashes at runtime with `TypeLoadException`.

So, the good news is that this PR solves one very specific problem where the user has `net8.0-ios18.0` installed via .NET SDK 8.0.402, but also has .NET SDK 9.0 installed which doesn't yet include `net8.0-ios18.0`; the user's machine will revert to using the installed .NET 8.0 SDK instead of the .NET 9.0 SDK. 

But it doesn't solve the problem where `CommunityToolkit.Maui` requires a specific version of a .NET MAUI Workload installed. 

I think our best move is to add a MSBuild task that checks the user's local environment and produces a build error with a help message letting them know which MAUI workload we require. I'll play around with adding something like this today.

bijington:
> So, the good news is that this PR solves one very specific problem where the user has `net8.0-ios18.0` installed via .NET SDK 8.0.402, but also has .NET SDK 9.0 installed which doesn't yet include `net8.0-ios18.0`; this PR ensures that the dev's machine will revert to using the installed .NET 8.0 SDK + net8.0-ios18.0 instead of the .NET 9.0 SDK.

Does it? Wasn't this the original issue because we were implicitly targeting `net8-ios18.0` in the last release (9.1.0) - surely this means we are just preserving that issue? Anyway I am trying to test that now.

filipnavara:
> this PR ensures that the dev's machine will revert to using the installed .NET 8.0 SDK + net8.0-ios18.0 instead of the .NET 9.0 SDK.

No, it doesn't fix that as is. You should change it back to target `ios17.5` to fix it.

brminnick:
@filipnavara To workaround your specific problem, I recommend adding a [`global.json`](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json) file to your project that prevents it from using .NET 9 until we add support for .NET 9. This will ensure that the .NET tooling only uses the .NET 8 SDK when you build your project.

## global.json
```
{
  "sdk": {
    "version": "8.0.402",
    "rollForward": "latestFeature",
    "allowPrerelease": false
  }
}
```

filipnavara:
> To workaround your specific problem, I recommend adding a [`global.json`](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json) file to your project that prevents it from using .NET 9 until we add support for .NET 9.

I am aware of that workaround (along few others). It's just that .NET 9 SDK builds are noticeably faster on some machines so our devs prefer to use it.

filipnavara:
Note that technically there should be no benefit for libraries to target `net8.0-ios18.0` over `net8.0-ios17.5` if you don't use any of the new APIs. It's a different story for apps where you need to target newer SDK due to the way the Xcode tooling is versioned, app store submissions, and compatibility with newer simulators for dev cycle.

brminnick:
Thanks for confirming the workaround works for you, @filipnavara. 

We'll close this PR and work on a way that we can leverage MSBuild Task to ensure developer's local environments are using the correct .NET SDK, .NET MAUI Workload and Xcode versions.

filipnavara:
> We'll close this PR and work on a way that we can leverage MSBuild Task to ensure developer's local environments are using the correct .NET SDK, .NET MAUI Workload and Xcode versions.

It would be nice to share more details about your plan.

I don't think libraries should force users to upgrade SDKs unless there's a compelling reason (eg. relying on new API, tooling or fix). Workloads are already tied to Xcode versions and they have the necessary checks.

Also, NuGet largely runs before most of the MSBuild evaluation takes place. You can detect that you are compiling for `net8.0-ios` TFM and that you incorrectly consumed the `net8.0` asset of your library but you cannot really fix it unless you bypass the the whole NuGet resolution and rely on MSBuild script and `Reference` item group. Essentially you can detect the mismatch and issue an error but to fully fix the problem you still need to downgrade the TFM inside the NuGet to target `net8.0-ios17.X`.

bijington:
> to fully fix the problem you still need to downgrade the TFM inside the NuGet to target `net8.0-ios17.X`.

As Brandon has pointed out to me offline of this discussion; downgrading the TFM and shipping a version would effectively break anyone targeting iOS 18.0 already. So I agree it best that we don't break those users especially if we have a workaround for this issue. 

The PR for the MSBuild task is #2266. It's frustrating that we have to guard this rather than Nuget doing it but that sadly appears to be some pain of relying on workloads

filipnavara:
> downgrading the TFM and shipping a version would effectively break anyone targeting iOS 18.0 already.

How? Lower version of library TFMs is still consumable by `net8.0-ios18.0`. That’s easy to verify if you create new iOS app with the latest workloads (specifies `net8.0-ios` which is implicitly transformed into `net8.0-ios18.0` by the workload) and reference CommunityToolkit.Maui 9.0.3 which shipped older TFM.

bijington:
You are correct, I have tested running our build targeting 17.5 from this PR against the latest SDK and workload and it behaves.

PR #2254, opened 2024-10-04
[Housekeeping] Update NuGet Packages

brminnick:
### Description of Change ###

This PR updates the NuGet Packages in `CommunityToolkit.Maui.Sample`.

It also adds `<WindowsSdkPackageVersion>10.0.19041.41</WindowsSdkPackageVersion>` to `CommunityToolkit.Maui.Sample.csproj` which is required by `CommunityToolkit.Mvvm v8.3.2`: 
> This version of the MVVM Toolkit requires 'Microsoft.Windows.SDK.NET.Ref' version '10.0.19041.38' or later. Please update to .NET SDK 8.0.109, 8.0.305 or 8.Comm0.402 (or later). Alternatively, use a temporary 'Microsoft.Windows.SDK.NET.Ref' reference, which can be done by setting the 'WindowsSdkPackageVersion' property in your .csproj file. For your project configuration, it is recommended to set the package version to '10.0.19041.41'.

 ### PR Checklist ###
 
In `azurepipelines.yml`, I've re-ordered the steps so that the `Run CommunityToolkit.Maui.Analyzers.UnitTests` step only runs once we've finished building all of the projects to avoid a build error arising in a `Run Unit Test` step.

### DIFF
--- a/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
+++ b/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
@@ -37,9 +37,9 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)' == 'Release' 
-                  AND $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'tizen'
-                  AND $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'android'
-                  AND $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'windows'">
+                 AND $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'tizen'
+                 AND $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'android'
+                 AND $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) != 'windows'">
     <!-- Ahead-of-time compilation is not yet supported net8.0-tizen -->
     <!-- Cross-OS native compilation is not supported net8.0-android -->
     <!-- Only iOS + MacCatalyst support AOT in .NET 8 -->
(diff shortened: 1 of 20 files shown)


samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -31,6 +31,8 @@
 
     <!-- Uncomment the line below if Android App installation fails -->
     <!--<EmbedAssembliesIntoApk>true</EmbedAssembliesIntoApk>-->
+
+    <WindowsSdkPackageVersion>10.0.19041.41</WindowsSdkPackageVersion>
```
bijington, line 35: Why do we need to do this when the MVVM toolkit project refers to an older package: https://github.com/CommunityToolkit/dotnet/blob/main/src/CommunityToolkit.Mvvm/CommunityToolkit.Mvvm.csproj

@Sergio0694 do you know which value is correct?

Sergio0694: .38 is the minimum to build, but if you're manually setting the package version you should set the latest available at that time that's compatible. It was .41 when I made the last MVVM Toolkit update. You can use .53 today. Also FYI, once you get the updated .NET SDK, you should be able to remove that property, as the .NET SDK should set the correct Windows SDK package version automatically 🙂

bijington: @Sergio0694 thank you for the speedy response! When should we get the updated .NET SDK? What version of the SDK are you referring to there?

Sergio0694: I think the new .NET SDK that went out in the October patch Tuesday should have it (8.0.403) 🤔

brminnick, line 35:
```suggestion
    <WindowsSdkPackageVersion>10.0.19041.53</WindowsSdkPackageVersion>
```

brminnick: Thanks Sergio! I've updated it to `10.0.19041.53`.

bijington (approved):

PR #2267, opened 2024-10-08
PopupService,Make parameter meaningful

BaY1251:
### Description of Change ###

 

 ### Proposal ###

 ### PR Checklist ###
 
### DIFF
--- a/src/CommunityToolkit.Maui/PopupService.cs
+++ b/src/CommunityToolkit.Maui/PopupService.cs
@@ -63,7 +63,7 @@ public class PopupService : IPopupService
 
 		var popup = GetPopup(typeof(TViewModel));
 
-		ValidateBindingContext<TViewModel>(popup, out _);
+		popup.BindingContext = viewModel;
 
 		ShowPopup(popup);
 	}
@@ -119,7 +119,7 @@ public class PopupService : IPopupService
 
 		var popup = GetPopup(typeof(TViewModel));
 
-		ValidateBindingContext<TViewModel>(popup, out _);
+		popup.BindingContext = viewModel;
 
 		return ShowPopupAsync(popup, token);
 	}
--- a/src/CommunityToolkit.Maui/PopupService.cs
+++ b/src/CommunityToolkit.Maui/PopupService.cs
@@ -63,7 +63,7 @@ public class PopupService : IPopupService
 
 		var popup = GetPopup(typeof(TViewModel));
 
-		popup.BindingContext = viewModel;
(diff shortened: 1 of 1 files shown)


src/CommunityToolkit.Maui/PopupService.cs
```diff
@@ -63,7 +63,7 @@ public void ShowPopup<TViewModel>(TViewModel viewModel) where TViewModel : INoti
 
 		var popup = GetPopup(typeof(TViewModel));
 
-		ValidateBindingContext<TViewModel>(popup, out _);
```
bijington, line 66: I think we should make sure that the BindingContext is not assigned before assigning it here to make sure we aren't replacing something already created. Mostly to help the developer know if they are making a mistake. What do you think?

bijington (changes requested):

src/CommunityToolkit.Maui/PopupService.cs
BaY1251: I agree with you and have modified the code

Pantheas: But isn't setting the `BindingContext` of the page in any case exactly the purpose of this method?
Imagine having a popup where the ViewModel is resolved using DI in most cases.
But for one specific case, we are manually resolving or creating the instance and want to pass it to the popup service.
The instance would never be set as `BindingContext` with this implementation.

bijington: That is correct but I think it best to bring this to the developers attention (potentially with an exception). If they wanted to use your scenario they could potentially register their view models as `AddScoped` and share the instance...

brminnick:
Hi @BaY1251! We plan on deprecating the `public void ShowPopup<TViewModel>(TViewModel)` API. 

I understand your desire, and we've discussed the pros + cons of this in the past in our monthly standups.

However, it's wrong for an API called `ShowPopup` to do any hidden variable assignment under the hood that the caller may be unaware of. It is also a breaking change because it breaks existing functionality for folks currently using the API as designed.

The recommended way to use `PopupService` is to add both the Popup and its ViewModel to `builder.Services` in `MauiProgram.cs`, then call `PopupService.ShowPopup<TViewModel>()`.

BaY1251:
@brminnick This is truly unfortunate news.
I need to use Popup to display the details of list items, so I can only write code like this in the ViewModel 
`Shell. Current.ShowPopup(new MyPopup { BindingContext = viewModel })`.

PR #2288, opened 2024-10-17
Fix animation behavior hot reload

BretJohnson:
### Description of Change ###

Fix https://developercommunity.visualstudio.com/t/Hot-reload-stops-working-when-using-Comm/10707756

This is a follow up to an earlier attempt to fix this issue with https://github.com/CommunityToolkit/Maui/pull/2174, which we decided not to take.

See that PR for more context. Here's the update since then:

Ideally, this would be fixed on the MAUI tooling side, making the VS Hot Reload code respect the
special rules around readonly `OneWayToSource` properties (which aren't supported in WPF but
are in MAUI). Evgeny and I discussed fixing there & started to scope, but we ran out of time to
get a fix in for VS 17.12 before it froze (and before we both went on vacation). I'm not sure
we'll be able to get a fix in for 17.13 either.

Given all that, it seems best to go back to applying this workaround on the CommunityToolkit side
to make Hot Reload work - with a public setter, it works. Updating the CommunityToolkit
also has the advantage that it works in all VS versions, including older ones.

I added a comment to the code to make it clearer that apps shouldn't use the setter directly, instead supplying a binding - the setter is only needed as a workaround for Hot Reload. @bijington - please review text of that comment.

Also, earlier we thought that making the setter `internal` may work, with an `InternalsVisibleTo`. But Evgeny clarified that
> Internal will not work. InternalsVisibleTo should include target assembly. For example, If you edit MyPage.xaml which is part of MyApp then CommynityToolkit should have `[assembly, InternalsVisibleTo("MyApp")]`. Which is clearly impossible

So `internal` in a no go & public with comment seems like the best option to get this working with XAML Hot Reload.

 ### Linked Issues ###
 

 - Fixes # https://developercommunity.visualstudio.com/t/Hot-reload-stops-working-when-using-Comm/10707756
 ### PR Checklist ###
 

src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
```diff
@@ -24,11 +24,19 @@ public class AnimationBehavior : EventToCommandBehavior
...
+	/// property by creating a binding, in XAML or C#. If done via C# use code like this:
+	/// "behavior.SetBinding(AnimationBehavior.AnimateCommandProperty, nameof(ViewModel.TriggerAnimationCommand));"
 	/// </summary>
 	/// <remarks>
 	/// <see cref="AnimateCommand"/> has a <see cref="Type"/> of Command&lt;CancellationToken&gt; which requires a <see cref="CancellationToken"/> as a CommandParameter. See <see cref="Command{CancellationToken}"/> and <see cref="System.Windows.Input.ICommand.Execute(object)"/> for more information on passing a <see cref="CancellationToken"/> into <see cref="Command{T}"/> as a CommandParameter"
 	/// </remarks>
-	public Command<CancellationToken> AnimateCommand => (Command<CancellationToken>)GetValue(AnimateCommandProperty);
+	public Command<CancellationToken> AnimateCommand {
+		get => (Command<CancellationToken>)GetValue(AnimateCommandProperty);
+		set => SetValue(AnimateCommandProperty, value);
```
bijington, line 38: Apologies for testing something out, I'm on the bus to work so relying on the pipeline builds to tell me whether this will work.

```suggestion
      [Obsolete("Do not use this setter, it only exists to enable XAML Hot reload support in your IDE.")]
		set => SetValue(AnimateCommandProperty, value);
```

bijington:
Thanks for the follow up @BretJohnson I have just tried marking the setter `[Obsolete]` in the hope it will make it clearer to a developer. Let's see how it builds :)

src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
```diff
@@ -24,11 +24,20 @@ public class AnimationBehavior : EventToCommandBehavior
...
+	/// property by creating a binding, in XAML or C#. If done via C# use code like this:
+	/// "behavior.SetBinding(AnimationBehavior.AnimateCommandProperty, nameof(ViewModel.TriggerAnimationCommand));"
 	/// </summary>
 	/// <remarks>
 	/// <see cref="AnimateCommand"/> has a <see cref="Type"/> of Command&lt;CancellationToken&gt; which requires a <see cref="CancellationToken"/> as a CommandParameter. See <see cref="Command{CancellationToken}"/> and <see cref="System.Windows.Input.ICommand.Execute(object)"/> for more information on passing a <see cref="CancellationToken"/> into <see cref="Command{T}"/> as a CommandParameter"
 	/// </remarks>
-	public Command<CancellationToken> AnimateCommand => (Command<CancellationToken>)GetValue(AnimateCommandProperty);
+	public Command<CancellationToken> AnimateCommand {
+		get => (Command<CancellationToken>)GetValue(AnimateCommandProperty);
+      [Obsolete("Do not use this setter, it only exists to enable XAML Hot reload support in your IDE.")]
```
bijington, line 38:
```suggestion
      [Obsolete(
"""
Do not use this setter, it only exists to enable XAML Hot reload support in your IDE.

Instead, apps should provide a value for this OneWayToSource property by creating a binding, in XAML or C#. If done via C# use code like this:

behavior.SetBinding(AnimationBehavior.AnimateCommandProperty, nameof(ViewModel.TriggerAnimationCommand));
""")]
```

src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
```diff
@@ -24,11 +24,20 @@ public class AnimationBehavior : EventToCommandBehavior
 
 	/// <summary>
 	/// Gets the Command that allows the triggering of the animation.
+	/// 
+	/// NOTE: Apps should not directly set this property, treating it as read only. The setter is only public because
+	/// that's currently needed to make XAML Hot Reload work. Instead, apps should provide a value for this OneWayToSource
+	/// property by creating a binding, in XAML or C#. If done via C# use code like this:
+	/// "behavior.SetBinding(AnimationBehavior.AnimateCommandProperty, nameof(ViewModel.TriggerAnimationCommand));"
```
bijington, line 31:
```suggestion
	/// <c>behavior.SetBinding(AnimationBehavior.AnimateCommandProperty, nameof(ViewModel.TriggerAnimationCommand));</c>
```

bijington (commented):
Two more quick tests

bijington:
Thanks again for this @BretJohnson  I just want to check what this looks like inside an IDE at some point before merging but otherwise it should be good.

bijington (approved):
Thanks @BretJohnson the result looks good!

BretJohnson:
[heart]         Bret Johnson reacted to your message:
________________________________

PR #2345, opened 2024-11-14
Change expander to use the trimming safe binding

pictos:
This PR updates `Expander` to use the new and AoT friendly Binding API introduced by .NET MAUI.

With this we can make sure our code is trimming safe. For reference you can watch this part of [Peppers presentation on .NET Conf](https://youtu.be/FntsxTqtQEU?t=21708)

### DIFF
--- a/src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
+++ b/src/CommunityToolkit.Maui/Views/Expander/Expander.shared.cs
@@ -12,7 +12,6 @@ namespace CommunityToolkit.Maui.Views;
 [BindableProperty<object>("CommandParameter")]
 [BindableProperty<ICommand>("Command")]
 [ContentProperty(nameof(Content))]
-[RequiresUnreferencedCode("Calls Microsoft.Maui.Controls.Binding.Binding(String, BindingMode, IValueConverter, Object, String, Object)")]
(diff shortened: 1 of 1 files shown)


brminnick (commented):
Have you tested this in `CommunityToolkit.Maui.Controls` v9.0.10? 

I tested it using `CommunityToolit.Maui.Controls` v9.0.0 and it caused our sample app to crash at runtime when trying to use `Expander` because the new [Interceptor](https://github.com/dotnet/roslyn/blob/main/docs/features/interceptors.md) the MAUI team wrote doesn’t find it, causing the app to crash when the binding is fired. I only know this because I learned the hard way. It’s another .NET MAUI bug we need to escalate if it hasn't already been fixed.

This compiled bindings bug is not a blocker for our .NET 9 release.

pictos:
@brminnick I didn't try out, but when we chated on discord you mentioned that error for TypedBinding... This one, I presume, is a different API. Peppers also mentioned that we should have a better XP on .net 9 GA version, compared to the RC and preview ones.

brminnick:
Thanks for confirming. 

It uses TypedBinding under the hood. 

I'll close this for now until the Compiled Bindings interceptor has been updated to intercept bindings defined out of scope like these.

PR #2215, opened 2024-09-18
Support .NET 9.0

bijington:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes https://github.com/CommunityToolkit/Maui/issues/1865
 - Fixes https://github.com/CommunityToolkit/Maui/issues/1804
 - Fixes https://github.com/CommunityToolkit/Maui/issues/2109
 - Fixes #2356
 - Fixes https://github.com/CommunityToolkit/Maui/issues/2354
 - Fixes https://github.com/CommunityToolkit/Maui/issues/2205
 - Fixes https://github.com/CommunityToolkit/Maui/issues/2370
 - Fixes https://github.com/CommunityToolkit/Maui/issues/2373
 - Fixes https://github.com/CommunityToolkit/Maui/issues/1783
 - Fixes https://github.com/CommunityToolkit/Maui/issues/2377
 - Fixes https://github.com/CommunityToolkit/Maui/issues/238
 - Fixes https://github.com/CommunityToolkit/Maui/issues/2385
 - Fixes https://github.com/CommunityToolkit/Maui/issues/2388

 ### PR Checklist ###
 
 - [ ] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [ ] Has tests (if omitted, state reason in description)
 - [ ] Has samples (if omitted, state reason in description)
 - [ ] Rebased on top of `main` at time of PR
 - [ ] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#contributing-code---best-practices)
 - [ ] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pulls 

### Breaking Changes

**Major Breaking Changes**
- .NET 8 no longer supported
- Xcode 16.1 Required 
- Stop automatically assigning `Behavior.BindingContext`
  - .NET MAUI Behaviors do not automatically inherit their parent's BindingContext (a breaking change from Xamarin.Forms)
  - Going forward, the .NET MAUI Community Toolkit will follow this implementation of Behaviors
- Increase minimum supported OS versions of `CommunityToolkit.Maui.MediaElement`
  - Android 26.0
  - iOS 15.0
  - MacCatalyst 15.0
- Increase minimum supported versions of `CommunityToolkit.Maui.Camera` to 15.0:
  -  iOS 15.0
  - MacCatalyst 15.0

**Minor Breaking Changes**
- `Expander` is not trim safe
[...]

PR #2348, opened 2024-11-17
Create locker.yml

VladislavAntonyuk:
Auto Lock issues 1 day after close


brminnick:
Thanks Vlad! What do you think about the Co Pilot review?

I'm cool with merging this despite the build errors since those will be fixed when we merger #2215.

VladislavAntonyuk:
@brminnick, I copied the file from the .NET MAUI repo. Let's merge it (and check if it works). I will also try to apply Copilot suggestions with the next PR.

PR #2413, opened 2024-12-26
Fix statusBar color changes on modal pages

### LINKED ISSUE #2370: [BUG] Net 9 Android - StatusBar Color overriden on modal pages

### Current Behavior

In MAUI Android, when adding StatusBarBehavior to change the color, and then going to a modal page - the status bar color gets replaced with the colorPrimary color from colors.xml.

This only happens on NET 9. In NET 8 the status bar color remained the same as the previous page.
I think its because int NET 9 in the App.cs you have to create a new Window for the app shell, but is there a solution to the status bar changing?

`protected override Window CreateWindow(IActivationState? activationState)
{
    var window = new Window(new AppShell());
    return window;
}`

### Expected Behavior

In the modal page, the status bar color should be the same as the previous normal page. This was the behaviour in NET 8

### Steps To Reproduce

1. Open repo
2. Run in NET 9 for Android and click on the button - see the status bar color change from orange to green/red.
3. Change to NET 8 - see the color stays orange throughout.

https://github.com/Zack-G-I-T/StatusBarIssue-Net9

### Environment

```markdown
- .NET MAUI CommunityToolkit:
- OS:
- .NET MAUI:
```

### Anything else?

_No response_

pictos:
### Description of Change ###

Created and register new service for handle the DialogFragment (Modal pages on Android).

 ### Linked Issues ###
 

 - Fixes #2370

 ### PR Checklist ###
 

src/CommunityToolkit.Maui/AppBuilderExtensions.shared.cs
```diff
@@ -26,6 +31,29 @@ public static MauiAppBuilder UseMauiCommunityToolkit(this MauiAppBuilder builder
...
+		}
+
+		builder.ConfigureLifecycleEvents(builder =>
+		{
+			builder.AddAndroid(androidBuilder =>
+			{
+				androidBuilder.OnCreate((activity, bundle) =>
+				{
+					if (activity is not AndroidX.AppCompat.App.AppCompatActivity componentActivity)
+					{
```
ne0rrmatrix, line 47: If this fails do we want to ouput a message to developer?

src/CommunityToolkit.Maui/Services/DialogFragmentService.android.cs
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	public void OnFragmentSaveInstanceState(FragmentManager fm, Fragment f, Bundle outState)
+	{
+	}
+
+	public void OnFragmentStarted(FragmentManager fm, Fragment f)
+	{
+		if (!IsDialogFragment(f, out var dialogFragment) || Platform.CurrentActivity is not AppCompatActivity activity)
+		{
```
ne0rrmatrix, line 55: If this fails do we want to output a message to developer?

brminnick:
Thanks Pedro! Is there any reason to keep the previous implementation?

I saw that you implemented it under a feature flag to give devs the option to opt-out. Would it be cool if we just ripped out the old code? Or is there a known-issue where devs would want to fall back to the previous implementation?

brminnick:
Thanks Pedro! Is there any reason to keep the previous implementation?

I saw that you implemented it under a feature flag to give devs the option to opt-out. Would it be cool if we just ripped out the old code? Or is there a known-issue where devs would want to fall back to the previous implementation?

src/CommunityToolkit.Maui/AppBuilderExtensions.shared.cs
```diff
@@ -26,6 +31,29 @@ public static MauiAppBuilder UseMauiCommunityToolkit(this MauiAppBuilder builder
...
+						return;
+					}
+
+					componentActivity.GetFragmentManager()?.RegisterFragmentLifecycleCallbacks(new MCTFragmentLifecycle(), false);
```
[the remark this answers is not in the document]
pictos: The `GetFragmentManager` is an extension method provided by .NET Maui and is used internally to get the correct fragment manager, so to keep everything aligned is better to keep it the way it's

pictos:
@brminnick, we should keep both implementations, this PR adds support for Modal pages in Android maui apps. 
In the short the reason is:

Our code for status bar looks for `Activity.Window` and change the color and style. `DialogFragment` has its own Window and that Window doesn't inherit anything from the `Activity.Window` so we need to handle that manually. Also, Android doesn't provide a way to know which DialogFragment is presented (imagine a scenario where you do 3 modal navigations, you will end up with 3 ModalPages and 3 DialogFragment, which one is on top and should change?)



Let me give the full context, so it will be clear why this is needed

For .NET 9 Shane and I [worked](https://github.com/dotnet/maui/pull/22869) to migrate all modal navigations to use `DialogFragment` to improve the general code and will be more aligned with android provides today, but it has some details that, at least for me, were unknown. So far, the way to intercept all the `DialogFragment`s is using the `Fragment.LifecycleCallback`.

MCT wasn't the only package that has problem with this change, I saw that Mpopups and a bottomSheet lib also faced issues due to this change in .NET Maui, and I believe we can face other issues in the future, like for our Popup implementation.

So, we land on this PR, where I choose to create a proxy for the `Fragment.LifecycleCallback`, making easy for devs to extend it and centralize it, if it works really well, I'll try to promote it to Maui for .NET 10.

Now answering your questions:

>  Is there any reason to keep the previous implementation?

Yes, because the old code handles all scenarios but Modal navigation. Sadly, there's not out of the box way to know if there's a `DialogFragment` in the live-tree and which one (can be more than one) is being presented to change its StatusBarColor. Maybe Android improve it in the near future.

> I saw that you implemented it under a feature flag to give devs the option to opt-out. Would it be cool if we just ripped out the old code? Or is there a known issue where devs would want to fall back to the previous implementation?

The idea of having it as a feature flag, is to give more power to devs. Right now, our implementation just handles StatusBarColor, but in their apps they want to work around/fix issues on other packages, so they can implement the interface and have full control over all Android Modal Navigation aspects.

Let me know if you've any question or consideration

src/CommunityToolkit.Maui/Services/DialogFragmentService.android.cs
pictos: No, if it fails is because the Fragment isn't a DialogFragment and we want to silently ignore it. The CurrentActivity check is just to make sure isn't null, I used the strongly typed to reuse it, and we already check and log when the Activity isn't the desired type

src/CommunityToolkit.Maui/Services/DialogFragmentService.android.cs
```diff
@@ -0,0 +1,146 @@
...
+		var window = dialog.Window;
+
+		bool isColorTransparent = platformColor == Android.Graphics.Color.Transparent;
+
+		if (OperatingSystem.IsAndroidVersionAtLeast(30))
+		{
+			var windowInsetsController = window.InsetsController;
+			var appearance = activity.Window.InsetsController?.SystemBarsAppearance;
+
+			Debug.Assert(windowInsetsController is not null);
```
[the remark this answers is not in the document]
pictos: chagend

brminnick:
Thanks Pedro!! Yup - that makes perfect sense. 

FYI - I tweaked some of the naming and moved all of the code to `CommunityToolkit.Maui.Core` because it doesn't use `Microsoft.Maui.Controls` and our existing `StatusBar` code (eg `CommunityToolkit.Maui.Core.Platform.Statusbar`) also lives in `CommunityToolkit.Maui.Core`. 

Could you open a  [Docs PR](https://github.com/MicrosoftDocs/CommunityToolkit/pulls) documenting this new option and include an example for devs who are interested in expanding upon our implementation?

src/CommunityToolkit.Maui.Core/Services/DialogFragmentService.android.cs
```diff
@@ -68,18 +65,16 @@ static void HandleStatusBarColor(DialogFragment dialogFragment, AppCompatActivit
...
 		var platformColor = new Android.Graphics.Color(statusBarColor);
-		var dialog = dialogFragment.Dialog;
-
-		Debug.Assert(dialog is not null);
-		Debug.Assert(dialog.Window is not null);
-
-		var window = dialog.Window;
+		if (dialogFragment.Dialog?.Window is not Window dialogWindow)
+		{
+			throw new InvalidOperationException("Dialog window cannot be null");
```
pictos, line 70: @brminnick  we shouldn't throw here, since there's no way for devs to catch this exception. I used `Debug.Assert` because it's what the runtime/sdk uses to _safe check for null_ . But here you're checking against a type which can be false and throw this exception.

pictos:
@brminnick I added the implementation outside `Core` in order to make it `not public`, the `Core` project is meant to not have all members public so other toolkit implementors can access everything.

src/CommunityToolkit.Maui.Core/Services/DialogFragmentService.android.cs
brminnick: Using [`Debug.Assert()`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debug.assert?view=net-9.0) is a bit of an anti-pattern because it only works when building in Debug configuration:

  > [Debug.Assert](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debug.assert?view=net-9.0) method works only in debug builds.

This means that when users consume our library and the value of `dialog.Window` is `null`, the following code in Release builds is guaranteed to throw an unhelpful `NullReferenceException`:

```cs
Debug.Assert(dialog is not null);
Debug.Assert(dialog.Window is not null);

var window = dialog.Window;
```

In what scenario would this code be executed called _and_ `dialog.Window` also be null? If this scenario exists, I agree that we should implement a solution.

brminnick:
> the Core project is meant to not have all members public so other toolkit implementors can access everything

Is this documented? I don't recall setting this rule when I created `CommunityToolkit.Maui.Core`. 

[CONTRIBUTING.md](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#where-to-make-your-changes) only includes one rule, that new code belongs in `CommunityToolkit.Maui.Core` if  it does not use `Microsoft.Maui.Controls`

I'll add the `needs-discussion` label to ask the the team in our January Standup. If you're able to join too, we'll bump this discussion to the top of the meeting!

src/CommunityToolkit.Maui.Core/Services/DialogFragmentService.android.cs
pictos: I think just returning is enough, the only scenario I can imagine that will happening is when the app is backgrounded or is moving into background.

pictos:
@brminnick , I would love to join, let me see here and I'll confirm with you and team on discord

src/CommunityToolkit.Maui.Core/Services/DialogFragmentService.shared.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Runtime.Versioning;
+
+namespace CommunityToolkit.Maui.Core.Services;
+
+/// <summary>
+/// An Android-specific service to support <see cref="CommunityToolkit.Maui.Core.Platform.StatusBar"/> on modal pages
+/// </summary>
+[EditorBrowsable(EditorBrowsableState.Never)]
+[SupportedOSPlatform("Android")]
+public sealed partial class DialogFragmentService
```
pictos, line 11: Just curious, why add a shared version if the feature is just for android?

brminnick: Good question! I added it when I was seeing if I could remove `#if ANDROID` from the shared code in this PR. 

I was ultimately unsuccessful in my attempt, but I decided to keep `DialogFragmentService.shared.cs` now that it's public since it could possibly help devs who want to reference it without an #IfDef:

```cs
if(OperatingSystem.IsAndroid())
{
  // Use `DialogFragmentService` without an #IfDef
} 
```

brminnick:
Thanks Pedro! Approved ✅

Could you open up a [Docs PR](https://github.com/MicrosoftDocs/CommunityToolkit/pulls) to document the following: 
1. Document the new option in `CommunityToolkit.Maui.Core.Options`
2. Document that it's `opt-in` by default (ie `Options.ShouldUseStatusBarBehaviorOnAndroidModalPage` is `true` by default`
3. Document how devs implement a custom `IDialogFragmentService`

src/CommunityToolkit.Maui.Core/Services/FragmentLifecycleManager.android.cs
```diff
@@ -0,0 +1,89 @@
+﻿using System.ComponentModel;
+using Android.Content;
+using FragmentManager = AndroidX.Fragment.App.FragmentManager;
+
+namespace CommunityToolkit.Maui.Core.Services;
+
+[EditorBrowsable(EditorBrowsableState.Never)]
```
pictos, line 7: Just realized that devs will need to access this one in order to implement their own `IDialogFragmentService`
```suggestion
```

TheCodeTraveler (approved):

PR #2418, opened 2024-12-28
Fix crash when multiple Android Service are active

### LINKED ISSUE #2419: [BUG] Using MediaElement in Android causes crash

### Current Behavior

The following XAML code produces a crash on my Android Tablet device:

```xaml
<toolkit:MediaElement
    x:Name="MediaElement"
    Grid.Row="1"
    ShouldAutoPlay="False"
    ShouldKeepScreenOn="True"
    ShouldShowPlaybackControls="True"
    Aspect="AspectFill"
    MetadataTitle="{Binding VideoTitle}"
    Source="{Binding VideoSource}" />
```

This happens on my case when the previous page *has* another running [Android Service](https://github.com/jfversluis/Plugin.Maui.ScreenRecording), but I think it has nothing to do with it, rather, the Android Notification in [MediaControlsService](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs#L117) setup seems to be modifying the Action object behavior even before the [Notification object](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs#L162-L171) has been configured.

There is a similar issue in [StackOverflow](https://stackoverflow.com/q/47920158/2304737), that seems to back up my claim on this and the solution was to add the Actions first before configuring it.

### Expected Behavior

When another Android Service is running, the `MediaControlsService` shouldn't crash the app. A simple code example like the code below should also be expected to run without issues:

### Page.xaml
```xaml
<toolkit:MediaElement
    x:Name="MediaElement"
    Grid.Row="1"
    ShouldAutoPlay="False"
    ShouldKeepScreenOn="True"
    ShouldShowPlaybackControls="True"
    Aspect="AspectFill"
    MetadataTitle="{Binding VideoTitle}"
    Source="{Binding VideoSource}" />
```

### MauiProgram.cs
```cs
public static class MauiProgram
{
    public static MauiApp CreateMauiApp()
    {
        builder
            .UseMauiApp<App>()
            .UseMauiCommunityToolkit()
            .UseMauiCompatibility()
            .UseScreenRecording()                     // The other Android Service
            .UseMauiCommunityToolkitMediaElement();   // MediaControlsService
        // ...
    }
}
```

### AndroidManifest.xml
```xml
    <!-- other things -->

    <!-- Video Player -->
    <service android:name="communityToolkit.maui.media.services" android:exported="false" android:enabled="true" android:foregroundServiceType="mediaPlayback">
      <intent-filter>
        <action android:name="android.intent.action.MEDIA_BUTTON" />
      </intent-filter>
      <intent-filter>
        <action android:name="androidx.media3.session.MediaSessionService"/>
      </intent-filter>
    </service>
    
  </application>
	<uses-permission android:name="android.permission.INTERNET" />
	<uses-permission android:name="android.permission.CAMERA" />
  <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
  <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
  <uses-permission android:name="android.permission.RECORD_AUDIO" />
  <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
  <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
  <uses-permission android:name="android.permission.MEDIA_CONTENT_CONTROL" />
  <!-- This one is only needed when targeting API 34 and up -->
  <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PROJECTION" />
```

### Steps To Reproduce

1. Install [ScreenRecording](https://github.com/jfversluis/Plugin.Maui.ScreenRecording)
2. Install [MediaElement](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/views/mediaelement)
3. Create a page that starts ScreenRecording and stops it as soon as it navigates to next page
4. Next page would be the one hosting `MediaElement`
5. Play a video file from file system (file system might not be necessary here)
6. As soon as it transitions to the page with `MediaElement`, it would crash

https://github.com/mr5z/Repro-Maui.MediaElement.ScreenRecording

### Environment

```markdown
- .NET MAUI CommunityToolkit: 10.0.0
- OS: Android 13
- .NET MAUI: 9.0.21
```

### Anything else?

mr5z:
### Description of Change ###

- Code refactors
  - Change of naming convention of constants
  - Removal of confusing code
  - Update Android Service name
- Fix crash that occurs in Android Service

 ### Linked Issues ###
 

 - Fixes #2419

 ### PR Checklist ###
 
### DIFF
--- a/src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
@@ -18,15 +18,20 @@ using Resource = Microsoft.Maui.Resource;
 namespace CommunityToolkit.Maui.Media.Services;
 
 [SupportedOSPlatform("Android26.0")]
-[Service(Exported = false, Enabled = true, Name = "communityToolkit.maui.media.services", ForegroundServiceType = ForegroundService.TypeMediaPlayback)]
+[Service(Exported = false, Enabled = true, Name = "CommunityToolkit.Maui.Media.Services", ForegroundServiceType = ForegroundService.TypeMediaPlayback)]
 class MediaControlsService : Service
 {
-	public const string ACTION_PLAY = "MediaAction.play";
-	public const string ACTION_PAUSE = "MediaAction.pause";
-	public const string ACTION_UPDATE_UI = "CommunityToolkit.Maui.Services.action.UPDATE_UI";
-	public const string ACTION_UPDATE_PLAYER = "CommunityToolkit.Maui.Services.action.UPDATE_PLAYER";
-	public const string ACTION_REWIND = "MediaAction.rewind";
-	public const string ACTION_FASTFORWARD = "MediaAction.fastForward";
+	public const string ActionPlay = "MediaAction.play";
+	public const string ActionPause = "MediaAction.pause";
+	public const string ActionUpdateUI = "CommunityToolkit.Maui.Services.action.UPDATE_UI";
+	public const string ActionUpdatePlayer = "CommunityToolkit.Maui.Services.action.UPDATE_PLAYER";
+	public const string ActionRewind = "MediaAction.rewind";
+	public const string ActionFastForward = "MediaAction.fastForward";
+
+	public const string NotificationChannelId = "Maui.MediaElement";
+	public const string NotificationChannelName = "Transport Controls";
+
+	public const int NotificationId = 2024;
 
 	bool isDisposed;
 
@@ -50,7 +55,7 @@ class MediaControlsService : Service
 
 		if (!string.IsNullOrEmpty(intent.Action) && receiveUpdates is not null)
 		{
-			BroadcastUpdate(ACTION_UPDATE_PLAYER, intent.Action);
+			BroadcastUpdate(ActionUpdatePlayer, intent.Action);
 		}
 
 		StartForegroundService(intent).AsTask().ContinueWith(t =>
@@ -69,7 +74,7 @@ class MediaControlsService : Service
 
 	static void CreateNotificationChannel(NotificationManager notificationMnaManager)
 	{
-		var channel = new NotificationChannel("1", "notification", NotificationImportance.Low);
+		var channel = new NotificationChannel(NotificationChannelId, NotificationChannelName, NotificationImportance.Low);
 		notificationMnaManager.CreateNotificationChannel(channel);
 	}
 
@@ -91,7 +96,7 @@ class MediaControlsService : Service
 		{
 			receiveUpdates = new ReceiveUpdates();
 			receiveUpdates.PropertyChanged += OnReceiveUpdatesPropertyChanged;
-			LocalBroadcastManager.GetInstance(this).RegisterReceiver(receiveUpdates, new IntentFilter(ACTION_UPDATE_UI));
+			LocalBroadcastManager.GetInstance(this).RegisterReceiver(receiveUpdates, new IntentFilter(ActionUpdateUI));
 		}
 
 		OnSetupAudioServices();
@@ -112,20 +117,15 @@ class MediaControlsService : Service
 		var style = new AndroidX.Media.App.NotificationCompat.MediaStyle();
 		style.SetMediaSession(token);
 
-		if (Build.VERSION.SdkInt >= BuildVersionCodes.S)
-		{
-			style.SetShowActionsInCompactView(0, 1, 2, 3);
-		}
-
 		if (Build.VERSION.SdkInt < BuildVersionCodes.Tiramisu
 			&& notification is null)
 		{
-			notification = new NotificationCompat.Builder(Platform.AppContext, "1");
+			notification = new NotificationCompat.Builder(Platform.AppContext, NotificationChannelId);
 			OnSetIntents();
 			await OnSetContent(mediaManagerIntent, cancellationToken).ConfigureAwait(false);
 		}
 
-		notification ??= new NotificationCompat.Builder(Platform.AppContext, "1");
+		notification ??= new NotificationCompat.Builder(Platform.AppContext, NotificationChannelId);
 
 		notification.SetStyle(style);
 		notification.SetSmallIcon(_Microsoft.Android.Resource.Designer.Resource.Drawable.exo_styled_controls_audiotrack);
@@ -142,13 +142,13 @@ class MediaControlsService : Service
 
 		if (OperatingSystem.IsAndroidVersionAtLeast(29))
 		{
-			StartForeground(1, notification.Build(), ForegroundService.TypeMediaPlayback);
+			StartForeground(NotificationId, notification.Build(), ForegroundService.TypeMediaPlayback);
 			return;
 		}
 
 		if (OperatingSystem.IsAndroidVersionAtLeast(26))
 		{
-			StartForeground(1, notification.Build());
+			StartForeground(NotificationId, notification.Build());
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
```diff
@@ -18,15 +18,20 @@
 namespace CommunityToolkit.Maui.Media.Services;
 
 [SupportedOSPlatform("Android26.0")]
-[Service(Exported = false, Enabled = true, Name = "communityToolkit.maui.media.services", ForegroundServiceType = ForegroundService.TypeMediaPlayback)]
+[Service(Exported = false, Enabled = true, Name = "CommunityToolkit.Maui.Media.Services", ForegroundServiceType = ForegroundService.TypeMediaPlayback)]
```
ne0rrmatrix, line 21: Changing the case to uppercase will break older API's compatibility and prevent app from running on Android 23 or below. We only support android 26+ now on media element but I do want to point out this would break some current apps. Also changing this here requires changing the string in the manifest to match which has not been done.

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
```diff
@@ -18,15 +18,20 @@
...
-	public const string ACTION_UPDATE_UI = "CommunityToolkit.Maui.Services.action.UPDATE_UI";
-	public const string ACTION_UPDATE_PLAYER = "CommunityToolkit.Maui.Services.action.UPDATE_PLAYER";
-	public const string ACTION_REWIND = "MediaAction.rewind";
-	public const string ACTION_FASTFORWARD = "MediaAction.fastForward";
+	public const string ActionPlay = "MediaAction.play";
+	public const string ActionPause = "MediaAction.pause";
+	public const string ActionUpdateUI = "CommunityToolkit.Maui.Services.action.UPDATE_UI";
+	public const string ActionUpdatePlayer = "CommunityToolkit.Maui.Services.action.UPDATE_PLAYER";
+	public const string ActionRewind = "MediaAction.rewind";
+	public const string ActionFastForward = "MediaAction.fastForward";
```
pictos, line 29: Why did you change the pattern? Can you revert and follow our pattern

pictos (changes requested):
Thanks for your PR, but I can't see what I should review. Where's the actual fix for the issue?
Can you also revert all code style changes that you made and follow ours?

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
```diff
@@ -160,16 +160,15 @@ void OnReceiveUpdatesPropertyChanged(object? sender, PropertyChangedEventArgs e)
...
 		{
 			notification.AddAction(actionPause);
 		}
-		if (receiveUpdates.Action is ACTION_PAUSE)
+		if (receiveUpdates.Action is ActionPause)
 		{
 			notification.AddAction(actionPlay);
 		}
 		notification.AddAction(actionNext);
-		notification.Build();
```
mr5z, line 172: This is an immutable operation. It basically does nothing so I think it's better to remove it.

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
```diff
@@ -18,15 +18,20 @@
...
-	public const string ACTION_REWIND = "MediaAction.rewind";
-	public const string ACTION_FASTFORWARD = "MediaAction.fastForward";
+	public const string ActionPlay = "MediaAction.play";
+	public const string ActionPause = "MediaAction.pause";
+	public const string ActionUpdateUI = "CommunityToolkit.Maui.Services.action.UPDATE_UI";
+	public const string ActionUpdatePlayer = "CommunityToolkit.Maui.Services.action.UPDATE_PLAYER";
+	public const string ActionRewind = "MediaAction.rewind";
+	public const string ActionFastForward = "MediaAction.fastForward";
+
+	public const string NotificationChannelId = "Maui.MediaElement";
```
mr5z, line 31: I have introduced a constant to easily spot the notification ID of this control. I think it's important to keep track of notification ID as we introduce more foreground Android services and I also think it's important to name them appropriately instead of just "1" as what was previously implemented.

mr5z, line 31: I have introduced a constant to easily spot the notification ID of this control. I think it's important to keep track of notification ID as we introduce more foreground Android services since the same notification ID could not co-exist in the same Android application. I also think it's important to name them appropriately instead of just "1" as what was previously implemented.

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
```diff
@@ -18,15 +18,20 @@
...
+	public const string NotificationChannelName = "Transport Controls";
```
mr5z, line 32: I have also introduced a const for channel name so we can easily track it. This is a user facing text so I think it's better to name it appropriately instead of just "notification". I am not sure about the exact text to put but I think title casing is the norm.

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
```diff
@@ -18,15 +18,20 @@
...
+
+	public const int NotificationId = 2024;
```
mr5z, line 34: 2024 I think is unique enough not to clash with other foreground service ID.

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
```diff
@@ -112,20 +117,15 @@ async ValueTask InitializeNotification(MediaSessionCompat mediaSession, Intent m
 		var style = new AndroidX.Media.App.NotificationCompat.MediaStyle();
 		style.SetMediaSession(token);
 
-		if (Build.VERSION.SdkInt >= BuildVersionCodes.S)
-		{
-			style.SetShowActionsInCompactView(0, 1, 2, 3);
```
mr5z, line 117: I think this operation must be called after setting actions to the notification object, but it's empty at this point and it's the main reason for crash issues reported here. I have also tried including this after the notification action has been set but it just completely hides the buttons in the notification when I tested it so I think it's better to remove this until we have a better reason to keep it and have it implement properly.

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
mr5z: Isn't PascalCasing the standard convention in C# to name consts?

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
mr5z: I really have no strong opinion on what naming convention should be done for android-net, but in native Android, it's `package.name.ServiceName` but it can be addressed relatively in AndroidManifest.xml as `.ServiceName`. Maybe we should follow that? I think it is still early for this library to think about the naming convention before the inconsistency spreads out even more, but if you think this is the convention we should follow moving forward, I'm happy to revert it.

src/CommunityToolkit.Maui.MediaElement/Services/MediaControlsService.android.cs
mr5z: @pictos I've left some comments on some of my code changes.

jfversluis:
Seems like there are conflicts now with the switch to Media3. Can you maybe try the latest release and see if this still happens and if needed apply the fix to the latest code changes? Thanks!

ne0rrmatrix:
I'm closing this PR. The changes in with the introduction of media 3 have removed all of the code that was changed or updated in this PR. I appreciate the effort that went into creating this.

PR #2452, opened 2025-01-17
[Housekeeping] Add Benchmarks to GitHub Actions

brminnick:
This PR adds the `Run Benchmarks` step to the GitHub Actions CI/CD Workflow.

**Additional Information**
This PR also adds the following housekeeping updates to the GitHub Actions CI/CD Workflow:
- `actions/setup-dotnet`
  - Update to v4
  - Add `dotnet-quality: 'ga'`
    - This ensures the build step does not install a preview version of .NET
- `Install .NET MAUI Workload`
  - Add `dotnet workload update` following `dotnet workload install maui`
    -  This is a workaround I've found to ensure pre-installed .NET workloads on the hosted agent are up-to-date
    - It's helped us fix CI/CD errors in the past
- `Set Xcode version`
  - Print the installed version of Xcode
    - This helps us identify when new versions of Xcode are supported in the hosted agents
    - This also helps us understand if this step fails when a hosted agent is running an old version of macOS that does not yet include the needed version of Xcode
- `build_library`
  - Add `env: VSTEST_TESTHOST_SHUTDOWN_TIMEOUT: 1100`
    - This fixes "The active test run was aborted. Reason: Test host process crashed "
- Build each project individually 
  - Yes, it's redundant since `dotnet pack` first builds the csproj file, however, building each project individually makes it easier for us to quickly identify which project is throwing a compiler error and whether the error is occurring on `dotnet build` step or on `dotnet pack`


.github/workflows/dotnet-build.yml
```diff
@@ -138,8 +150,47 @@ jobs:
       - name: Display dotnet info
         run: dotnet --info
 
-      - name: Build CommunityToolkit.Maui
-        run: dotnet build -c Release ${{ env.PathToLibrarySolution }}
+      - name: 'Build CommunityToolkit.Maui.Analyzers'
```
VladislavAntonyuk, line 153: It was useful before net 9, now the new output (--tl) provides all information. Also with migration to GitHub actions I focused on pipeline execution time. Building solution faster than each csproj. Also I would say in 90% we change only Core and MAUI project and tests. Analyzers remain the same.
Also in case we have a new project we don't need to change pipeline.

.github/workflows/dotnet-build.yml
```diff
@@ -138,8 +150,47 @@ jobs:
...
+        run: dotnet build ${{ env.PathToCommunityToolkitMapsCsproj }} -c Release -p:PackageVersion=${{ env.NugetPackageVersionMaps }} -p:Version=${{ env.NugetPackageVersion }}
+
+      - name: 'Build CommunityToolkit.Maui.Core'
+        run: dotnet build ${{ env.PathToCommunityToolkitCoreCsproj }} -c Release -p:PackageVersion=${{ env.NugetPackageVersion }} -p:Version=${{ env.NugetPackageVersion }}
+
+      - name: 'Build CommunityToolkit.Maui'
+        run: dotnet build ${{ env.PathToCommunityToolkitCsproj }} -c Release -p:PackageVersion=${{ env.NugetPackageVersion }} -p:Version=${{ env.NugetPackageVersion }}
+
+      - name: 'Run CommunityToolkit.Maui.Analyzers.UnitTests'
+        run: 'dotnet test ${{ env.PathToCommunityToolkitAnalyzersUnitTestCsproj }} -c Release'
```
VladislavAntonyuk, line 193: We run test for the whole solution. This can be removed

.github/workflows/dotnet-build.yml
```diff
@@ -268,3 +321,52 @@ jobs:
           --source https://api.nuget.org/v3/index.json
           --api-key ${{ secrets.NUGET_PACKAGE_PUSH_TOKEN }}
           --skip-duplicate
+
+  run_benchmarks:
```
VladislavAntonyuk, line 325: Can we run benchmarks as part of the main pipeline? I want the pipeline runs as fast as possible. Also do we need benchmarks on every pipeline? Maybe it can be a manual step?

.github/workflows/dotnet-build.yml
```diff
@@ -85,6 +90,8 @@ jobs:
   build_library:
     name: Build Library
     runs-on: ${{ matrix.os }}
+    env:
+      VSTEST_TESTHOST_SHUTDOWN_TIMEOUT: 1100 # Fixes "The active test run was aborted. Reason: Test host process crashed"
```
VladislavAntonyuk, line 94: Do we have this issue on new pipeline?
I also curious if this variable is specific to Azure Pipelines or exists in Github.

.github/workflows/dotnet-build.yml
brminnick: Good point! I've edited [our Ruleset](https://github.com/CommunityToolkit/Maui/settings/rules/2541778) so that the Benchmark workflow is not required to merge the PR. 

Screenshot 2025-01-19 at 8 22 38 AM

They'll still run, giving us benchmark data that we can leverage for Analyzer updates, but we don't have to wait on them to finish to merge.

.github/workflows/dotnet-build.yml
brminnick: Haha - I wish I knew the answer to that! 

I don't have direct evidence that supports whether we still need it in GitHub Actions or not, but I'd kinda rather keep it just in case since it doesn't introduce any adverse effects to the pipeline and it may save us headaches in the future.

.github/workflows/dotnet-build.yml
brminnick: Ah - good catch! Removed ✔️

.github/workflows/dotnet-build.yml
brminnick: Hey Vlad! Yup - I LOVE that new feature in the dotnet CLI 💯

The goal here is to make things easier / simpler for us maintainers, reducing our cognitive load when reviewing PRs. By separating each build step, we're able to quickly see at a glance which project has a build error. Separating the build steps also makes it easier for new devs contributing to the Toolkit to understand what they broke. And then, if needed, we can always click into that step to see the error message. In other words, sifting through CI/CD logs sucks and we'll be more efficient reviewing PRs by seeing this information at a higher level. 

Here's an example that quickly shows the `Build CommunityToolkit.Maui.Analyzers` step failed. If the PR I'm reviewing doesn't touch `CommunityToolkit.Maui.Analyzers`, I know we have a problem in somewhere else in our solution not caused by this PR. And the developer from the community who submitted this PR also knows that the failure isn't their fault. 

image

If we combine every build step, we will always have to click into the failed build step and sift through the logs to understand what broke. 

In other words, separating the build steps may increase our build times by a few seconds, but us maintainers will gain much more productivity!

VladislavAntonyuk: what do you think if we move it to a separate yaml file?
I see we can have 3 files
- library.yaml - installs .net 9.0.102 and have sign/release steps
- Samples.yaml - installs latest versions, do not need any logic except building samples app
- benchmarks.yaml - can be executed manually or as part of pipeline.

Such separation allows us to remove confusion in variables we have at the beginning of the file.

.github/workflows/dotnet-build.yml
```diff
@@ -138,10 +150,46 @@ jobs:
...
+
+      - name: 'Build CommunityToolkit.Maui.Maps'
+        run: dotnet build ${{ env.PathToCommunityToolkitMapsCsproj }} -c Release -p:PackageVersion=${{ env.NugetPackageVersionMaps }} -p:Version=${{ env.NugetPackageVersion }}
 
-      - name: Run CommunityToolkit.Maui.UnitTests
+      - name: 'Build CommunityToolkit.Maui.Core'
+        run: dotnet build ${{ env.PathToCommunityToolkitCoreCsproj }} -c Release -p:PackageVersion=${{ env.NugetPackageVersion }} -p:Version=${{ env.NugetPackageVersion }}
+
+      - name: 'Build CommunityToolkit.Maui'
+        run: dotnet build ${{ env.PathToLibrarySolution }} -c Release -p:PackageVersion=${{ env.NugetPackageVersion }} -p:Version=${{ env.NugetPackageVersion }}
```
VladislavAntonyuk, line 190: this should be a library, not solution

brminnick: I used the Library SLN on purpose to help ensure the pipeline always builds all of the projects. 

In the future, a PR may add a new library but forget to update the pipeline. This ensures that the pipeline will still compile the new code until we remember to add it as a separate step.

VladislavAntonyuk: Sorry, in that case I still don't see the benefit of building each project individually to build the whole solution later.
I have no strong objections merging the PR, but I still voting keeping it as simple as possible (less code). Once we get the issue with build investigation we can back to the changes.

PR #2479, opened 2025-01-29
[Housekeeping] Fix Unit Test Failure: `Catastrophic failure: System.ArgumentOutOfRangeException`

TheCodeTraveler:
### Description of Change ###

This PR fixes the [`ArgumentOutOfRangeException` occurring in  the `Run All Unit Tests` step of our `Build Library (windows-latest)` pipeline](https://github.com/CommunityToolkit/Maui/actions/runs/13037793620/job/36372408491): 
> [xUnit.net 00:00:29.07] Catastrophic failure: System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')

The exception was happening because both `MockPopup` and `MockPopupHandler` were both calling `HandlerCompleteTCS.SetResult();` in `OnClosed` causing a race condition where our test would sometimes fail and sometimes pass. 

 ### PR Checklist ###
 
This PR updates the PopupTests to use `Assert.ThrowsAnyAsync<OperationCanceledException>()` which will pass when either an `OperationCanceledException` is thrown or an `TaskCanceledException` is thrown since `TaskCanceledException` derives from `OperationCanceledException`; `Assert.ThrowsAnyAsync<T>()` allows for derived types, whereas `Assert.ThrowsAsync<T>()` does not allow derived types.

This PR fixes an edge case where `await popupDismissedTaskCompletionSource.Task.WaitAsync(token)` does not throw a `TaskCanceledException` despite the token being canceled when `HandlerCompleteTCS.Task` had already completed; the fix is to manually add ` token.ThrowIfCancellationRequested();` in `Popup.OnClosed()`.
 
This PR changes `readonly IPopup poup` -> `readonly MockPopup popup` because of the discrepancy between `void IPopup.OnDismissedByTappingOutsideOfPopup()` and the awaitable `Task MockPopup.OnDismissedByTappingOutsideOfPopup()`. I also added the missing `await` keyword when the tests call `MockPopup.OnDismissedByTappingOutsideOfPopup()`.

This PR adds `readonly MockPopupHandler popupHandler` to `PopupTests`, and assigns this handler to `readonly MockPopup popup` in the constructor: `popupHandler = CreateElementHandler<MockPopupHandler>(popup);`

This PR also adds [GitHubActionsTestLogger](https://github.com/Tyrrrz/GitHubActionsTestLogger) to provide us with more comprehensive logs when tests fail. This will help us solve test errors in the future.

### DIFF
--- a/.github/workflows/dotnet-build.yml
+++ b/.github/workflows/dotnet-build.yml
@@ -190,7 +190,7 @@ jobs:
         run: dotnet build ${{ env.PathToLibrarySolution }} -c Release -p:PackageVersion=${{ env.NugetPackageVersion }} -p:Version=${{ env.NugetPackageVersion }} 
 
       - name: Run All Unit Tests
-        run: dotnet test -c Release ${{ env.PathToLibrarySolution }} --settings ".runsettings" --collect "XPlat code coverage" --logger trx --results-directory ${{ runner.temp }}
+        run: dotnet test -c Release ${{ env.PathToLibrarySolution }} --settings ".runsettings" --collect "XPlat code coverage" --logger trx --results-directory ${{ runner.temp }} --logger GitHubActions
 
       - name: Publish Test Results
         if: runner.os == 'Windows'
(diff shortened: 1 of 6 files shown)


TheCodeTraveler:
@dotnet-policy-service agree

ne0rrmatrix (approved):
Looks like this solves a huge bug. Good job! It is nice to see this bug getting quashed. I had no idea what had been causing my builds to fail.

PR #2486, opened 2025-02-02
XUnit V3

VladislavAntonyuk:
- Replaced all CancellationToken.None with new TestContext.Current.CancellationToken
- Fixed token validation for cancellation
- Removed old VSTest and replaced with new Microsoft Testing Platform https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-platform-intro?tabs=dotnetcli
- Publish Test artifacts and prevent too many concurrent builds running
- Renamed azure-pipelines.yml as we will remove it soon.


ne0rrmatrix (commented):
This is amazing! I am currently on iteration 28 with zero errors! But the one issue still persisting in tests is line 63 `public async Task InvalidValue_InvalidStyle()` in ValidationBehaviorTests.cs is failing randomly. My tests have that commented out. If you can figure out that one test failing this PR will be even better. I am waiting for the local run to complete a 1000 round test. I am going to wait to see how it runs but i think this PR is very close to being ready. It is 100 percent better than current tests now. But that one test is still failing randomly.

VladislavAntonyuk:
@ne0rrmatrix do you see the test (InvalidValue_InvalidStyle) fails randomly on your local env?

ne0rrmatrix (approved):
This looks great. I am happy with the results of this change. I see a 5X increase in average reliability after extensive testing in local environment.

Issue #2443, opened 2025-01-15
[BUG] Touchbehavior Keyboard Access Click not working on Android


plewm:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [x] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

When you use an App on Android via Keyboard access/navigation and you try to hit a button via space bar for example, the animation is executed, but the click event is not raised or executed

### Expected Behavior

You would expect the Click to be executed normally, like it does on iOS

### Steps To Reproduce

1. Just add the Touchbehavior to any component
2. try using the App with Keyboard 
3. navigate to that component
4. press space bar to initiate click

In the Example App:
1. Navigate via Tab on keyboard to the Border with TouchBehavior
2. Press Space bar to Click -> nothing happens
3. Touch normally -> works
4. activate TalkBack
5. Navigate again by Tab on Keyboard
6. Press Space bar -> works

### Link to public reproduction project repository

https://github.com/plewm/TouchbehaviorExample

### Environment

```markdown
- .NET MAUI CommunityToolkit: 9.1.0
- OS: macOS 15.1 (24B83)
- .NET MAUI: 8.0.402.1
```

### Anything else?

From my debugging I found the issue to be the accessibilityManager and bool IsAccessibilityMode. it seems the accessibilityManager is null and accessibility mode is just false if the App is only used with keyboard. It will only be set if TalkBack is activated on the device. The Keyboard actions than also work, when TalkBack is activated as well. But the App should be usable with only keyboard without TalkBack as well. The Problem is, that the bool IsAccessibilityMode is checked in the OnClick method and is not reaching the HandleTouchEnded if AccessibilityMode is false.:

`void UpdateClickHandler()
	{
		if (view is null || !view.IsAlive())
		{
			return;
		}

		view.Click -= OnClick;
		if (IsAccessibilityMode || (IsEnabled && (Element?.IsEnabled ?? false)))
		{
			view.Click += OnClick;
			return;
		}
	}

void OnClick(object? sender, EventArgs args)
	{
		if (!IsEnabled)
		{
			return;
		}

		if (!IsAccessibilityMode)
		{
			return;
		}

		IsCanceled = false;
		HandleTouchEnded(TouchStatus.Completed);
	}`

[...]

PR #2497, opened 2025-02-06
fix TouchBehavior Android KeyBoard activation

plewm:
### Description of Change ###

 

 ### Linked Issues ###
 
https://github.com/CommunityToolkit/Maui/issues/2443

 - Fixes https://github.com/CommunityToolkit/Maui/issues/2443

 ### PR Checklist ###
 

plewm:
> @dotnet-policy-service agree

@dotnet-policy-service agree

plewm:
Can someone please verify those builds?

plewm:
do you have any idea why that one Test is Failing. It should not have anything todo with my change or a I mistaken?

ne0rrmatrix:
> @ne0rrmatrix  do you have any idea why that one Test is Failing. It should not have anything todo with my change or am I mistaken?

That test failing is completely unrelated to your PR. You can safely ignore it.

plewm:
okay so, can anyone approve that than?

plewm:
So how is this proceeding now? What can I do to get that thing merged finally?

bijington:
Your original comment said this solution isn't perfect does this mean you still want to improve it?

I do wonder whether we should add a mechanism to configure this new functionality in case keyboard support for a touch behavior isn't desired in all scenarios. What do you think?

plewm:
Well the reason for introducing this at all is a new ruling of the EU that requires accessibility for all Apps starting middle of the year. Our App will also be tested by a specific 3rd party because of that. So in general, KeyBoard access needs to be available 100% in my opinion.
For the overall usecase the solution I introduced, would be fine for most cases. An ideal solution would also provide a long touch functionality for the touchbehavior, because that is a feature of the normal touch functionality. 
I tried that out, but the keyDown event is just getting fired over and over and the properties of the event to detect if it was hold or pushed again are not working as expected. 

In regards to that my suggestion is, take that solution here for now, merge that to be fine for the general accessibility usecase and create another thread/issue for the long touch.

bijington:
Thanks. I'm not debating the value this adds to apps. I was debating whether developers would expect and always want something called `TouchBehavior` to provide keyboard support. To be honest I'd prefer a name like `InteractionBehavior` but that is a big breaking change.

My thought was we could introduce a property `EnableKeyboardSupport`, default it to `true` and handle your event subscriptions based on that property value. It would give developers a way to turn it off in case they don't need it.

As for the KeyDown being called continuously I have seen this recently while dealing with keyboard support for something else. Was this on iOS or macOS that you saw it? Let me see if I can dig out the issue

plewm:
Well the KeyBoard will not interfere with the standard behavior, so it should not create any issues for those cases I think, but we can still discuss that usecase. What do you think would be a proper case where you absolutely want to forbid KeyBoard usage for such view component?

The KeyDown problem I found was happening on Android for me, but build on Mac. I did not research the iOS implementation deeper, cause the KeyBoard access there just worked as it was supposed to.

bijington:
As a quick further follow-up have you tried using any of the accessibility options on Android before trying this fix? I am not entirely familiar with TouchBehavior but I can see there is explicit support in the Android implementation to make use of the [AccessibilityManager](https://developer.android.com/reference/android/view/accessibility/AccessibilityManager)

plewm:
you can take look into the bug ticket. I already explained the problem with the current implementation there:

https://github.com/CommunityToolkit/Maui/issues/2443

The problem results exactly from that accessibilityManager and how it is used.

bijington:
> you can take look into the bug ticket. I already explained the problem with the current implementation there:
> 
> #2443
> 
> The problem results exactly from that accessibilityManager and how it is used.

Sorry I missed that detail in the original report. If I am reading this correctly is the addition of this keyboard handling the right fix or should we be looking to improve how we are using the accessibilityManager? Also you mentioned that keyboard control works if TalkBack is enabled, what happens when you have both your keyboard handling and TalkBack enabled? You don't get 2 attempts to handle the interaction?

ne0rrmatrix:
> > you can take look into the bug ticket. I already explained the problem with the current implementation there:
> > #2443
> > The problem results exactly from that accessibilityManager and how it is used.
> 
> Sorry I missed that detail in the original report. If I am reading this correctly is the addition of this keyboard handling the right fix or should we be looking to improve how we are using the accessibilityManager? Also you mentioned that keyboard control works if TalkBack is enabled, what happens when you have both your keyboard handling and TalkBack enabled? You don't get 2 attempts to handle the interaction?

@bijington I believe the issue is about using the accessibility manager with touch. Looking at the bug report and the linked google docs it looks like touch should not use accessibility manager the way we are.

plewm:
You will also not have 2 (clicks) registered. Its either if you have TalkBack on a double tap needed or with keyboard the space button. The google docs say to my understanding, that you should not depend on the accessibilityManager state at all to init different workflows.
My fix removes the keyBoard action from the accessibilityManager workflow. I think that is the more correct approach.

bijington:
@plewm thank you for this. Apologies for the delay in getting back to you on this.

Also it was great to meet you the other day at MAUI Day! I hope you enjoyed the day and managed to travel back safely.

bijington (approved):
Thank you!

plewm:
@bijington yes it was a nice event and also good to meet some people working on that framework here in person. Some of us stayed a little longer in London as well. We had a lot of fun :)

regarding the pipeline issue again...we cant merge yet because of the failing test of the "validationBehavior". How are fixing that? I would really like to get that thing on track here :D

bijington:
We have a flaky test :(. I have kicked off another build but don't worry it won't impact this being merged

plewm:
@bijington  uh that looks perfect now. we are ready to merge 🎉 

who will be initiating the merge process? I can't do that it seems. I still have no button for that. 
I first thought that the failing test was  responsible, that I did not see the merge button. 

anyways, thanks for bearing with me here :)

PR #2518, opened 2025-02-14
Rename azure-pipelines.yml-old to azure-pipelines.yml

jfversluis:
Adding back the pipeline file for now until we are really sure that the GH Action works. Until then, we will need this as a fallback for releasing

I added an override in Azure Pipelines that the Pipeline builds should not be triggered for CI and PRs so it should only still kick in for when I add a tag to release (I hope...). For this PR it will probably still build.


bijington (approved):

jfversluis:
Build errors unrelated to this change, so merging this for now!

PR #2567, opened 2025-03-03
Fix: Animation behavior TapGestureRecognizer

### LINKED ISSUE #2489: [BUG] AnimationBehavior remove TapGestureRecognizer

### Current Behavior

I have quite a simple animation that rings the notification bell if a new notification comes in -
https://github.com/user-attachments/assets/21a09e58-6417-4437-9f3d-76c7d8ac4383

That works by just calling the `AnimateCommand` as documented here - https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/behaviors/animation-behavior#programmatically-triggering-the-animation

My issue is randomly pressing the notification button triggers the animation, and for it not too open up the notification list as it should be.

Looking at the code, it adds a `TapGestureRecognizer` to the control unless its triggered by an event - https://github.com/CommunityToolkit/Maui/blob/d8cf48db2900835dac424f1d7245708295c516fc/src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs#L87C3-L90C67

That `TapGestureRecognizer` then triggers the animation -  https://github.com/CommunityToolkit/Maui/blob/d8cf48db2900835dac424f1d7245708295c516fc/src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs#L108C3-L108C43

How do I remove the `TapGestureRecognizer` when triggering it via `AnimateCommand`?

### Expected Behavior

For there to be no `TapGestureRecognizer` added

### Steps To Reproduce

See above

n/a

### Environment

```markdown
- .NET MAUI CommunityToolkit: 10.0.0
- OS: More of an issue on iOS
- .NET MAUI: Latest
```

### Anything else?

_No response_

IeuanWalker:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #2489

 ### PR Checklist ###
 
Adds a new property `AnimateOnTap`, that is used to determine if a TapGestureRecognizer should be added to the control.

 Current behaviour is a bit strange, if you use an event to trigger the animation then the TapGestureRecognizer isn't added, but if you use a command to trigger the animation it is added. 

The current [documentation](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/behaviors/animation-behavior) just says -

> By default a TapGestureRecognizer is attached to the VisualElement and triggers the associated animation when that recognizer detects that the user has tapped or clicked on the VisualElement.

Which doesn't indicate a difference between using events/ commands.

Also with my use case _(mentioned in the linked issue)_, I'm using a command and I don't want the TapGestureRecognizer added, but with the current implementation that's not possible unless I assign an event.

> This is a breaking change for people who trigger an animation via a command and also want it to animate on tap. For those users they would need to update their code to include `AnimateOnTap="True"`. I can't see a way of fixing this without a breaking change due to the inconsistency between events/ commands.

image

### DIFF
--- a/src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
@@ -34,7 +34,7 @@ public partial class AnimationBehavior : EventToCommandBehavior
 	/// Backing BindableProperty for the <see cref="AnimateOnTap"/> property.
 	/// </summary>
 	public static readonly BindableProperty AnimateOnTapProperty =
-		BindableProperty.Create(nameof(AnimationType), typeof(bool), typeof(AnimationBehavior));
+		BindableProperty.Create(nameof(AnimateOnTap), typeof(bool), typeof(AnimationBehavior));
 
 	TapGestureRecognizer? tapGestureRecognizer;
 
--- a/src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
@@ -30,11 +30,17 @@ public partial class AnimationBehavior : EventToCommandBehavior
 	public static readonly BindableProperty AnimateCommandProperty =
 		BindableProperty.CreateReadOnly(nameof(AnimateCommand), typeof(Command<CancellationToken>), typeof(AnimationBehavior), default, BindingMode.OneWayToSource, propertyChanging: OnAnimateCommandChanging, defaultValueCreator: CreateAnimateCommand).BindableProperty;
 
+	/// <summary>
+	/// Backing BindableProperty for the <see cref="AnimateOnTap"/> property.
+	/// </summary>
+	public static readonly BindableProperty AnimateOnTapProperty =
+		BindableProperty.Create(nameof(AnimationType), typeof(bool), typeof(AnimationBehavior));
+
 	TapGestureRecognizer? tapGestureRecognizer;
 
 	/// <summary>
 	/// Gets the Command that allows the triggering of the animation.
-	/// 
+	///
 	/// NOTE: Apps should not directly set this property, treating it as read only. The setter is only public because
 	/// that's currently needed to make XAML Hot Reload work. Instead, apps should provide a value for this OneWayToSource
 	/// property by creating a binding, in XAML or C#. If done via C# use code like this:
@@ -63,37 +69,48 @@ public partial class AnimationBehavior : EventToCommandBehavior
 		get => (BaseAnimation?)GetValue(AnimationTypeProperty);
 		set => SetValue(AnimationTypeProperty, value);
 	}
+	/// <summary>
+	/// Whether a TapGestureRecognizer is added to the control or not
+	/// </summary>
+	public bool AnimateOnTap
+	{
+		get => (bool)GetValue(AnimateOnTapProperty);
+		set => SetValue(AnimateOnTapProperty, value);
+	}
 
 	/// <inheritdoc/>
 	protected override void OnAttachedTo(VisualElement bindable)
 	{
 		base.OnAttachedTo(bindable);
 
-		if (!string.IsNullOrWhiteSpace(EventName))
+		if(!string.IsNullOrWhiteSpace(EventName))
 		{
 			return;
 		}
 
-		if (bindable is ITextInput)
+		if(bindable is ITextInput)
 		{
 			throw new InvalidOperationException($"Animation Behavior can not be attached to {nameof(ITextInput)} without using the EventName property.");
 		}
 
-		if (bindable is not IGestureRecognizers gestureRecognizers)
+		if(bindable is not IGestureRecognizers gestureRecognizers)
 		{
 			throw new InvalidOperationException($"VisualElement does not implement {nameof(IGestureRecognizers)}.");
 		}
 
-		tapGestureRecognizer = new TapGestureRecognizer();
-		tapGestureRecognizer.Tapped += OnTriggerHandled;
+		if(AnimateOnTap)
+		{
+			tapGestureRecognizer = new TapGestureRecognizer();
+			tapGestureRecognizer.Tapped += OnTriggerHandled;
 
-		gestureRecognizers.GestureRecognizers.Add(tapGestureRecognizer);
+			gestureRecognizers.GestureRecognizers.Add(tapGestureRecognizer);
+		}
 	}
 
 	/// <inheritdoc/>
 	protected override void OnDetachingFrom(VisualElement bindable)
 	{
-		if (tapGestureRecognizer is not null)
+		if(tapGestureRecognizer is not null)
 		{
 			tapGestureRecognizer.Tapped -= OnTriggerHandled;
 			tapGestureRecognizer = null;
@@ -118,7 +135,7 @@ public partial class AnimationBehavior : EventToCommandBehavior
 
 	static void OnAnimateCommandChanging(BindableObject bindable, object oldValue, object newValue)
 	{
-		if (newValue is not Command<CancellationToken>)
+		if(newValue is not Command<CancellationToken>)
 		{
 			var animationBehavior = (AnimationBehavior)bindable;
 			throw new InvalidOperationException($"{nameof(AnimateCommand)} must of Type {animationBehavior.AnimateCommand.GetType().FullName}");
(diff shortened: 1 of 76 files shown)


src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
```diff
@@ -63,31 +69,27 @@ public BaseAnimation? AnimationType
...
 
-		if (!string.IsNullOrWhiteSpace(EventName))
+		if (AnimateOnTap && bindable is IGestureRecognizers gestureRecognizers)
 		{
-			return;
-		}
+			tapGestureRecognizer = new TapGestureRecognizer();
+			tapGestureRecognizer.Tapped += OnTriggerHandled;
 
-		if (bindable is ITextInput)
```
IeuanWalker, line 77: Wasnt sure why this check is needed

src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
```diff
@@ -63,31 +69,27 @@ public BaseAnimation? AnimationType
...
-		{
-			throw new InvalidOperationException($"Animation Behavior can not be attached to {nameof(ITextInput)} without using the EventName property.");
+			gestureRecognizers.GestureRecognizers.Add(tapGestureRecognizer);
 		}
```
IeuanWalker, line 92: Wasnt sure why this check was needed

TheCodeTraveler: Have you confirmed that `AnimationBehavior` can now be attached to an `ITextInput` without using the `EventName` property? 

I'm cool removing this as long as we have confirmation on each platform (iOS, Android, Windows, MacCatalyst and Tizen).

IeuanWalker: @TheCodeTraveler tested windows, android and iOS. All seem to work fine.

I don't work with MacCatalyst or Tizen, so I have updated the sample app, hoping someone else can test those platforms.

Only thing added to the page is the following _(VS is auto formatting the rest)_
```xaml
<Entry Margin="16,0" Placeholder="Animate on tap">
    <Entry.Behaviors>
        <mct:AnimationBehavior AnimateOnTap="True">
            <mct:AnimationBehavior.AnimationType>
                <behaviorPages:SampleScaleAnimation Easing="{x:Static Easing.Linear}" Length="100" />
            </mct:AnimationBehavior.AnimationType>
        </mct:AnimationBehavior>
    </Entry.Behaviors>
</Entry>
```

https://github.com/user-attachments/assets/46081c82-a024-4a0b-8a52-b5ebf2dc2c74

IeuanWalker:
If this PR is all good, im happy to look at updating the docs.
Just let me know :)

src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
```diff
@@ -63,31 +69,27 @@ public BaseAnimation? AnimationType
...
+		set => SetValue(AnimateOnTapProperty, value);
+	}
 
 	/// <inheritdoc/>
 	protected override void OnAttachedTo(VisualElement bindable)
 	{
 		base.OnAttachedTo(bindable);
 
-		if (!string.IsNullOrWhiteSpace(EventName))
+		if (AnimateOnTap && bindable is IGestureRecognizers gestureRecognizers)
```
bijington, line 86: I see `AnimateOnTap` is a set-able property, I think that means we need to react to its value changing and not just refer to it in `OnAttached`?

IeuanWalker: Could look at adding it, is there an of that somewhere?

_Previous code wouldn't add/ remove the tap gesture when an event was added/ removed_

bijington: good point on the `EventName` property. We should also fix! Maybe we do it in this PR or a follow-up one but I think if we add in a property that provides the support for changing whether animation is supported then we should respond to the property value changing.

An example of handling this is inside the base class `EventToCommandBehavior`

property changed handling
https://github.com/CommunityToolkit/Maui/blob/fea67c2ff72f89ed118be044a82d8a572e405bf0/src/CommunityToolkit.Maui/Behaviors/EventToCommandBehavior.shared.cs#L17

and the associated method
https://github.com/CommunityToolkit/Maui/blob/fea67c2ff72f89ed118be044a82d8a572e405bf0/src/CommunityToolkit.Maui/Behaviors/EventToCommandBehavior.shared.cs#L93C1-L116C3

Does that help? Happy to discuss it further or provide any further assistance if you need?

IeuanWalker: @bijington I've updated the code to have a property changed event, but can't get it to work with a switch to test it. The updated code is on the branch `OnAnimateOnTapPropertyChanged`. The binding should be correct but it's not triggering the property changed event -
image

> Also your comment about a follow up PR to fix the eventName property, isn't neccerrary after this PR. As this PR fixes it by moving all the logic under this new boolean.

IeuanWalker: All works with a button clicked, and changing it in C# - 
image
image

bijington: Good work! Thanks for the update. I don't see the commits, did you push them?

As for your binding not working I suspect that is because the current default binding mode for the `AnimateOnTapProperty` is set to `BindingMode.OneWayToSource` but that is the opposite direction to what you want. I think you want `BindingMode.OneWay`.

IeuanWalker: The default on the property is `OneWay`, i  did try overriding it like this -
image

And changing the default - 
image

Neither worked. I've merged all the changes now though

IeuanWalker: Great thanks

bijington: I have pushed a commit that should fix it. It needed an x:DataType to be added to the binding

IeuanWalker: ahh cool, thanks :)

bijington:
@IeuanWalker thanks again for this. I think we could add some unit test and update the docs and then we should be good to merge. Are you happy to either or both? I'm here to assist with them if not

IeuanWalker:
@bijington unit tests added and here is a link to the doc PR - https://github.com/MicrosoftDocs/CommunityToolkit/pull/538

Not sure why the tests are failing on mac os pipeline, they're working fine locally for me on windows

bijington:
Thanks for the speedy work! I'll try and take a look in a bit.

bijington (approved):
@IeuanWalker thanks for this. It looks good to me!

src/CommunityToolkit.Maui/Behaviors/AnimationBehavior.shared.cs
```diff
@@ -30,11 +30,17 @@ public partial class AnimationBehavior : EventToCommandBehavior
 	public static readonly BindableProperty AnimateCommandProperty =
 		BindableProperty.CreateReadOnly(nameof(AnimateCommand), typeof(Command<CancellationToken>), typeof(AnimationBehavior), default, BindingMode.OneWayToSource, propertyChanging: OnAnimateCommandChanging, defaultValueCreator: CreateAnimateCommand).BindableProperty;
 
+	/// <summary>
+	/// Backing BindableProperty for the <see cref="AnimateOnTap"/> property.
+	/// </summary>
+	public static readonly BindableProperty AnimateOnTapProperty =
+		BindableProperty.Create(nameof(AnimateOnTap), typeof(bool), typeof(AnimationBehavior), propertyChanged: OnAnimateOnTapPropertyChanged);
```
bijington, line 37: Sorry to bring something back up but I just thought, the original behaviour was to add the gesture recogniser right? I wonder if we should default this property to true which means this won't break existing users. What do you think?

IeuanWalker: @bijington it was mixed behavior before.

If a user used an event, it would never add a tap gesture.
If a user used a command, it would always add a tap gesture.

So, making it default true, would change the behavior for people using events. And, defaulting it to false changes the behavior for people using commands

bijington: I see. OK then we can just make it clear that it is opt-in on the docs. We'll have to break someone so this is fine.

Thanks I'll merge it now

bijington (approved):

Issue #2463, opened 2025-01-21
[BUG] OfflineSpeechToText Failure: LanguageNotSupported


MattiaDurli:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [x] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

Using the unmodified sample provided, I try OfflineSpeechToText and receive a Failure in speech engine - LanguageNotSupported

### Expected Behavior

Expect the sample to work and listen to speech

### Steps To Reproduce

Press the button "StartListenAs" and the error appears in Language Output label

### Link to public reproduction project repository

https://github.com/CommunityToolkit/Maui

### Environment

```markdown
Latest Community toolkit sample 11.0.0, tried on Android 13, 14 and 15.
```

### Anything else?

Italian language devices, language pack downloaded in italian and english.
Tried:
		await speechToText.StartListenAsync(new SpeechToTextOptions
		{
			Culture = new System.Globalization.CultureInfo("en-US"),
			ShouldReportPartialResults = true
		}, CancellationToken.None);

to force "en-US" and the default Culture = CultureInfo.CurrentCulture, that picks "it-IT"

MattiaDurli:
Ok, if I use:

await speechToText.StartListenAsync(new SpeechToTextOptions
{
	Culture = new System.Globalization.CultureInfo("it_IT"),
	ShouldReportPartialResults = true
}, CancellationToken.None);

or 

await speechToText.StartListenAsync(new SpeechToTextOptions
{
	Culture = new System.Globalization.CultureInfo("en_US"),
	ShouldReportPartialResults = true
}, CancellationToken.None);

works.
But I had to use en_US and not en-US, same it_IT instead of it-IT.
But on my device Culture = CultureInfo.CurrentCulture returns it-IT

VladislavAntonyuk:
as I see the solution is to replace - with _. It seems like it is only on Android, on iOS culture should be dash separated.

wmassalski-arp:
Hey, i've experienced the same issue. However after trying out "en_US" it prompts me to download English language package update (even tho i installed it beforehand). After pressing download it continues but throws "Failure in speech engine - NoMatch".

alex3696:
[...]

PR #2578, opened 2025-03-11
[PR] Return support Android 33 offline speech recognition

alex3696:
Nothing complicated, just returned the support of Android 33.
Corrected check placement: "Culture ... is not supported". 
Added check, that the language is pending downloading (including delayed through Wi-Fi)
Added check, that avoids endless expectation if the load will be scheduled on WiFi

 ### Description of Change ###

 

 ### Linked Issues ###
 

https://github.com/CommunityToolkit/Maui/issues/2463#event-16631788933
 - Fixes #
 ### PR Checklist ###
 

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/OfflineSpeechToTextImplementation.android.cs
```diff
@@ -81,22 +81,46 @@ async Task InternalStartListening(SpeechToTextOptions options, CancellationToken
...
 		speechRecognizer.CheckRecognitionSupport(recognizerIntent, new Executor(), new RecognitionSupportCallback(recognitionSupportTask));
 		var recognitionSupportResult = await recognitionSupportTask.Task;
+
+		if (!recognitionSupportResult.SupportedOnDeviceLanguages.Contains(options.Culture.Name))
+		{
+			throw new NotSupportedException($"Culture '{options.Culture.Name}' is not supported");
+		}
+		if (recognitionSupportResult.PendingOnDeviceLanguages.Contains(options.Culture.Name))
+		{
+			throw new Exception($"Speech Recognition {options.Culture.Name} pending download. Loading can be delayed if WiFi only");
```
VladislavAntonyuk, line 91: Exception is not correct in this case. Developer needs to be notified when download is completed.
The correct behavior is waiting for the download.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/OfflineSpeechToTextImplementation.android.cs
```diff
@@ -81,22 +81,46 @@ async Task InternalStartListening(SpeechToTextOptions options, CancellationToken
...
+				TryDownloadOfflineRecognizer33(recognizerIntent);
 			}
-			
-			var downloadLanguageTask = new TaskCompletionSource();
-			speechRecognizer.TriggerModelDownload(recognizerIntent, new Executor(), new ModelDownloadListener(downloadLanguageTask));
-			await downloadLanguageTask.Task.WaitAsync(token);
 		}
 		
 		speechRecognizer.SetRecognitionListener(listener);
 		speechRecognizer.StartListening(recognizerIntent);
```
VladislavAntonyuk, line 106: We should not start recognition if language is not installed.

VladislavAntonyuk (changes requested):

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/OfflineSpeechToTextImplementation.android.cs
alex3696: The load can be delayed for an indefinite period or not at all if there is no connection by WiFi. For example, in my case, the download occurred only at night when connecting to home wifi.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/OfflineSpeechToTextImplementation.android.cs
alex3696: Yes, I thought about it, it would be possible to do so:
```
			else if (OperatingSystem.IsAndroidVersionAtLeast(33))
			{
				TryDownloadOfflineRecognizer33(recognizerIntent);
				throw new Exception($"Speech Recognition {options.Culture.Name} pending download. Loading can be delayed if WiFi only");
			}
```
but there is a chance that the package will load before the recognition launch.
I decided not to complicate the code now.

src/CommunityToolkit.Maui.Core/Essentials/SpeechToText/OfflineSpeechToTextImplementation.android.cs
pictos: Is there a way to notify the user and ask to download using mobile data? Like store does when you try to download some app?

alex3696: > Is there a way to notify the user and ask to download using mobile data? Like store does when you try to download some app?
I did not find such an API to switch to mobile data. I also did not find an API, which method (WiFi or mobile) is now configured.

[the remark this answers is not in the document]
alex3696: 1) This code already uses generic Exception `new Exception($"Failure in speech engine - {error}"))`
2 )There is no suitable Exception to notify the scheduled download, I decided not to complicate code.

VladislavAntonyuk (commented):
I have concerns about user behavior with Android 33. Recognition starts before language is downloaded. Getting infinite exception in case language is pending download (as there is no way to notify user when it is finished), more code with checks for different API versions.
All these issues are resolved in Android 34 (just one version up), so I am voting for closing this PR and keeping minimum version = 34.

alex3696:
> I have concerns about user behavior with Android 33. Recognition starts before language is downloaded. Getting infinite exception in case language is pending download (as there is no way to notify user when it is finished), more code with checks for different API versions. All these issues are resolved in Android 34 (just one version up), so I am voting for closing this PR and keeping minimum version = 34.

I don't understand your concerns. 
There are no infinite exceptions: only one check block on start and one exception - recognition will not begin if there is no language or it is in the process of loading.
API Android 34 does not solve anything in this case, only adds the ability to monitor the progress of loading onProgress, [onScheduled ](https://developer.android.com/reference/android/speech/ModelDownloadListener#onScheduled()) . 
Please note that you do not process this onScheduled, this leads to endless expectation if the load will be scheduled on WiFi.
This check avoids such behavior if the load is already scheduled.
```
if (recognitionSupportResult.PendingOnDeviceLanguages.Contains(options.Culture.Name))
{
	throw new Exception($"Speech Recognition {options.Culture.Name} pending download. Loading can be delayed if WiFi only");
}
```
Посмотрите внимательно ещё раз, возможно нас может рассудить кто-то третий.

Another one reason to Android 33 - no matter how interesting it, but I can’t help you, I do not have device with android 34 and above ;-)

alex3696:
We must add check, that`s avoids endless expectation if the load will be scheduled on WiFi.
```
public void OnProgress(int completedPercent)
{
    downloadLanguageTask.SetException(new Exception($"Speech Recognition {options.Culture.Name} pending download. Loading can be delayed if WiFi ");
}
```

VladislavAntonyuk:
How do you plan to use the API?
```csharp
do
{
	try
	{
		await speechRecognizer.StartListening(...);
	}
	catch 
    {
	    
	}
}
while (speechRecognizer.State != State.Listening)
```

Android 34 displays the popup and shows a download progress in the notification.

pictos:
> You can use it like that:
> 
> ```
> try
> {
>     do
>     {
>         await speechRecognizer.StartListening(...);
>     }
>     while (speechRecognizer.State != State.Listening)
> }
> catch 
> {
> 	    
> }
> ```
> 
> I want to make voice control in my application. The prototype of API using is not yet reached. In theory, yes, a continuous listening is needed. Maybe we will create DownloadScheduledException for cases when a language package loading or scheduled? Then it will be possible to catch and react in the right way.
> 
> my Android 33(S20+) keep silent when loading

I don't like this API, is error prone and not intuitive at all. I advocate to rethink it or add some way to extend what we've today, then devs can implement how they want to handle this feature in older versions

alex3696:
> I don't like this API

my PR - not about the API and not about use. This PR about the fact that if it is impossible to immediately start recognition for any reason, then you need to give out exception, +android 33.

alex3696:
@VladislavAntonyuk, Greetings. Good idea, I like it.
Please explain: I need to close PR or change my PR, where the check will be virtual?
virtual Task TryDownloadOfflineRecognizer(...)

VladislavAntonyuk:
you can revert changes in this PR and add virtual to `async Task TryDownloadOfflineRecognizer34Async(Intent recognizerIntent, CancellationToken token)`

pictos:
I don't think it makes sense to make the method virtual, since we are using the interface approaches, if the user wants a custom implementation is just to replace the `Default` service with their own. Mixing virtual and interfaces looks like an anti-pattern to me

VladislavAntonyuk:
> @VladislavAntonyuk 
> 
> I don't think it makes sense to make the method virtual, since we are using the interface approaches, if the user wants a custom implementation is just to replace the `Default` service with their own. Mixing virtual and interfaces looks like an anti-pattern to me.
> 
> 
> 
> With that we can close this PR and guide devs to implement the interface them self
> 
> 

I partially agree, because class is rather big and requires a lot of code to be copied. But I see class is sealed and users can't inherit. Maybe for this case it's really worth just close the PR.
@CommunityToolkit/maui what do you think?

bijington:
It would be great to make it easy to tweak the implementation detail to enable developers to override bits and support this scenario. I can't see an easy change that would enable this right now though.

bijington:
Thanks. Yes I couldn't see how they could reduce this down to a smaller change.

PR #2613, opened 2025-04-02
Pin WindowsSdkPackageVersion

jfversluis:
### Description of Change ###

Also see #2612, but this one should be merged to main.

This change pins the `WindowsSdkPackageVersion` in the `Directory.Build.props` file to `10.0.19041.44`. Additionally, it adds a note about keeping this in sync with .NET MAUI.

When we don't do that, depending on what is installed on the build agent or local machines or how the stars are aligned, we can get different versions of the WindowsAppSDK. That in turn might bring in different versions of transitive packages like WinRT.Runtime.

This was the case I was hitting. I tried to update the Toolkit dependency for the .NET MAUI templates, but it wouldn't build because the Toolkit was (transitively) referencing WinRT.Runtime 2.2.0 (because we were building against a newer WindowsAppSDK) and the templates were referencing WinRT.Runtime 2.1.0, because .NET MAUI pins the WindowsAppSDK to 10.0.19041.44 (at time of writing).

To overcome this happening by surprise lets pin the WindowsSdkPackageVersion in the Toolkit to the same version as .NET MAUI.

I'm doing this on a weirdish branch that is branched of the 11.1.0 tag and will be released as a 11.1.1 version. This is needed because 11.2.0 has a minimum dependency on .NET MAUI 9.0.50. The templates will be inserted in Visual Studio together with version 9.0.40, so I will need a version of the Toolkit with this bug fixed _and_ that is able to be used with 9.0.40, hence this extra release.

This change also needs to be ported to the main branch so that we also do this moving forward.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,6 +22,12 @@
     <SkipValidateMauiImplicitPackageReferences>true</SkipValidateMauiImplicitPackageReferences>
     <MauiEnableXamlCBindingWithSourceCompilation>true</MauiEnableXamlCBindingWithSourceCompilation>
 
+    <!-- WindowsAppSDK
+      We should keep this in sync with .NET MAUI https://github.com/dotnet/maui/blob/8ee00e00f603cd8a420a3467daddaf3935c8c587/Directory.Build.props#L142 
+      This will prevent transitive dependencies from causing conflicts when they're not aligned.
+    -->
+    <WindowsSdkPackageVersion>10.0.19041.44</WindowsSdkPackageVersion>
+
     <!-- https://learn.microsoft.com/dotnet/core/deploying/native-aot/?tabs=net8plus%2Cwindows -->
     <StripSymbols>false</StripSymbols>
     <TrimmerSingleWarn>false</TrimmerSingleWarn>


crwsolutions:
Hi @jfversluis  I saw you talking in the standup about this issue, just some minutes ago. I struggled with this. I am not really sure whether it is related. But when I had these two in my project:

```xml
<WindowsSdkPackageVersion>10.0.19041.41</WindowsSdkPackageVersion>
<!--<WindowsPackageType>None</WindowsPackageType>-->
```

I get this error:
```
InnerException = {"Could not load file or assembly 'WinRT.Runtime, Version=2.2.0.0, Culture=neutral, PublicKeyToken=99ea127f02d97709'. The system cannot find the file specified.":"WinRT.Runtime, Version=2.2.0.0, Culture=neutral, PublicKeyToken=99ea127f02d97709"}
```

And this is solved when I put this in my project:

```xml
<!--<WindowsSdkPackageVersion>10.0.19041.41</WindowsSdkPackageVersion>-->
<WindowsPackageType>MSIX</WindowsPackageType>
```

launchSettings.json : 
 `"commandName": "MsixPackage",`
 
 Hopefully this is somehow helpful 🙂

jfversluis:
@crwsolutions thanks for watching and the suggestion! While in the same area I don't think its related... It might actually even be caused by the Toolkit...? We're building this from the library perspective and you are probably talking about a .NET MAUI app project. Our library doesn't have the concept of MSIX etc. but I'd be curious if we merge this, or maybe if you use the 11.1.1 version I just released if that changes this behavior. If you are willing and able to test that out I'm happy to hear the results!

crwsolutions:
@jfversluis 
>  but I'd be curious if we merge this, or maybe if you use the 11.1.1 version I just released if that changes this behavior

Yes, v11.1.1 also fixes it 👍

So about this line:

```xml
<WindowsSdkPackageVersion>10.0.19041.41</WindowsSdkPackageVersion>
```

I think I added it in the past to resolve some problem, then I upgraded a lot (.net 9 etc), and moved it out to get it working. Can you recommend something? (This is a private project, I just want it to compile and work)

jfversluis:
You probably resolved the same problem only I'm resolving it in the Toolkit now.

I'm not sure what I need to recommend? What is the problem you still need help with? The fix I did for 11.1.1 will also go into 11.2.1 (or whatever version we will do next), so that will probably then fix it for you?

crwsolutions:
I Checked my history: I think I added it because of this issue: https://github.com/CommunityToolkit/dotnet/issues/923 . But I will leave it out then and hope for the best with a future upgrade.

PR #2625, opened 2025-04-10
Fix to SelectAllTextBehavior.android.cs

Transis-Felipe:
Fix for the error
Java.Lang.IllegalArgumentException
end should be < than charSequence length

### DIFF
--- a/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.android.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.android.cs
@@ -12,7 +12,10 @@ public class SelectAllTextBehavior : BasePlatformBehavior<InputView, EditText>
 	{
 		base.OnAttachedTo(bindable, platformView);
 
-		platformView.SetSelectAllOnFocus(true);
+		if (!string.IsNullOrEmpty(platformView.Text))
+		{
+			platformView.SetSelectAllOnFocus(true);
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


pictos:
@Transis-Felipe, thanks for your PR. Is there an open issue for this error? Before we accept PRs we first try to reproduce the error

Transis-Felipe:
@dotnet-policy-service agree

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.android.cs
```diff
@@ -12,7 +12,10 @@ protected override void OnAttachedTo(InputView bindable, EditText platformView)
 	{
 		base.OnAttachedTo(bindable, platformView);
 
-		platformView.SetSelectAllOnFocus(true);
+		if (!string.IsNullOrEmpty(platformView.Text))
```
pictos, line 15: this doesn't look right, if you do this and then the text isn't null or empty the effect will not apply. You need to implement something that will that in account

pictos (changes requested):

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/SelectAllText/SelectAllTextBehavior.android.cs
Transis-Felipe: You are right

Transis-Felipe:
Looks like the problem was EmojiCompatEnabled, so I'm closing this.

Issue #2520, opened 2025-02-14
[BUG] CameraView on Windows crashes app when the camera is in use


MFinkBK:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [x] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

The CommunityToolkit.Maui sample app (as well as my internally developed .NET MAUI app) crashes when the camera is already in use. The `CameraView` control sends a `MediaCapturedFailed` event but crashes anyway, as the exception reported is rethrown, with no possibility to handle it.

### Expected Behavior

The `CameraView` sends a `MediaCapturedFailed` event when the camera is already in use, but doesn't crash the app.

### Steps To Reproduce

1. Start the CommunityToolkit.Maui sample app on Windows and navigate to the "Views > CameraView Page". Observe the live camera preview.
2. Press the "StopCameraPreview" button
3. Start the Camera app that comes with Windows and adjust the selected camera to be the same as used in 1.
4. Press the "StartCameraPreview" button
5. Observe a black preview image
6. Press the "Capture Image" button
7. (the `CameraViewPage.xaml.cs` currently has no `MediaCaptureFailed` event handler, but when it would have one, you could observe the exception, with the `FailureReason` text: "Hardware MFT failed to start streaming due to lack of hardware resources."
8. The app crashes; when the debugger is attached, it breaks in the .NET MAUI Windows app's standard `UnhandledException` handler.

Alternative steps to reproduce would be:
9. Start the Camera app that comes with Windows
10. Start the CommunityToolkit.Maui sample app on Windows and navigate to the "Views > CameraView Page". Observe a gray area where the live camera preview would be.
11. Continue with 4., like above

I also tested using Android, and it's a bit harder to reproduce the error, but it at least doesn't crash the app. Basically if you put the CommunityToolkit.Maui sample app in split mode with the Android devices' camera app, you can let the camera app "steal" the camera preview from the CommunityToolkit.Maui app and still press "Capture Image". The `MediaCaptureFailed` event handler is called again, but the exception isn't rethrown. Inspecting the code where `ICameraView.OnMediaCapturedFailed()` is called suggests that the error only occurs on Windows.

### Link to public reproduction project repository

https://github.com/CommunityToolkit/Maui

### Environment

```markdown
[...]

PR #2655, opened 2025-05-06
Refactor `CameraView.CaptureImage`, `CameraView.StartCameraPreview` and `CameraView.StopCameraPreview`

TheCodeTraveler:
### Description of Change ###

This PR refactors the following public methods in `CameraView`:
- `CameraView.CaptureImage(CancellationToken)`, 
- `CameraView.StartCameraPreview(CancellationToken)` 
- `CameraView.StopCameraPreview()`

Previously, these methods deferred to the Handler to invoke the underlying platform-specfic implemetnation via `Handler.Invoke()`. The problem with this approach is that that `Handler.Invoke` is not `await`'able, yet was invoking asynchronous methods using `async void`. 

Using `async void` lead to issues where developers are unable to catch valid Exceptions thrown by the Operating System, as noted in https://github.com/CommunityToolkit/Maui/issues/2520.

With this PR, developers will be able to wrap these methods, like `await CaptureImage(token)`, in a try/catch block to gracefully handle any exception thrown.

 ### Linked Issues ###
 

 - Fixes https://github.com/CommunityToolkit/Maui/issues/2520

 ### PR Checklist ###
 
In https://github.com/CommunityToolkit/Maui/issues/2520 we discussed promoting the underlying `CameraManager` class to be `public`. However, after further investigation, because each `CameraView` requires a unique instance of `CameraManager`, it is best to keep `CameraManager` internal.

This PR also demotes `CameraView.OnMediaCaptured` and `CameraView.OnMediaCapturedFailed` from `public` to `internal`. These methods fire `event MediaCaptured` and `event MediaCaptureFailed`, respectively, and we should not allow library users to manually fire these events.


VladislavAntonyuk (approved):

TheCodeTraveler:
Thanks Vlad! I'll merge this once I've updated the Docs 🙌

PR #2658, opened 2025-05-07
Add support to make AndroidForegroundService optional to MediaElement

### LINKED ISSUE #2657: [Proposal] Add MediaElement Foreground Service to Media Options
### Feature name

Media Element Foreground Service Options

### Link to discussion

https://github.com/CommunityToolkit/Maui/discussions/2225

### Progress tracker

- [ ] Android Implementation
- [ ] iOS Implementation
- [ ] Windows Implementation
- [ ] Tizen Implementation
- [ ] Unit Tests
- [ ] Samples
- [ ] Documentation

### Summary

Add support to disable Media Element Android service as a Builder option in MediaElement Options class

### Motivation

Allow developers to choose whether they want to have a foreground service and if they want to have notifications. Not all use case scenarios need the service or notifications.

### Detailed Design

API:

`MediaElementOptions`
```csharp
/// <summary>
/// Set Android Foreground Service for MediaElement on construction
/// </summary>
internal static bool AndroidForeServiceEnabled { get; private set; } = true;
```

### Usage Syntax

```markdown
Here is an example of `AndroidForegroundServiceEnabled`

XAML:

 <toolkit:MediaElement
     x:Name="MediaElement"
     ShouldAutoPlay="True"
     AndroidForegroundServiceEnabled="True"
     Source="{x:Static constants:StreamingVideoUrls.BuckBunny}"
     />
```

### Drawbacks

None. If users choose to ignore setting option it will default to current behavior. The current behavior is service is enabled by default.

### Alternatives

If we do not add this it will continue to always be on and leave less choice for developers. I have seen many use cases where the current always on service may not be optimal.

### Unresolved Questions

_No response_

ne0rrmatrix:
- Feature/Proposal

 ### Description of Change ###

Updated `MediaElement` to include `AndroidForegroundServiceEnabled` property for better control over media playback on Android. Adjusted related methods and classes to handle this new property, ensuring the foreground service is utilized appropriately during media playback. Removed unnecessary properties and streamlined the popup media element configuration.
 ### Linked Issues ###

 - Fixes #2225
 - Fixes #2657

 ### PR Checklist ###

This PR is for the community to see how making android service optional could be implemented.

`MediaElement.shared.cs`

```csharp
/// <summary>
/// Read the MediaElementOptions set in on construction, cannot be changed after construction
/// </summary>
public AndroidViewType AndroidViewType { get; init; } = MediaElementOptions.DefaultAndroidViewType;
```

`MediaElementOptions`
```csharp
/// <summary>
/// Set Android Foreground Service for MediaElement on construction
/// </summary>
internal static bool AndroidForeServiceEnabled { get; private set; } = true;

/// <summary>
/// Set Android Foreground Service for MediaElement on construction
/// </summary>
/// <param name="androidForegroundServiceEnabled"></param>
public void SetDefaultAndroidForegroundService(bool androidForegroundServiceEnabled) => AndroidForeServiceEnabled = androidForegroundServiceEnabled;
```

### DIFF
--- a/src/CommunityToolkit.Maui.MediaElement/MediaElementOptions.shared.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/MediaElementOptions.shared.cs
@@ -18,7 +18,18 @@ public class MediaElementOptions()
 	internal static AndroidViewType DefaultAndroidViewType { get; private set; } = AndroidViewType.SurfaceView;
 
 	/// <summary>
+	/// Set Android Foreground Service for MediaElement on construction
+	/// </summary>
+	internal static bool AndroidForeServiceEnabled { get; private set; } = true;
+
+	/// <summary>
 	/// Set Android View type for MediaElement as SurfaceView or TextureView on construction
 	/// </summary>
 	public void SetDefaultAndroidViewType(AndroidViewType androidViewType) => DefaultAndroidViewType = androidViewType;
+
+	/// <summary>
+	/// Set Android Foreground Service for MediaElement on construction
+	/// </summary>
+	/// <param name="androidForegroundServiceEnabled"></param>
+	public void SetDefaultAndroidForegroundService(bool androidForegroundServiceEnabled) => AndroidForeServiceEnabled = androidForegroundServiceEnabled;
 }
--- a/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
@@ -20,6 +20,7 @@ namespace CommunityToolkit.Maui.Core.Views;
 
 public partial class MediaManager : Java.Lang.Object, IPlayerListener
 {
+	bool androidForegroundServiceEnabled;
 	const int bufferState = 2;
 	const int readyState = 3;
 	const int endedState = 4;
@@ -129,11 +130,11 @@ public partial class MediaManager : Java.Lang.Object, IPlayerListener
 	/// <returns>The platform native counterpart of <see cref="MediaElement"/>.</returns>
 	/// <exception cref="NullReferenceException">Thrown when <see cref="Context"/> is <see langword="null"/> or when the platform view could not be created.</exception>
 	[MemberNotNull(nameof(Player), nameof(PlayerView), nameof(session))]
-	public (PlatformMediaElement platformView, PlayerView PlayerView) CreatePlatformView(AndroidViewType androidViewType)
+	public (PlatformMediaElement platformView, PlayerView PlayerView) CreatePlatformView(AndroidViewType androidViewType, bool androidForegroundServiceEnabled)
 	{
 		Player = new ExoPlayerBuilder(MauiContext.Context).Build() ?? throw new InvalidOperationException("Player cannot be null");
 		Player.AddListener(this);
-
+		this.androidForegroundServiceEnabled = androidForegroundServiceEnabled;
 		if (androidViewType is AndroidViewType.SurfaceView)
 		{
 			PlayerView = new PlayerView(MauiContext.Context)
@@ -352,7 +353,7 @@ public partial class MediaManager : Java.Lang.Object, IPlayerListener
 			return;
 		}
 
-		if (connection is null)
+		if (connection is null && androidForegroundServiceEnabled)
 		{
 			StartService();
 		}
--- a/samples/CommunityToolkit.Maui.Sample/Pages/Views/MediaElement/MediaElementPage.xaml
+++ b/samples/CommunityToolkit.Maui.Sample/Pages/Views/MediaElement/MediaElementPage.xaml
@@ -30,6 +30,7 @@
             <toolkit:MediaElement
                 x:Name="MediaElement"
                 ShouldAutoPlay="True"
+                AndroidForegroundServiceEnabled="True"
                 Source="{x:Static constants:StreamingVideoUrls.BuckBunny}"
                 MetadataArtworkUrl="https://lh3.googleusercontent.com/pw/AP1GczNRrebWCJvfdIau1EbsyyYiwAfwHS0JXjbioXvHqEwYIIdCzuLodQCZmA57GADIo5iB3yMMx3t_vsefbfoHwSg0jfUjIXaI83xpiih6d-oT7qD_slR0VgNtfAwJhDBU09kS5V2T5ZML-WWZn8IrjD4J-g=w1792-h1024-s-no-gm"
                 MetadataTitle="Big Buck Bunny"
--- a/samples/CommunityToolkit.Maui.Sample/Pages/Views/MediaElement/MediaElementPage.xaml.cs
+++ b/samples/CommunityToolkit.Maui.Sample/Pages/Views/MediaElement/MediaElementPage.xaml.cs
@@ -250,7 +250,7 @@ public partial class MediaElementPage : BasePage<MediaElementViewModel>
 		{
 			AndroidViewType = AndroidViewType.SurfaceView,
 			Source = MediaSource.FromResource("AppleVideo.mp4"),
-			MetadataArtworkUrl = botImageUrl,
+			AndroidForegroundServiceEnabled = false,
 			HeightRequest = 600,
 			WidthRequest = 600,
 			ShouldAutoPlay = true,
--- a/src/CommunityToolkit.Maui.MediaElement/Handlers/MediaElementHandler.android.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/Handlers/MediaElementHandler.android.cs
@@ -24,7 +24,7 @@ public partial class MediaElementHandler : ViewHandler<MediaElement, MauiMediaEl
 								VirtualView,
 								Dispatcher.GetForCurrentThread() ?? throw new InvalidOperationException($"{nameof(IDispatcher)} cannot be null"));
 
-		var (_, playerView) = MediaManager.CreatePlatformView(VirtualView.AndroidViewType);
+		var (_, playerView) = MediaManager.CreatePlatformView(VirtualView.AndroidViewType, VirtualView.AndroidForegroundServiceEnabled);
 		return new(Context, playerView);
 	}
 
--- a/src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/MediaElement.shared.cs
@@ -226,6 +226,11 @@ public partial class MediaElement : View, IMediaElement, IDisposable
 	public AndroidViewType AndroidViewType { get; init; } = MediaElementOptions.DefaultAndroidViewType;
 
 	/// <summary>
+	/// Gets or sets whether the Android Foreground Service is enabled.
+	/// </summary>
+	public bool AndroidForegroundServiceEnabled { get; init; } = MediaElementOptions.AndroidForeServiceEnabled;
+
+	/// <summary>
 	/// Gets or sets whether the media should start playing as soon as it's loaded.
 	/// Default is <see langword="false"/>. This is a bindable property.
 	/// </summary>


bijington (commented):
I think the general approach looks good. I am surprised it takes such little code change.

I think we should add an `Is` prefix to the property to follow typical Boolean naming conventions (e.g. `IsAndroidForegroundServiceEnabled`)

I understand that you wanted to have Android as the initial prefix here. Why did we do this for the AndroidViewType property in stead of making it a property that only exists for Android? Was it to simplify the code that needs to be written?

src/CommunityToolkit.Maui.MediaElement/MediaElementOptions.shared.cs
```diff
@@ -17,8 +17,19 @@ internal MediaElementOptions(in MauiAppBuilder builder) : this()
 	/// </summary>
 	internal static AndroidViewType DefaultAndroidViewType { get; private set; } = AndroidViewType.SurfaceView;
 
+	/// <summary>
+	/// Set Android Foreground Service for MediaElement on construction
+	/// </summary>
+	internal static bool AndroidForeServiceEnabled { get; private set; } = true;
```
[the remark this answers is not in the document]
ne0rrmatrix: Yes that was a typo. I agree it should not be shortened. Changing it to `IsAndroidForeGroundServiceEnabled` looks better and fits our naming scheme better. I will make the switch

ne0rrmatrix:
> I understand that you wanted to have Android as the initial prefix here. Why did we do this for the AndroidViewType property in stead of making it a property that only exists for Android? Was it to simplify the code that needs to be written?

I honestly have no idea how or when we came up with that. I am simply using the same convention used in texture view as it is has already been established and keeping it the same made sense to me.

mrucker:
Have you checked the Android Manifest file that is created at build time to make sure your code changes prevent the foreground service permission from being requested? That is the real problem, because that is what Google checks when you submit to the app store.

Elinares-82:
> @Elinares-82 there's an easy work around for this in the meantime. You just need to remove like 3 lines of code from the MAUI project, build it, and then include the package you just built in your app instead of the Nuget package. It took me like an hour to get a version in my app without the foreground permission requirement.

Thanks!, I will it a try.

bijington:
> Hi all, we just got rejected out app since the MediaElement control added this permission. We only use this control to display a video in the Splash screen. When can we get this new CommunityToolkit nuget version? In the meantime, we will appeal the rejection, however, this is just to see if they understand the situation. I would appreciate it if we could move a little faster. Thanks

I think a PR opened within a few hours is moving pretty quickly. If you would like things to move more quickly please free to get involved and help out

Elinares-82:
> > Hi all, we just got rejected out app since the MediaElement control added this permission. We only use this control to display a video in the Splash screen. When can we get this new CommunityToolkit nuget version? In the meantime, we will appeal the rejection, however, this is just to see if they understand the situation. I would appreciate it if we could move a little faster. Thanks
> 
> I think a PR opened within a few hours is moving pretty quickly. If you would like things to move more quickly please free to get involved and help out

Sure, give me the permissions and I click on Approve button. The Play Store is rejecting our apps since you decided to add permissions by default, why? Let the developers add or not the permissions.

pictos (commented):
I see that you added the property inside the `MediaElement`, how is the behavior if I have two media elements in different pages, one setting foreground to `true` and other to `false`, what will be the behavior? I didn't see code to tear down the `ForegroundService` if it changes to `false` in fly.

But I think it should only be something that you set on AppBuilder and only once

Elinares-82:
That I could see is prevent to start the service here
image
I'm taking the @ne0rrmatrix branch a create an assembly in the meantime, since we need our app get approved ASAP.

ne0rrmatrix:
> I see that you added the property inside the `MediaElement`, how is the behavior if I have two media elements in different pages, one setting foreground to `true` and other to `false`, what will be the behavior? I didn't see code to tear down the `ForegroundService` if it changes to `false` in fly.
> 
> But I think it should only be something that you set on AppBuilder and only once

The behavior if one media element is using the service and the other is not is that one will have access to the service and the other will not. This PR was meant to allow input and let ppl see how it could be done. It is marked as draft and I explicitly created it so I could gather feedback and make changes. I was hoping for this exact sort of input. 

Do we want to have it as init only and apply to any media elements that launch? Or do we want to allow each media element the option at runtime to decide?

If one media element is already using the service and the second one is not using it it will not tear down the service. I do not believe that is possible without tearing down the media element using the service. The service is tightly integrated with media session and is as close to default implementation as I could create based on platform limitations. The only thing I could do to make it more like the google example is to move the Exoplayer Builder into the service itself. That was the original goal but do to various factors in the original media2 to media3 PR it never happened.

I would prefer to keep this as simple as possible for maintainability while allowing developers the choice to do what they want. My goal here is to add the option to opt out of the service at runtime. Do you want it set in builder only? Or does anyone have any better suggestions? I am not looking at adding anything complex or doing something new. I simply want to add developer choice.

Anyone who has an opinion please speak up. I still think of this as being in design phase. This PR was purely to allow ppl to see the idea I had for design.

Elinares-82:
> Do we want to have it as init only and apply to any media elements that launch? Or do we want to allow each media element the option at runtime to decide?

What if we can have both?, We can have an initial flag to disable it for all the media elements, and force with the new IsAndroidForegroundServiceEnabled to start individually? I do not know if this make sense. 

I the other hand, I know that each team have processes, I don't like to sound rude at all. I'd just like to know when this version might be coming out.

Thanks,

mrucker:
I was reading back through old discussions because I've been confused. I think there are two different issues being discussed. I'm going to call them issue 1 and issue 2 here.

Issue 1 -- MediaElement should only turn on a foreground service if it is requested
Issue 2 -- MediaElement should only add foreground permissions to the Android Manifest if requested

I think this PR will likely fix what I call issue 1, but it won't fix what I call issue 2.

@ne0rrmatrix Is this PR intended to fix both issue 1 and 2? Or are you only trying to fix issue 1?

@Elinares-82 I think you have an issue 2 problem (which is what I had when I joined these threads several months ago). Unless ne0rrmatrix says otherwise, I don't think this PR is going to fix your issue.

ne0rrmatrix:
@mrucker ty for the clarification. I will update the PR to address both issues. I appreciate all the help and input.

mrucker:
(Also, I seem to be the one to blame for the confusion. I hijacked discussion #2225, which was initially about issue 1, and started talking about issue 2 because I didn't understand the difference at the time.)

Elinares-82:
> MediaElementOptions

You are totally right; I think this will not fix my issue. Google rejected our app since in the Manifest are several permissions like FOREGROUND_SERVICE_MEDIA_PLAYBACK and FOREGROUND_SERVICE, but we just display a promo video in the splash screen. We are stuck at this point. The only way that I can see if use an old version or use an alternative like Blazor component.

ne0rrmatrix:
Updating the service to address both issues will require the developer to add permissions manually to manifest and will only allow the service to run if permissions and foreground service are turned on. My questions is how do we want to deal with that. The reason we set the permissions ourselves was to fix the issue with constant bug reports where the solution was to simply add the service and intent filter to the manifest.

mrucker:
> Updating the service to address both issues will require the developer to add permissions manually to manifest and will only allow the service to run if permissions and foreground service are turned on. My questions is how do we want to deal with that. The reason we set the permissions ourselves was to fix the issue with constant bug reports where the solution was to simply add the service and intent filter to the manifest.

Yes... That's kind of the problem we never solved in an old discussion about it... Which is why I've never offered my own PR... There didn't seem to be a great solution.

Elinares-82:
> Updating the service to address both issues will require the developer to add permissions manually to manifest and will only allow the service to run if permissions and foreground service are turned on. My questions is how do we want to deal with that. The reason we set the permissions ourselves was to fix the issue with constant bug reports where the solution was to simply add the service and intent filter to the manifest.

I think, if this is documented all of us have to read and follow the guidance. For me is normal to add the permissions that the controls need. But, is there a possibility to disabled to add the permissions to the Manifest if the new Flag is turn it on? In a complex scenario use a project property to disable this in the source generator (I'm assuming that these permissions are being added using a source generator)

mrucker:
> > MediaElementOptions
> 
> You are totally right; I think this will not fix my issue. Google rejected our app since in the Manifest are several permissions like FOREGROUND_SERVICE_MEDIA_PLAYBACK and FOREGROUND_SERVICE, but we just display a promo video in the splash screen. We are stuck at this point. The only way that I can see is use an old version or use an alternative like Blazor component.

It is easy to remove the permission requirements if you are able to build the MediaElement yourself and create your own Nuget Package. It's not ideal, but if you're really up against it will allow you to release your app. I'd send you my team's custom Nuget package we've been using to remove this permission, but handing out code to strangers on the internet doesn't seem like a best practice 😄.

pictos:
> > MediaElementOptions
> 
> You are totally right; I think this will not fix my issue. Google rejected our app since in the Manifest are several permissions like FOREGROUND_SERVICE_MEDIA_PLAYBACK and FOREGROUND_SERVICE, but we just display a promo video in the splash screen. We are stuck at this point. The only way that I can see is use an old version or use an alternative like Blazor component.

@Elinares-82 there are tools that you allow you to edit the manifest on android app, so when you build your project and it generates the apk or aab you can unzip it, remove the permissions that you don't want and zip it again to ship into stores. I already did that for many apps that I worked and never had issues with that, and since are tools for it, I would say, isn't something unusual

Elinares-82:
> @Elinares-82
> 
> > You are totally right; I think this will not fix my issue. Google rejected our app since in the Manifest are several permissions like FOREGROUND_SERVICE_MEDIA_PLAYBACK and FOREGROUND_SERVICE, but we just display a promo video in the splash screen. We are stuck at this point. The only way that I can see is use an old version or use an alternative like Blazor component.
> 
> It is easy to remove the permission requirements if you are able to build the MediaElement yourself and create your own Nuget Package. It's not ideal, but if you're really up against it will allow you to release your app. I'd send you my team's custom Nuget package we've been using to remove this permission, but handing out code to strangers on the internet doesn't seem like a best practice 😄.

I think these are the lines that I have to comment out. I'm trying to use the generated "custom" DLL  but I'm getting several compiler errors so I will try the @pictos suggestion.  

image

Thanks to all for your help and recommendations.

ne0rrmatrix:
Give me an hour or two and I will update this PR to address issues brought up. I have an idea of how we can do this and have everyone get what they want.

mrucker:
@Elinares-82 Those two and you'd also need to get these two

image

ne0rrmatrix:
> @Elinares-82 Those two and I think you'd also need to get these two (it's been a while since I made the change)
> 
> image

I was unable to remove those two items. It simply will not work without them. If you can test the latest PR in your project and see if not adding anything in manifest and then disabling the service using builder works for you I would appreciate it.

Elinares-82:
> > @Elinares-82 Those two and I think you'd also need to get these two (it's been a while since I made the change)
> > image
> 
> I was unable to remove those two items. It simply will not work without them. If you can test the latest PR in your project and see if not adding anything in manifest and then disabling the service using builder works for you I would appreciate it.

Thanks, I'm trying to compile and use the assembly, however, in my app several compilation errors come up and this is because I just need the Android part. I'm using the conditional references as well like

  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0-ios'">
    <PackageReference Include="CommunityToolkit.Maui.MediaElement" Version="6.0.1" />
  </ItemGroup>

  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0-android'">
    <Reference Include="CommunityToolkit.Maui.MediaElement">
      <HintPath>..\..\..\..\packages\CommunityToolkit.Maui.MediaElement.dll</HintPath>
    </Reference>
  </ItemGroup>

Elinares-82:
> > > MediaElementOptions
> > 
> > 
> > You are totally right; I think this will not fix my issue. Google rejected our app since in the Manifest are several permissions like FOREGROUND_SERVICE_MEDIA_PLAYBACK and FOREGROUND_SERVICE, but we just display a promo video in the splash screen. We are stuck at this point. The only way that I can see is use an old version or use an alternative like Blazor component.
> 
> @Elinares-82 there are tools that you allow you to edit the manifest on android app, so when you build your project and it generates the apk or aab you can unzip it, remove the permissions that you don't want and zip it again to ship into stores. I already did that for many apps that I worked and never had issues with that, and since are tools for it, I would say, isn't something unusual

I tried this, but, the installer is not be able to run.

ne0rrmatrix:
I created a blank Maui app targeting dotnet 9. I then added the package after building and creating a nuget. I did not add anything to android manifest. I did add this to builder:
```csharp
.UseMauiCommunityToolkitMediaElement(options => options.SetDefaultAndroidForegroundService(false));
```
It works as expected and service is disabled. Note that the rich media notifications do not work without the service. This is one of the tradeoffs.

Elinares-82:
> I then added the package after building and creating a nuget.

Thanks a lot @ne0rrmatrix , I was trying to create the nuget like yours, however, I do not know how, then I was researching about it.

mrucker:
I will test out this PR as well tomorrow. Thank you for your work @ne0rrmatrix!

src/CommunityToolkit.Maui.MediaElement/MediaElementOptions.shared.cs
```diff
@@ -17,8 +17,19 @@ internal MediaElementOptions(in MauiAppBuilder builder) : this()
 	/// </summary>
 	internal static AndroidViewType DefaultAndroidViewType { get; private set; } = AndroidViewType.SurfaceView;
 
+	/// <summary>
+	/// Set Android Foreground Service for MediaElement on construction
+	/// </summary>
+	internal static bool AndroidForegroundServiceEnabled { get; private set; } = true;
```
bijington, line 23: I think we should make this property opt-in rather than opt-out. That way developers only need to add the permissions if they want to use the foreground service, they don't want to have to add them by default.

bijington: Plus this should also include the `Is` prefix to the name

ne0rrmatrix: As for the Is prefix I will update to fix that.

If we choose to make the service opt in that would create a situation where on Mac, iOS, and windows behavior would be different than android. It would create confusion and core behavior to be differentiated per platform. If we do this on android we should have an opt in for all devices? As for other device types there has been no requests for this on anything but android.

The service is tightly tied to rich media notifications. Untying and removing the service might be technically possible but it would be against best practices for for media 3 API. We would have to decide if we want to either have platform specific behavior that is different on each platform as a default or add this to every device type when it is only really relevant for a single platform.

If anyone sees this from a different angle or can suggest an alternative idea or way of thinking about his I would appreciate input.

Elinares-82:
> I created a blank Maui app targeting dotnet 9. I then added the package after building and creating a nuget. I did not add anything to android manifest. I did add this to builder:
> 
> ```cs
> .UseMauiCommunityToolkitMediaElement(options => options.SetDefaultAndroidForegroundService(false));
> ```
> 
> It works as expected and service is disabled. Note that the rich media notifications do not work without the service. This is one of the tradeoffs.

@ne0rrmatrix  I just tested the PR, I was able to create the nuget and load it to my DevOps Artifacts. The only additional thing that I had to do was downgrade the Microsoft.Maui.Controls to 9.0.40, since the 9.0.50 and 9.0.60 has and issue with the Keyboard and the Scroll, when the Keyboard appear the scroll of the view does not work as expected and the user is not able to scroll the view to see other controls. We are currently loading the aab file to the Google Store, I let you know if the app is approved. Thanks so much.

Elinares-82:
> > I created a blank Maui app targeting dotnet 9. I then added the package after building and creating a nuget. I did not add anything to android manifest. I did add this to builder:
> > ```cs
> > .UseMauiCommunityToolkitMediaElement(options => options.SetDefaultAndroidForegroundService(false));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It works as expected and service is disabled. Note that the rich media notifications do not work without the service. This is one of the tradeoffs.
> 
> @ne0rrmatrix I just tested the PR, I was able to create the nuget and load it to my DevOps Artifacts. The only additional thing that I had to do was downgrade the Microsoft.Maui.Controls to 9.0.40, since the 9.0.50 and 9.0.60 has and issue with the Keyboard and the Scroll, when the Keyboard appear the scroll of the view does not work as expected and the user is not able to scroll the view to see other controls. We are currently loading the aab file to the Google Store, I let you know if the app is approved. Thanks so much.

Hi, our app was rejected. The permission android.permission.FOREGROUND_SERVICE is added to our AndroidManifest, we are reviewing why since, we actually did not add it. I think is because of the Service definition (image below), the other permission was not added, I mean the ForegroundServiceMediaPlayback.

image

pictos:
@Elinares-82 this issue with scroll view and keyboard is reported on Maui repo? If don't, do you mind to it?

Elinares-82:
> > > I created a blank Maui app targeting dotnet 9. I then added the package after building and creating a nuget. I did not add anything to android manifest. I did add this to builder:
> > > ```cs
> > > .UseMauiCommunityToolkitMediaElement(options => options.SetDefaultAndroidForegroundService(false));
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > It works as expected and service is disabled. Note that the rich media notifications do not work without the service. This is one of the tradeoffs.
> > 
> > 
> > @ne0rrmatrix I just tested the PR, I was able to create the nuget and load it to my DevOps Artifacts. The only additional thing that I had to do was downgrade the Microsoft.Maui.Controls to 9.0.40, since the 9.0.50 and 9.0.60 has and issue with the Keyboard and the Scroll, when the Keyboard appear the scroll of the view does not work as expected and the user is not able to scroll the view to see other controls. We are currently loading the aab file to the Google Store, I let you know if the app is approved. Thanks so much.
> 
> Hi, our app was rejected. The permission android.permission.FOREGROUND_SERVICE is added to our AndroidManifest, we are reviewing why since, we actually did not add it. I think is because of the Service definition (image below), the other permission was not added, I mean the ForegroundServiceMediaPlayback.
> 
> image

Apparently is a rejected error was for a previous version of the app. We do not know why Google is rejecting the app for previous versions that have those permissions. We send an update that override the old versions (This was made in the Play Console)

Elinares-82:
> @Elinares-82 this issue with scroll view and keyboard is reported on Maui repo? If don't, do you mind to it?

@pictos sure, I'll check is this issue was already reported, in case not, I'll do it!

Elinares-82:
> Finally, our apps were approved by Google. Thanks. Now I'm reviewing the issue regarding the Scroll and Microsoft.Maui.Controls 9.0.50 and 60. and log it into the repo.

The issue is already logged 

https://github.com/dotnet/maui/issues/19014

2023 almost 2 years ago.

src/CommunityToolkit.Maui.MediaElement/MediaElementOptions.shared.cs
ne0rrmatrix: Do we want core behavior to be different on a per platform basis? Or am I over thinking this?

ne0rrmatrix: No. Not at all. If we do this it would be a breaking change. I am not against it. It would make my life so much easier and solve many issues. I just want to make sure you were aware that this would mean that on windows, ios, and mac it will show notifications. But on Android you would need to opt in which is completely different behavior than other platforms. Disabling the service means no notifications at all from media element on android.

ne0rrmatrix: Unfortunately we are also using an bound android service in conjunction with a c# service. It is very complicated and took a while to figure out. I can add that service to the builder. I in fact missed it and had forgotten about it. It is required for it to work. I will modify code so that is not run when we are not using it. Good catch!

As for what you were saying I see no way to do this that does not have the user set the permissions. I do believe there is a way to do this without users/devs having to do anything. It has been mentioned specifically as to how we can do that before in the discussion about this. But to do that would require a complete rewrite of android implementation. My solution is simple and deals with all existing complaints about developers either not needing service or it actually causing it to rejected by play store because they don't want or need notifications at all.

My solution using builder means that if developer adds no permissions and sets service disabled it is not include at all in app and will pass certification by play store. One dev has already done this and is actively using it. Just scroll up and you will see that. I don't recommend anyone use a PR in production but it looks like they did do that.

Currently in main we set all services and permissions ourselves. If you can figure out how we might use a builder to set `[assemby]` and/or `[intent]`, `[service]` then it would be trivial for us to do everything and no developer action would be required. That would be best case. Unfortunately I don't know how to do that. Do you have any ideas?

pictos: > Currently in main we set all services and permissions ourselves. If you can figure out how we might use a builder to set [assemby] and/or [intent], [service] then it would be trivial for us to do everything and no developer action would be required. That would be best case. Unfortunately I don't know how to do that. Do you have any ideas?

Here I can see that you're trying to protect devs from themself, and that (in 99% of the cases) are very unproductive, the best way is to let devs add permissions themself, you may argue that causes them to open a lot of issues because them don't read the docs, but at the end this isn't our fault. Devs should read docs before using something.

But if still you want to protect devs from themself, you can have a SG that will read the changes on the MauiAppBuilder code and add the attributes when it sees the dev option to use the Foreground service. That will make it trimmable and will add the lines on manifest for them

ne0rrmatrix: @pictos this PR adds back the need for devs to set permissions themselves. At the moment in main we do everything ourselves. This was a mistake I think. I was happy with it for a while because it drastically cut down bug reports. But it limits our options in the long run. I agree with what you are saying.

In this PR we created a builder option to add dev choice as to whether they want the android service.  This was all discussed previously in our monthly meeting and in multiple discussions. It was just brought up again. I am going to add it to our monthly meeting schedule as I would like a final resolution as to what way we want to go. 

I have multiple reviews where the suggestions are essentially polar opposite ideas. I am hoping to gain some clarity when we all get together at the beginning of the month.

The idea I have and want to go with is have devs set the permissions and point them at the docs when we need to. It adds a choice in builder that defaults to ON as that unifies cross platform behavior and minimizes code changes. The main thing we are doing here is allowing developer choice as I have found out there are many use cases where rich media notifications are not wanted or needed. This was a much requested feature. This is all fully implemented and just needs one more fix before It is done. 

I am hoping we can decide this soon. Since I joined the team this is so far the most upvoted and requested item I have come across relating to media element.

Elinares-82: > SG

Totally agree. We as a developers have to read the documentation. Add an exclamation or important note in the docs saying, "Permissions Required!"

pictos: @ne0rrmatrix I'm sorry if I wasn't clear, english isn't my primary language.

I wasn't talking about the PR as a hole, but of what I understand from the quoted paragraph. Any path that you choose I'll be happy with, just wanted to add my 2 cents

ne0rrmatrix:
I am thinking we should switch to using just a builder option with no ability to toggle the service off and on. ATM switching back and forth does work but the images are not loading after turning off service and then turning it back on. Can see this bug live by loading popup in media element page. In this PR I have the service disabled when you load the popup to test this out. I will look at why the image is not loading. This may be caused by having two media elements loaded where one has it turned off and the other has it on. I will investigate the behavior.

Does anyone need or want the ability to switch at runtime? Or is a builder option ok? This PR currently has both. We could end up with issues when using multiple media elements where one is turned on and one is off? How do we want to handle that? It would be much simpler to have just the builder and no runtime switch. This would simplify implementation and make it easier to maintain. My bad for overcomplicating things.

bijington:
I say just go with the builder option and keep things simple

JoacimWall:
Hi 
is this fixed now and then in what version? and are they instruction on how to disable foreground service?
Regards Joacim

bijington:
@JoacimWall this hasn't merged yet so is not yet available.

@ne0rrmatrix where are we with this? I think we have a solution but it is breaking? Do you think there is anything we could mark as `Obsolete` in an upcoming release that would indicate to developers they needed to opt-in to the foreground service in future releases? I am suspecting it wouldn't be easy given the developer didn't have to call any foreground specific code?

ne0rrmatrix:
The main changes are having to add service registration and permissions to Android manifest. We had removed that and added it directly. With this PR developers will be required to do so again. Also they need to opt out of service. i don't see how to use obsolete as intended with how this works.

VladislavAntonyuk (dismissed):

Elinares-82:
Hi all, Does this improvement have a ETA? I built my own version of this component as @mrucker suggested, however, the idea is to have the official nuget. 

Regards,

dainius-r:
@ne0rrmatrix @bijington was is blocking this PR for merging? We would like not to declare of foreground service permissions on Android if not used.

bijington:
I've been thinking about this and how we handle the breaking change for users - I know this breaks the existing pattern but what if we move the Boolean property from the options class and make it a parameter on `UseMauiCommunityToolkitMediaElement` this would make it clear to developers at compile time that something has changed. Then it's an easy fix as they will get a clear error from the compiler, otherwise changing the defaults that they only discover at runtime feels wrong. Thoughts?

ne0rrmatrix:
> I've been thinking about this and how we handle the breaking change for users - I know this breaks the existing pattern but what if we move the Boolean property from the options class and make it a parameter on `UseMauiCommunityToolkitMediaElement` this would make it clear to developers at compile time that something has changed. Then it's an easy fix as they will get a clear error from the compiler, otherwise changing the defaults that they only discover at runtime feels wrong. Thoughts?

I am happy to do that. Can you give me an example of what that looks like. I can't actually visualize or imagine what it is exactly that you mean.
```csharp
 .UseMauiCommunityToolkitMediaElement(static options =>
								{
									options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
									options.SetDefaultAndroidForegroundService(true);
								}) 
``` 
This is new behavior? Can you provide example of how you would suggest we change it?

bijington:
Yes of course - sorry I realise my statement was a little abstract. I am thinking this:

```csharp

.UseMauiCommunityToolkitMediaElement(
    static options =>
    {
        options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
    }, 
    enableForegroundService: true);

```

ne0rrmatrix:
> Yes of course - sorry I realise my statement was a little abstract. I am thinking this:
> 
> ```cs
> .UseMauiCommunityToolkitMediaElement(
>     static options =>
>     {
>         options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
>     }, 
>     enableForegroundService: true);
> ```

That looks like a great idea. I will look into how to implement that.

bijington:
> > Yes of course - sorry I realise my statement was a little abstract. I am thinking this:
> 
> > 
> 
> > ```cs
> 
> > .UseMauiCommunityToolkitMediaElement(
> 
> >     static options =>
> 
> >     {
> 
> >         options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
> 
> >     }, 
> 
> >     enableForegroundService: true);
> 
> > ```
> 
> 
> 
> That looks like a great idea. I will look into how to implement that.

@ne0rrmatrix did you have a look at this approach?

ne0rrmatrix:
@bijington I got sidetracked and forgot about this PR. I am looking at it now.

bijington:
> @bijington I got sidetracked and forgot about this PR. I am looking at it now.

No rush! I just thought I'd check as I saw the update from GitHub just now

bijington:
@ne0rrmatrix sorry I am struggling to follow what the result of this PR will mean for developers. I see foreground now defaults to true but what does it mean for permissions? They will now need to add permissions in themselves which is fine. If a developer upgrades from a version to this one will the permissions fail early enough for a developer to catch it?

ne0rrmatrix:
> @ne0rrmatrix sorry I am struggling to follow what the result of this PR will mean for developers. I see foreground now defaults to true but what does it mean for permissions? They will now need to add permissions in themselves which is fine. If a developer upgrades from a version to this one will the permissions fail early enough for a developer to catch it?

The purpose of this PR is to allow users to opt out of of using Rich Media Notifications and not use a an android service. It would require everyone who currently uses MediaElement to start adding all the permissions and everything we were handling ourselves to make life easier. This PR is much requested as loads of developers have use cases where they neither want nor need a long lasting service or do they want to use notifications. I see many downsides to adding this. The only upside is it will clear out a few bug reports about being unable to disable the service. The big downside will be a lot of bug reports that will repeat many times over the next year or two. They will repeat and never go away. That is the down side.

ne0rrmatrix:
> @ne0rrmatrix sorry I am struggling to follow what the result of this PR will mean for developers. I see foreground now defaults to true but what does it mean for permissions? They will now need to add permissions in themselves which is fine. If a developer upgrades from a version to this one will the permissions fail early enough for a developer to catch it?

They will need to add permissions that are not currently added by us. It is complicated. It will fail unexpectedly otherwise. They need to add all the permissions we currently add. If they do not it will result in unexpected behavior depending on state of app. I have done nothing to mitigate that at this time.

ne0rrmatrix:
The issue with permissions is if the app is currently installed and it is upgraded it may not fail if permissions are not set correctly. That is a huge issue. There would be no error and no unexpected behavior. The app does not request permissions from user. It does not need to for rich media notifications. As for developer if they fail to include them it fails silently and the notifications are just not shown at all.

bijington:
Thanks. I completely agree with moving towards an opt-in approach that this PR brings. The bit that I want to cover is how seamless the upgrade path is to help developers and ultimately avoid introducing bug reports. 

We can add details to the release notes but that won't stop it entirely. I was just hoping there is something we could do at compile time to help here

ne0rrmatrix:
For developers that currently expect it to just work they might have issues. Everyone who uses default options will need to add all the permissions. Those that want to opt out will not have to add anything. The biggest issue we have faced with mediaElement is people not reading and/or following directions for permissions. After automating that the number of bugs went down to close to zero.

Can we add an analyzer that checks for permissions in manifest and refuse to compile if they are missing? Maybe? Not sure if htat would work. But it is an idea I could look into? Maybe have analyzer check boolean for options and if it sees it in default or true then check if any permissions are missing?

bijington:
That might be a nice future addition, for now maybe we just make it clear in the extension method when a developer registers to use it in the foreground. I still think I prefer the idea of forcing developers to specify whether they want it foreground or background when initialising the toolkit as it covers th scenario of existing developers upgrading - they will receive a compiler error and be forced to understand why

JKNAUG:
Will this be merged soon, by any chance?

bijington:
@ne0rrmatrix where are we at with this PR? Is the following a suitable summary?

The foreground service is still enabled by default and can be disabled by developers (opt-out).

There is a breaking change that requires developers to add permission to their manifest file?

ne0rrmatrix:
@bijington Yes that would perfectly describe the current state of this PR.

bijington:
@ne0rrmatrix thanks! Is this what we want? I'm asking with an emphasis on upgrade path for existing devs, I would like to make sure this is as smooth as possible.

Currently as I understand it we have the following scenarios:

1. An existing developer upgrades and needs to provide the permissions but this isn't obvious at time of upgrade or even at compilation time.

2. A new developer not wanting foreground service support, installs the package and needs to disable the feature.

3. A new developer wanting foreground service support, installs the package and has to add permission.

I don't think the above are ideal and are likely to introduce bug reports because there is friction in all cases. I personally think we should move to an opt-in model rather than opt-out. This would mean that point 2 is already taken care of. For points 1 and 3 I propose we remove the method from the options class and instead make it a required parameter on the `UseMauiCommunityToolkitMedia` method. This would mean that when a developer upgrades they will receive a compilation error and we can make it clear with XML docs what the requirement is.

I guess I should state that my proposed solution is neither opt-in or out, the developer must make the choice when registering the library.

What do you think?

ne0rrmatrix:
@bijington it is a breaking change, yes. It will cause all users that use notifications a huge headache. I 100 percent agree that is a problem. We moved to the current model where we manage permissions specifically to alleviate this issue. The way permissions are currently added is at compile time. It is something we either add or not. There have been many users who do not want or need these permissions and actively do not want media notifications at all. 

In order to allow the developer to choose what permissions they use we need to first allow the service to be turned off. If we do not the app will crash. If you have any alternative to this I am happy to look at it. I have looked into it and have found no way to let us maintain control of said permissions and make them optional. If you know a way to add this:
```
[assembly: UsesPermission(Android.Manifest.Permission.ForegroundServiceMediaPlayback)]
[assembly: UsesPermission(Android.Manifest.Permission.ForegroundService)]
[assembly: UsesPermission(Android.Manifest.Permission.MediaContentControl)]
[assembly: UsesPermission(Android.Manifest.Permission.PostNotifications)]
```
The above code adds those permissions into the Android manifest directly. If we can make this optional I can figure out a solution to all of this that would not be such a huge issue. Also when the option is set it cannot be changed at runtime. As it is hard-coded into the manifest. The above code just edits the manifest and essentially inserts that code directly into it.

ne0rrmatrix:
@bijington I am happy to make it opt it. That would fix a lot of issues. I will update it to do that!

LGE-AUGMENSYS:
Hi, do you plan to merge this soon? If so, will it apply to the dotnet 9.0 version as well?
Our app got rejected because of the media playback foreground service since last month. We made an appeal to Google but only got the same reasons for rejection repeated to us again.

I understand the concern on your side about this change. But IMHO, it is a lot more inconvenient for developers to deal with this store rejection issue (by potentially building and maintaining their own MediaElement library) than for developers to add the permission when they actually need it.
I hope you could consider this in your internal decision making process. Thanks.

ne0rrmatrix:
@LGE-AUGMENSYS We only support the current version of dotnet Maui which is on dotnet 10.x. We do not support any older versions of dotnet or Maui. So no support for dotnet 9.x. I would suggest updating your app to use dotnet 10.x or continue to use an older version of MediaElement that does not use the service.

bijington:
@LGE-AUGMENSYS I think this is pretty close to being ready. If we can make the changes to simplify the toolkit usage as per https://github.com/CommunityToolkit/Maui/pull/2658#issuecomment-3769374211 then I think this could get merged. I think @ne0rrmatrix is looking at it so I'm hopeful we could get it into the next release. As James said it will only support .NET 10 though

ne0rrmatrix:
@bijington I updated it as you suggested. Tests are updated too. I still need to update the docs after you look at this and are happy with the changes.

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
```diff
@@ -21,11 +21,16 @@ public static class AppBuilderExtensions
 	/// </summary>
 	/// <param name="builder"><see cref="MauiAppBuilder"/> generated by <see cref="MauiApp"/>.</param>
 	/// <param name="options"><see cref="MediaElementOptions"/>.</param>
+	/// <param name="enableForegroundService">Enable Android Foreground Service for MediaElement.</param>
 	/// <returns><see cref="MauiAppBuilder"/> initialized for <see cref="MediaElement"/>.</returns>
-	public static MauiAppBuilder UseMauiCommunityToolkitMediaElement(this MauiAppBuilder builder, Action<MediaElementOptions>? options = null)
+	public static MauiAppBuilder UseMauiCommunityToolkitMediaElement(this MauiAppBuilder builder, Action<MediaElementOptions>? options = null, bool enableForegroundService = false)
```
bijington: This is looking good and gets us much closer to a smoother upgrade path for developers. One further thought I have is - should we remove the default value for the parameter and force a developer to provide it? This will mean that when they upgrade their code will no longer compile but I think this is a good thing, a developer just needs to read the compiler error and see 'oh I need to provide the enableForegroundService parameter'. It should be clear which they are expecting. Then we can pile all the details about setting the permissions inside the XML docs.

What do you think?

ne0rrmatrix: Good idea. I have no idea how to do that. I will try to figure it out.

bijington: Let me know how you get on as there are some rules around it which I can help explain if you need

bijington:
Hey James, the AppBuilderExtensions looks good to me. One thing I am unsure of is what value does the new analyser bring? The compiler will already force a developer to supply the `enableForegroundService` parameter. If we keep the analyser I think it should check whether said parameter is true and also verify that permissions have been set. I am also ok with dropping the analyser if what I just wrote is too difficult to create and potentially maintain. What do you think?

LGE-AUGMENSYS:
Thanks for all the updates.

Since we can't just update to dotnet 10 during our current app release phase, I tried using older versions of the MediaElement library. Here's some hints if you face the same situation.

- Version 6.0.0 is the last one where this FOREGROUND_SERVICE_MEDIA_PLAYBACK permission is NOT automatically added. However, it still seems to spawn a service, or at least creates the music play UI in the notification area. And I suspect Playstore could still reject a non-player app exhibiting such behaviors;
- Version 5.0.0 has neither the permission added, nor the service/notification created;
- The only breaking change I noticed was the moving of the namespace for MediaElementState. In case you refer to that in your code, you have to resolve the using statement(s).

ne0rrmatrix:
@bijington I removed the analyzer and associated tests. I spent many hours investigating the possibility of scanning the android manifest xml file and was unable to resolve the issues I found with testing to see if permissions and service were missing. I'm sure it is possible to test for that and we should do that. But I think maybe we should do that as a separate PR? Anyways I hope we can move forward now with what we have. We can always discuss if we want to add the analyzer back and with what options later.

src/CommunityToolkit.Maui.MediaElement/MediaElementOptions.shared.cs
```diff
@@ -26,8 +26,19 @@ internal MediaElementOptions(in MauiAppBuilder builder) : this()
 	/// </summary>
 	internal static AndroidViewType DefaultAndroidViewType { get; private set; } = AndroidViewType.SurfaceView;
 
+	/// <summary>
+	/// Set whether Android Foreground Service is enabled for MediaElement
+	/// </summary>
+	internal static bool IsAndroidForegroundServiceEnabled { get; private set; }
```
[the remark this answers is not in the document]
ne0rrmatrix: We have decided to default to false and make this a breaking change

bijington:
> @bijington I removed the analyzer and associated tests. I spent many hours investigating the possibility of scanning the android manifest xml file and was unable to resolve the issues I found with testing to see if permissions and service were missing. I'm sure it is possible to test for that and we should do that. But I think maybe we should do that as a separate PR? Anyways I hope we can move forward now with what we have. We can always discuss if we want to add the analyzer back and with what options later.

Yes that makes sense to me. The analyser approach might not be worth the maintenance effort anyway so I think this approach is good. I'll take a quick review now

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
```diff
@@ -20,12 +20,17 @@ public static class AppBuilderExtensions
 	/// Initializes the .NET MAUI Community Toolkit MediaElement Library
 	/// </summary>
 	/// <param name="builder"><see cref="MauiAppBuilder"/> generated by <see cref="MauiApp"/>.</param>
-	/// <param name="options"></param>
+	/// <param name="options"><see cref="MediaElementOptions"/>.</param>
+	/// <param name="enableForegroundService">Enable Android Foreground Service for MediaElement.</param>
```
bijington: I think we should try to include more detail in this documentation because this is where the developers will first encounter the option either through an upgrade or by first adding it.

What do you think to making it something like:

```csharp
/// <param name="enableForegroundService">Enable Android Foreground Service for MediaElement.
/// When providing <c>true</c> the required permissions must be included in your application, see the example for more details. 
/// <example>
/// When providing <c>true</c> the required permissions must be included in your application the following permissions must be added to your application:
/// <code>
/// [assembly: UsesPermission(Android.Manifest.Permission.MediaContentControl)]
/// [assembly: UsesPermission(Android.Manifest.Permission.PostNotifications)]
/// </code>
/// </example>
/// </param>
```

I am not sure that my initial suggestion is the best and you probably know more detail about what exactly is involved as well as permissions. It would mean that an IDE should show something like:

Image

What do you think?

ne0rrmatrix: Good Idea. I will Update it. We do have documentation but adding additional details here does make some sense. It will help cut down on the number of bug reports I hope.

bijington: Yeah that is my thinking. The more signs we hold up for people to read should increase they actually read them

ne0rrmatrix: I updated the comment to include a direct link to relevant part of documentation on Microsoft learn for configuring the manifest for service module.

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
```diff
@@ -20,12 +20,20 @@ public static class AppBuilderExtensions
 	/// Initializes the .NET MAUI Community Toolkit MediaElement Library
 	/// </summary>
 	/// <param name="builder"><see cref="MauiAppBuilder"/> generated by <see cref="MauiApp"/>.</param>
-	/// <param name="options"></param>
+	/// <param name="options"><see cref="MediaElementOptions"/>.</param>
+	/// <param name="enableForegroundService"> Enable Android Foreground Service for MediaElement.
+	/// When providing <c>true</c> the required permissions and service must be included in android manifest.
+	/// For more information, see the <see href=" https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/views/mediaelement?tabs=android">MediaElement Documentation.</see> for details.
```
bijington: We don't want to include a locale code in the url so the documentation will load in the developers preferred language

```suggestion
	/// For more information, see the <see href="https://learn.microsoft.com/dotnet/communitytoolkit/maui/views/mediaelement?tabs=android">MediaElement Documentation.</see> for details.
```

bijington (changes requested):
If we fix the url in the xml comment then I think this PR is good

src/CommunityToolkit.Maui.UnitTests/Views/MediaElement/MediaElementOptionsTests.cs
```diff
@@ -0,0 +1,85 @@
...
+namespace CommunityToolkit.Maui.UnitTests.Views;
+
+public class MediaElementOptionsTests : BaseViewTest
+{
+	[Fact]
+	public void MediaElementOptions_HasExpectedDefaults()
+	{
+		// Assert defaults on the options class
+		MediaElementOptions.DefaultAndroidViewType.Should().Be(AndroidViewType.SurfaceView);
+		MediaElementOptions.IsAndroidForegroundServiceEnabled.Should().BeTrue();
```
bijington: I think this should be verifying that the default is false?

bijington (changes requested):
I also think some of the copilot reviews are worth checking. Things like method name consistency we should action

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
ne0rrmatrix: Updated the URL as suggested

src/CommunityToolkit.Maui.UnitTests/Views/MediaElement/MediaElementOptionsTests.cs
ne0rrmatrix: Should there be a default? If we are forcing the developer to make that choice why would we set a default? If we set a default would that not prevent an error from being thrown? I believe it will show whatever we set in builder as default atm which is true in samples app. I should probably remove that test as we are not actually setting a default? 

What do you think?

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
```diff
@@ -62,9 +63,8 @@ public void OnPlaybackParametersChanged(PlaybackParameters? playbackParameters)
 
 	public void UpdateNotifications()
 	{
-		if (connection?.Binder?.Service is null)
+		if (connection?.Binder?.Service is null || !isAndroidServiceEnabled)
 		{
```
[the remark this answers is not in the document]
ne0rrmatrix: This will trigger many times during playback and could appear as spam. I have seen complaints about this sort of trace writeline before. It has been posted as a bug many times for various reasons.

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
```diff
@@ -20,12 +20,20 @@ public static class AppBuilderExtensions
...
-	public static MauiAppBuilder UseMauiCommunityToolkitMediaElement(this MauiAppBuilder builder, Action<MediaElementOptions>? options = null)
+	public static MauiAppBuilder UseMauiCommunityToolkitMediaElement(this MauiAppBuilder builder, bool enableForegroundService, Action<MediaElementOptions>? options = null)
 	{
+		// Create MediaElementOptions and set the foreground service setting
+		var mediaElementOptions = new MediaElementOptions(builder);
+		mediaElementOptions.SetAndroidForegroundServiceEnabled(enableForegroundService);
+		
 		// Update the default MediaElementOptions for MediaElement if Action is not null
-		options?.Invoke(new MediaElementOptions(builder));
+		options?.Invoke(mediaElementOptions);
```
TheCodeTraveler: Please revert these changes. They are unnecessary.

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
```diff
@@ -20,12 +20,20 @@ public static class AppBuilderExtensions
 	/// Initializes the .NET MAUI Community Toolkit MediaElement Library
 	/// </summary>
 	/// <param name="builder"><see cref="MauiAppBuilder"/> generated by <see cref="MauiApp"/>.</param>
-	/// <param name="options"></param>
+	/// <param name="options"><see cref="MediaElementOptions"/>.</param>
+	/// <param name="enableForegroundService"> Enable Android Foreground Service for MediaElement.
+	/// When providing <c>true</c> the required permissions and service must be included in android manifest.
+	/// For more information, see the <see href="https://learn.microsoft.com/dotnet/communitytoolkit/maui/views/mediaelement?tabs=android">MediaElement Documentation.</see> for details.
+	/// </param>
```
TheCodeTraveler: Please revert these changes. They are unnecessary. 


```suggestion
```

src/CommunityToolkit.Maui.UnitTests/Extensions/AppBuilderExtensionsTests.cs
```diff
@@ -148,7 +148,7 @@ void HandleShouldUseStatusBarBehaviorOnAndroidModalPageOptionCompleted(object? s
 	public void UseMauiCommunityToolkitMediaElement_ShouldUseSurfaceViewByDefault()
 	{
 		var builder = MauiApp.CreateBuilder();
-		builder.UseMauiCommunityToolkitMediaElement();
+		builder.UseMauiCommunityToolkitMediaElement(enableForegroundService: true);
```
TheCodeTraveler: Please update this to use the proper syntax:


```suggestion
		builder.UseMauiCommunityToolkitMediaElement(static options =>
		{
			options.SetAndroidForegroundServiceEnabled(true);
		});
```

src/CommunityToolkit.Maui.UnitTests/Extensions/AppBuilderExtensionsTests.cs
```diff
@@ -159,12 +159,28 @@ public void UseMauiCommunityToolkitMediaElement_ShouldSetDefaultAndroidViewType(
 		MediaElementOptions.DefaultAndroidViewType.Should().Be(AndroidViewType.SurfaceView);
 
 		var builder = MauiApp.CreateBuilder();
-		builder.UseMauiCommunityToolkitMediaElement(static options =>
+		builder.UseMauiCommunityToolkitMediaElement(enableForegroundService: true, static options =>
```
TheCodeTraveler: Please update this to use the proper syntax:
```suggestion
		builder.UseMauiCommunityToolkitMediaElement(static options =>
		{
			options.SetAndroidForegroundServiceEnabled(true);
		});
```

src/CommunityToolkit.Maui.UnitTests/Extensions/AppBuilderExtensionsTests.cs
```diff
@@ -159,12 +159,28 @@ public void UseMauiCommunityToolkitMediaElement_ShouldSetDefaultAndroidViewType(
...
 		});
 
 		MediaElementOptions.DefaultAndroidViewType.Should().Be(AndroidViewType.TextureView);
 	}
+
+	[Fact]
+	public void UseMauiCommunityToolkitMediaElement_ShouldSetAndroidServiceByDefault()
+	{
+		var builder = MauiApp.CreateBuilder();
+		builder.UseMauiCommunityToolkitMediaElement(enableForegroundService: true);
```
TheCodeTraveler: Please update this to use the proper syntax
```suggestion
		builder.UseMauiCommunityToolkitMediaElement(static options =>
		{
			options.SetAndroidForegroundServiceEnabled(true);
		});
```

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
```diff
@@ -68,10 +68,12 @@ public static MauiApp CreateMauiApp()
...
-								.UseMauiCommunityToolkitMediaElement(static options =>
-								{
-									options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
-								})
+								.UseMauiCommunityToolkitMediaElement(
+									enableForegroundService: true,
+									static options =>
+									{
+										options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
+									})
```
TheCodeTraveler: Please update theist use the proper syntax


```suggestion
								.UseMauiCommunityToolkitMediaElement(static options =>
								{
									options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
									options.SetAndroidForegroundServiceEnabled(true);
								})
```

src/CommunityToolkit.Maui.Analyzers.UnitTests/UseCommunityToolkitMediaElementInitializationAnalyzerTests.cs
```diff
@@ -34,7 +34,7 @@ public static MauiApp CreateMauiApp()
 					{
 						var builder = MauiApp.CreateBuilder();
 						builder.UseMauiApp<Microsoft.Maui.Controls.Application>()
-							.UseMauiCommunityToolkitMediaElement()
+							.UseMauiCommunityToolkitMediaElement(enableForegroundService: true)
```
TheCodeTraveler: It is not necessary here to set the value of the Android Foreground Service. It doesn't add value to the test `VerifyNoErrorsWhenUseMauiCommunityToolkitMediaElement`

 
```suggestion
							.UseMauiCommunityToolkitMediaElement()
```

It would add value to create four new Unit Tests in this file that include the different permutations of the `MediaElementOptions`:

```cs
		builder.UseMauiCommunityToolkitMediaElement(static options =>
		{
			options.SetAndroidForegroundServiceEnabled(true);
			options.SetDefaultAndroidViewType(AndroidViewType.SurfaceView);
		});
```

```cs
		builder.UseMauiCommunityToolkitMediaElement(static options =>
		{
			options.SetAndroidForegroundServiceEnabled(false);
			options.SetDefaultAndroidViewType(AndroidViewType.SurfaceView);
		});
```

```cs
		builder.UseMauiCommunityToolkitMediaElement(static options =>
		{
			options.SetAndroidForegroundServiceEnabled(true);
			options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
		});
```

```cs
		builder.UseMauiCommunityToolkitMediaElement(static options =>
		{
			options.SetAndroidForegroundServiceEnabled(false);
			options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
		});
```

bijington:
> Hey James! Please update this PR to properly use `MediaElementOptions`:
> 1. Remove the `bool enableForegroundService` parameter from `UseMauiCommunityToolkitMediaElement()`
> 2. Update `MediaElementOptions.SetAndroidForegroundServiceEnabled` to `public`
> 
> Developers should be setting `IsAndroidForegroundServiceEnabled`  the same way they set `DefaultAndroidViewType`:
> 
> ```cs
> .UseMauiCommunityToolkitMediaElement(static options =>
> {
> 	options.SetDefaultAndroidViewType(AndroidViewType.TextureView);
> 	options.SetAndroidForegroundServiceEnabled(true);
> })
> ```

Please take a look through the discussion I had with James. The issue with doing it the suggested way is the way it hides the breaking changes from developers. We will receive unnecessary bug reports. This approach while it doesn't fit all models will force developers to consider the breaking change the compiler will highlight to them

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.android.cs
```diff
@@ -27,6 +27,7 @@ public partial class MediaManager : Java.Lang.Object, IPlayerListener
 
 	static readonly HttpClient client = new();
 	readonly SemaphoreSlim seekToSemaphoreSlim = new(1, 1);
+	bool isAndroidServiceEnabled = false;
```
TheCodeTraveler: Please update the name of this variable to make it more specific:

```suggestion
	bool isAndroidForegroundServiceEnabled = false;
```

src/CommunityToolkit.Maui.MediaElement.SourceGenerators/Generators/AndroidMediaElementForegroundServiceConfigurationGenerator.cs
```diff
@@ -0,0 +1,271 @@
...
+			{
+				GeneratePermissions(spc);
+			}
+		});
+	}
+
+	static bool IsMediaElementOptionsClass(SyntaxNode node)
+	{
+		return node is ClassDeclarationSyntax classDecl &&
+			   classDecl.Identifier.Text.Contains(mediaElementOptionsClassName);
```
bijington, line 69: Why does this use `Contains`? Shouldn't we be explicitly checking for the CommunityToolkit class?

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
```diff
@@ -20,12 +20,17 @@ public static class AppBuilderExtensions
...
+	/// <param name="options"><see cref="MediaElementOptions"/>.</param>
 	/// <returns><see cref="MauiAppBuilder"/> initialized for <see cref="MediaElement"/>.</returns>
-	public static MauiAppBuilder UseMauiCommunityToolkitMediaElement(this MauiAppBuilder builder, Action<MediaElementOptions>? options = null)
+	public static MauiAppBuilder UseMauiCommunityToolkitMediaElement(this MauiAppBuilder builder, bool isAndroidForegroundServiceEnabled, Action<MediaElementOptions>? options = null)
 	{
 		// Update the default MediaElementOptions for MediaElement if Action is not null
-		options?.Invoke(new MediaElementOptions(builder));
+		var mediaElementOptions = new MediaElementOptions(builder);
+		options?.Invoke(mediaElementOptions);
+		mediaElementOptions.UpdateIsAndroidForegroundServiceEnabledWithUseMauiCommunityToolkitMediaElementParameter(isAndroidForegroundServiceEnabled);
```
bijington, line 33: Why call this additional internal method and not just call `SetIsAndroidForegroundServiceEnabled`?

bijington: In fact why don't we mark `SetIsAndroidForegroundServiceEnabled` as internal for now and avoid needing the additional method?

src/CommunityToolkit.Maui.slnx
```diff
@@ -6,23 +6,28 @@
     <Project Path="CommunityToolkit.Maui.Camera.Analyzers/CommunityToolkit.Maui.Camera.Analyzers.csproj" />
     <Project Path="CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes/CommunityToolkit.Maui.MediaElement.Analyzers.CodeFixes.csproj" />
     <Project Path="CommunityToolkit.Maui.MediaElement.Analyzers/CommunityToolkit.Maui.MediaElement.Analyzers.csproj" />
+    <Project Path="CommunityToolkit.Maui.MediaElement.SourceGenerators/CommunityToolkit.Maui.MediaElement.SourceGenerators.csproj" />
     <Project Path="CommunityToolkit.Maui.SourceGenerators/CommunityToolkit.Maui.SourceGenerators.csproj" />
   </Folder>
   <Folder Name="/Benchmarks/">
     <Project Path="CommunityToolkit.Maui.Analyzers.Benchmarks/CommunityToolkit.Maui.Analyzers.Benchmarks.csproj" />
-    <Project Path="CommunityToolkit.Maui.SourceGenerators.Benchmarks/CommunityToolkit.Maui.SourceGenerators.Benchmarks.csproj" />
+    <Project Path="CommunityToolkit.Maui.SourceGenerators.Benchmarks/CommunityToolkit.Maui.SourceGenerators.Benchmarks.csproj" Id="cc195bb1-9cac-410e-81fe-23665f99f294" />
```
bijington: This guid looks a little out of place in the new tidy slnx file. Is it really needed?

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
TheCodeTraveler: Yea, it's a bit of a duplicate, for sure.

My thought here was a bit of future proofing. In the future, devs will be forced to use `SetIsAndroidForegroundServiceEnabled(bool)`, so why not let them use it now. And when we remove the `bool isAndroidForegroundServiceEnabled` parameter when we release .NET 11, then their code will still work.

I've already configured the Source Generator to cover the four current possible cases:
-  `SetIsAndroidForegroundServiceEnabled(true)` and `isAndroidForegroundServiceEnabled = false`
-  `SetIsAndroidForegroundServiceEnabled(false)` and `isAndroidForegroundServiceEnabled = false`
- `SetIsAndroidForegroundServiceEnabled(true)` and `isAndroidForegroundServiceEnabled = true`
- `SetIsAndroidForegroundServiceEnabled(false)` and `isAndroidForegroundServiceEnabled = true`

This should make it easier for devs when we deprecate the `isAndroidForegroundServiceEnabled` parameter. They'll be able to more-easily transition when we remove it in the future.

What do you think?

src/CommunityToolkit.Maui.slnx
TheCodeTraveler: I was wondering the same thing! My guess is that it was auto-added by Visual Studio.

If you're cool with removing it, I am too. I verify that everything still compiles when I do.

src/CommunityToolkit.Maui.MediaElement/AppBuilderExtensions.shared.cs
bijington: Ok that's fair enough

src/CommunityToolkit.Maui.slnx
bijington:
```suggestion
    <Project Path="CommunityToolkit.Maui.SourceGenerators.Benchmarks/CommunityToolkit.Maui.SourceGenerators.Benchmarks.csproj" />
```

bijington (dismissed):

PR #1581, opened 2023-11-29
Initial musings on showing popups without having to subclass Popup

bijington:
### Description of Change ###

This is primarily aimed at playing around but also with the aim of exposing the implementation to see what everything thinks.

 

 ### Linked Issues ###
 

 - Fixes #

 ### PR Checklist ###
 
 - [ ] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [ ] Has tests (if omitted, state reason in description)
 - [ ] Has samples (if omitted, state reason in description)
 - [ ] Rebased on top of `main` at time of PR
 - [ ] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#contributing-code---best-practices)
 - [ ] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pulls 

 ### Additional information ###

PR #2661, opened 2025-05-09
Add Support for Shell + `IQueryAttributable` for PopupV2

TheCodeTraveler:
### Description of Change ###

This PR augments https://github.com/CommunityToolkit/Maui/pull/1581, adding support for `Shell` and `IQueryAttributable`. I.e., this PR will be merged into the open [PopupV2 PR](https://github.com/CommunityToolkit/Maui/pull/1581), not the `main` branch.

Specifically, this PR was created in response to @bijington's review noting that users will no longer be able to pass data from ViewModel to ViewModel after removing the `onPresenting` parameter: https://github.com/CommunityToolkit/Maui/pull/1581#discussion_r2079810384

This PR is currently in Draft because it still requires Unit Tests.

I updated `SimplePopup` and `CsharpBindingPopup` to demonstrate using `IQueryAttributable` in the Sample App.

### DIFF
--- a/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
+++ b/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
@@ -8,40 +8,31 @@ using Microsoft.Maui.Controls.Shapes;
 
 namespace CommunityToolkit.Maui.Views;
 
-sealed partial class PopupPage<T>(Popup<T> popup, IPopupOptions popupOptions, IDictionary<string, object>? shellParameters)
-	: PopupPage(popup, popupOptions, shellParameters)
+sealed partial class PopupPage<T>(Popup<T> popup, IPopupOptions popupOptions)
+	: PopupPage(popup, popupOptions)
 {
-	public PopupPage(View view, IPopupOptions popupOptions, IDictionary<string,object>? shellParameters)
-		: this(view as Popup<T> ?? CreatePopupFromView<Popup<T>>(view), popupOptions, shellParameters)
+	public PopupPage(View view, IPopupOptions popupOptions)
+		: this(view as Popup<T> ?? CreatePopupFromView<Popup<T>>(view), popupOptions)
 	{
-		if (view is IQueryAttributable queryAttributableView && shellParameters is not null)
-		{
-			queryAttributableView.ApplyQueryAttributes(shellParameters);
-		}
 	}
 
 	public Task Close(PopupResult<T> result, CancellationToken token = default) => base.Close(result, token);
 }
 
-partial class PopupPage : ContentPage
+partial class PopupPage : ContentPage, IQueryAttributable
 {
 	readonly Popup popup;
 	readonly IPopupOptions popupOptions;
 	readonly Command tapOutsideOfPopupCommand;
 	readonly WeakEventManager popupClosedEventManager = new();
 
-	public PopupPage(View view, IPopupOptions popupOptions, IDictionary<string,object>? shellParameters)
-		: this(view as Popup ?? CreatePopupFromView<Popup>(view), popupOptions, shellParameters)
+	public PopupPage(View view, IPopupOptions popupOptions)
+		: this(view as Popup ?? CreatePopupFromView<Popup>(view), popupOptions)
 	{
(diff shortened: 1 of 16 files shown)


bijington:
@TheCodeTraveler I love this! I hope you don't mind I pushed some XML docs improvements and will follow up with more on the other public areas if that's ok?

The only question I will ask (and I am happy to be shot down) is whether we could/should include something similar for non-Shell based navigation? Basically including something like the following into the `INavigation` based extension methods:

```csharp
if (view is IQueryAttributable queryAttributable)
{
    queryAttributable.ApplyQueryAttributes(shellParameters);
}
else if (view.BindingContext is IQueryAttributable queryAttributableBindingContext)
{
    queryAttributableBindingContext.ApplyQueryAttributes(shellParameters);
}
```

I understand that `IQueryAttributable` is related to `Shell` and is promoted by MS but if there are people out there using `INavigation` they could also benefit. I am also happy to defer this as a possible improvement if people ask. What do you think?

TheCodeTraveler:
Ok, @bijington - I've added Unit Tests, squishing some bugs along the way, and this is now ready for review!!

src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
```diff
@@ -25,13 +30,18 @@ partial class PopupPage : ContentPage
...
 	readonly WeakEventManager popupClosedEventManager = new();
 
-	public PopupPage(View view, IPopupOptions popupOptions)
-		: this(view as Popup ?? CreatePopupFromView<Popup>(view), popupOptions)
+	public PopupPage(View view, IPopupOptions popupOptions, IDictionary<string,object>? shellParameters)
+		: this(view as Popup ?? CreatePopupFromView<Popup>(view), popupOptions, shellParameters)
 	{
 		ArgumentNullException.ThrowIfNull(view);
+
+		if (view is IQueryAttributable queryAttributableView && shellParameters is not null)
```
bijington, line 38: Why do we need to apply this is we already pass the parameters into `Shell.Current.GoToAsync()`?

bijington (approved):
Great work @TheCodeTraveler! I just have one question which more to help me understand the change rather than question whether we need it.

src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
TheCodeTraveler: Good catch! Long story short, I dug into the MAUI source code and found that the Shell navigation engine only fires `IQueryAttributable.ApplyQueryAttributes()` when `IQueryAttributable` is inherited from the `ContentPage` or the `ContentPage.BindingContext` (View Model). 

This is great for us because the Shell Navigation engine will automatically fire `IQueryAttributable.ApplyQueryAttributes()` on the Popup View Model because we propogate its `BindingContext` to `PopupPage.BindingContext`.

However, because the Popup/View given to us from the user is not used as the underlying `ContentPage` (we use our internal `PopupPage` as the underlying `ContentPage`), we have to manually check for `IQueryAttributable` and call `IQueryAttributable.ApplyQueryAttributes()`.

TheCodeTraveler: All that said - this conversation inspired me to just come up with a new mechanism for manually calling `IQueryAttributable.ApplyQueryAttributes()` **from** IQueryAttributable.ApplyQueryAttributes()! 

Instead of doing it in the constructor here, and requiring us to pass in `IDictionary<string,object>` into the Constructor of `PopupPage`, we can instead let the Shell Navigation engine provide the `IDictionary<string,object>` to us by having `PopupPage` inherit from `IQueryAttributable` (`PopupPage :IQueryAttributable`).

Now, when `PopupPage.ApplyQueryAttributes(IDictionary<string, object>)` fires we can check if the `Popup` or `Popup.Content` is inheriting from `IQueryAttributable` and pass in the `IDictionary<string, object>` provided to us from the Shell Navigation engine.

bijington: Thanks for explaining. Yes I like the idea of implementing the interface and doing it your proposed way

PR #2692, opened 2025-06-09
[Housekeeping] GitHub Actions: Set `fail-fast: true`

TheCodeTraveler:
This PR updates the CI/CD pipeline, setting `fail-fast: true`.

The goal of this PR is to prevent our CI/CD pipelines from hanging once completed.


TheCodeTraveler:
Confirmed the bug still exists

PR #2705, opened 2025-06-12
Speech Recognition Windows Ensure unsubscribe from events on Stop Recording

VladislavAntonyuk:
Speech Recognition Windows Ensure unsubscribe from events on Stop Recording


pictos (approved):

VladislavAntonyuk:
@TheCodeTraveler, it is already in the docs:
image

TheCodeTraveler:
Nice! Could you also add a section that recommends using `IDisposable` or `IAsyncDisposable`?

VladislavAntonyuk:
I don't think it's recommend. For example users may register their service as Singleton.

PR #2714, opened 2025-06-13
Don't break us please oh mighty MCT

mattleibow:
This pull request introduces updates to the `ValidationBehavior` class and its associated `ValidationFlags` enum in the `CommunityToolkit.Maui` project. The changes include adding new obsolete flags for validation behavior and updating the file's imports.

### Updates to `ValidationFlags` enum:

* Added two obsolete flags, `ValidateOnFocusing` and `ValidateOnUnfocusing`, which map to existing flags `ValidateOnFocused` and `ValidateOnUnfocused`. These changes aim to provide backward compatibility while encouraging the use of newer flag names.

### Codebase maintenance:

* Imported the `System` namespace in `ValidationBehavior.shared.cs`, possibly to support the `[Obsolete]` attribute or other added functionality.

### DIFF
--- a/src/CommunityToolkit.Maui/Behaviors/Validators/ValidationBehavior.shared.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/Validators/ValidationBehavior.shared.cs
@@ -1,3 +1,4 @@
+using System;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
@@ -21,6 +22,13 @@ public enum ValidationFlags
 	ValidateOnValueChanged = 8,
 	/// <summary>Force make valid when focused</summary>
 	ForceMakeValidWhenFocused = 16
+
+	/// <summary> Validate on focusing</summary>
+ 	[Obsolete("Use ValidateOnFocused instead.")]
+	ValidateOnFocusing = ValidateOnFocused,
+	/// <summary>Validate on unfocusing</summary>
+ 	[Obsolete("Use ValidateOnUnfocused instead.")]
+	ValidateOnUnfocusing = ValidateOnUnfocused,
 }
 
 /// <summary>


TheCodeTraveler:
Hi Matt! Could you help me understand why we'd add two new flags and immediately mark them obsolete?

mattleibow:
Because you broke the world @TheCodeTraveler! You broke my world.

lol

The MCT renamed the enum values, and this is a breaking change that affects everyone updating from 9.x I think. 

@jfversluis did we have an issue for the people that were shattered by this betrayal?

jfversluis:
https://github.com/dotnet/maui/issues/29983

https://github.com/dotnet/maui/pull/28510/files#diff-85b0572dd0ecb27cbcbcdc5148f81829d8d2b16e52cd85551530b225cc4a60ecR65-R70

TheCodeTraveler:
Yes, I understand it was a breaking change.

We highlighted this change in our documentation and in our release notes. 

Is there a blocker for you or other developers to modify the values of this enum when upgrading to a new major version? Are there developers unable to modify these enum values? Did this change somehow affect the underlying .NET MAUI framework?

Respectfully, we don't get paid to work on this project and this PR adds technical debt to our library. Unless a developer is blocked, I do not advocate we add `Obsolete` code to our code base.

mattleibow:
I think we can work aaround it to use different values in our templates based on the version of MCT/.NET version we are creating for. I think it was like .NET 9 was using MCT 9 and .NET 10 is using MCT 12, which resulted in the code not working out the box.

But Gerald has a fix so this is not so much a blocker but more a sad face emoji type of thing.

But yeah, no need to re-mortgage the house on this.

VladislavAntonyuk:
For the specific version in template it works without issues. If users want to update to the next major version of MCT, they can see breaking changes in release notes and respectively apply changes to their code.

VladislavAntonyuk:
Luckily you don't have Popup in template 😃

Issue #2703, opened 2025-06-12
[BUG] Popup V2 - Android - ShowPopupAsync(INavigation) returns a Task that never completes


mfeingol:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [x] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

When I call ShowPopupAsync and then call CloseAsync() from inside my popup, the popup closes but ShowPopupAsync's task never completes so the result is never available.

### Expected Behavior

I expect the task to complete when I close the popup.

### Steps To Reproduce

1. Derive a popup from Popup<bool?>
2. Call OpenPopupAsync() on it, and close the popup from within using CloseAsync()
3. Note that await OpenPopupAsync() never returns

### Link to public reproduction project repository

https://github.com/mfeingol/repros/tree/master/CommunityToolkit_2703_repro

### Environment

```markdown
- .NET MAUI CommunityToolkit: 12.0.0
- OS: Windows 11
- .NET MAUI: 9.0.71
```

Cristian-Azeta:
Example in Android ...

**TestPage.xaml**

```
<?xml version="1.0" encoding="utf-8" ?>
<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:tk="http://schemas.microsoft.com/dotnet/2022/maui/toolkit"
             x:Class="TEST.TestPage"
             Title="TEST">

	<Label Text="TEST" />

</ContentPage>

```

**TestPage.xaml.cs**

```
namespace TEST;

using CommunityToolkit.Maui;
using CommunityToolkit.Maui.Core;
using CommunityToolkit.Maui.Extensions;
using CommunityToolkit.Mvvm.Messaging;

public partial class TestPage : ContentPage
{
    #region Constructor

    public TestPage()
    {
        InitializeComponent();
    }

    #endregion

    #region Event

    protected override Boolean OnBackButtonPressed()
    {
        return true;
    }

    protected override async void OnAppearing()
    {
        base.OnAppearing();
        IsBusy = true;
        try
        {
		IPopupResult<Boolean> result = await this.ShowPopupAsync<Boolean>(new TestPopup(), new PopupOptions
		{
[...]

PR #2716, opened 2025-06-13
[Popup] Properly handle Android Back Button pressed

TheCodeTraveler:
### Description of Change ###

This PR updates `protected override bool PopupPage.OnBackButtonPressed()` to ensure we properly close the Popup using `CloseAsync()` when the Android Back Button is tapped by the user.

This ensures that the `Task` returned by `ShowPopupAsync()` completes when the Android Back Button is tapped.

 ### Linked Issues ###
 

This fixes the Issue highlighted by @crwsolutions in their comment here: https://github.com/CommunityToolkit/Maui/issues/2703#issuecomment-2970573356. 

**Note:** This does not close resolve the issue #2703. We currently unable to reproduce that Issue and are awaiting a sample reproduction to help us solve that Issue.

 ### PR Checklist ###
 
**SafeFireAndForget**
Because `protected override bool OnBackButtonPressed()` does not return a `Task`, we are unable to use the `await` keyword. Since `PopupPage.Close` returns a `Task`, we use `.SafeFireAndForget()` to ensure we follow best practices and await its `Task` under the hood.

**PopupResult**
For the `PopupResult`, I am setting `WasDismissedByTappingOutsideOfPopup` to `true` for two reasons:
1. By tapping the Android Back button, the user did tap outside of the popup to close it
1. `WasDismissedByTappingOutsideOfPopup = false` requires `PopupResult<T>.Result` to be a value provided by the developer when programmatically closing the popup

### DIFF
--- a/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
+++ b/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
@@ -43,11 +43,6 @@ partial class PopupPage : ContentPage, IQueryAttributable
 		// Only set the content if the parent constructor hasn't set the content already; don't override content if it already exists
 		base.Content ??= new PopupPageLayout(popup, popupOptions);
 
-		if (Shell.Current is Shell shell)
-		{
-			Shell.SetPresentationMode(shell, PresentationMode.ModalNotAnimated);
-		}
-
 		tapOutsideOfPopupCommand = new Command(async () =>
 		{
 			popupOptions.OnTappingOutsideOfPopup?.Invoke();
@@ -114,9 +109,10 @@ partial class PopupPage : ContentPage, IQueryAttributable
 	{
 		if (popupOptions.CanBeDismissedByTappingOutsideOfPopup)
 		{
+			popupClosedEventManager.HandleEvent(this, new PopupResult(true), nameof(PopupClosed));
 			return base.OnBackButtonPressed();
 		}
-
+		
 		return true;
 	}
 
--- a/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
+++ b/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
@@ -2,6 +2,7 @@ using System.ComponentModel;
 using System.Globalization;
 using CommunityToolkit.Maui.Converters;
 using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.Extensions;
 using Microsoft.Maui.Controls.PlatformConfiguration;
 using Microsoft.Maui.Controls.PlatformConfiguration.iOSSpecific;
 using Microsoft.Maui.Controls.Shapes;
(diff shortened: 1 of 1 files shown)


src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
```diff
@@ -114,9 +110,9 @@ protected override bool OnBackButtonPressed()
 	{
 		if (popupOptions.CanBeDismissedByTappingOutsideOfPopup)
 		{
-			return base.OnBackButtonPressed();
+			CloseAsync(new PopupResult(true), CancellationToken.None).SafeFireAndForget();
```
bijington, line 113: Why do we provide a value for the result? Do we want this to mimic tapping outside where there is no Result?

Also I can't comment on line 108 but I think that comment conflicts with this behaviour?

TheCodeTraveler: >  Do we want this to mimic tapping outside where there is no Result?

Yes. And we do.

> Why do we provide a value for the result?

I hopefully answered your question in the description of this PR:
> PopupResult
> For the PopupResult, I am setting WasDismissedByTappingOutsideOfPopup to true for two reasons:
>
> - By tapping the Android Back button, the user did tap outside of the popup to close it
> - `WasDismissedByTappingOutsideOfPopup = false` requires `PopupResult<T>.Result` to be a value provided by the developer when programmatically closing the popup

Instantiating `new PopupResult(true)` just sets its `WasDismissedByTappingOutsideOfPopup` property to `true`; it does not provide a value to `PopupResult<T>.Result`. In other words, `new PopupResult(true)` sets `PopupResult<T>.Result` to `null` because the user tapped outside of the popup.

TheCodeTraveler: > Also I can't comment on line 108 but I think that comment conflicts with this behaviour?

I'm not sure what you mean here. The comment in Line 108 is accurate:
```cs
// Prevent the Android Back Button from dismissing the Popup if CanBeDismissedByTappingOutsideOfPopup is true
```

The method signature for `OnBackButtonPressed()` is a bit unintuitive given it returns a `bool`. In reality, the `bool` return result just indicates whether or not we manually handled the navigation programmatically or not.

It'd make more sense if this method was called `bool DidManuallyHandleNavigationWhenAndroidBackButtonPressed()`.

Here's a screenshot from its intellisence:

bijington: Thanks man! Sorry I jumped straight into the code rather than reading the description.

As for the comment: shouldn't we be preventing the Android Back Button from dismissing the popup if the property is **false**?

bijington (dismissed):

src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
TheCodeTraveler: > shouldn't we be preventing the Android Back Button from dismissing the popup if the property is false

Yup 100%! That's why we only call `CloseAsync()` when `popupOptions.CanBeDismissedByTappingOutsideOfPopup == true`

I'll update the comment to add more clarity.

TheCodeTraveler: Done ✅

Let me know the comments make more sense now. If so, I'll turn on auto-merge and once you approve it again we can merge!

PR #2740, opened 2025-06-17
Make it possible to pass parameters over to INavigation based popups

bijington:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #2696

 ### PR Checklist ###
 
### DIFF
--- a/src/CommunityToolkit.Maui.UnitTests/Extensions/PopupExtensionsTests.cs
+++ b/src/CommunityToolkit.Maui.UnitTests/Extensions/PopupExtensionsTests.cs
@@ -215,7 +215,7 @@ public class PopupExtensionsTests : BaseHandlerTest
 		var shellNavigation = Shell.Current.Navigation;
 
 		// Act
-		shell.ShowPopup(viewWithQueryable, shellParameters: shellParameters);
+		shell.ShowPopup(viewWithQueryable, parameters: shellParameters);
 
 		// Assert
 		Assert.Single(shellNavigation.ModalStack);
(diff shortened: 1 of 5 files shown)


TheCodeTraveler:
Thanks Shaun! 

I advocate against implementing this because there are already ways for developers to pass in values to their popup via Dependency Injection, DataBinding, Messaging Services, etc. 

Yes, I know this feature existed in Popup v1, but to be honest it is a bit of anti-pattern implementing a messaging service into Popup and we should have never supported it. I am not aware of any developer who is blocked without this feature. 

If a developer can provide a sample app demonstrating how they are blocked without this feature, I'm happy to readdress this feature.

bijington:
This scenario only introduces the ability to pass data in at the point of creation/navigating so I don't think DataBinding or Messaging Services are suitable alternatives.

As for developers being blocked there are some over here: https://github.com/CommunityToolkit/Maui/discussions/2696#discussioncomment-13500610

VladislavAntonyuk:
@bijington could you please explain how they will use the parameters? PopupPage is internal and I don't know how it replaces onPresenting. 

I agree with @TheCodeTraveler. At least they can subscribe to Opened event and set parameters they need.

bijington:
> @bijington could you please explain how they will use the parameters? PopupPage is internal and I don't know how it replaces onPresenting. 
> 
> I agree with @TheCodeTraveler. At least they can subscribe to Opened event and set parameters they need. 

Happy to... this doesn't introduce any new concepts it just makes the use of Popups in non-Shell based apps symmetrical with Shell based apps. The parameters would work as per our documentation https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/views/popup-service#passing-data-to-a-popup-view-model only for the methods that accept `INavigation` and not only `Shell`.

My intention was never to replace `onPresenting`, I would have left that in. `IQueryAttributable` was a compromise but the initial implementation only supported Shell based apps

VladislavAntonyuk:
Yes, but you cast popupPage to IQueryAttributable. Maybe you should do it with ViewModel or View?

bijington:
> Yes, but you cast popupPage to IQueryAttributable. Maybe you should do it with ViewModel or View?

Oh I see your point. `PopupPage` is a wrapper for the content so it delegates down to the content view/view model. I've removed the explicit implementation of `IQueryAttributable` so we don't need to cast

Issue #2747, opened 2025-06-19
[BUG] Padding and Margin are not applied to the Popup class of Popup2 by Style attribute


usausa:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [x] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

Margin and Padding are not applied to Popup in the Style attribute.
Property, the value is applied.

### Expected Behavior

Margin and Padding are applied in the Style attribute as well as in the property.

### Steps To Reproduce

1. Create a Style to be applied to the Popup and set Padding and Margin.
2. Apply Style to Popup

## by Style

```xml
<?xml version="1.0" encoding="utf-8" ?>
<toolkit:Popup
    x:Class="Popup2App.NoBorderPropertyPopup"
    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
    xmlns:toolkit="http://schemas.microsoft.com/dotnet/2022/maui/toolkit"
    Style="{StaticResource NoBorderPopup}"
    WidthRequest="200">

    <toolkit:Popup.Resources>
        <ResourceDictionary>
            <Style x:Key="NoBorderPopup" TargetType="toolkit:Popup">
                <Setter Property="Padding" Value="0" />
                <Setter Property="Margin" Value="0" />
                <Setter Property="BackgroundColor" Value="Blue" />
            </Style>
        </ResourceDictionary>
    </toolkit:Popup.Resources>

    <VerticalStackLayout>
        <Button Clicked="OnCloseClick" Text="Close" />
    </VerticalStackLayout>

</toolkit:Popup>
```

![Image](https://github.com/user-attachments/assets/df530484-388a-4145-aaa8-61cd0973634a)

* by Property

```xml
<?xml version="1.0" encoding="utf-8" ?>
<toolkit:Popup
    x:Class="Popup2App.NoBorderStylePopup"
    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
    xmlns:toolkit="http://schemas.microsoft.com/dotnet/2022/maui/toolkit"
    Margin="0"
    Padding="0"
    BackgroundColor="Blue"
[...]

PR #2759, opened 2025-06-22
[Popup] Add `DefaultPopupOptionsSettings` and `DefaultPopupSettings` to `.UseMauiCommunityToolkit(Options)`

TheCodeTraveler:
### Description of Change ###

This PR grants developers the capability to set global default values for every popup in their app using `DefaultPopupOptionsSettings` and `DefaultPopupSettings`  in`.UseMauiCommunityToolkit(Options)`:

```cs
.UseMauiCommunityToolkit(static options =>
{
  options.SetPopupDefaults(new DefaultPopupSettings
  {
	  CanBeDismissedByTappingOutsideOfPopup = true,
	  BackgroundColor = Colors.Orange,
	  HorizontalOptions = LayoutOptions.End,
	  VerticalOptions = LayoutOptions.Start,
	  Margin = 72,
	  Padding = 4
  });
  options.SetPopupOptionsDefaults(new DefaultPopupOptionsSettings
  {
	  CanBeDismissedByTappingOutsideOfPopup = true,
	  OnTappingOutsideOfPopup = () => hasOnTappingOutsideOfPopupExecuted = true,
	  PageOverlayColor = Colors.Orange,
	  Shadow = null,
	  Shape = null
  });
})
```

 ### PR Checklist ###
 
 - [ ] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/574

This PR also fixes a bug where `View.Padding` was improperly being passed through to `PopupBorder.Padding`


src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
```diff
@@ -198,23 +196,22 @@ public PopupPageLayout(in Popup popupContent, in IPopupOptions options, in IComm
...
 			{
-				BackgroundColor = popupContent.BackgroundColor ??= PopupDefaults.BackgroundColor,
+				BackgroundColor = popupContent.BackgroundColor ??= Options.DefaultPopupSettings.BackgroundColor,
 				Content = popupContent
 			};
 
 			// Bind `Popup` values through to Border using OneWay Bindings 
-			PopupBorder.SetBinding(Border.MarginProperty, static (Popup popup) => popup.Margin, source: popupContent, mode: BindingMode.OneWay);
-			PopupBorder.SetBinding(Border.PaddingProperty, static (Popup popup) => popup.Padding, source: popupContent, mode: BindingMode.OneWay);
+			PopupBorder.SetBinding(Border.MarginProperty, static (Popup popup) => popup.Margin, source: popupContent, mode: BindingMode.OneWay, converter: new MarginConverter());
```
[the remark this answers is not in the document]
TheCodeTraveler: The `Popup.Padding` should've never propagated through to `PopupBorder `. This was a bug, removed on purpose.

src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
```diff
@@ -27,19 +27,18 @@ partial class PopupPage : ContentPage, IQueryAttributable
...
-	public PopupPage(Popup popup, IPopupOptions popupOptions)
+	public PopupPage(Popup popup, IPopupOptions? popupOptions)
 	{
 		ArgumentNullException.ThrowIfNull(popup);
-		ArgumentNullException.ThrowIfNull(popupOptions);
-
+		
 		this.popup = popup;
-		this.popupOptions = popupOptions;
+		this.popupOptions = popupOptions ??= Options.DefaultPopupOptionsSettings;
```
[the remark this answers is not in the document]
TheCodeTraveler: > `??=` can be confusing

Haha - what?

bijington:
I know I initiated this thought process but I wanted to play devils advocate a little before we go further. I do like the idea of allowing developers to define defaults but I want to float some thoughts by you all to see what you think:

- We should allow developers to control values through a `Style` which I believe is currently broken (#2747). I am not suggesting we fix it in this PR but I don't want this PR to make it any more difficult to fix
- ~Perhaps a way to reduce the complexity of multiple possible values for things like `CanDismiss...` is to accept `PopupOptions` when passing in a `View` to the `ShowPopup` methods and not when passing in a `Popup`? I know this breaks some symmetry but with `Popup` but that might not be a bad thing.~ - I originally considered this but have struck it through because I think it causes more issues than it solves.
- I would like us to make it possible for properties like `BackgroundColor` and `PageOverlayColor` to be set using an `AppThemeBinding` - yes it would be possible via a `Style` but should it be possible as a default too?

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
```diff
@@ -55,6 +55,8 @@ public static MauiApp CreateMauiApp()
 								.UseMauiCommunityToolkit(static options =>
 								{
 									options.SetShouldEnableSnackbarOnWindows(true);
+									options.SetPopupDefaults(new DefaultPopupSettings());
```
bijington, line 58: This is just me probably being lazy and not checking properly but do developers need to do this for empty options? Or is this just highlighting its existence?

src/CommunityToolkit.Maui/Views/Popup/Popup.shared.cs
```diff
@@ -30,17 +30,18 @@ public partial class Popup : ContentView
...
 	/// <summary>
 	/// Initializes Popup
 	/// </summary>
 	public Popup()
 	{
-		Margin = PopupDefaults.Margin;
-		Padding = PopupDefaults.Padding;
-		HorizontalOptions = PopupDefaults.HorizontalOptions;
-		VerticalOptions = PopupDefaults.VerticalOptions;
+		Margin = Options.DefaultPopupSettings.Margin;
```
bijington, line 40: I have dragged my heels reviewing this to see if I could find a clean way of avoiding setting these values in the constructor. I had hoped to override more of the `BindableProperty` definitions above in order to control the default values. Sadly I could only do it with providing more of a definition of the `MarginProperty` etc. which introduces more risk of things breaking in the future so I think we should just go ahead with this and see how it plays with styling in the next PR.

bijington (approved):
Hey @TheCodeTraveler good work on this! I have added 2 comments - one to help me understand what needs to be documented and another mostly just explaining why I took so long to review. Feel free to resolve the comments and merge.

samples/CommunityToolkit.Maui.Sample/MauiProgram.cs
TheCodeTraveler: Haha, yup - it just highlights its existence. This example in the sample app is the equivalent of doing nothing.

src/CommunityToolkit.Maui/Views/Popup/Popup.shared.cs
TheCodeTraveler: Agreed. Let's see if we can crack this nut in your PR that adds support for Styles 💯

PR #2770, opened 2025-06-25
Move default values into a style to support style overriding

### LINKED ISSUE #2747: [BUG] Padding and Margin are not applied to the Popup class of Popup2 by Style attribute

bijington:
### Description of Change ###

 

I haven't fully tested this but from what I can see it certainly seems to solve the lack of Style support for the Popup properties. I suspect there may be a more efficient way to do this too, perhaps a single instance of the default style?

 ### Linked Issues ###
 

 - Fixes #2747 

 ### PR Checklist ###
 
### DIFF
--- a/src/CommunityToolkit.Maui/Views/Popup/Popup.shared.cs
+++ b/src/CommunityToolkit.Maui/Views/Popup/Popup.shared.cs
@@ -8,39 +8,26 @@ namespace CommunityToolkit.Maui.Views;
 public partial class Popup : ContentView
 {
 	/// <summary>
-	/// Bindable property to set the margin between the <see cref="Popup"/> and the edge of the window
-	/// </summary>
-	public static new readonly BindableProperty MarginProperty = View.MarginProperty;
-
-	/// <summary>
-	/// Bindable property to set the padding between the <see cref="Popup"/> border and the <see cref="Popup"/> content
-	/// </summary>
-	public static new readonly BindableProperty PaddingProperty = ContentView.PaddingProperty;
-
-	/// <summary>
-	/// Bindable property to set the horizontal position of the <see cref="Popup"/> when displayed on screen
-	/// </summary>
-	public static new readonly BindableProperty HorizontalOptionsProperty = View.HorizontalOptionsProperty;
-
-	/// <summary>
-	/// Bindable property to set the vertical position of the <see cref="Popup"/> when displayed on screen
-	/// </summary>
-	public static new readonly BindableProperty VerticalOptionsProperty = View.VerticalOptionsProperty;
-
-	/// <summary>
 	///  Backing BindableProperty for the <see cref="CanBeDismissedByTappingOutsideOfPopup"/> property.
 	/// </summary>
 	public static readonly BindableProperty CanBeDismissedByTappingOutsideOfPopupProperty = BindableProperty.Create(nameof(CanBeDismissedByTappingOutsideOfPopup), typeof(bool), typeof(Popup), PopupDefaults.CanBeDismissedByTappingOutsideOfPopup);
-
+	
 	/// <summary>
 	/// Initializes Popup
 	/// </summary>
 	public Popup()
 	{
-		Margin = PopupDefaults.Margin;
-		Padding = PopupDefaults.Padding;
-		HorizontalOptions = PopupDefaults.HorizontalOptions;
-		VerticalOptions = PopupDefaults.VerticalOptions;
+		var style = new Style(typeof(Popup))
+		{
+			ApplyToDerivedTypes = true
+		};
+
+		style.Setters.Add(new Setter { Property = Popup.HorizontalOptionsProperty, Value = PopupDefaults.HorizontalOptions });
+		style.Setters.Add(new Setter { Property = Popup.VerticalOptionsProperty, Value = PopupDefaults.VerticalOptions });
+		style.Setters.Add(new Setter { Property = Popup.MarginProperty, Value = PopupDefaults.Margin });
+		style.Setters.Add(new Setter { Property = Popup.PaddingProperty, Value = PopupDefaults.Padding });
+		
+		Resources.Add(style);
 	}
 
 	/// <summary>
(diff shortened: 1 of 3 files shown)


bijington:
In fact I think this currently fails for `ImplicitStylePopup`

src/CommunityToolkit.Maui/Views/Popup/Popup.shared.cs
```diff
@@ -7,40 +7,27 @@ namespace CommunityToolkit.Maui.Views;
...
-	/// <summary>
-	/// Bindable property to set the horizontal position of the <see cref="Popup"/> when displayed on screen
-	/// </summary>
-	public static new readonly BindableProperty HorizontalOptionsProperty = View.HorizontalOptionsProperty;
-
-	/// <summary>
-	/// Bindable property to set the vertical position of the <see cref="Popup"/> when displayed on screen
-	/// </summary>
-	public static new readonly BindableProperty VerticalOptionsProperty = View.VerticalOptionsProperty;
-
```
TheCodeTraveler: Do we need to delete these? I'd prefer to keep 'em. 

These exist to update the XML comments on each existing View `BindableProperty` to improve intelligence.

src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.Maui.Controls.PlatformConfiguration;
 using Microsoft.Maui.Controls.PlatformConfiguration.iOSSpecific;
 using Microsoft.Maui.Controls.Shapes;
+using static Microsoft.Maui.Controls.View;
```
TheCodeTraveler: Just a personal opinion - I'm not a fan of using `using static` for a large class like `Microsoft.Maui.Controls.View` because it can cause confusion and be bit overwhelming when intellisense provides suggestions when writing future code. 

I prefer to nix the `using static`, but it's not a blocker for me to approve this PR.

TheCodeTraveler (commented):
Thanks Shaun! I hadn't considered `Style` when working on Popup v2 and boy was that a glaring mistake. 

Is this PR ready, or does it need more work for implicit styles like you mentioned in your comment above?

src/CommunityToolkit.Maui/Views/Popup/Popup.shared.cs
bijington: Happy to put them back

src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
bijington: That's fair, I just let Rider go with its suggestions. I'll remove this

bijington:
> Thanks Shaun! I hadn't considered `Style` when working on Popup v2 and boy was that a glaring mistake. 
> 
> 
> 
> Is this PR ready, or does it need more work for implicit styles like you mentioned in your comment above?

I can't decide if the behaviour is correct or not. Currently the ImplicitStylePopup will fill full screen because it doesn't define a value for `HorizontalOptions` or `VerticalOptions`. It does provide `HeightRequest` and `WidthRequest` which has no effect because of the first two that I mentioned.

I am not thinking that this behaviour is correct because this implicit style overrides our implicit style in the toolkit - there can only be one implicit style to use right?

TheCodeTraveler:
> I can't decide if the behaviour is correct or not. Currently the ImplicitStylePopup will fill full screen because it doesn't define a value for HorizontalOptions or VerticalOptions. It does provide HeightRequest and WidthRequest which has no effect because of the first two that I mentioned.

We don't want it to fill the screen by default; it should be centered by default.

I haven't had time to dig into the code - just a thought - this may be related to how we're propagating the `HorizontalOptions` and `VerticalOptions` via Bindings from the `View` that the user gives us onto the underlying `Popup` and underlying `PopupBorder` in `PopupPage. CreatePopupFromView<T>()` and `PopupPage.PopupPageLayout()`:


### `PopupPage. CreatePopupFromView<T>()`
https://github.com/CommunityToolkit/Maui/blob/28e030c24f370c73e6fb0490ba4111ad7c74f205/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs#L127-L148


### `PopupPage.PopupPageLayout()`
https://github.com/CommunityToolkit/Maui/blob/28e030c24f370c73e6fb0490ba4111ad7c74f205/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs#L186-L220

mos379:
I really would like this change to be considered...

PR #2862, opened 2025-08-23
fix camera-preview size issue on iOS (#2296)

### LINKED ISSUE #2296: [BUG] Captured image in CameraView is Larger than camera preview

### Current Behavior

When using `CommunityToolkit.CameraView`, the image captured is larger than the preview displayed on the camera page. This results in a mismatch between the visible area in the `CameraView` and the actual captured image.

### CameraView:
drawing

### Result:
drawing

### Expected Behavior

The captured image should match the preview area displayed in the `CameraView`.

### Steps To Reproduce

1. Implement the `CameraView` as shown below: 
**XAML:**
```xaml
<toolkit:CameraView x:Name="CameraView"
                    VerticalOptions="Fill"
                    HorizontalOptions="Fill"
                    CameraFlashMode="Off"
                    MediaCaptured="OnMediaCaptured" />

```

**C# Code Behind:**
```csharp
protected override async void OnAppearing()
{
    base.OnAppearing();
    await _cameraProvider.RefreshAvailableCameras(CancellationToken.None);

    CameraView.SelectedCamera = _cameraProvider.AvailableCameras
        .FirstOrDefault(c => c.Position == CameraPosition.Rear);
}

private async void OnCaptureButtonClicked(object sender, EventArgs e)
{
    if (!CameraView.IsAvailable)
    {
        await DisplayAlert("Camera Unavailable", "Camera is not available.", "OK");
        return;
    }

    try
    {
        await CameraView.CaptureImage(CancellationToken.None);
    }
    catch
    {
        // Ignore exceptions
    }
}

private async void OnMediaCaptured(object? sender, MediaCapturedEventArgs e)
{
    try
    {
        using var stream = (MemoryStream)e.Media;
        _stateService.CapturedPhotos.Add(stream.ToArray());
    }
    catch
    {
        // Ignore exceptions
    }
}
```

2. Capture an image using the `CameraView`.

‌

### Environment

- .NET MAUI CommunityToolkit: 9.1.0
- .NET MAUI CommunityToolkit.Camera: 1.0.5
- OS: MacOS Sequoia 15
- .NET MAUI: 9.0.0-rc.2.24503.2

### Anything else?

_No response_

MoienTajik:
### Description of Change ###
This PR fixes an issue with `CameraView` where the camera preview showed an image of a different size (smaller/larger based on selected resolution) than the one actually captured.

 - Fixes #2296 

 ### PR Checklist ###
 
 - [x] Has a linked Issue, and the Issue has been `approved` (bug) => not approved yet


bijington:
Thank you for this. I think the better fix is to allow developers to provide this value. Either by an options setting or a property on the CameraView, or both.

MoienTajik:
Thanks for the review @bijington! To confirm, I’ll add a cross-platform preview-scaling enum on `CameraView` (mapped to iOS `AVLayerVideoGravity` and Android `PreviewView.ScaleType`), expose it as a bindable property and via options, and set this new behavior as the default since the current default is incorrect. Good to proceed?

bijington:
Sorry I have questions:
1. If we have a cross platform enum is there an equivalent for Windows too?
2. The original PR only modified iOS does the Android implementation already set the desired behaviour?
3. Depending on the answer to the above I'm not sure we should change the default for iOS as that is breaking. Or we at least need to make sure it's what we want and label/advertise it correctly

CoreMyrkai:
Hello there,

is there any chance that we get this fix?

MoienTajik:
> Sorry I have questions:
> 
> 1. If we have a cross platform enum is there an equivalent for Windows too?
> 2. The original PR only modified iOS does the Android implementation already set the desired behaviour?
> 3. Depending on the answer to the above I'm not sure we should change the default for iOS as that is breaking. Or we at least need to make sure it's what we want and label/advertise it correctly

1. Currently, no enum exists. We need to introduce one and handle the internal mapping for each platform. I haven’t reviewed the Windows implementation yet.

2. From my testing, Android behaves correctly by default. It’s working fine, although I noticed in the issue thread that someone reported it for Android as well, but I couldn’t reproduce it.

3. The current iOS behavior is incorrect and seems to be a bug. However, fixing it would result in a breaking change.

@bijington

Myrkai:
Hello again,

we have been waiting for this important bug fix for quite some time.
Is there any chance it will be released soon, or are there any workarounds available?

Issue #2852, opened 2025-08-15
[BUG] CollectionView SelectionChanged event does not fire unless held for 3 seconds, only on iOS and only with 12.2.0 update


mccoym98:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [x] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

CollectionViews within toolkit:popup windows do not fire the SelectionChanged event unless the item has been held for 3 seconds. This issue is only on iOS, not Android, and was introduced with the 12.2.0 Update. Rolling back one fixed the problem.

### Expected Behavior

Immediate click recongition on CollectionView items within a toolkit:popup window as in previous versions.

### Steps To Reproduce

1. Open and run solution from repository.
2. Click to open popup.
3. Clicking any popup item should immediately close the popup (within the SelectionChanged event)
4. It will on android, but you must hold 3 seconds on iOS to fire SelectionChanged

### Link to public reproduction project repository

https://github.com/mccoym98/ToolkitCollectionViewBugRepro.git

### Environment

```markdown
- .NET MAUI CommunityToolkit: 12.2.0
- OS: iOS 14, 15 & 16
- .NET MAUI: current
```

### Anything else?

MAUI Community Toolkit Update 12.2.0 introduced a bug into my application only when run on iOS. I use CollectionViews inside of toolkit:popups. These CollectionViews are composed of borders and other similar elements within the DataTemplate. Those all now require a solid 3-second press in order to be activated as measured by the SelectionChanged event from the CollectionView. From what I have tested, CollectionViews not being used on a popup page are unaffected. I have rolled back NuGet updates one-by-one and pinned the issue to this toolkit on the 12.2.0 version. Buttons on the popup work fine as expected. Android works fine as well. This is an iOS only issue.

MichaelShapiro:
I haven't updated to 12.2.0 yet (still on 12.1.0) so can't confirm the above behavior and have no desire to :)   But it sure sounds like the #2727 that I reported a short while back in 12.0.0 and that was since fixed. @bijington is it a regression in 12.2.0?

bijington:
I thought we fixed this as part of the 12.2.0 release so I'm stumped

MichaelShapiro:
[...]

PR #2741, opened 2025-06-17
Prevent a tap inside the content from closing a popup

bijington:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #2730
 - Fixes #2727

 ### PR Checklist ###
 
 - [x] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [ ] Has tests (if omitted, state reason in description)
 - [ ] Has samples (if omitted, state reason in description)
 - [x] Rebased on top of `main` at time of PR
 - [x] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#contributing-code---best-practices)
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pulls 

 ### Additional information ###

Issue #2727, opened 2025-06-16
[BUG] [iOS] CollectioView hosted in PopupV2 doesn't select an item unless you hold a row for longer than 4 second


MichaelShapiro:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [x] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

CollectionView hosted inside the PopupV2 doesn't acknowledge item selection tap unless **you tap and hold the row for more than 4 seconds**. Android works fine. I tested with "old" CollectionView and with CV2 -the same behavior.  Video is attached.

### Expected Behavior

Tapping on a CollectionView with SelectionMode="Multiple" or "Single" must actually select the tapped item (same as on Android as shown in the attached video)

### Steps To Reproduce

Follow the repository link, run the project "PopupV2Bugs", tap the button "CV2 selection bug"

------------------------
Host a CollectionView inside the PopupV2
Populate CollectionView with some data
Set SelectionMode="Multiple" (or "Single")
Show popup
on iOS device - Tap any item in the CollectionView

Expected result:  
- item remains selected (highlighted)
- SelectionChanged event fires
 

Actual:
- item remains unselected
- SelectionChanged doesn't fie
- only when you hold an item for longer than 4 seconds the above two actions happen
 

https://github.com/user-attachments/assets/ee318a82-6af4-434e-bff9-a43933210550

### Link to public reproduction project repository

https://github.com/MichaelShapiro/MauiBugs/tree/ad48ce23ab1548c10e90b55ccdef65264a9f2ab7/PopupV2_CV2_Selection

### Environment

```markdown
----------------------
OS: Windows 10
Visual Studio 2022 v 17.14.5

----------------------
nugets installed:
<PackageReference Include="CommunityToolkit.Maui" Version="12.0.0" />
<PackageReference Include="Microsoft.Maui.Controls" Version="9.0.70" />

----------------------
dotnet --list-sdks

9.0.301 [C:\Program Files\dotnet\sdk]


--------------------
dotnet workload list

Installed Workload Id      Manifest Version       Installation Source
---------------------------------------------------------------------
android                    35.0.61/9.0.100        VS 17.14.36203.30
ios                        18.4.9288/9.0.100      VS 17.14.36203.30
[...]

PR #2871, opened 2025-09-01
Using PointerGestureRecognizer to avoid blocking taps within Popup on iOS

kramer-e:
### Description of Change ###

The `TapGestureRecognizer` to check for taps outside the popup hijacks the taps for inner (child) controls. Changed the `TapGestureRecognizer` to a `PointerGestureRecognizer` that does not block/intercept the tap gesture. The bug was introduced in this PR https://github.com/CommunityToolkit/Maui/pull/2741 Tested on Android and iOS, the issue was only present on iOS.

 ### Linked Issues ###

 - Fixes [2852](https://github.com/CommunityToolkit/Maui/issues/2852)
 - Fixes [2852](https://github.com/CommunityToolkit/Maui/issues/2852)
- Fixes [2727](https://github.com/CommunityToolkit/Maui/issues/2727)

 ### PR Checklist ###
 
Added `CollectionView` with selectable items to `ComplexPopup` to test tapping a `CollectionView` item within the Popup. When tapping an item, it sets the value of the ReturnText `Entry`.


kramer-e:
@dotnet-policy-service agree

bijington:
Thank you for this, I haven't had a chance to test it but I am a little worried about the Microsoft documentation on this:

image

For some reason it doesn't claim to officially support iOS which either feels wrong from the docs or a potentially risky thing to include. Have you tested on iOS?

To follow-up with an additional question, is the `PointerGestureRecognizer` designed to not swallow the taps or is this just by chance? I haven't seen anything that explains either way but I'm just intrigued

kramer-e:
As far as I can see, iOS is supported in code, I just tested it and gave the exact same coordinates as the `TapGestureRecogniser`. I think it is not fully supported, because there is no "hover" event on iOS https://github.com/dotnet/maui/issues/9141

The funny thing is, that for Android `PointerGestureRecognizer` is officially is supported, but is broken because of a bug https://github.com/dotnet/maui/issues/20849 I just tested it on Android and it does not return a coordinate when using `PointerGestureRecognizer`.

So, to get this working with both platforms, `TapGestureRecognizer` must be used for Android and `PointerGestureRecognizer` for iOS.

kramer-e:
The `PointerGestureRecogniser` on iOS uses a custom implementation of the `UIGestureRecognizer` and is low level:
https://github.com/dotnet/maui/blob/936db46aa2c06c94b4eda7f288585c24b478d058/src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.iOS.cs#L410
This detects pointer entered/released and is not really tied to a specific UIView I think. 

The `TapGestureRecogniser` on iOS uses `UITapGestureRecognizer`  under the hood. 
https://github.com/dotnet/maui/blob/936db46aa2c06c94b4eda7f288585c24b478d058/src/Controls/src/Core/Platform/GestureManager/GesturePlatformManager.iOS.cs#L482
So I guess there is more involved there to detect a tap/double tap and it is tied to a certain UIView, so more risk of the tap being swallowed by this implementation.

For detecting a tap outside of the Popup I would use the more low level `PointerGestureRecogniser`, because you only need the coordinates.

bijington:
Or perhaps we go back to the idea of placing a view behind that can handle the tap detection. BoxView caused styling issues but we could create our own PopupBackground class that would allow developers to style it if they wanted (although it would fight with PopupOptions). This would save dealing with the platform differences that you mentioned. What do you think?

kramer-e:
I don't think styling the background will be needed in most usecases, having PageOverlayColor is good enough. Maybe it is also more complex, if you need to take safe-area's into account and it adds another layer in the UI. I updated my code to Popup v2 and hope this issue can be fixed soon, or use a workaround for the time being. 

The Android `PointerGestureRecognizer` should be fixed soon* (I hope), so maybe that is an option.


Thanks for your help Shaun!

TheCodeTraveler:
Just FYI - we discussed this PR in our [December Standup](https://www.youtube.com/watch?v=sJP9ZwAQ958).

tl;dw - While using `PointerGestureRecognizer` works, it feels like a bit of a tacky work-around that could bite us in the future if/when MAUI adds support for it on iOS. 

@CliffAgius volunteered to team up with @bijington to dig into the root cause of this issue to see what is actually causing the problem and if we can fix it at that lower-level instead of relying on `PointerGestureRecognizer`.

MichaelShapiro:
> Just FYI - we discussed this PR in our [December Standup](https://www.youtube.com/watch?v=sJP9ZwAQ958).
> 
> tl;dw - While using `PointerGestureRecognizer` works, it feels like a bit of a tacky work-around that could bite us in the future if/when MAUI adds support for it on iOS.
> 
> @CliffAgius volunteered to team up with @bijington to dig into the root cause of this issue to see what is actually causing the problem and if we can fix it at that lower-level instead of relying on `PointerGestureRecognizer`.

@TheCodeTraveler  Thank you so much for looking into this - it's a show stopper for my project unfortunately, and I'm watching it closely with lots of hope:)  So, just to clarify- do you still plan to release a fix with `PointerGestureRecognizer`  sometime soon? or you'll hold on until a real cause of the issue is found ? Thank you so much.

bijington:
@MichaelShapiro im planning on taking a look at the preferred fix over the next week

TheCodeTraveler:
Closed in favor of https://github.com/CommunityToolkit/Maui/pull/2997/

PR #2932, opened 2025-11-07
Update `[BindableProperty]` Support for Additional Scenarios, Add `CommunityToolkit.Maui.SourceGenerators.Internal.UnitTests`, Implement `[BindableProperty]` for Behaviors

TheCodeTraveler:
### Description of Change ###

As Discussed in our [November 2025 Standup](https://www.youtube.com/watch?v=TaIZww5HoCU), this PR implements the `[BindableProperty]` attribute, replacing the manually created BindableProperties.

This PR also adds support to `[BindableProperty]` for the following scenarios:
- Property names using a reserved keyword, e.g. `@class`
- Properties inside Generic Classes
- Properties inside Nested Classes
- Properties using Nullable reference types
- Properties using `Nullable<T>`
- Properties using Array types
- Properties inside classes using the global namespace
- Properties inside an inherited class that also has a property using `[BindableProperty]`
- Multiple Classes in the same file using `[BindableProperty]`
- All default value types, e.g. `bool`, `char`, `string` and `enum`

This PR also implements `CommunityToolkit.Maui.SourceGenerators.Internal.UnitTests`.

 ### PR Checklist ###
 
I've added the `breaking change` label because the generated code from `[BindableProperty]` fixes a few typos in `BindableProperty` names.


src/CommunityToolkit.Maui.SourceGenerators.Internal/Range.cs
```diff
@@ -0,0 +1,242 @@
+﻿// https://github.com/dotnet/runtime/blob/419e949d258ecee4c40a460fb09c66d974229623/src/libraries/System.Private.CoreLib/src/System/Index.cs
+// https://github.com/dotnet/runtime/blob/419e949d258ecee4c40a460fb09c66d974229623/src/libraries/System.Private.CoreLib/src/System/Range.cs
+
+using System.Runtime.CompilerServices;
+
+namespace System
```
pictos: wouldn't be better to use this lib instead?
 - https://github.com/Sergio0694/PolySharp

TheCodeTraveler: Oh nice! Done ✅

ne0rrmatrix (changes requested):
It looks very well done. Lets spin up `CommunityToolkit.Maui.SourceGenerators.Internal.UnitTests` as you suggested and test all the scenarios for `Create()` and `DefaultValue`

TheCodeTraveler:
> It looks very well done. Lets spin up CommunityToolkit.Maui.SourceGenerators.Internal.UnitTests as you suggested and test all the scenarios for Create() and DefaultValue

Thanks @ne0rrmatrix! I've added `CommunityToolkit.Maui.SourceGenerators.Internal.UnitTests`. 

It was a ton of hard work, but I was able to catch a few bugs in the source generator, particularly around edge cases. Thanks to these Unit Tests, the `[BindableProperty]` now supports all of the following scenarios:

- Property names using a reserved keyword, e.g. `@class`
- Properties inside Generic Classes
- Properties inside Nested Classes
- Properties using Nullable reference types
- Properties using `Nullable<T>`
- Properties using Array types
- Properties inside classes using the global namespace
- Properties inside an inherited class that also has a property using `[BindableProperty]`
- Multiple Classes in the same file using `[BindableProperty]`
- All default value types, e.g. `bool`, `char`, `string` and `enum`

ne0rrmatrix (approved):
This looks like you have thought about just about everything I could want or hope for with this. It is amazing! Approved!

PR #2981, opened 2025-12-01
Use `Copy` instead of `Move` in FileSaver UI on iOS

### LINKED ISSUE #2336: [BUG] FileSaver iOS displays 'Move' instead of 'Save'

### Current Behavior

Using CommunityToolkit.Maui.Storage.FileSaver.SaveAsync() on iOS, when the file dialog is raised it shows 'Move' in the upper-right corner of the interface instead of 'Save'.  This interface error happens even when saving a newly created file for the first time.  

The file is successfully saved to the selected location when Move is tapped, however the 'Move' prompt is simply confusing for a user that has created a brand-new file.  If the file already exists and Move is tapped, the user is properly prompted to replace or rename the file.  

This issue is reproducable using the CommunityToolkit Samples app, or by creating your own app using FileSaver.

### Expected Behavior

The iOS FileSaver dialog should show 'Save' when saving a newly created file. 

### Steps To Reproduce

1. Open and run the CommunityToolkit.Maui.Sample solution.
2. Tap the hamburger menu, select essentials, select FileSaver.
3. Select any of the FileSaver instantiation methods.

https://github.com/CommunityToolkit/Maui.git

### Environment

```markdown
- .NET MAUI CommunityToolkit: 4.0.0
- OS: iOS 17.2
- .NET MAUI: net8.0-ios
```

### Anything else?

_No response_

matt-goldman:
### Description of Change ###

 

Updates the `FileSaver` maciOS implementation to use the new `asCopy` argument in the `UIDocumentPickerViewController` to tell the OS that you're saving a new file (not copying) when moving the temp file out of the sandobx to the users' desired location.

 ### Linked Issues ###
 

 - Fixes #2336

 ### PR Checklist ###
 
 Only affects macOS and iOS. Screenshots show the resultant change on a physical iOS device:

IMG_3596

IMG_3598

### DIFF
--- a/src/CommunityToolkit.Maui.Core/Essentials/FolderPicker/FolderPickerImplementation.macios.cs
+++ b/src/CommunityToolkit.Maui.Core/Essentials/FolderPicker/FolderPickerImplementation.macios.cs
@@ -56,7 +56,7 @@ public sealed partial class FolderPickerImplementation : IFolderPicker
 			var path = e.Urls[0].Path;
 			if (path is null)
 			{
-				tcs.TrySetException(new FileSaveException("File path cannot be null."));
+				tcs.TrySetException(new FolderPickerException("File path cannot be null."));
 				return;
 			}
 


bijington (dismissed):
Thanks Matt. LGTM!

src/CommunityToolkit.Maui.Core/Essentials/FileSaver/FileSaverImplementation.macios.cs
```diff
@@ -35,7 +35,7 @@ async Task<string> InternalSaveAsync(string initialPath, string fileName, Stream
 		taskCompetedSource?.TrySetCanceled(CancellationToken.None);
 		var tcs = taskCompetedSource = new(cancellationToken);
 
-		documentPickerViewController = new([fileUrl])
+		documentPickerViewController = new([fileUrl], true)
```
VladislavAntonyuk: Should we delete temp file in this case if we copy it instead of move?

matt-goldman: I think so. I started refactoring this, I wanted to see if I could make it a more logical flow, and added a Cleanup method that handled that and the dispose. But I didn't want to introduce too many changes.

VladislavAntonyuk: Ideally if we don't use the temp file at all and write directly to the target location.

To make it simple for now you can wrap the code in try/finally block and delete temp file in finally section if it exists. It should not add complexity and many changes.

matt-goldman: > Ideally if we don't use the temp file at all and write directly to the target location.

I've spent a bit of time looking into this and I don't think it's practical. The only way to write directly outside the app's sandbox is using a security scoped URL and something like `NSFileCoordinator`. The problem is, this requires the user to grant access to it _first_, so it's not something we can do with a new file.

Additionally, we would have to track that the user has granted access and have safety checks that fall back to the temp file approach if not. And the problem with the first part is that it requires app-wide tracking and coordination. It's not impossible (far from it) but it's outside the scope of the vertical slice of this API.

The current approach (creating a temp file, then moving) is the canonical approach with Apple's API.

> To make it simple for now you can wrap the code in try/finally block and delete temp file in finally section if it exists. It should not add complexity and many changes.

Agreed 🙂 Done in 47ddfe5bceef5ec7f4773aa861d496369f3835bc. Please let me know if this is ok, or if you want me to make any other changes.

VladislavAntonyuk: I made a tiny change to ensure the temp file is always deleted. Please let me know if it looks good

src/CommunityToolkit.Maui.Core/Essentials/FileSaver/FileSaverImplementation.macios.cs
```diff
@@ -29,30 +29,41 @@ async Task<string> InternalSaveAsync(string initialPath, string fileName, Stream
...
-		if (currentViewController is not null)
-		{
-			currentViewController.PresentViewController(documentPickerViewController, true, null);
+			var currentViewController = Platform.GetCurrentUIViewController();
+			if (currentViewController is not null)
+			{
+				currentViewController.PresentViewController(documentPickerViewController, true, () =>
+				{
+					fileManager.Remove(tempDirectoryPath, out _);
+				});
```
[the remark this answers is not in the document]
matt-goldman: @VladislavAntonyuk what are your thoughts on this? I had the cleanup method, which could be called in a `finally` or as part of `InternalDispose`, that will always check for the file and remove it if it exists. If we restore that, it's always safe to call, and can be called from any catch or finally block, from dispose, or from the event handlers suggested here.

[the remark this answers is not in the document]
VladislavAntonyuk: moved back to finally block

src/CommunityToolkit.Maui.Core/Essentials/FileSaver/FileSaverImplementation.macios.cs
matt-goldman: Looks good to me. I made a couple more changes, I updated the exception message wording a little, and moved the controller and event handler registration to address the comment below. I also updated the finally to delete the temp directory instead of the file, to avoid accumulation of orphaned guid-named directories.

VladislavAntonyuk (dismissed):

PR #2995, opened 2025-12-08
Remove `BindablePropertyAttribute.DefaultValue`

TheCodeTraveler:
### Description of Change ###

This PR removes the `DefaultValue` property from `[BindableProperty]` for the following reasons now that it is not longer required thanks to https://github.com/CommunityToolkit/Maui/pull/2987:
  1. Setting the `DefaultValue` has been the largest source of pain and workarounds in `BindablePropertySourceGenerator`
      - You can see how the [helper method creating the generated code](https://github.com/CommunityToolkit/Maui/blob/20ab3fdc1edf5f7dd5149d53d342efddc845ab36/src/CommunityToolkit.Maui.SourceGenerators.Internal/Helpers/AttributeExtensions.cs#L21-L60) continues to grow every time we discover a new not-yet-supported edge case
      - We can now remove [that nasty helper method](https://github.com/CommunityToolkit/Maui/blob/20ab3fdc1edf5f7dd5149d53d342efddc845ab36/src/CommunityToolkit.Maui.SourceGenerators.Internal/Helpers/AttributeExtensions.cs#L21-L60), saving maintenance costs and avoid future bugs from yet-to-be encountered edge cases
      - Initializers already support **every** edge case
  2. The `DefaultValue` property only supported compile-time constants, while partial property initializer can use static readonly values
      - For example, `[BindableProperty(DefaultValue = Colors.Transparent)]` would generate a compiler error because `Colors.Transparent` is `public static readonly`
  3. Improve the developer experience
      - It's a bit confusing to give developers two ways to initialize / set the default value of the bindable property
      - If we keep `BindablePropertyAttribute.DefaultValue`, we would need to write an additional Analyzer to generate a compiler error when we detect a developer is using both `DefaultValue` and the partial property initializer

 ### PR Checklist ###
 
### `TouchBehavior` Refactoring
I refactored TouchBehavior's properties to be `null` by default so that we can determine whether or not a developer as set each property. 

I was forced into this refactor because the underlying `GestureManager` was relying on `BindableProeprety.IsSet` to determine whether or not a developer had set a value. This logic only works with `BindableProperty.DefaultValue`. Now that partial property initializers use `DefaultValueCreator`, `BindableProperty.IsSet` always returned `false` until .NET MAUI only calls `DefaultValueCreator` when the `BindableProperty` is first accessed (e.g. `BindableProperty.GetValue(DefaultOpacityProperty)`.

This refactor of TouchBehavior should not cause any breaking changes for developers, as noted by the fact that I did not need to modify any of the following code:
- TouchBehaviorPage
- TouchBehaviorTests 
  - (other than testing the Default Values of its properties)

### ImageTouchBehavior Refactoring

I also refactored `ImageTouchBehavior` (same reasons as `TouchBehavior`).

This refactor required adding an OnPropertyChanged method for `DefaultImageSource`, `HoveredImageSource` and `PressedImageSource` to support the scenario where the `ImageSource?` was set to `null` when the image was active. 

This refactor of TouchBehavior should not cause any breaking changes for developers, as noted by the fact that I did not need to modify any of the following code:
- ImageTouchBehaviorPage
- ImageTouchBehaviorTests 
  - I added additional Assertions to `ImageTouchBehaviorTests .VerifyImageSourceStateMachine`
  - I created `ImageTouchBehaviorTests .VerifyImageSourceStateMachineWhenImageSourceSetToNullWhilstActive`
  - I modified the DefaultValues test

### DIFF
--- a/src/CommunityToolkit.Maui.UnitTests/Behaviors/ImageTouchBehaviorTests.cs
+++ b/src/CommunityToolkit.Maui.UnitTests/Behaviors/ImageTouchBehaviorTests.cs
@@ -198,7 +198,7 @@ public class ImageTouchBehaviorTests() : BaseBehaviorTest<ImageTouchBehavior, Vi
 		imageTouchBehavior.HandleHover(HoverStatus.Entered);
 		Assert.Equal(imageTouchBehavior.PressedImageSource, image.Source);
 
-		imageTouchBehavior.PressedImageSource = ImageSource.FromUri(new Uri("https://www.google.com/images/branding/googlelogo/2x/googlelogo_dark_color_272x92dp.png"));
+		imageTouchBehavior.PressedImageSource = ImageSource.FromUri(new Uri("https://www.google.com/images/branding/googlelogo/2x/googlelogo_light_color_272x92dp.png"));
 		Assert.Equal(imageTouchBehavior.PressedImageSource, image.Source);
 
 		imageTouchBehavior.PressedImageSource = null;
@@ -208,7 +208,7 @@ public class ImageTouchBehaviorTests() : BaseBehaviorTest<ImageTouchBehavior, Vi
 		imageTouchBehavior.HandleTouch(TouchStatus.Completed);
 		Assert.Equal(imageTouchBehavior.HoveredImageSource, image.Source);
 
-		imageTouchBehavior.HoveredImageSource = ImageSource.FromUri(new Uri("https://www.google.com/images/branding/googlelogo/2x/googlelogo_dark_color_272x92dp.png"));
+		imageTouchBehavior.HoveredImageSource = ImageSource.FromUri(new Uri("https://www.google.com/images/branding/googlelogo/1x/googlelogo_dark_color_272x92dp.png"));
 		Assert.Equal(imageTouchBehavior.HoveredImageSource, image.Source);
 
 		imageTouchBehavior.HoveredImageSource = null;
@@ -218,7 +218,7 @@ public class ImageTouchBehaviorTests() : BaseBehaviorTest<ImageTouchBehavior, Vi
 		imageTouchBehavior.HandleHover(HoverStatus.Exited);
 		Assert.Equal(imageTouchBehavior.DefaultImageSource, image.Source);
 
-		imageTouchBehavior.DefaultImageSource = ImageSource.FromUri(new Uri("https://www.google.com/images/branding/googlelogo/2x/googlelogo_dark_color_272x92dp.png"));
+		imageTouchBehavior.DefaultImageSource = ImageSource.FromUri(new Uri("https://www.google.com/images/branding/googlelogo/1x/googlelogo_light_color_272x92dp.png"));
 		Assert.Equal(imageTouchBehavior.DefaultImageSource, image.Source);
 
 		imageTouchBehavior.DefaultImageSource = null;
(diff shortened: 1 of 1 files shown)


ne0rrmatrix:
Do we want to add tests for most common system types that we will see used? Like `DateTime` and others? Or add them as we update our sample app?

TheCodeTraveler:
> Do we want to add tests for most common system types that we will see used? Like DateTime and others?

@ne0rrmatrix Done! I've added these to this Unit Test: `GenerateBindableProperty_WithComplexDefaultValues_GeneratesCorrectCode`

src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/Touch/GestureManager.shared.cs
```diff
@@ -278,56 +291,64 @@ static async Task SetImageSource(TouchBehavior touchBehavior, TouchState touchSt
...
 		}
 
+		ImageSource? updatedImageSource = imageTouchBehavior.DefaultImageSource ?? (ImageSource?)ImageTouchBehaviorDefaults.DefaultBackgroundImageSource;
+		Aspect? updatedImageAspect = imageTouchBehavior.DefaultImageAspect ?? ImageTouchBehaviorDefaults.DefaultBackgroundImageAspect;
+
 		switch (touchState, hoverState)
 		{
 			case (TouchState.Pressed, _):
-				if (imageTouchBehavior.IsSet(ImageTouchBehavior.PressedImageAspectProperty))
+				if (imageTouchBehavior.PressedImageAspect is not null)
```
ne0rrmatrix, line 300: Can `ImageTouchBehavior.PressedImageAspect` be null? If so are we handling that? Should we throw an error if it can be null and is? Or is this redundant? There are many other examples in this class with same question. If this is irrelevant just ignore this.

TheCodeTraveler: Yes, it can! That's why I went ahead and added these `OnPropertyChanged` methods to our ImageTouchBehavior properties:

https://github.com/CommunityToolkit/Maui/blob/f829660650638dddae9a04bcdf01076382ff6780/src/CommunityToolkit.Maui/Behaviors/PlatformBehaviors/ImageTouch/ImageTouchBehavior.shared.cs#L52-L108

ne0rrmatrix: Wow I'm blind. I was staring at that code 30 min ago thinking it was super neat. I am now fully happy with this PR. It is crazy complicated in some areas but actually after working around some of these issues in Camera and MediaElement I am happy someone created this solution!

TheCodeTraveler: These `OnPropertyChanged` methods are the workaround to handle the `null` scenario you mention.

They also add a new feature where `ImageTouchBehavior` will now change the `ImageSouce` immediately when the BindableProperty changes. Previously, we would wait to change the image until the `CurrentTouchState` or `CurrentHoverState` changed.

TheCodeTraveler: And, because this existing Unit Test for the `ImageSource` State Machine is nearly unchanged, we know that everything is still working as expected compared to the existing `GestureManager` implementation: 

https://github.com/CommunityToolkit/Maui/blob/c12938460a571c7a28ce50fa7c1169ce696bfe59/src/CommunityToolkit.Maui.UnitTests/Behaviors/ImageTouchBehaviorTests.cs#L228-L285

TheCodeTraveler: And to test the new feature where we will update the Image when it changes without waiting for the state to change, I added this additional Unit Test: 

https://github.com/CommunityToolkit/Maui/blob/c12938460a571c7a28ce50fa7c1169ce696bfe59/src/CommunityToolkit.Maui.UnitTests/Behaviors/ImageTouchBehaviorTests.cs#L180-L226

PR #3016, opened 2025-12-19
Refactor Layouts to use [BindableProperty] partial properties

ne0rrmatrix:
### Description of Change ###

This adds support for `[BindableProperties]` to `Layouts` in CommunityToolkitMaui. 
This PR is blocked until support is added for `BindableProperties.AddAttached` as it is required to complete this PR. There are several classes `StateContainer`, `StateContainerController` and `StateView` that require addition updates to our internal source generator to complete this. 

 - Fixes #

 ### PR Checklist ###
 
 - [ ] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [ ] Has tests (if omitted, state reason in description)
 - [ ] Has samples (if omitted, state reason in description)
 - [x] Rebased on top of `main` at time of PR
 - [x] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#contributing-code---best-practices)
 - [ ] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pulls 

 ### Additional information ###

PR #3024, opened 2026-01-02
Create `[AttachedBindableProperty<T>]`

TheCodeTraveler:
### Description of Change ###

This PR implements `[AttachedBindableProperty<T>("PropertyName")]`:

The attribute can be applied to a Class, like so:

```cs
using CommunityToolkit.Maui;
using Microsoft.Maui.Controls;

namespace TestNamespace

[AttachedBindableProperty<string>("Text", // Name of the Attached Property
			      IsNullable = true, // Should generate a nullable type for T
			      DefaultValue = "Hello World", // The default value for the property
			      DefaultBindingMode = default, // The BindingMode to use on SetBinding() if no BindingMode is given. Default is <see cref="Microsoft.Maui.Controls.BindableProperty.Default"/>
			      ValidateValueMethodName = null, // Method name for <see cref="Microsoft.Maui.Controls.ValidateValueDelegate"/> to be run when a value is set. Default value is null
			      PropertyChangedMethodName = null, // Method name for <see cref="Microsoft.Maui.Controls.BindingPropertyChangedDelegate"/> to be run when a value has changed. Default value is null
			      PropertyChangingMethodName = null, // Method name for <see cref="Microsoft.Maui.Controls.BindingPropertyChangingDelegate"/> to be run when a value is set. Default value is null
			      CoerceValueMethodName = null, // Method name for <see cref="Microsoft.Maui.Controls.CoerceValueDelegate"/> used to coerce the range of a value. Default value is null
			      DefaultValueCreatorMethodName = null, // Method name for <see cref="Microsoft.Maui.Controls.CreateDefaultValueDelegate"/> used to initialize default value for reference types. Default value is null
			      BindablePropertyXmlDocumentation = null, // Custom XML Comments added to BindableProperty
			      GetterMethodXmlDocumentation = null, // Custom XML Comments added to Get{PropertyName}(BindableProperty bindable)
			      SetterMethodXmlDocumentation = null, // Custom XML Comments added to <c>Set{PropertyName}(BindableProperty bindable, T value)
			      BindablePropertyAccessibility = AccessModifier.Public, // The access modifier applied to the generated field <see cref="Microsoft.Maui.Controls.BindableProperty"/>
			      GetterAccessibility = AccessModifier.Public, // The access modifier applied to the generated method Get{PropertyName}(BindableProperty bindable)
			      SetterAccessibility = AccessModifier.Public)] // The access modifier applied to the generated method Set{PropertyName}(BindableProperty bindable, T value)
public partial class TestClass : View
{
    // Alternatively, it can be placed on the constructor, for example:
    // [AttachedBindableProperty<string>("Text")]
    public TestClass()
    {
    }
}
```

Both usages will generate the following source code:
```cs
// <auto-generated>
// See: CommunityToolkit.Maui.SourceGenerators.Internal.AttachedBindablePropertyAttributeSourceGenerator
#pragma warning disable
#nullable enable
namespace TestNamespace;
public partial class TestClass
{
	/// <summary>
	/// Attached BindableProperty for the Text property.
	/// </summary>
	public static readonly global::Microsoft.Maui.Controls.BindableProperty TextProperty = global::Microsoft.Maui.Controls.BindableProperty.CreateAttached("Text", typeof(string), typeof(TestNamespace. TestClass),(string?)"Hello World", (Microsoft.Maui.Controls.BindingMode)0, null, null, null, null, null);
	/// <summary>
	/// Gets Text for the <paramref = "bindable"/> child element.
	/// </summary>
	public static string? GetText(global::Microsoft.Maui.Controls.BindableObject bindable) => (string)bindable.GetValue(TextProperty);
	/// <summary>
	/// Sets Text for the <paramref = "bindable"/> child element.
	/// </summary>
	public static void SetText(global::Microsoft.Maui.Controls.BindableObject bindable, string? value) => bindable.SetValue(TextProperty, value);
}
```

 ### PR Checklist ###
 
This PR is still in draft as some unit tests are still failing and need to be resolved.

This PR unblocks https://github.com/CommunityToolkit/Maui/pull/3016 and https://github.com/CommunityToolkit/Maui/pull/3000

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,7 +22,7 @@ https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitati
     <NuGetAuditMode>all</NuGetAuditMode>
 
     <!-- MAUI Specific -->
-    <MauiPackageVersion>10.0.20</MauiPackageVersion>
+    <MauiPackageVersion>10.0.30</MauiPackageVersion>
     <MauiStrictXamlCompilation>true</MauiStrictXamlCompilation>
     <SkipValidateMauiImplicitPackageReferences>true</SkipValidateMauiImplicitPackageReferences>
     <MauiEnableXamlCBindingWithSourceCompilation>true</MauiEnableXamlCBindingWithSourceCompilation>
--- a/src/CommunityToolkit.Maui.UnitTests/Behaviors/BaseBehaviorTest.cs
+++ b/src/CommunityToolkit.Maui.UnitTests/Behaviors/BaseBehaviorTest.cs
@@ -66,8 +66,17 @@ public abstract class BaseBehaviorTest<TBehavior, TView> : BaseTest
 		Assert.True(wasSuccessful);
 		Assert.Equal(view.BindingContext, attachedBehavior.BindingContext);
 	}
+	
+	protected override void Dispose(bool isDisposing)
+	{
+		base.Dispose(isDisposing);
+		
+		view.Behaviors.Clear();
+	}
 
-	protected class MockValidationBehavior : ValidationBehavior<string>
+	sealed class MockViewModel;
+	
+	protected sealed class MockValidationBehavior : ValidationBehavior<string>
 	{
 		public string? ExpectedValue { get; init; }
 		public bool SimulateValidationDelay { get; init; } = false;
@@ -81,10 +90,5 @@ public abstract class BaseBehaviorTest<TBehavior, TView> : BaseTest
 
 			return value == ExpectedValue;
 		}
-	}
-
-	class MockViewModel
-	{
-
 	}
 }
--- a/src/CommunityToolkit.Maui.UnitTests/Behaviors/BaseBehaviorTests.cs
+++ b/src/CommunityToolkit.Maui.UnitTests/Behaviors/BaseBehaviorTests.cs
@@ -1,10 +1,11 @@
+using CommunityToolkit.Maui.Behaviors;
 using CommunityToolkit.Maui.UnitTests.Mocks;
 using FluentAssertions;
 using Xunit;
 
 namespace CommunityToolkit.Maui.UnitTests.Behaviors;
 
-public class BaseBehaviorTests
+public class BaseBehaviorTests() : BaseBehaviorTest<MockBehavior, Label>(new MockBehavior(), new Label())
 {
 	[Fact]
 	public void AttachAndDetachCallsShouldCorrectlyAssignView()
--- a/src/CommunityToolkit.Maui/Behaviors/Validators/ValidationBehavior.shared.cs
+++ b/src/CommunityToolkit.Maui/Behaviors/Validators/ValidationBehavior.shared.cs
@@ -330,15 +330,24 @@ public abstract partial class ValidationBehavior : BaseBehavior<VisualElement>, 
 
 		if (view is not null)
 		{
-			UpdateStyle(view, IsValid);
+			TryUpdateStyle(view, IsValid);
 		}
 	}
 
-	void UpdateStyle(in VisualElement view, bool isValid)
+	bool TryUpdateStyle(in VisualElement view, bool isValid)
 	{
-		VisualStateManager.GoToState(view, isValid ? ValidVisualState : InvalidVisualState);
+		try
+		{
+			VisualStateManager.GoToState(view, isValid ? ValidVisualState : InvalidVisualState);
 
-		view.Style = (isValid ? ValidStyle : InvalidStyle) ?? view.Style;
+			view.Style = (isValid ? ValidStyle : InvalidStyle) ?? view.Style;
+
+			return true;
+		}
+		catch (ArgumentException)
+		{
+			return false;
+		}
 	}
 
 	void ResetValidationTokenSource(CancellationTokenSource? newTokenSource)


src/CommunityToolkit.Maui.SourceGenerators.Internal/Generators/AttachedBindablePropertyAttributeSourceGenerator.cs
```diff
@@ -0,0 +1,582 @@
...
+		  // See: CommunityToolkit.Maui.SourceGenerators.Internal.AttachedBindablePropertySourceGenerator
+
+		  #pragma warning disable
+		  #nullable enable
+		  namespace CommunityToolkit.Maui;
+
+		  [global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
+		  [global::System.AttributeUsage(global::System.AttributeTargets.Class | global::System.AttributeTargets.Constructor, AllowMultiple = true, Inherited = false)]
+		  [global::System.Diagnostics.CodeAnalysis.Experimental("{{BindablePropertyDiagnostic.BindablePropertyAttributeExperimentalDiagnosticId}}")]
+		  sealed partial class AttachedBindablePropertyAttribute<T>(string propertyName) : global::System.Attribute
```
pictos: Don't we want this to be public? So projects that reference the assembly that has the SG will also benefit from it.

src/CommunityToolkit.Maui.SourceGenerators.Internal/Generators/AttachedBindablePropertyAttributeSourceGenerator.cs
```diff
@@ -0,0 +1,582 @@
...
+			.Collect();
+
+		context.RegisterSourceOutput(classProvider, ExecuteAllValues);
+		context.RegisterSourceOutput(constructorProvider, ExecuteAllValues);
+	}
+
+	static void ExecuteAllValues(SourceProductionContext context, ImmutableArray<AttachedBindablePropertySemanticValues> semanticValues)
+	{
+		// Pre-allocate dictionary with expected capacity
+		var groupedValues = new Dictionary<(string, string, string, string), List<AttachedBindablePropertySemanticValues>>(semanticValues.Length);
```
pictos: I would say to create a record type for the key

src/CommunityToolkit.Maui.SourceGenerators.Internal/Generators/AttachedBindablePropertyAttributeSourceGenerator.cs
```diff
@@ -0,0 +1,582 @@
...
+				if (values.Count is 0 || string.IsNullOrEmpty(className) || string.IsNullOrEmpty(containingNamespace))
+				{
+					continue;
+				}
+
+				// Flatten attached bindable properties without SelectMany allocation
+				var attachedBindablePropertiesCount = 0;
+				foreach (var value in values)
+				{
+					foreach (var abp in value.BindableProperties.AsImmutableArray())
```
pictos: Why the usage of `AsImmutableArray`?

src/CommunityToolkit.Maui.SourceGenerators.Internal/Generators/AttachedBindablePropertyAttributeSourceGenerator.cs
```diff
@@ -0,0 +1,582 @@
...
+					{
+						if (attachedBindablePropertiesCount >= attachedBindablePropertiesBuffer.Length)
+						{
+							var newBuffer = System.Buffers.ArrayPool<AttachedBindablePropertyModel>.Shared.Rent(attachedBindablePropertiesBuffer.Length * 2);
+							Array.Copy(attachedBindablePropertiesBuffer, newBuffer, attachedBindablePropertiesBuffer.Length);
+							System.Buffers.ArrayPool<AttachedBindablePropertyModel>.Shared.Return(attachedBindablePropertiesBuffer);
```
pictos: We aren't cleaning up the  array when returning to the pool, this will make this array to hold values there, which can lead  in bugs if we use the full `Length` of the array instead of the right number of BindableProperties

src/CommunityToolkit.Maui.SourceGenerators.Internal/Generators/AttachedBindablePropertyAttributeSourceGenerator.cs
```diff
@@ -0,0 +1,582 @@
...
+							attachedBindablePropertiesBuffer = newBuffer;
+						}
+						attachedBindablePropertiesBuffer[attachedBindablePropertiesCount++] = abp;
+					}
+				}
+
+				var attachedBindableProperties = ImmutableArray.Create(attachedBindablePropertiesBuffer, 0, attachedBindablePropertiesCount);
```
pictos: Why not create the EquatableArray directly? Here we're allocating an ImmutableArray just to use it to allocate an EquatableArray. I see the EquatableArray ctor only allows ImmutableArray, maybe we could add an overload that accepts an Array. Since it holds the collection as an Array anyway

src/CommunityToolkit.Maui.SourceGenerators.Internal/Generators/AttachedBindablePropertyAttributeSourceGenerator.cs
TheCodeTraveler: Good call! I'll update it.

src/CommunityToolkit.Maui.Core/Primitives/Defaults/StateViewDefaults.shared.cs
```diff
@@ -0,0 +1,6 @@
+namespace CommunityToolkit.Maui.Core;
+
+static class StateViewDefaults
+{
+	public const string StateKey = "";
```
VladislavAntonyuk, line 5: We can use string.Empty.

src/CommunityToolkit.Maui.SourceGenerators.Internal.UnitTests/CommunityToolkit.Maui.SourceGenerators.Internal.UnitTests.sln
```diff
@@ -0,0 +1,24 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
```
VladislavAntonyuk: Let's use slnx file

TheCodeTraveler: Ah - good catch! I'll delete this. 

Visual Studio Code auto generates a `.sln` file. I try to delete them before committing, but this one slipped through the cracks.

src/CommunityToolkit.Maui.Core/Primitives/Defaults/StateViewDefaults.shared.cs
TheCodeTraveler: I actually tried that first! But `string.Empty` is still not a compile-time constant 
Image

src/CommunityToolkit.Maui.SourceGenerators.Internal.UnitTests/CommunityToolkit.Maui.SourceGenerators.Internal.UnitTests.sln
TheCodeTraveler: Done ✅

src/CommunityToolkit.Maui.SourceGenerators.Internal/Generators/AttachedBindablePropertyAttributeSourceGenerator.cs
TheCodeTraveler: Apologies - that was some AI slop code that was totally unnecessary. I've deleted that code because it's redundant.

src/CommunityToolkit.Maui.SourceGenerators.Internal/Generators/AttachedBindablePropertyAttributeSourceGenerator.cs
TheCodeTraveler: Sure! I'll update the code to use `EquatableArray<T>` for `EquatableArray<AttachedBindablePropertyModel>` attachedBindablePropertiesArray.

The good news is that there's no additional allocation when creating an `EquatableArray` from an `ImmutableArray`; both are `struct`s and both hold the same reference to `T[]` under the hood. 

When `EquatableArray` is created from an existing `ImmutableArray`, it takes that same reference to the `ImmutableArray`'s underlying array. And since both are `struct`s, there's no allocation on the heap.

src/CommunityToolkit.Maui.SourceGenerators.Internal/Generators/AttachedBindablePropertyAttributeSourceGenerator.cs
TheCodeTraveler: I'm not sure I understand what's missing. What do you recommend for the fix?

TheCodeTraveler:
> Looks good to go. Very well done. The tests look very comprehensive and complete.

Thanks @ne0rrmatrix! Could you give it one more review? I've just added support for `BindableProperty.CreateAttachedReadOnly()` in this commit: https://github.com/CommunityToolkit/Maui/pull/3024/commits/8294bf8ba87fcdda52e0cf845c6b63640312ff05

stephenquan:
Some alternate syntaxes we can consider:

## Option 1: Attach the new attribute to dummy properties (alternatively, consider a static or const)

 - Pro: a very simple syntax
 - Con: creates bloated unused properties
 - Con: issue with XML docs

```c#
public partial class TestClass
{
	// Attached the new attribute to a dummy property
	[AttachedBindableProperty]
	public partial string? Text { get; set; } = "Hello World";
}
```

Proposed generated source code for option 1:

```c#
// Use the dummy property to extract type and name and generate the attached property
public partial class TestClass
{
	public partial string? Text
	{
		get => field;
		set => field = value;
	}
	/// <summary>
	/// Attached BindableProperty for the Text property.
	/// </summary>
	public static readonly global::Microsoft.Maui.Controls.BindableProperty TextProperty = global::Microsoft.Maui.Controls.BindableProperty.CreateAttached("Text", typeof(string), typeof(TestNamespace. TestClass),(string?)"Hello World", (Microsoft.Maui.Controls.BindingMode)0, null, null, null, null, null);
	/// <summary>
	/// Gets Text for the <paramref = "bindable"/> child element.
	/// </summary>
	public static string? GetText(global::Microsoft.Maui.Controls.BindableObject bindable) => (string)bindable.GetValue(TextProperty);
	/// <summary>
	/// Sets Text for the <paramref = "bindable"/> child element.
	/// </summary>
	public static void SetText(global::Microsoft.Maui.Controls.BindableObject bindable, string? value) => bindable.SetValue(TextProperty, value);
}
```

## Option 2: Attach the new attribute to the real getter/setter, e.g.

 - Pro: Developers has to provide their GetText/SetText in this style anyway
 - Pro: Developers don't need to remember how to write the attached property
 - Pro: Has a better XML doc integration
 - Pro: Potentially addresses the readonly scenario because that will be the difference of one method versus two

```c#
public partial class TestClass
{
	[AttachedBindableProperty]
	public static string? GetText(BindableObject view)
	{
		return (string?)view.GetValue(TextProperty);
	}
	[AttachedBindableProperty]
	public static void SetText(BindableObject view, string? value)
	{
		view.SetValue(TextProperty, value);
	}
}
```

Proposed generated source code for option 2:

```c#
// Generate the attached property
public partial class TestClass
{
    /// <summary>
    /// Attached BindableProperty for the Text property.
    /// </summary>
    public static readonly global::Microsoft.Maui.Controls.BindableProperty TextProperty = global::Microsoft.Maui.Controls.BindableProperty.CreateAttached("Text", typeof(string), typeof(TestNamespace. TestClass),(string?)"Hello World", (Microsoft.Maui.Controls.BindingMode)0, null, null, null, null, null);}
```

pictos:
@stephenquan, thanks for your suggestions... But I've to say that I don't like that design.

1. It will create two methods (getter and setter) on the assembly, it will increase the app size in an order of O(n) for nothing. Also, it will create a property that will confuse devs, they may think they can get/set the value from that property and will fail.
2. For the method, I'm not sure about it, the idea of the SG is to avoid that boilerplate, the methods also are a boilerplate since they always will be the same.

But, looking at your suggestions for the methods gave me an idea on something that's missing. Right now, the SG doesn't provide an easy way for devs to insert behavior on the generated methods.

@TheCodeTraveler I think we should add it, and we can use the MVVM toolkit as a reference, it uses `partial methods` to allow devs to add behavior. I think we should do the same. I ain't 100% sure about the API design, but the general idea would be



```cs
// SG Code
	public static void SetText(BindableObject view, string? value)
	{
                BeforeSet(view, ref value);
		view.SetValue(TextProperty, value);
                AfterSet(view, value);
	}
        partial void BeforeSet(BindableObject view, ref string? value);
        partial void AfterSet(BindableObject view, string? value);

// User code
partial void BeforeSet(BindableObject view, string? value)
{
     if (!decimal.TryParse(value, out var money))
            value = string.Empty;
    
     value = $"R${money}";
}
```

bijington:
@pictos that looks good. Re: naming should we follow the `Changing` (BeforeSet) and `Changed` (AfterSet) approach?  We could even pass in the new value to changing if we think it helps

TheCodeTraveler:
Love the idea of adding partial methods! But, does it differ enough from the existing `void PropertyChanged(Bindable, object, object)` and `void PropertyChanging(Bindable,object,object)` delegates in `BindableProperty` that devs can hook into? I think we'd just be this existing functionality, no?

pictos:
@TheCodeTraveler oh yeah, I forgot those, yeah so I believe it's good to go

stephenquan:
I had another crack at this and came up with more syntaxes to consider:

### Option 3 user stubs out delegates:

```c#
public partial class TestClass
{
    [AttachedBindableProperty]
    public static partial Func<BindableObject, string?> GetText { get; }
    [AttachedBindableProperty]
    public static partial Action<BindableObject, string?> SetText { get; }
}
```

SG produces:

```c#
public partial class TestClass
{
    public static readonly global::Microsoft.Maui.Controls.BindableProperty TextProperty = global::Microsoft.Maui.Controls.BindableProperty.CreateAttached("Text", typeof(string), typeof(TestNamespace.TestClass), (string?)"Hello World", (Microsoft.Maui.Controls.BindingMode)0, null, null, null, null, null);
    public static partial Func<BindableObject, string?> GetText { get => field; } = new((bindable) => (string)bindable.GetValue(TextProperty));
    public static partial Action<BindableObject, string?> SetText { get => field; } = new((bindable, value) => bindable.SetValue(TextProperty, value));
}
```

### Option 4 (Option 3 refactored with easier-to-remember delegates)

```c#
public partial class TestClass
{
    [AttachedBindableProperty]
    public static partial AttachedBindablePropertyGetter<string?> GetText { get; }
    [AttachedBindableProperty]
    public static partial AttachedBindablePropertySetter<string?> SetText { get; }
}
```

SG produces:

```c#
public partial class TestClass
{
    public static readonly global::Microsoft.Maui.Controls.BindableProperty TextProperty = global::Microsoft.Maui.Controls.BindableProperty.CreateAttached("Text", typeof(string), typeof(TestClass), (string?)default);
    public static partial AttachedBindablePropertyGetter<string?> GetText { get => field; } = new((BindableObject view) =>
    {
        return (string?)view.GetValue(TextProperty);
    });
    public static partial AttachedBindablePropertySetter<string?> SetText { get => field; } = new((bindable, value) =>
    {
        bindable.SetValue(TextProperty, value);
    });
}
```

We define the helpers as delegates, i.e.

```c#
public delegate T AttachedBindablePropertyGetter<T>(BindableObject view);
public delegate void AttachedBindablePropertySetter<T>(BindableObject view, T value);
```

PR #3033, opened 2026-01-09
Fix StateContainer SourceGen mishandling Xaml with invalid cast.

ne0rrmatrix:
### Summary ###

This PR fixes a System.InvalidCastException occurring when defining StateViews in XAML. The crash was caused by the XAML compiler attempting to directly cast a VerticalStackLayout to IList<Microsoft.Maui.Controls.View> when assigned as the immediate child of the StateContainer.StateViews attached property.

 ### Description of Change ###

The root cause was a mismatch between the XAML usage and how the code generator handles attached properties of collection types.

StateContainer.StateViewsProperty is defined as an IList<View>. When a single layout (like VerticalStackLayout) is placed directly inside the <mct:StateContainer.StateViews> tags, the XAML generator emits code that attempts to assign that single element directly to the property via a cast:

SetValue(StateContainer.StateViewsProperty, (IList<View>)verticalStackLayout).

Because VerticalStackLayout does not implement IList<View> (its Children property is an IList<IView>), this explicit cast fails at runtime. By wrapping the content or explicitly defining the collection items, we ensure the XAML generator treats the elements as items to be added to the list rather than the list instance itself.

Technical Details

- Issue: System.InvalidCastException during XAML initialization.

- Fix: Updated XAML structure to ensure proper collection population, avoiding the invalid cast emitted by the XAML compiler.

- Note: This is a XAML usage/codegen mismatch rather than a logic bug within the StateContainer implementation.
 ### Linked Issues ###
 

 - Fixes #

 ### PR Checklist ###
 
We switched on xaml source generation in a previous PR. The error when navigating to StateContainer page is possibly a bug in Maui and I will be contacting someone who can take a look at it hopefully. I will update this PR with info as I get it.


TheCodeTraveler (changes requested):
Thanks James! I confirm that this fixes the runtime crash uncovered by enabling the XAML Source Generator. 

A big concern I have is that the [StateContainer Documentation](https://learn.microsoft.com/dotnet/communitytoolkit/maui/layouts/statecontainer) also recommends putting a `VerticalStackLayout` directly into `<mct:StateContainer.StateViews>` which was causing the runtime crash in our sample app after enabling XAML Source Generator.

Could you please work with @bijington to determine if this is a documentation issue or whether we have a deeper problem in our implementation of `StateContainer.StateViews`?

I'm not a XAML expert, but perhaps there's a solution where we use `[ContentProperty]`? I have no idea, so I'll defer to you guys who know more XAML than me! 

For now, I'm marking my review as `Request Changes` until we either have a PR submitted to update the Docs or we do something in code to make `StateContainer.StateViews` more compatible with our documented recommendations.

### StateContainer Docs

Here's where we are currently recommending putting a `VerticalStackLayout` directly into `<mct:StateContainer.StateViews>` in our docs: https://learn.microsoft.com/en-us/dotnet/communitytoolkit/maui/layouts/statecontainer

Screenshot 2026-01-09 at 4 02 54 PM

VladislavAntonyuk:
Does the issue can be fixed if we change `IList<View>`to `IList<IView>`?

ne0rrmatrix:
> Does the issue can be fixed if we change `IList<View>`to `IList<IView>`?

I tried to do that. I had no luck making that work.

bijington:
Does anyone know why this change fixes the issue?

Issue #3064, opened 2026-02-04
[BUG] Regression in 14.0.0: `DefaultPopupSettings` Broken for Padding


dartasen:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [x] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

I use popup in my application with the IPopupService, after upgrading to the latest version of the toolkit they don't behave the same.

Padding and margin seems broken (Defaults are set in MauiProgram). Whether I set it in the popup xaml or in the PopupDefault.

Screenshots are from an Android 16 Device (Pixel 9a), but can repro as well on iOS 18.4 & 26

Toolkit 13 :
<img width="578" height="691" alt="Image" src="https://github.com/user-attachments/assets/194bbff5-d04a-4d03-95fc-7d4eb065f73d" />

Toolkit 14 : 

<img width="680" height="613" alt="Image" src="https://github.com/user-attachments/assets/88143e60-64d5-4aff-8926-63b1ce3da4cf" />

### Expected Behavior

Works as previous version

### Steps To Reproduce

```cs
.UseMauiCommunityToolkit(static options =>
{
 options.SetPopupOptionsDefaults(new DefaultPopupOptionsSettings()
 {
     CanBeDismissedByTappingOutsideOfPopup = false,
     PageOverlayColor = Color.FromRgba(0, 0, 0, 200),
     Shape = new RoundRectangle()
     {
         CornerRadius = new CornerRadius(20, 20, 20, 20),
         StrokeThickness = 0d
     },
     Shadow = null,
     OnTappingOutsideOfPopup = null
 });

 options.SetPopupDefaults(new DefaultPopupSettings()
 {
     CanBeDismissedByTappingOutsideOfPopup = false,
     BackgroundColor = Color.FromArgb("22314E"),
     HorizontalOptions = LayoutOptions.Center,
     VerticalOptions = LayoutOptions.Center,
     Margin = new Thickness(12),
     Padding = new Thickness(10)
 });
})
``` 
> Note : Margin/Padding doesn't apply if adding those properties to the popup xaml
 
* Using toolkit 13, display a popup with Margin/Padding
* Upgrade to toolkit 14, same popup layout will break

### Link to public reproduction project repository

[...]

PR #3074, opened 2026-02-08
Fix Popup Padding Regression Introduced in v14.0.0

TheCodeTraveler:
### Description of Change ###

 

This PR fixes three problems:

1. In `PopupPage`, use `value.IsEmpty || value.IsNaN` instead of `value == default` for `PopupPage.PaddingConverter`
    - For certain elements, like `Button`, the value of its `Padding` will be `NaN` until the View has been laid out on the screen 
    - We now ensure that our converters check for both NaN and Thickness.Zero
2. In `Popup`, remove `[BindableProperty]` from `Padding`, passthrough its value to the underlying `ContentView.Padding` property
    - The `Popup.Padding` property is used for the Popup's Padding between itself and the `PopupBorder` which maintains how `ContentView` implements `Padding`; we don't need to override its functionality
    - We keep the `new` keyword on the `Popup.Padding` property to override the XML comments to explain to the developer how we are using `Popup.Padding` under the hood, passing through its value to the underlying `ContentView.Padding` property
    - In comparison, we keep `[BindableProperty]` on `Popup.Margin`, `Popup.HorizontalOptions` and `Popup.VerticalOptions` because these values do **not** propagate to the underlying `ContentView`, rather, we use these properties to assign `PopupBorder.Margin`, `PopupBorder.HoriztonalOptions` and `PopupBorder.VerticalOptions` respectively
    - I've added new unit tests to `DefaultPopupSettingsTests` that demonstrate this usage
3. In `Popup`, remove default property initializers
     - Move initialization to constructor

 ### Linked Issues ###
 

 - Fixes https://github.com/CommunityToolkit/Maui/issues/3064#issuecomment-3865760350

 ### PR Checklist ###
 
### DIFF
--- a/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
+++ b/src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
@@ -282,21 +282,21 @@ partial class PopupPage : ContentPage, IQueryAttributable
 		{
 			public override LayoutOptions DefaultConvertReturnValue { get; set; } = Options.DefaultPopupSettings.HorizontalOptions;
 
-			public override LayoutOptions ConvertFrom(LayoutOptions value, CultureInfo? culture) => value == LayoutOptions.Fill ? Options.DefaultPopupSettings.HorizontalOptions : value;
+			public override LayoutOptions ConvertFrom(LayoutOptions value, CultureInfo? culture) => value == LayoutOptions.Fill ? DefaultConvertReturnValue : value;
 		}
 
 		sealed partial class VerticalOptionsConverter : BaseConverterOneWay<LayoutOptions, LayoutOptions>
 		{
 			public override LayoutOptions DefaultConvertReturnValue { get; set; } = Options.DefaultPopupSettings.VerticalOptions;
 
-			public override LayoutOptions ConvertFrom(LayoutOptions value, CultureInfo? culture) => value == LayoutOptions.Fill ? Options.DefaultPopupSettings.VerticalOptions : value;
+			public override LayoutOptions ConvertFrom(LayoutOptions value, CultureInfo? culture) => value == LayoutOptions.Fill ? DefaultConvertReturnValue : value;
 		}
 
 		sealed partial class BackgroundColorConverter : BaseConverterOneWay<Color?, Color>
 		{
 			public override Color DefaultConvertReturnValue { get; set; } = Options.DefaultPopupSettings.BackgroundColor;
 
-			public override Color ConvertFrom(Color? value, CultureInfo? culture) => value ?? Options.DefaultPopupSettings.BackgroundColor;
+			public override Color ConvertFrom(Color? value, CultureInfo? culture) => value ?? DefaultConvertReturnValue;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
```diff
@@ -304,13 +304,13 @@ sealed partial class PaddingConverter : BaseConverterOneWay<Thickness, Thickness
...
-		public override Thickness ConvertFrom(Thickness value, CultureInfo? culture) => value == default ? Options.DefaultPopupSettings.Padding : value;
+		public override Thickness ConvertFrom(Thickness value, CultureInfo? culture) => value.IsEmpty || value.IsNaN ? Options.DefaultPopupSettings.Padding : value;
 	}
 
 	sealed partial class MarginConverter : BaseConverterOneWay<Thickness, Thickness>
 	{
 		public override Thickness DefaultConvertReturnValue { get; set; } = Options.DefaultPopupSettings.Margin;
 
-		public override Thickness ConvertFrom(Thickness value, CultureInfo? culture) => value == default ? Options.DefaultPopupSettings.Margin : value;
+		public override Thickness ConvertFrom(Thickness value, CultureInfo? culture) => value.IsEmpty || value.IsNaN ? Options.DefaultPopupSettings.Margin : value;
```
VladislavAntonyuk: Value is empty looks like a valid value for margin and padding, should we return our default value instead?

Should we return DefaultConvertReturnValue?

VladislavAntonyuk (dismissed):

src/CommunityToolkit.Maui/Views/Popup/PopupPage.shared.cs
TheCodeTraveler: Oh - good idea! I'll make `Margin` and `Padding` nullable. Then we'll just do a `null` check of using `default(Thickness)` / `value.IsEmpty`.

And yea, we can return `DefaultConvertReturnValue`. That property is technically never used. It only exists because I used `BaseConverterOneWay<>` instead of `IValueConverter`. But, it does make sense not to use it.

TheCodeTraveler: Done ✅

TheCodeTraveler: Whoops - I forgot that making `Popup.Padding` nullable doesn't do anything because `PaddingConverter` is only used when a `Popup` is created from a `View` passed into the `PopupPage` constructor. When a `Popup` is passed into the `PopupPage` constructor, we don't create a binding that uses `PaddingConverter`.

Long story short, we'll need to keep `value.IsEmpty || value.IsNaN ? DefaultConvertReturnValue : value;` for `PaddingConverter`

TheCodeTraveler: On second thought- let's not make the Popup properties nullable, because the cascading logic in PopupPage introduces too many breaking changes to Popup.

I'll revert the changes I made and we'll keep `value.IsEmpty || value.IsNaN ? DefaultConvertReturnValue : value` for both `PaddingConverter` and `MarginConverter`.

PR #3137, opened 2026-03-10
Add New Enum Description Converter

BillyMartin1964:
This converter converts Enum values into readable text so they display nicely in the UI. When an Enum value is passed in, the converter looks for attributes that provide a more user-friendly name. If one is found, it uses that value; otherwise it simply uses the Enum’s name. This ensures that Enum values can be displayed clearly while still working correctly even when no attributes are defined. Unit tests are included to verify the expected behavior.

 ### Description of Change ###

Added new converter

 ### Linked Issues ###
 

 - Fixes #

 ### PR Checklist ###
 
### Example use case:
 
```
 Text="{Binding SelectedModeName, Converter={StaticResource EnumDescriptionConverter}}"

public enum ModeName
{
    [Display(Name = "Light  Mode")]
    LightMode,
    [Display(Name = "Dark  Mode")]
    DarkMode,
    [Display(Name = "System")]
    System
}
```


BillyMartin1964:
There is a PR for the documentation at:

https://github.com/MicrosoftDocs/CommunityToolkit/pull/622

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+    public override string ConvertFrom(Enum value, CultureInfo? culture = null)
+    {
+        ArgumentNullException.ThrowIfNull(value);
+        var fieldInfo = value.GetType().GetField(value.ToString());
+
+        // Check for DisplayAttribute first (common in data annotations)
+        var displayAttr = fieldInfo?.GetCustomAttribute<DisplayAttribute>();
+        if (displayAttr is not null && !string.IsNullOrWhiteSpace(displayAttr.Name))
+        {
+            return displayAttr.Name;
```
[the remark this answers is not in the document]
BillyMartin1964: Fixed

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+        var descriptionAttr = fieldInfo?.GetCustomAttribute<DescriptionAttribute>();
+        if (descriptionAttr is not null && !string.IsNullOrWhiteSpace(descriptionAttr.Description))
+        {
+            return descriptionAttr.Description;
+        }
+
+        return value.ToString(); // Fallback to enum name if no attribute found
+    }
+
+    // ConvertBackTo is sealed in BaseConverterOneWay and cannot be overridden.
```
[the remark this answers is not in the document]
BillyMartin1964: Fixed

src/CommunityToolkit.Maui.UnitTests/Converters/EnumDescriptionConverterTests.cs
```diff
@@ -0,0 +1,109 @@
+﻿using Xunit;
+
+using CommunityToolkit.Maui.Converters;
+
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+
+namespace CommunityToolkit.Maui.UnitTests.Converters;
+
+public class EnumDescriptionConverterTests
```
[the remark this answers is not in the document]
BillyMartin1964: Fixed

src/CommunityToolkit.Maui.UnitTests/Converters/EnumDescriptionConverterTests.cs
```diff
@@ -0,0 +1,109 @@
...
+		WhitespaceDisplay,
+		[Description("")]
+		EmptyDescription,
+	}
+
+  [Fact]
+	public void ConvertFrom_ThrowsArgumentNullException_WhenNull()
+	{
+		var converter = new EnumDescriptionConverter();
+		Assert.Throws<ArgumentNullException>(() => converter.ConvertFrom(null!));
```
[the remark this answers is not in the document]
BillyMartin1964: Fixed

BillyMartin1964:
@dotnet-policy-service agree
________________________________

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+	/// <returns>
+	/// The value of the <see cref="DisplayAttribute.Name"/> if defined;
+	/// otherwise the value of the <see cref="DescriptionAttribute.Description"/> if defined;
+	/// otherwise the enum name as a string.
+	/// </returns>
+	public override string ConvertFrom(Enum value, CultureInfo? culture = null)
+	{
+		ArgumentNullException.ThrowIfNull(value);
+		var fieldInfo = value.GetType().GetField(value.ToString());
+
```
TheCodeTraveler, line 21: Is this possible without reflection?

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+namespace CommunityToolkit.Maui.Converters;
+
+/// <summary>
+/// Converts an <see cref="Enum"/> value to its display string using <see cref="DisplayAttribute"/> or <see cref="DescriptionAttribute"/>.
+/// </summary>
+[AcceptEmptyServiceProvider]
+public partial class EnumDescriptionConverter : BaseConverterOneWay<Enum, string>
+{
+	/// <inheritdoc/>
+	public override string DefaultConvertReturnValue { get; set; } = string.Empty;
```
TheCodeTraveler, line 15:
```suggestion
	public override string? DefaultConvertReturnValue { get; set; }
```

It would make sense for this converter to return a nullable string and return `null` when an enum does not contain a Description attribute.

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Globalization;
+using System.Reflection;
+
+namespace CommunityToolkit.Maui.Converters;
+
+/// <summary>
+/// Converts an <see cref="Enum"/> value to its display string using <see cref="DisplayAttribute"/> or <see cref="DescriptionAttribute"/>.
+/// </summary>
+[AcceptEmptyServiceProvider]
+public partial class EnumDescriptionConverter : BaseConverterOneWay<Enum, string>
```
TheCodeTraveler, line 12:
```suggestion
public partial class EnumDescriptionConverter : BaseConverterOneWay<Enum, string?>
```

It would make sense for this converter to return a nullable string, `BaseConverterOneWay<Enum, string?>` and return `null` when an enum does not contain a Description attribute.

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+	/// Converts an <see cref="Enum"/> value to its display string.
+	/// </summary>
+	/// <param name="value">The enum value to convert.</param>
+	/// <param name="culture">The culture to use for the conversion (not used).</param>
+	/// <returns>
+	/// The value of the <see cref="DisplayAttribute.Name"/> if defined;
+	/// otherwise the value of the <see cref="DescriptionAttribute.Description"/> if defined;
+	/// otherwise the enum name as a string.
+	/// </returns>
+	public override string ConvertFrom(Enum value, CultureInfo? culture = null)
```
TheCodeTraveler, line 18:
```suggestion
	public override string? ConvertFrom(Enum value, CultureInfo? culture = null)
```

It would make sense for this converter to return a nullable string and return `null` when an enum does not contain a Description attribute.

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		// Fallback to DescriptionAttribute
+		var descriptionAttr = fieldInfo?.GetCustomAttribute<DescriptionAttribute>();
+		if (descriptionAttr is not null && !string.IsNullOrWhiteSpace(descriptionAttr.Description))
+		{
+			return descriptionAttr.Description;
+		}
+
+		return value.ToString(); // Fallback to enum name if no attribute found
```
TheCodeTraveler:
```suggestion
		return value.ToString(); // Fallback to enum name if no attribute found
```

It would make sense for this converter to return a nullable string and return `null` when an enum does not contain a Description attribute.

TheCodeTraveler (changes requested):
Thanks @BillyMartin1964. 

I'm a bit confused by the scope of this Converter and would like to avoid using Reflection. 

### Scope

Does it get the `DescriptionAttribute` or does it get `DisplayAttribute.Name`? 

To put yourself in another developer's shoes, if you saw a converter named `EnumDescriptionConverter` as you're scrolling through intellisense, what would you assume it would do? Devs don't always read docs, so naming + scope is critically important. 

It's named the `EnumDescriptionConverter` which implies it will return the `DescriptionAttribute`, but this current implementation may return three different results without providing any control to the developer over which results comes back:
- `DescriptionAttribute`
- `DisplayAttribute.Name`
- Enum name

We should not return the enum name as that will cause confusion to the developer who expected its description. Instead, it should return `null` to indicate to the developer that no attribute was found, or throw an InvalidOperationException if we decide this converter does not support enums without descriptions. 

After that fix, I see two paths forward for this new converter:

1. Separate this into two separate converters:     
    - `EnumDescriptionConverter : BaseConverterOneWay<Enum, string?>`
    - `EnumDisplayNameConverter : BaseConverterOneWay<Enum, string?>`
2. Create an `enum` that allows the developer to choose which value is returned
  - This would be similar to our implementation for the `CompareConverter` `StringToListConverter` and how it allows developers customization via its enum `OperatorType`

### Reflection

I don't love that the current implementation is using reflection which limits the abilities for AOT compiling. Is this possible without Reflection? 

Have you confirmed this runs works on an iOS Device built in Release configuration without the Mono Interpreter? And have you confirmed this works in an AOT'd code base?

### New Feature Proposal Process

In the future, before submitting a PR implementing a feature and before writing the docs for it, I recommend following our [process for `Submitting a New Feature` ](https://github.com/CommunityToolkit/Maui?tab=readme-ov-file#submitting-a-new-feature). You've skipped the Proposal phase where we would've had this discussion about its implementation and saved time and re-work.

BillyMartin1964:
I wrote this converter for my own use, when I realized there wasn't  one in the toolkit. It's really quite handy, and I thought I would share it with the community. I did open a discussion, and asked James Montemagno to champion it, but he said to just send the pull request.


Here's the discussion link:

https://github.com/CommunityToolkit/Maui/discussions/3136


I don't really see the problem  with a tiny converter. I tried to cover all the bases. Some devs use display name, some use description, and some use localization. We could make three converters if you want. 

It really doesn't matter to me as long as it ends up in the toolkit. I think it's  quite a useful converter. Also, if you want to finish it up, that's  fine too. Just let me know. Happy to help.

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
BillyMartin1964: I originally had a null return, but the copilots reviewer said we should return  the key if the user was localizing. What do you think?

BillyMartin1964:
> We should not return the enum name as that will cause confusion to the developer who expected its description. Instead, it should return `null` to indicate to the developer that no attribute was found, or throw an InvalidOperationException if we decide this converter does not support enums without descriptions.

I'm not sure that you are right on this one. I usually don't put descriptions for one-word enum members.

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
BillyMartin1964: @TheCodeTraveler , It is possible without reflection. It requires running a CodeGenerator. I'm working on it.

BillyMartin1964:
Sample app page is completed. Works fine without reflection.

image

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
```diff
@@ -0,0 +1,301 @@
...
+[Generator]
+public class EnumDescriptionGenerator : IIncrementalGenerator
+{
+	// Using fields instead of const to avoid naming rule violations
+	static readonly string displayAttributeName = "System.ComponentModel.DataAnnotations.DisplayAttribute";
+	static readonly string descriptionAttributeName = "System.ComponentModel.DescriptionAttribute";
+
+	public void Initialize(IncrementalGeneratorInitializationContext context)
+	{
+		IncrementalValuesProvider<INamedTypeSymbol?> enumDeclarations = context.SyntaxProvider
```
pictos: please use the `context.SyntaxProvider.ForAttributeWithMetadataName` method instead, it's designed to be used with attributes

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
```diff
@@ -0,0 +1,301 @@
...
+		return sb.ToString();
+	}
+
+   static string? GetDescriptionFromDescriptionAttribute(IFieldSymbol member)
+	{
+     // Check for DescriptionAttribute
+		AttributeData? descAttr = member.GetAttributes()
+			.FirstOrDefault(attr => attr.AttributeClass?.ToDisplayString() == descriptionAttributeName);
+
+		if (descAttr != null && descAttr.ConstructorArguments.Length > 0)
```
pictos: we do use pattern matching for null check, please change here and other places that this comment applies


```suggestion
		if (descAttr is not null && descAttr.ConstructorArguments.Length > 0)
```

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
```diff
@@ -0,0 +1,301 @@
...
+	public void Initialize(IncrementalGeneratorInitializationContext context)
+	{
+		IncrementalValuesProvider<INamedTypeSymbol?> enumDeclarations = context.SyntaxProvider
+			.CreateSyntaxProvider(
+				predicate: static (node, _) => node is EnumDeclarationSyntax,
+				transform: static (ctx, _) => ctx.SemanticModel.GetDeclaredSymbol(ctx.Node) as INamedTypeSymbol)
+         .Where(static enumSymbol => enumSymbol?.TypeKind == TypeKind.Enum)
+			.Where(static enumSymbol => enumSymbol is not null && IsAccessibleFromNamespace(enumSymbol));
+
+		context.RegisterSourceOutput(enumDeclarations, static (spc, enumSymbol) =>
```
pictos: Do not use Roslyn types here, they will break the incremental behavior. Use the pipeline steps to convert the info that you need into your own model. For more info look into Source Generators that we have done

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
```diff
@@ -0,0 +1,301 @@
...
+       sb.AppendLine("using CommunityToolkit.Maui.Converters;");
+		sb.AppendLine();
+
+		if (!string.IsNullOrEmpty(ns))
+		{
+			sb.AppendLine($"namespace {ns};");
+			sb.AppendLine();
+		}
+
+      sb.AppendLine($"internal static class {initializerClassName}");
```
pictos: please format this file

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
```diff
@@ -0,0 +1,301 @@
...
+			{
+				hash ^= c;
+				hash *= 16777619;
+			}
+
+			return hash.ToString("x8", System.Globalization.CultureInfo.InvariantCulture);
+		}
+	}
+
+	static bool IsAccessibleFromNamespace(INamedTypeSymbol enumSymbol)
```
pictos: can you refactor those methods into a helper class?

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
```diff
@@ -0,0 +1,301 @@
...
+			spc.AddSource(hintName, SourceText.From(code, Encoding.UTF8));
+		});
+	}
+
+	static string GenerateCode(INamedTypeSymbol enumSymbol)
+	{
+		string ns = enumSymbol.ContainingNamespace.ToDisplayString();
+		string enumName = enumSymbol.Name;
+		string enumQualifiedName = enumSymbol.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat);
+		string initializerClassName = $"{enumName}DescriptionInitializer_{GetStableHash(enumQualifiedName)}";
```
pictos: There's no need for this hash on the class name. I think using the full qualifier name is way better than this

pictos (changes requested):

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
BillyMartin1964: Done

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
BillyMartin1964: Done

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
BillyMartin1964: Done

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
BillyMartin1964: Done

BillyMartin1964:
Works great in debug, but crashing immediately in release. I think it may be my profiles. Can one of you try it in release mode, please?

src/CommunityToolkit.Maui/Converters/EnumDescriptionConverter.cs
BillyMartin1964: Done

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
BillyMartin1964: Done

BillyMartin1964:
@TheCodeTraveler , @pictos 
Reflection is the easiest, but won't work with AOT. The generator will work, but it needs to be a separate package that is a dependency of toolkit. I asked Copilot about it, and it said this:

1. Can the generator be put into the toolkit package?
Yes, but with a caveat:
•	You can include the source generator in the toolkit package.
•	However, the generator will only run on code inside the toolkit package itself (not on developer app code) unless you package it as an analyzer and reference it in the developer’s app project.
Best practice:
•	Ship the generator as a separate analyzer package (or as an analyzer in the toolkit package).
•	Ensure it is referenced in the developer’s app project so it can process developer enums.
---
2. Can the generator automatically register the enums?
Yes, if the generator runs in the developer’s app project:
•	The generator can scan all enums in the app project at build time.
•	It can generate code to register those enums and their descriptions automatically (e.g., via a module initializer or static constructor).
•	No manual registration is needed; it’s fully automatic.
---
3. Will there be a performance hit?
No significant runtime performance hit:
•	Source generators run at build time, not at runtime.
•	The generated code is just static registration (dictionary initialization, etc.).
•	At runtime, enum description lookup is a simple dictionary access—very fast and efficient.
Build time:
•	There is a small build-time cost for running the generator, but this is negligible for most projects.
AOT:
•	No reflection is used, so it’s fully compatible and fast in AOT scenarios.
---
Summary
•	Generator can be in the toolkit, but must run in the developer’s app project to see their enums.
•	Automatic registration is possible and recommended.
•	No runtime performance hit; dictionary lookup is fast.

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
```diff
@@ -0,0 +1,146 @@
...
+			   if (member.ResourceType is not null)
+			   {
+				   // Cannot safely resolve localized resource access at compile-time
+				   sb.AppendLine($"        dict[\"{member.Name}\"] = \"{EnumDescriptionGeneratorHelper.EscapeString(fallbackDescription)}\";");
+				   continue;
+			   }
+			   sb.AppendLine($"        dict[\"{member.Name}\"] = \"{EnumDescriptionGeneratorHelper.EscapeString(member.DisplayName)}\";");
+			   continue;
+		   }
+		   sb.AppendLine($"        dict[\"{member.Name}\"] = \"{EnumDescriptionGeneratorHelper.EscapeString(fallbackDescription)}\";");
```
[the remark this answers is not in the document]
TheCodeTraveler: @BillyMartin1964 - What are your thoughts on this comment from CoPilot?

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
```diff
@@ -0,0 +1,146 @@
...
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Text;
+
+namespace CommunityToolkit.Maui.Analyzers;
+
+[Generator]
+public class EnumDescriptionGenerator : IIncrementalGenerator
```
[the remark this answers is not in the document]
TheCodeTraveler: @BillyMartin1964 - Could you please fix the namespace?

src/CommunityToolkit.Maui.SourceGenerators/Generators/EnumDescriptionGenerator.cs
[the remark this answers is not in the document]
BillyMartin1964: To be honest, @TheCodeTraveler , I'm getting lost. After making all the changes that you guys wanted, it quit working. Copilot says that the generator is only generating for the enums that are actually in the Toolkit and not the enums in the user's app.

I know you're busy, but I could really use another set of eyes. This is so close to just throw it away now.

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -87,6 +86,18 @@
     <ProjectReference Include="..\..\src\CommunityToolkit.Maui.Maps\CommunityToolkit.Maui.Maps.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
pictos, line 89: can you revert those changes?

src/CommunityToolkit.Maui.SourceGenerators/Helpers/EnumDescriptionGeneratorHelper.cs
```diff
@@ -0,0 +1,175 @@
...
+			if (descAttr.ConstructorArguments[0].Value is string description && !string.IsNullOrWhiteSpace(description))
+			{
+				return description;
+			}
+		}
+
+		return null;
+	}
+
+	public static bool TryGetDisplayInfo(IFieldSymbol member, string displayAttributeName, out string? displayName, out INamedTypeSymbol? resourceType)
```
pictos, line 27: from the code, there are a lot of `if/else` to fill the `displayName` and `resourceType`. And with that I've a felling that those can be null even if it returns true. Can you add some assurance that the values will not be null if it returns true?

also, add `[NotNullWhen(true)]` on those `out` parameters.

src/CommunityToolkit.Maui.SourceGenerators/Models/EnumDescriptionRecords.cs
```diff
@@ -0,0 +1,17 @@
+using System.Collections.Generic;
+
+namespace CommunityToolkit.Maui.SourceGenerators.Models;
+
+public record EnumDescriptionModel(
+    string EnumName,
+    string Namespace,
+    string QualifiedName,
+    IReadOnlyList<EnumMemberModel> Members
```
pictos, line 9: For this kind of model you should use [EquatableArray](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.SourceGenerators/Helpers/EquatableArray%7BT%7D.cs).

Collections doesn't implement _value equals_ it uses the `ReferenceEquals`, which will always be false on your implementation, causing this generator to not be incremental.

PR #3169, opened 2026-03-20
Uri media source http headers

Kaaybi:
### Description of Change ###

Add HTTP headers support to MediaElement

 ### Linked Issues ###
 

 - Fixes #1889 and #3154 

 ### PR Checklist ###
 
 - [X] Has tests (if omitted, state reason in description)
 - [X] Has samples (if omitted, state reason in description)
 - [X] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/639

Thank you for reading through!

As documented, Tizen does not support custom HTTP headers.

I tested this on Windows with Mockoon. On HLS for instance, custom headers are propagated to both manifest and segments.
Feel free to test this by using the sample and setting custom headers.

screenshot2026-03-20 192916
screenshot2026-03-20 192949

### DIFF
--- a/src/CommunityToolkit.Maui.MediaElement/MediaSource/UriMediaSource.shared.cs
+++ b/src/CommunityToolkit.Maui.MediaElement/MediaSource/UriMediaSource.shared.cs
@@ -42,7 +42,7 @@ public sealed partial class UriMediaSource : MediaSource
 	/// </summary>
 	/// <remarks>
 	/// Use this to provide authentication tokens (e.g. <c>Authorization: Bearer &lt;token&gt;</c>) or other custom HTTP headers.
-	/// Setting this property triggers a source update on the underlying platform player.
+	/// Mutating the contents of the returned dictionary triggers a source update on the underlying platform player.
 	/// Not supported on Tizen.
 	/// </remarks>
 	public IDictionary<string, string> HttpHeaders { get; } = new ObservableDictionary<string, string>();


kaybi-gh:
@dotnet-policy-service agree

src/CommunityToolkit.Maui.MediaElement/MediaSource/UriMediaSource.shared.cs
```diff
@@ -39,6 +41,24 @@ public Uri? Uri
...
+	/// Gets or sets the HTTP headers to include in the request when loading the media from <see cref="Uri"/>.
+	/// </summary>
+	/// <remarks>
+	/// Use this to provide authentication tokens (e.g. <c>Authorization: Bearer &lt;token&gt;</c>) or other custom HTTP headers.
+	/// Setting this property triggers a source update on the underlying platform player.
+	/// Not supported on Tizen.
+	/// </remarks>
+	public IDictionary<string, string> HttpHeaders
+	{
+		get => httpHeaders ??= new Dictionary<string, string>();
```
pictos: I believe you can use the `field` keyword here instead

src/CommunityToolkit.Maui.MediaElement/MediaSource/MediaSource.shared.cs
```diff
@@ -78,6 +78,28 @@ internal event EventHandler SourceChanged
...
 	}
 
+	/// <summary>
+	/// Creates a <see cref="UriMediaSource"/> from an absolute URI with custom HTTP headers.
+	/// </summary>
+	/// <param name="uri">Absolute URI to load.</param>
+	/// <param name="httpHeaders">HTTP headers to include in the request (e.g. Authorization).</param>
+	/// <returns>A <see cref="UriMediaSource"/> instance.</returns>
+	/// <exception cref="ArgumentException">Thrown if <paramref name="uri"/> is not an absolute URI.</exception>
+	public static MediaSource? FromUri(Uri? uri, IDictionary<string, string>? httpHeaders)
```
pictos: I believe we should work with `IReadOnlyDictionary` instead of `IDictionary`

src/CommunityToolkit.Maui.MediaElement/Primitives/HttpRandomAccessStream.windows.cs
```diff
@@ -0,0 +1,113 @@
...
+		this.httpClient = httpClient;
+		this.requestUri = requestUri;
+		this.size = size;
+	}
+
+	/// <inheritdoc/>
+	public ulong Size
+	{
+		get => size;
+		set => throw new NotSupportedException("Cannot set size on a read-only HTTP stream.");
```
pictos, line 32: if you can't set the size why implement a setter?

src/CommunityToolkit.Maui.MediaElement/Views/HttpRandomAccessStream.windows.cs
```diff
@@ -0,0 +1,113 @@
...
+
+	/// <inheritdoc/>
+	public ulong Size
+	{
+		get => size;
+		set => throw new NotSupportedException("Cannot set size on a read-only HTTP stream.");
+	}
+
+	/// <inheritdoc/>
+	public ulong Position => currentPosition;
```
pictos: any reason to not use only the property?

```suggestion
	public ulong Position {get; private set;}
```

src/CommunityToolkit.Maui.MediaElement/Views/HttpRandomAccessStream.windows.cs
```diff
@@ -0,0 +1,113 @@
...
+	/// <inheritdoc/>
+	public IAsyncOperationWithProgress<IBuffer, uint> ReadAsync(IBuffer buffer, uint count, InputStreamOptions options)
+	{
+		return AsyncInfo.Run<IBuffer, uint>(async (cancellationToken, _) =>
+		{
+			using var request = new HttpRequestMessage(HttpMethod.Get, requestUri);
+			var rangeEnd = currentPosition + count - 1;
+			request.Headers.TryAppendWithoutValidation("Range", $"bytes={currentPosition}-{rangeEnd}");
+
+			using var response = await httpClient.SendRequestAsync(request, HttpCompletionOption.ResponseHeadersRead).AsTask(cancellationToken).ConfigureAwait(false);
```
pictos: change for this and remove the `ConfigureAwait(false)` for other methods
```suggestion
			using var response = await httpClient.SendRequestAsync(request, HttpCompletionOption.ResponseHeadersRead).AsTask(cancellationToken).ConfigureAwait(ConfigureAwaitOptions.ForceYielding);
```

src/CommunityToolkit.Maui.MediaElement/Primitives/HttpRandomAccessStream.windows.cs
```diff
@@ -0,0 +1,113 @@
...
+using HttpMethod = Windows.Web.Http.HttpMethod;
+using HttpCompletionOption = Windows.Web.Http.HttpCompletionOption;
+
+namespace CommunityToolkit.Maui.Core.Views;
+
+/// <summary>
+/// An <see cref="IRandomAccessStream"/> implementation backed by HTTP Range requests,
+/// enabling progressive streaming of media content with custom HTTP headers without buffering the entire file in memory.
+/// </summary>
+sealed partial class HttpRandomAccessStream : IRandomAccessStream
```
pictos, line 15: Not sure if this class should live inside `Views`

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
```diff
@@ -319,6 +332,78 @@ protected virtual async partial ValueTask PlatformUpdateSource()
...
 	}
 
+	async Task SetUriSourceWithHeaders(Uri uri, IDictionary<string, string> headers)
+	{
+		if (Player is null)
+		{
+			return;
+		}
+
+		var httpClient = new HttpClient();
```
pictos: we shouldn't create a new httpClient on every call, it should be reused

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
```diff
@@ -319,6 +332,78 @@ protected virtual async partial ValueTask PlatformUpdateSource()
...
+		Trace.WriteLine($"MediaElement [Windows]: Applying {headers.Count} custom HTTP header(s) to media request.");
+		foreach (var header in headers)
+		{
+			Trace.WriteLine($"MediaElement [Windows]: Header '{header.Key}' set.");
+			httpClient.DefaultRequestHeaders.TryAppendWithoutValidation(header.Key, header.Value);
```
pictos: and how about the unhappy path? Shouldn't we handle that?

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
```diff
@@ -319,6 +332,78 @@ protected virtual async partial ValueTask PlatformUpdateSource()
...
+		}
+
+		var adaptiveResult = await AdaptiveMediaSource.CreateFromUriAsync(uri, httpClient);
+		Trace.WriteLine($"MediaElement [Windows]: AdaptiveMediaSource status: {adaptiveResult.Status}");
+
+		if (adaptiveResult.Status is AdaptiveMediaSourceCreationStatus.Success && adaptiveResult.MediaSource is not null)
+		{
+			var adaptiveSource = adaptiveResult.MediaSource;
+			adaptiveSource.DownloadRequested += async (sender, args) =>
```
pictos: this will create a closure capture, which may cause a memory-leak. Can you refactor this in a way the delegate can unsubscribe?

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
```diff
@@ -319,6 +332,78 @@ protected virtual async partial ValueTask PlatformUpdateSource()
...
+			await Dispatcher.DispatchAsync(() =>
+			{
+				Player.AutoPlay = MediaElement.ShouldAutoPlay;
+				Player.Source = WinMediaSource.CreateFromStream(stream, string.Empty);
+			});
+			Trace.WriteLine("MediaElement [Windows]: HttpRandomAccessStream set successfully.");
+		}
+		catch (Exception ex)
+		{
+			Trace.WriteLine($"MediaElement [Windows]: HttpRandomAccessStream failed ({ex.Message}), falling back to SetUriSource without headers.");
```
pictos: wondering if we should throw this exception in a way that devs can catch and handle it in code

kaybi-gh:
Thank you pictos for your detailed review, I will address your points asap! 
edit: Done!

src/CommunityToolkit.Maui.MediaElement/MediaSource/UriMediaSource.shared.cs
kaybi-gh: Yes indeed, fixed!

src/CommunityToolkit.Maui.MediaElement/MediaSource/MediaSource.shared.cs
kaybi-gh: I considered this, but I kept `IDictionary<string, string>` intentionally for consistency: the `HttpHeaders` property on `UriMediaSource` is typed as `IDictionary<string, string>` because consumers may need to mutate headers in-place (e.g. source.HttpHeaders["Authorization"] = "Bearer token"). Since FromUri assigns directly to that property, using IReadOnlyDictionary for the parameter would require an unnecessary copy to convert. `Dictionary<TKey, TValue>` implements both interfaces anyway, so callers aren't restricted.

src/CommunityToolkit.Maui.MediaElement/Primitives/HttpRandomAccessStream.windows.cs
kaybi-gh: The setter is required by the `RandomAccessStream` interface contract, it defines `Size` as `{ get; set; }`. Since this is a read-only HTTP stream, throwing `NotSupportedException` in the setter is, _from what I've searched_, the standard .NET pattern for read-only implementations of read-write interface members (same pattern as `ReadOnlyCollection<T>.Add()` etc.).

src/CommunityToolkit.Maui.MediaElement/Views/HttpRandomAccessStream.windows.cs
kaybi-gh: No indeed, nice catch, fixed!

src/CommunityToolkit.Maui.MediaElement/Views/HttpRandomAccessStream.windows.cs
kaybi-gh: May I ask why? `ConfigureAwait(false)` seems to be the established convention the codebase for library code right?

src/CommunityToolkit.Maui.MediaElement/Primitives/HttpRandomAccessStream.windows.cs
kaybi-gh: Not sure either.. Though it's an internal sealed class exclusively used by `MediaManager.windows.cs`. I can move it elsewhere if needed!

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
kaybi-gh: Oh yes my bad, I totally agree. Fixed!

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
kaybi-gh: Yes you're right, I added `TryAppendWithoutValidation` value check and made a `Trace.WriteLine` on failure.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
kaybi-gh: Mh yes again that's right! The lambda is now a named method `OnAdaptiveMediaSourceDownloadRequested` and the `AdaptiveMediaSource` is stored as a field so the handler can be unsubscribed in both `PlatformUpdateSource` and `Dispose`.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
kaybi-gh: You're right, I initially assumed the fallback would provide a better user experience, but it should be a developer decision, especially when headers contain authentication tokens (the server would just return 401/403 anyway). I removed the fallback, the exception will now propagate and trigger `MediaFailed`.

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.windows.cs
kaybi-gh: Following the [discussion about not failing silently](https://github.com/CommunityToolkit/Maui/pull/3169#discussion_r2968412692), I finally now throw an `InvalidOperationException` instead of just logging with `Trace.WriteLine`, so developers get immediate feedback if a header name is invalid.

test-error.txt
```diff
@@ -0,0 +1,11 @@
+Unhandled exception. System.ArgumentException: An item with the same key has already been added. Key: Microsoft.Maui.Controls.BindableProperty
```
TheCodeTraveler: Please remove this file from the Pull Request

kaybi-gh: Oh yes sorry, I'll remove that asap

kaybi-gh: Done!

src/CommunityToolkit.Maui.MediaElement/MediaSource/UriMediaSource.shared.cs
```diff
@@ -39,6 +39,24 @@ public Uri? Uri
...
+	/// Use this to provide authentication tokens (e.g. <c>Authorization: Bearer &lt;token&gt;</c>) or other custom HTTP headers.
+	/// Setting this property triggers a source update on the underlying platform player.
+	/// Not supported on Tizen.
+	/// </remarks>
+	public IDictionary<string, string> HttpHeaders
+	{
+		get => field ??= new Dictionary<string, string>();
+		set
+		{
+			field = value;
```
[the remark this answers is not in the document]
kaybi-gh: Fixed!

src/CommunityToolkit.Maui.MediaElement/Primitives/HttpRandomAccessStream.windows.cs
```diff
@@ -0,0 +1,111 @@
...
+
+	/// <inheritdoc/>
+	public IAsyncOperationWithProgress<IBuffer, uint> ReadAsync(IBuffer buffer, uint count, InputStreamOptions options)
+	{
+		return AsyncInfo.Run<IBuffer, uint>(async (cancellationToken, _) =>
+		{
+			using var request = new HttpRequestMessage(HttpMethod.Get, requestUri);
+			var rangeEnd = Position + count - 1;
+			request.Headers.TryAppendWithoutValidation("Range", $"bytes={Position}-{rangeEnd}");
+
```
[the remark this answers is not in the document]
kaybi-gh: Fixed!

src/CommunityToolkit.Maui.MediaElement/Views/MediaManager.macios.cs
```diff
@@ -229,7 +230,26 @@ protected virtual async partial ValueTask PlatformUpdateSource()
...
+					Trace.WriteLine($"MediaElement [iOS/macCatalyst]: Applying {headers.Count} custom HTTP header(s) to AVUrlAsset.");
+					foreach (var header in headers)
+					{
+						Trace.WriteLine($"MediaElement [iOS/macCatalyst]: Header '{header.Key}' set.");
+					}
+
+					var nativeHeaders = NSDictionary.FromObjectsAndKeys(
+						headers.Values.ToArray(),
+						headers.Keys.ToArray());
+					var options = new NSDictionary("AVURLAssetHTTPHeaderFieldsKey", nativeHeaders);
```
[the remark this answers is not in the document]
kaybi-gh: Fixed!

src/CommunityToolkit.Maui.UnitTests/Views/MediaElement/UriMediaSourceTests.cs
```diff
@@ -0,0 +1,135 @@
...
+		var headers = new Dictionary<string, string>
+		{
+			["Authorization"] = "Bearer abc123"
+		};
+
+		var source = MediaSource.FromUri(uri, headers);
+
+		source.Should().BeOfType<UriMediaSource>();
+		var uriSource = (UriMediaSource)source!;
+		uriSource.Uri.Should().Be(uri);
```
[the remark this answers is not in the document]
kaybi-gh: Fixed!

src/CommunityToolkit.Maui.MediaElement/Primitives/HttpRandomAccessStream.windows.cs
```diff
@@ -0,0 +1,111 @@
...
+	public void Seek(ulong position) => Position = position;
+
+	/// <inheritdoc/>
+	public IRandomAccessStream CloneStream() => throw new NotSupportedException();
+
+	/// <inheritdoc/>
+	public IInputStream GetInputStreamAt(ulong position)
+	{
+		Position = position;
+		return this;
```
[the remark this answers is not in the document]
kaybi-gh: I looked into this, the only consumer here is `WinMediaSource.CreateFromStream`, which reads sequentially and doesn't call `CloneStream()`. I'd prefer to keep things simple for now and add this later if a real need comes up than rather push up dead code.

src/CommunityToolkit.Maui.MediaElement/Views/HttpRandomAccessStream.windows.cs
pictos: Sure, for context this library exists before the addition of `ConfigureAwaitOptions`, so that's used way more.

The way that `ConfigureAwait(false)` works, you just need it on the first method call that will complete asynchronous, all the others aren't needed. Adding a lot of them pollute the code. So adding the `.ConfigureAwait(ConfigureAwaitOptions.ForceYielding)` at first one is enough and all others `ConfigureAwait` can be deleted, since they'll change nothing

kaybi-gh: Thanks for taking time to explain! 🙏 
I've made the change 👌

src/CommunityToolkit.Maui.MediaElement/Primitives/HttpRandomAccessStream.windows.cs
pictos: probably we should have a folder called `primitives` or something like that and move this implementation

ne0rrmatrix: There is a Primitives directory and the `namespace` we are using in it is: `CommunityToolkit.Maui.Core`. Should be able to move it there.

TheCodeTraveler: Done ✅

PR #3188, opened 2026-04-06
Prevent race conditions in BindableProperty tests and benchmarks

### LINKED ISSUE #3187: [BUG] Tests are Fragile

### Current Behavior

Tests and builds have some fragility and and do not seem to repeatedly, reliably run with deterministic outcomes.

It looks like there are a couple of issues converging to cause this problem. The first is race conditions in tests and benchmarks, particularly around parallel benchmark runs that attempt to build artifacts that are locked by other runs. The second is that bindable properties created with the source generator were using `ConfigureAwait(false)` which combined with `async void` was exacerbating the above mentioned race conditions.

Note that I have investigated only the apparent failures and don't have a good understanding of the full context, so will appreciate some input and guidance here.

### Expected Behavior

Tests should deterministically pass or fail based on expected conditions, not ephemeral environmental conditions.

### Steps To Reproduce

1. Open a PR with some code changes
2. Allow tests and builds to run
3. Re-run tests and builds and compare results

https://github.com/CommunityToolkit/Maui

### Environment

```markdown
- .NET MAUI CommunityToolkit: 14.1.0
- OS: Windos and macOS
- .NET MAUI: 10.0.100
```

### Anything else?

_No response_

matt-goldman:
### Description of Change ###

 

This PR addresses two compounding issues in tests resulting in non-deterministic behaviour:

- Concurrency of builds in benchmarks
- Thread-pool context in `ValidationBehvaiour` property change handler

To resolve the first, build and run steps have been separated. The alternative would be to disallow parallelism which would add significant runtime. Separating the build allows multiple run instances without DLL locking issues.

The second is addressed by removing `ConfigurAwait(false)` from the `UpdateAsync` call in the property changed handler, as well as adding a `try`/`catch`.

Additionally, I added `try`/`finally` to the bindable property source generator when creating default values, to better mitigate these kinds of issues more broadly. For default value creation this should be safe as the issue is only caused by the static method, and throwing an exception on these failures is not a sensible default.

 ### Linked Issues ###
 

 - Fixes #3187 

 ### PR Checklist ###
 
 - [ ] Documentation created or updated: No documentation changes as it is not a public facing API change, and I don't think there are docs that cover this yet anyway.

### DIFF
--- a/src/CommunityToolkit.Maui.SourceGenerators/Generators/BindablePropertyAttributeSourceGenerator.cs
+++ b/src/CommunityToolkit.Maui.SourceGenerators/Generators/BindablePropertyAttributeSourceGenerator.cs
@@ -142,8 +142,8 @@ public class BindablePropertyAttributeSourceGenerator : IIncrementalGenerator
 
 		sb.Append(value.ClassInformation.DeclaredAccessibility).Append(" partial class ").Append(classNameWithGenerics).Append("\n{\n\n");
 
-		// Prepare helper builder for file-static class members (static flags + default creators)
-		var fileStaticClassStringBuilder = new StringBuilder(256);
+		// Prepare helper builder for nested static class members (default value creators only)
+		var helperClassStringBuilder = new StringBuilder(256);
 		var helperNames = new HashSet<string>();
 
 		// Build fully-qualified declaring type name for helper method casts (include containing types)
@@ -151,10 +151,19 @@ public class BindablePropertyAttributeSourceGenerator : IIncrementalGenerator
 			? classNameWithGenerics
 			: string.Concat(value.ClassInformation.ContainingTypes, ".", classNameWithGenerics);
 
-		var bindablePropertyInitHelpersClassName = $"__{value.ClassInformation.ClassName}BindablePropertyInitHelpers";
+		var bindablePropertyInitHelpersClassName = "__BindablePropertyInitHelpers";
 
 		foreach (var info in value.BindableProperties)
 		{
+			if (info.ShouldUsePropertyInitializer)
+			{
+				// Emit the initializing flag as an instance field on the partial class
+				if (helperNames.Add(info.InitializingPropertyName))
+				{
+					AppendInstanceInitializingField(sb, in info);
+				}
+			}
+
 			if (info.IsReadOnlyBindableProperty)
 			{
 				GenerateReadOnlyBindableProperty(sb, in info, bindablePropertyInitHelpersClassName);
@@ -164,33 +173,26 @@ public class BindablePropertyAttributeSourceGenerator : IIncrementalGenerator
 				GenerateBindableProperty(sb, in info, bindablePropertyInitHelpersClassName);
 			}
 
+			GenerateProperty(sb, in info, bindablePropertyInitHelpersClassName);
+
 			if (info.ShouldUsePropertyInitializer)
 			{
-				// Generate only references within the class; actual static field and creator method
-				// will be placed inside the file static helper class below.
-				if (helperNames.Add(info.InitializingPropertyName))
-				{
-					AppendHelperInitializingField(fileStaticClassStringBuilder, in info);
-				}
+				// Collect the CreateDefault method for the nested helper class
 				if (helperNames.Add(info.EffectiveDefaultValueCreatorMethodName))
 				{
-					AppendHelperDefaultValueMethod(fileStaticClassStringBuilder, in info, fullDeclaringType);
+					AppendHelperDefaultValueMethod(helperClassStringBuilder, in info, fullDeclaringType);
 				}
 			}
-
-			GenerateProperty(sb, in info, bindablePropertyInitHelpersClassName);
 		}
 
-		// If we generated any helper members and the declaring class is generic,
-		// emit the helper class nested inside the generated partial class so
-		// generic type parameters are in scope for casts used by the helper.
-		if (fileStaticClassStringBuilder.Length > 0 && !string.IsNullOrEmpty(value.ClassInformation.GenericTypeParameters))
+		// Always emit the helper class nested inside the partial class
+		if (helperClassStringBuilder.Length > 0)
 		{
 			sb.Append("[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]");
 			sb.Append("\n");
-			sb.Append("private static class ").Append(bindablePropertyInitHelpersClassName).Append("\n{");
+			sb.Append("static class ").Append(bindablePropertyInitHelpersClassName).Append("\n{");
 			sb.Append("\n");
-			sb.Append(fileStaticClassStringBuilder.ToString());
+			sb.Append(helperClassStringBuilder.ToString());
 			sb.Append("}\n\n");
 		}
 
@@ -206,15 +208,6 @@ public class BindablePropertyAttributeSourceGenerator : IIncrementalGenerator
 			}
 		}
 
-		// If we generated any helper members and the declaring class is not generic,
-		// emit a file static class with them. Generic types have their helpers emitted
-		// nested inside the class above to ensure type parameter scope.
-		if (fileStaticClassStringBuilder.Length > 0 && string.IsNullOrEmpty(value.ClassInformation.GenericTypeParameters))
-		{
-			sb.Append("\n\nfile static class ").Append(bindablePropertyInitHelpersClassName).Append("\n{\n");
-			sb.Append(fileStaticClassStringBuilder.ToString());
-			sb.Append("}\n");
-		}
 
 		return sb.ToString();
 	}
@@ -346,8 +339,8 @@ public class BindablePropertyAttributeSourceGenerator : IIncrementalGenerator
 		{
 			if (info.ShouldUsePropertyInitializer)
 			{
-				// Now reference the static flag on the file static helper class
-				sb.Append(bindablePropertyInitHelpersClassName).Append(".").Append(info.InitializingPropertyName);
+				// Reference the instance field directly on the partial class
+				sb.Append(info.InitializingPropertyName);
 			}
 			else
 			{
@@ -591,48 +584,52 @@ public class BindablePropertyAttributeSourceGenerator : IIncrementalGenerator
 	}
 
 	/// <summary>
-	/// Appends the initializing flag into the file-static helper class.
+	/// Appends the initializing flag as an instance field on the generated partial class.
+	/// Using an instance field avoids cross-thread / cross-instance interference that
+	/// occurred when the flag was static.
 	/// </summary>
-	/// <param name="fileStaticClassStringBuilder">Helper StringBuilder used to collect helper members.</param>
+	/// <param name="sb">StringBuilder for the partial class body.</param>
 	/// <param name="info">Property model.</param>
-	static void AppendHelperInitializingField(StringBuilder fileStaticClassStringBuilder, in BindablePropertyModel info)
+	static void AppendInstanceInitializingField(StringBuilder sb, in BindablePropertyModel info)
 	{
-		// Use [ThreadStatic] so each thread has its own flag, preventing cross-thread
-		// interference when multiple instances resolve default values concurrently.
-		fileStaticClassStringBuilder.Append("[global::System.ThreadStatic]\npublic static bool ")
+		sb.Append("[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]\n")
+			.Append("bool ")
 			.Append(info.InitializingPropertyName)
 			.Append(";\n");
 	}
 
 	/// <summary>
-	/// Appends a default value creator method into the file-static helper class.
-	/// The method sets the thread-static initializing flag, reads the property's initializer value by casting the bindable
-	/// to the declaring type, then clears the flag and returns the value.
+	/// Appends a default value creator method into the nested helper class.
+	/// The method sets the instance-level initializing flag on the bindable object, reads the property's initializer value
+	/// by casting the bindable to the declaring type, then clears the flag and returns the value.
 	/// </summary>
-	/// <param name="fileStaticClassStringBuilder">Helper StringBuilder used to collect helper members.</param>
+	/// <param name="helperClassStringBuilder">Helper StringBuilder used to collect helper methods.</param>
 	/// <param name="info">Property model.</param>
 	/// <param name="fullDeclaringType">Declaring type including containing types and generic parameters.</param>
-	static void AppendHelperDefaultValueMethod(StringBuilder fileStaticClassStringBuilder, in BindablePropertyModel info, string fullDeclaringType)
+	static void AppendHelperDefaultValueMethod(StringBuilder helperClassStringBuilder, in BindablePropertyModel info, string fullDeclaringType)
 	{
 		var sanitizedPropertyName = IsDotnetKeyword(info.PropertyName) ? string.Concat("@", info.PropertyName) : info.PropertyName;
 
-		fileStaticClassStringBuilder.Append("public static object ")
+		helperClassStringBuilder.Append("public static object ")
 			.Append(info.EffectiveDefaultValueCreatorMethodName)
 			.Append("(global::Microsoft.Maui.Controls.BindableObject bindable)\n")
 			.Append("{\n")
+			.Append("((")
+			.Append(fullDeclaringType)
+			.Append(")bindable).")
 			.Append(info.InitializingPropertyName)
 			.Append(" = true;\n")
-			.Append("try\n{\n")
 			.Append("var defaultValue = ((")
 			.Append(fullDeclaringType)
 			.Append(")bindable).")
 			.Append(sanitizedPropertyName)
 			.Append(";\n")
-			.Append("return defaultValue;\n")
-			.Append("}\nfinally\n{\n")
+			.Append("((")
+			.Append(fullDeclaringType)
+			.Append(")bindable).")
 			.Append(info.InitializingPropertyName)
 			.Append(" = false;\n")
-			.Append("}\n")
+			.Append("return defaultValue;\n")
 			.Append("}\n\n");
 	}
 }
--- a/src/CommunityToolkit.Maui.SourceGenerators/Generators/BindablePropertyAttributeSourceGenerator.cs
+++ b/src/CommunityToolkit.Maui.SourceGenerators/Generators/BindablePropertyAttributeSourceGenerator.cs
@@ -597,15 +597,16 @@ public class BindablePropertyAttributeSourceGenerator : IIncrementalGenerator
 	/// <param name="info">Property model.</param>
 	static void AppendHelperInitializingField(StringBuilder fileStaticClassStringBuilder, in BindablePropertyModel info)
 	{
-		// Make the flag public static so it can be referenced from the generated partial class in the same file.
-		fileStaticClassStringBuilder.Append("public static volatile bool ")
+		// Use [ThreadStatic] so each thread has its own flag, preventing cross-thread
+		// interference when multiple instances resolve default values concurrently.
+		fileStaticClassStringBuilder.Append("[global::System.ThreadStatic]\npublic static bool ")
 			.Append(info.InitializingPropertyName)
-			.Append(" = false;\n");
+			.Append(";\n");
 	}
 
 	/// <summary>
 	/// Appends a default value creator method into the file-static helper class.
-	/// The method sets the static initializing flag, reads the property's initializer value by casting the bindable
+	/// The method sets the thread-static initializing flag, reads the property's initializer value by casting the bindable
 	/// to the declaring type, then clears the flag and returns the value.
 	/// </summary>
 	/// <param name="fileStaticClassStringBuilder">Helper StringBuilder used to collect helper members.</param>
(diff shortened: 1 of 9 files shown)


TheCodeTraveler:
Hey Matt! You kinda came to the same conclusion that I did when I was playing around with this a few months ago, adding a try/catch block in `ValidationBehavior`.

I decided not to go this route because I don't want to catch _every_ exception. Plus, it's more of a work-around than a fix of the root cause.

However, I do kinda like CopIlot's suggestion. We could probably set `Options.ShouldSuppressExceptionsInBehaviors = true` for our Unit Tests that aren't expecting an Exception.

matt-goldman:
> Hey Matt! You kinda came to the same conclusion that I did when I was playing around with this a few months ago.
> 
> I decided not to go this route because I don't want to catch _every_ exception. Plus, it's more of a work-around than a fix of the root cause.
> 
> However, I do kinda like CopIlot's suggestion. We could probably set `Options.ShouldSuppressExceptionsInBehaviors = true` for our Unit Tests that aren't expecting an Exception.

Absolutely agree! I wanted to wait for the checks to finish before I applied it though to confirm they're deterministic though.

Should we be worried about that benchmarks run going for 35 minutes? I've been checking on the output and it does seem to be churning through, but not sure it should be taking that long.

TheCodeTraveler:
Sounds good! 

> Should we be worried about that benchmarks run going for 35 minutes?

Nah, don't worry about those. We just use those to generate the test benchmark artifacts. The Benchmark Pipeline completing isn't required for merging PRs.

TheCodeTraveler:
Thanks @matt-goldman! Are you confident that both adding `[ThreadStatic]` and removing `.ConfigureAwait(false)` are needed?

What are the results of only removing `.ConfigureAwait(false)`? And what are the results of only adding `[ThreadStatic]`?

I haven't tested each individually, however these two changes, along with adding the try/catch block, seem more akin to a "throw everything at the wall and see what sticks" strategy. 

Is it possible to be more surgical and strategic? Are all 3 changes necessary? What did you find is the root cause of the problem causing the failing unit tests?

matt-goldman:
> Thanks @matt-goldman! Are you confident that both adding `[ThreadStatic]` and removing `.ConfigureAwait(false)` are needed?
> 
> What are the results of only removing `.ConfigureAwait(false)`? And what are the results of only adding `[ThreadStatic]`?
> 
> I haven't tested each individually, however these two changes, along with adding the try/catch block, seem more akin to a "throw everything at the wall and see what sticks" strategy.
> 
> Is it possible to be more surgical and strategic? Are all 3 changes necessary? What did you find is the root cause of the problem causing the failing unit tests?

Hey @TheCodeTraveler, sorry slightly long-winded response here! (TLDR at the bottom).

I think adding both `[ThreadStatic]` and removing `ConfigureAwait(false)` are correct for this case. Removing the try/catch though is probably valid, it masks symptoms (in the source generator I mean; I think we should keep it in the `ValidationBehavior`, especially with Copilot's suggestion applied).

In terms of root cause, the problem is race conditions in the tests, which while less likely to appear in real world use, are still possible. To clarify what I mean there, the race conditions themselves aren't the problem but they are guaranteed to occur in tests while unlikely to occur in real world use.

Looking at the stack trace from the failing tests we can see:

```bash
Unhandled exception. System.ArgumentException: An item with the same key has already been added. Key: Microsoft.Maui.Controls.BindableProperty
at System.Collections.Generic.Dictionary2.TryInsert(TKey key, TValue value, InsertionBehavior behavior) at Microsoft.Maui.Controls.BindableObject.GetValue(BindableProperty property) at CommunityToolkit.Maui.Behaviors.ValidationBehavior.get_IsValid() in /_/src/CommunityToolkit.Maui/obj/Release/net10.0/CommunityToolkit.Maui.SourceGenerators/CommunityToolkit.Maui.SourceGenerators.BindablePropertyAttributeSourceGenerator/ValidationBehavior.g.cs:line 12 at CommunityToolkit.Maui.Behaviors.ValidationBehavior.UpdateStateAsync(VisualElement view, ValidationFlags flags, Boolean isForced, Nullable1 parentToken) in //src/CommunityToolkit.Maui/Behaviors/Validators/ValidationBehavior.shared.cs:line 333
at CommunityToolkit.Maui.Behaviors.ValidationBehavior.OnValidationPropertyChanged(BindableObject bindable, Object oldValue, Object newValue) in //src/CommunityToolkit.Maui/Behaviors/Validators/ValidationBehavior.shared.cs:line 237
at System.Threading.Tasks.Task.<>c.<throwasync>b__124_1(Object state)
at System.Threading.QueueUserWorkItemCallback.Execute()
at System.Threading.ThreadPoolWorkQueue.Dispatch()
at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
at System.Threading.Thread.StartCallback()
Test application process didn't exit gracefully, exit code is '7'</throwasync>
Error: Process completed with exit code 1.
 ```

We can see it's failing on `GetValue`. In `BindableObject` this goes down the chain through get or create logic, which eventually hits the `_properties` dictionary (see: https://github.com/dotnet/maui/blob/b43bdad1449ecb3f95ed0d4fe19c607a31467d52/src/Controls/src/Core/BindableObject.cs#L42). `Dictionary` is not thread safe, so having two calls to `GetValue` for a `BindableProperty` which hasn't been initialised yet will result in both trying to insert the value, which leads to the error that an item with the same key already exists. And we can see from the stack trace it has called `System.Threading.ThreadPoolWorkQueue.Dispatch()`.

I'll defer to you on this one (because you're literally the expert here!) but I believe `ConfigureAwait(false)` shouldn't be used on UI bound operations, which is fundamentally what a bindable property is. So I think that removing `ConfigureAwait(false)` is the correct approach.

Regarding `[ThreadStatic]`, that's to properly handle the initialization flag. `IsInitializingIsValid` for example is generated as static in a helper class, but `GetValue` on `BindableProperty` is not, so after the first call on any instance, any other instance, which may not have been initialised, will get `true` and will attempt to get the value from the field. Or the alternative case we seem to be hitting here is that multiple concurrent calls see it as _not_ initialised and attempt to insert the same default value.

Regarding more surgical test results:

- I removed the try/catch from the generated code altogether (but kept it in the `ValidationBehavior`, with the additional flag as suggested by Copilot)
- Keeping `ConfigureAwait(false)` but adding `[ThreadStatic]` allows all tests to pass
- Removing `ConfigureAwait(false)` and not adding `[ThreadStatic]` still results in failing tests

You can see how these results relate to different commits on my fork here: https://github.com/matt-goldman/Maui-1/pull/2

However, I still think both are correct changes for different reasons and in different contexts. Essentially `ConfigureAwait(false)` is correct for a UI bound operation (although in the tests there's no synchronisation context which is why this alone doesn't fix it), and fixes any consumers of the generated bindable property, and `[ThreadStatic]` prevents default value bleeding across threads. The original code had the `volatile` keyword which I think was intended to solve this problem, but I think it was actually making it worse. `volatile` guarantees that calls across different threads can see a write whereas what we need here is the opposite, ensuring calls from different threads only see their own.

---

With all that being said, I actually think there's another approach worth considering. Why not make the init flag a field on the bindable object rather than the helper? The instance is already passed to the `CreateDefault` method (as the `bindable` arg). In that case this generated code:

```csharp
public static volatile bool IsInitializingText = false;
public static object CreateDefaultText(global::Microsoft.Maui.Controls.BindableObject bindable)
{
	IsInitializingText = true;
	var defaultValue = ((TestView)bindable).Text;
	IsInitializingText = false;
	return defaultValue;
}
```

it could just be:

```csharp
public static object CreateDefaultText(global::Microsoft.Maui.Controls.BindableObject bindable)
{
	var instance = (TestView)bindable;
	instance._isInitializingText = true;
	var defaultValue = instance.Text;
	instance._isInitializingText = false;
	return defaultValue;
}
```

This pretty much gets rid of any need for volatile or ThreadStatic altogether. You could also even remove the helper classes and just have this method in the generated partial class.

---

**TL;DR:**
So I guess short answer here is:

1. Removing `ConfigureAwait(false)` doesn't fix the tests but is correct for UI bound operations
2. Removing `try`/`catch` is from the generated code is probably the right call to not mask errors
3. Adding `[ThreadStatic]` fixes the issue because each instance needs its own init flag
4. We could eliminate 3 as an issue altogether rather than fixing it by making the flag an instance field instead of a static property on a helper class

For that alternative approach, I haven't tested it yet, just thinking out loud. Before I go down that rabbit hole I wanted to run it by you as there's obviously a lot of decision history and context that I don't have.

What do you think?

TheCodeTraveler:
> With all that being said, I actually think there's another approach worth considering. Why not make the init flag a field on the bindable object rather than the helper? The instance is already passed to the CreateDefault method (as the bindable arg). In that case this generated code:

You read my mind! I had the same thought when reviewing the `[ThreadStatic]` code. I think this is the fix for the root cause. 

To do it, let's nest `static class __BindablePropertyInitHelpers` inside the generated `partial class` with `EditorBrowsableState.Never`, and then make `bool __isInitializingText = false;` a field with `EditorBrowsableState.Never` also in the generated `partial class`. 

```cs
public partial class TestClass : View
{
	[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
	bool __isInitializingText = false;
	
	/// <summary>
	/// BindableProperty for the <see cref = "Text"/> property.
	/// </summary>
	public static readonly global::Microsoft.Maui.Controls.BindableProperty TextProperty = global::Microsoft.Maui.Controls.BindableProperty.Create("Text", typeof(string), typeof(TestClass), null, (Microsoft.Maui.Controls.BindingMode)0, null, null, null, null, __BindablePropertyInitHelpers.CreateDefaultText);

	public partial string Text
	{
		get => __isInitializingText ? field : (string)GetValue(TextProperty);
		set => SetValue(TextProperty, value);
	}
	
	[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
	static class __BindablePropertyInitHelpers
	{
		public static object CreateDefaultText(global::Microsoft.Maui.Controls.BindableObject bindable)
		{
			((TestClass)bindable).__isInitializingText = true;
			var defaultValue = ((TestClass)bindable).Text;
			((TestClass)bindable).__isInitializingText = false;
			return defaultValue;
		}
	}
}
```

matt-goldman:
> > With all that being said, I actually think there's another approach worth considering. Why not make the init flag a field on the bindable object rather than the helper? The instance is already passed to the CreateDefault method (as the bindable arg). In that case this generated code:
> 
> You read my mind! I had the same thought when reviewing the `[ThreadStatic]` code. I think this is the fix for the root cause.
> 
> To do it, let's nest `static class __BindablePropertyInitHelpers` inside the generated `partial class` with `EditorBrowsableState.Never`, and then make `bool __isInitializingText = false;` a field with `EditorBrowsableState.Never` also in the generated `partial class`.
> 
> ```cs
> public partial class TestClass : View
> {
> 	[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
> 	bool __isInitializingText = false;
> 	
> 	/// <summary>
> 	/// BindableProperty for the <see cref = "Text"/> property.
> 	/// </summary>
> 	public static readonly global::Microsoft.Maui.Controls.BindableProperty TextProperty = global::Microsoft.Maui.Controls.BindableProperty.Create("Text", typeof(string), typeof(TestClass), null, (Microsoft.Maui.Controls.BindingMode)0, null, null, null, null, __BindablePropertyInitHelpers.CreateDefaultText);
> 
> 	public partial string Text
> 	{
> 		get => __isInitializingText ? field : (string)GetValue(TextProperty);
> 		set => SetValue(TextProperty, value);
> 	}
> 	
> 	[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
> 	static class __BindablePropertyInitHelpers
> 	{
> 		public static object CreateDefaultText(global::Microsoft.Maui.Controls.BindableObject bindable)
> 		{
> 			((TestClass)bindable).__isInitializingText = true;
> 			var defaultValue = ((TestClass)bindable).Text;
> 			((TestClass)bindable).__isInitializingText = false;
> 			return defaultValue;
> 		}
> 	}
> }
> ```

Awesome @TheCodeTraveler ! I'll take a look at this over the weekend :)

src/CommunityToolkit.Maui.SourceGenerators/Generators/BindablePropertyAttributeSourceGenerator.cs
```diff
@@ -591,41 +584,49 @@ static string GetFormattedReturnType(ITypeSymbol typeSymbol)
...
-	/// <param name="fileStaticClassStringBuilder">Helper StringBuilder used to collect helper members.</param>
+	/// <param name="sb">StringBuilder for the partial class body.</param>
 	/// <param name="info">Property model.</param>
-	static void AppendHelperInitializingField(StringBuilder fileStaticClassStringBuilder, in BindablePropertyModel info)
+	static void AppendInstanceInitializingField(StringBuilder sb, in BindablePropertyModel info)
 	{
-		// Make the flag public static so it can be referenced from the generated partial class in the same file.
-		fileStaticClassStringBuilder.Append("public static volatile bool ")
+		sb.Append("[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]\n")
+			.Append("bool ")
```
TheCodeTraveler, line 596: Let's keep `volatile`


```suggestion
			.Append("volatile bool ")
```

TheCodeTraveler (commented):
This is looking good!! Assuming the unit tests keep passing, I'm ready to approve + merge this once we add the `volatile` keyword.

src/CommunityToolkit.Maui.SourceGenerators/Generators/BindablePropertyAttributeSourceGenerator.cs
```diff
@@ -591,41 +584,49 @@ static string GetFormattedReturnType(ITypeSymbol typeSymbol)
...
-		fileStaticClassStringBuilder.Append("public static object ")
+		helperClassStringBuilder.Append("public static object ")
 			.Append(info.EffectiveDefaultValueCreatorMethodName)
 			.Append("(global::Microsoft.Maui.Controls.BindableObject bindable)\n")
 			.Append("{\n")
+			.Append("((")
+			.Append(fullDeclaringType)
+			.Append(")bindable).")
 			.Append(info.InitializingPropertyName)
 			.Append(" = true;\n")
```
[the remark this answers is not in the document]
TheCodeTraveler: Nah, we don't want the overhead of a try/finally block. The scenario of property initializer throwing an Exception is such a small edge case that the additive cost of a try/finally block isn't worth it (or really, necessary).

src/CommunityToolkit.Maui.UnitTests/Views/LazyView/LazyViewTests.cs
```diff
@@ -20,7 +20,10 @@ public async Task LoadViewAsync_CancellationTokenExpired()
 		var lazyView = new LazyView<Button>();
 
 		// Ensure CancellationToken Expired
-		await Task.Delay(100, TestContext.Current.CancellationToken);
+		while (!cts.Token.IsCancellationRequested)
+		{
+			await Task.Delay(1, TestContext.Current.CancellationToken);
+		}
```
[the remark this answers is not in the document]
TheCodeTraveler: Irrelevant and completely unnecessary

src/CommunityToolkit.Maui.SourceGenerators/Generators/BindablePropertyAttributeSourceGenerator.cs
matt-goldman: Happy to do that, but seems redundant now that it's not `static` doesn't it?

matt-goldman:
Hey @TheCodeTraveler, I'll work through the Copilot feedback, happy to add `volatile` back in too if that's your preference.

I was thinking about it and there are a couple of things I would probably have done differently:

* The nested class feels unnecessary. The helper methods could just live in the top level generated class, maybe it's a stylistic preference but they are directly manipulated instance fields rather than being extensions, and it feels to me like the additional layer adds more complexity and I'm not sure what the benefit is.
* I added the `EditorBrowsableState.Never` as you suggested, but it worries me a little bit. The double underscore is a good convention to signal this is source generated and not something you need to worry about or touch as a consumer, but making it not discoverable or hidden might make troubleshooting or debugging more difficult.

Minor and stylistic obviously so happy to go with what you prefer. And of course follow the styles and patterns of the library :) But just wondering what your thoughts are here?

TheCodeTraveler:
Dang. The unit tests have failed in CI Pipeline with the same error.

matt-goldman:
Huh. Weird. I'm re-running on my fork now and as soon as I get the chance I'll take another look at this. Stepping through locally obviously won't help, but what I might do is see if I can pass a test ID through and log it so we can see where the collision occurs. But obviously this has failed my main goal of making these test runs deterministic.

Stand by for updates!

EDIT: Fails with unrelated issue on my fork. Windows build passes, macOS fails a different test entirely.

I'm going to update my main branch, then rebase this branch and hard reset to a common ancestor, then re-introduce these changes just to eliminate any merge issues. Then I'll add the tracing I mentioned above. I'll also see if I can figure out why these are not reliably giving the same results every time.

matt-goldman:
Hey @TheCodeTraveler, I've got another approach I'm trying out at the moment. I think the problem is that we're still hitting a dictionary in the underlying .NET MAUI code, and with no synchronization context we're getting the same race condition, irrespective of what we do here. No matter what we do with the initialization flag, we're still going to hit `GetValue` in `BindableProperty`. In the typical/intended use case that's not a problem (UI and sync context). But `defaultValueCreator` introduces a lazy init, which under the build and test conditions, with concurrent access, surfaces this issue.

We can try to anticipate every possible use case and protect against these issues, but perhaps it's reasonable to put some level of responsibility onto consumers. The easiest approach is to just run the tests consecutively and document that generated code is not thread safe, but I think there's a reasonable middle ground.

The current logic is a universal default value initializer, but it's not really necessary. For any property where the initialiser can be evaluated automatically, we can just use the default value. Passing the initializer directly instead of using `defaultValueCreator` eliminates the lazy init for nearly all properties, and for everything else (which is pretty much just collection expressions and lambdas) we can still keep the option to provide an initialisation method by name.

If we document that, it would be up to the consumer to understand that if they're using a mutable reference type and using the source generator to generate the bindable properties for it, and _don't_ provide an initializer, then _if_ they're running in a non-threadsafe context (i.e., essentially not using it for the intended use case), they _might_ hit a problem. Later we could also add an analyzer that emits a warning if you're using the source generator for a mutable reference type and don't supply an initializer. 

That's a lot of ifs and buts and maybes, which is why I think trying to defensively protect against any hypothetical scenario is unnecessary. Hand written bindable properties don't hit the problem because you would declare a concrete initial value. For the properties in the library and the tests where this would be an issue, we provide a sensible default value (e.g. something like `defaultValue: Enumerable.Empty<View>()` instead of inferring from `=[]`).

I'm testing this now, and it still needs some refinement, but before I go too far down this rabbit hole, just wondering what you think?

In short, I'm suggesting get rid of the initialisation flags, get rid of the automatic default value creator by default, infer the default value, and document the cases where the consumer will _potentially_ hit thread safety problems if they don't specify a sensible instance default.

src/CommunityToolkit.Maui.SourceGenerators/Generators/BindablePropertyAttributeSourceGenerator.cs
TheCodeTraveler: > Happy to do that, but seems redundant now

Thanks! It's still valid. Multiple threads can access the BindableProperty simultaneously and `volatile` prevents .NET from cacheing its value. This ensures that each thread does not used a cached, out-of-date, value of the `bool`.

TheCodeTraveler:
Hey Matt! I can't exactly picture what you're suggesting, but I'm happy to review the code once you've updated the PR. 

> The easiest approach is to just run the tests consecutively and document that generated code is not thread safe

I've already disabled parallelization via [`xunit.runnner.json`](https://github.com/CommunityToolkit/Maui/blob/main/src/CommunityToolkit.Maui.UnitTests/xunit.runner.json).  Here's the link to those xUnit docs: https://xunit.net/docs/config-xunit-runner-json

Looking at additional [xUnit docs](https://xunit.net/docs/running-tests-in-parallel.html#changing-default-behavior), perhaps we also need to add `[assembly: CollectionBehavior(MaxParallelThreads = 1)]`?

> The current logic is a universal default value initializer, but it's not really necessary. For any property where the initialiser can be evaluated automatically, we can just use the default value

I am assuming you're suggesting passing the partial property initializer into `BindableProperty.DefaultValue`? This is only allowed when the partial property initializer is a static or a const.  

I'm not sure that it's worth the added complexity to `BindablePropertySourceGenerator` to support two different scenarios here, given that we still need to support the scenario where a user uses a static method as a partial property initializer: 

```cs
[BindableProperty]
public ICommand Command { get; set; } = CreateCommand();
```

That said, I'm happy to take a look at your implementation!

matt-goldman:
@TheCodeTraveler ok great :) I've now re-run these pipelines on my fork and I've gotten 5x consecutive, consistently successful build and test runs across all workflows. I'm getting intermittent failures on benchmarks due to file lock/concurrency issues; I'm happy to include a fix for that too but as it's a separate issue want to check with you.

Re your question, I'm not removing `defaultValueCreator`, just adding logic to resolve the default value directly. It's still there as the fallback for anything the generator can't resolve at compile time (like `CreateCommand()`, `= []`, `lambdas`). The change is just that for initializers we can resolve (`literals`, enum values, `TimeSpan.FromSeconds(30)`, `string.Empty`, etc.), we pass them directly as `defaultValue:` instead of wrapping them in a `defaultValueCreator`. Unresolvable expressions fall through to the existing pattern.

I've been testing this on a different branch and a different PR back to my own main so I can run the tests. I can update this branch and PR or could open another so you can compare side by side, any preference on that?

TheCodeTraveler:
Awesome!! Yea, let's open a new PR to start fresh and close this one

matt-goldman:
Superseded by #3204

PR #2841, opened 2025-08-03
Initial attempt at platform based processing from the CameraView

bijington:
### Description of Change ###

 

 ### Linked Issues ###
 

 - Fixes #

 ### PR Checklist ###
 
 - [ ] Has a linked Issue, and the Issue has been `approved`(bug) or `Championed` (feature/proposal)
 - [ ] Has tests (if omitted, state reason in description)
 - [ ] Has samples (if omitted, state reason in description)
 - [ ] Rebased on top of `main` at time of PR
 - [ ] Changes adhere to [coding standard](https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#contributing-code---best-practices)
 - [ ] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pulls 

 ### Additional information ###

PR #3274, opened 2026-07-26
Introduce the ability to extend camera functionality

bijington:
## Current status

This it the remaining TODO list

- [ ] test out multiple scenarios together
- [ ] Tested on macOS for both - Shared page crashes
- [ ] Needs to be built and tested on Android
- [ ] Needs to be built and tested on Windows

Introduce the ability to extend camera functionality for concepts such as Barcode scanning.

Replaces the original PR attempt at #2841.

 ### Description of Change ###

 

### Camera Scenarios Usage Guide

The Community Toolkit for MAUI CameraView allows extending its functionality through "Scenarios". There are two primary ways to implement a scenario: `PlatformCameraScenario` and `FrameBasedCameraScenario`.

#### PlatformCameraScenario

`PlatformCameraScenario` is designed for scenarios where you want to leverage platform-specific high-performance APIs (e.g., Google ML Kit on Android, Apple Vision framework on iOS, or Windows Media Capture APIs).

##### When to use it
- When performance is critical.
- When you want to use a feature already provided by the operating system.
- When you need fine-grained control over the platform-specific camera pipeline.

##### Implementation Example (Windows)
In the sample project, `PlatformBarcodeScanningScenario` on Windows uses the `MediaFrameReader` to capture frames and decode them using `ZXing.Net` within the platform-specific code.

```csharp
// Platforms/Windows/PlatformBarcodeScanningScenario.cs
public class PlatformBarcodeScanningScenario : PlatformCameraScenario
{
    // ... Windows specific implementation using MediaFrameReader
}
```

#### FrameBasedCameraScenario

`FrameBasedCameraScenario` is a specialized version of `PlatformCameraScenario` that abstracts the frame capture process. It provides a platform-agnostic `OnFrameReceived` method that gives you access to a `CameraFrame` containing raw byte data.

##### When to use it
- When you want to share the same processing logic across all platforms.
- When you have a cross-platform library (like `ZXing.Net`) that can process raw image data.
- When you don't need the absolute maximum performance of platform-native ML frameworks.

##### Implementation Example (Shared)
In the sample project, `SharedBarcodeScanningScenario` demonstrates how to implement barcode scanning in shared code.

```csharp
// SharedBarcodeScanningScenario.cs
public class SharedBarcodeScanningScenario : FrameBasedCameraScenario
{
    public override void OnFrameReceived(CameraFrame frame)
    {
        // frame.Data contains the raw byte array
        // frame.Width and frame.Height provide dimensions
        
        var luminanceSource = new RGBLuminanceSource(frame.Data, frame.Width, frame.Height);
        var result = barcodeReader.Decode(luminanceSource);

        if (result is not null)
        {
            // Handle result
        }
    }
}
```

#### Key Differences

| Feature | PlatformCameraScenario | FrameBasedCameraScenario |
| --- | --- | --- |
| **Logic Location** | Platform-specific folders | Shared code |
| **Performance** | Highest (Native APIs) | Good (Shared processing) |
| **Complexity** | Higher (Needs platform knowledge) | Lower (Abstraction handled for you) |
| **Data Access** | Native objects (e.g., `ImageProxy`, `CMSampleBuffer`) | `CameraFrame` (Raw bytes) |

#### Usage in XAML

Both types of scenarios are used in the same way with the `CameraView`.

```xml
<toolkit:CameraView>
    <toolkit:CameraView.Scenarios>
        <!-- Use either a platform-specific scenario -->
        <sample:PlatformBarcodeScanningScenario Command="{Binding OnCodeDetectedCommand}" />
        
        <!-- Or a shared frame-based scenario -->
        <sample:SharedBarcodeScanningScenario Command="{Binding OnCodeDetectedCommand}" />
    </toolkit:CameraView.Scenarios>
</toolkit:CameraView>
```

 ### Linked Issues ###
 

 - Fixes #

 ### PR Checklist ###
 
Example of it working on macOS

https://github.com/user-attachments/assets/025229f6-54da-49c2-808e-677afe09dc04


src/CommunityToolkit.Maui.Camera/CameraManager.tizen.cs
CliffAgius, line 1: Are we still supporting Tizen? it's not possible to build a MAUI app on tizen as it's stuck on Net6...

bijington: Good point! I needed to patch this for the builds to pass but it won't do anything

src/CommunityToolkit.Maui.Camera/Primitives/CameraFrame.shared.cs
```diff
@@ -0,0 +1,35 @@
...
+/// </summary>
+public sealed class CameraFrame
+{
+	/// <summary>
+	/// Initializes a new instance of the <see cref="CameraFrame"/> class.
+	/// </summary>
+	/// <param name="data">The raw frame data.</param>
+	/// <param name="width">The width of the frame.</param>
+	/// <param name="height">The height of the frame.</param>
+	public CameraFrame(byte[] data, int width, int height)
```
peruchali: @bijington Is there any format that these bytes will conform to? Is it rgba etc? or will it be platform specific?

bijington: I was wondering about this over the weekend. I think this will need to be more intelligent, I'll have a play this week

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.android.cs
```diff
@@ -0,0 +1,83 @@
...
+		public void Analyze(IImageProxy image)
+		{
+			var planes = image.GetPlanes();
+			if (planes.Length == 0)
+			{
+				image.Close();
+				return;
+			}
+
+			var buffer = planes[0].Buffer;
```
peruchali: @bijington I think this code will get only the Y plane (luminance) data, which will effectively just be grayscale. I think we should be getting 1 (U) and 2 (V) as well?

src/CommunityToolkit.Maui.Camera/Primitives/CameraFrame.shared.cs
peruchali: One option can be to add an enum for the Format for the common native formats (YUV*, etc), set it appropriately for different platforms and leave it up to the handler to process as they wish, reduces any perf overhead.

bijington: Yes I like this idea, or possibly provide a default converter and let the developer override, then we don't have to worry about support multiple formats?

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.android.cs
```diff
@@ -0,0 +1,83 @@
...
+			if (buffer is null)
+			{
+				image.Close();
+				return;
+			}
+
+			var data = new byte[buffer.Remaining()];
```
peruchali: @bijington Not sure if this is out of scope of this scenario, but allocating these byte arrays can end up in the LOH and cause GC slowdown with repeated allocations and releases (if it's anything like decoding video files on disk 😔).
Is it possible to augment FrameBasedCameraScenario to allow a consumer to optionally handle memory management?
For eg. FrameBasedCameraScenario can have a 
`protected virtual byte[] Allocate(int size) => new byte[size];`
which can be plumbed through in FrameAnalyzer.Analyze rather than directly allocating?
That way, a consumer can override FrameBasedCameraScenario to use ArrayPool/provide some preallocated/reusable buffer, etc or any other kind of memory management for these bytes (by coordinating with the use of CameraFrame) to avoid GC slowdown.
Didn't want to re-implement the entire FrameBasedCamera scenario on all platforms just for memory management from PlatformCameraScenario when all the other pieces exist...

bijington:
@peruchali it'll be great to hear your thoughts. I have made a couple of changes here based on your feedback:

- provide better memory management in the base scenario and allow developers to override the defaults
- provide an enum for common camera frame formats and make it possible for developers to provide custom conversions
- allow developers to set a preferred format to avoid having to convert post frame capture

I should add most of this is currently untested now but I am keen to iterate over the API and then we can get the underlying implementation working

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.android.cs
pictos: > For eg. FrameBasedCameraScenario can have a
protected virtual byte[] Allocate(int size) => new byte[size];
which can be plumbed through in FrameAnalyzer.Analyze rather than directly allocating?

I like the idea of giving extension points to our users. But it's better to have a good solution inside the library; most devs will shoot themselves in the foot trying to manage that. So we do provide a better solution, than `new byte [size]` and allow devs to implement their own if needed.

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.shared.cs
```diff
@@ -0,0 +1,68 @@
...
+	public virtual void OnFrameReceived(CameraFrame frame)
+	{
+	}
+
+	/// <summary>
+	/// Allocates a new byte array of the specified size.
+	/// </summary>
+	/// <param name="size">The size of the byte array to allocate.</param>
+	/// <returns>A new byte array of the specified size.</returns>
+	protected virtual byte[] Allocate(int size) => ArrayPool<byte>.Shared.Rent(size);
```
pictos, line 30: We can't do it, users may not know this is an `ArrayPool` array, and may not return it into the pool. So it can be worse than doing `new byte[size];`

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.shared.cs
```diff
@@ -0,0 +1,68 @@
...
+
+	/// <summary>
+	/// Frees the specified byte array.
+	/// </summary>
+	/// <param name="data">The byte array to free.</param>
+	protected virtual void Free(byte[] data) => ArrayPool<byte>.Shared.Return(data);
```
pictos, line 36: this is unsafe, if the returned array isn't an array from `ArrayPool`, this will throw an exception. We can't control if the user will reasign this array or something like. For example:

```cs
var data = myFrame.Allocate(1024);
data = data.Where( x => x != 0).ToArray(); // this will allocate a new array
myFrame.Free(data); //this will throw, because the array doesn't belong into the pool
```

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.shared.cs
pictos: Also, arrays from `ArrayPool` have their length as a power of 2 base. So if you ask for 1000, it will give back an array with a length of 1024. And it can be with garbage (when you return or rent an array into the pool it isn't zeroed by default), so users can't `foreach` over it.

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.android.cs
```diff
@@ -0,0 +1,137 @@
...
+using AndroidX.Camera.Core.ResolutionSelector;
+using Java.Util.Concurrent;
+
+namespace CommunityToolkit.Maui.Core;
+
+public abstract partial class FrameBasedCameraScenario
+{
+	ImageAnalysis? imageAnalysis;
+	
+	public override Task OnAttached(IExecutorService? cameraExecutor, ResolutionSelector? resolutionSelector)
```
pictos, line 12: I believe this should be `ValueTask` instead of `Task`

src/CommunityToolkit.Maui.Camera/Primitives/CameraScenario.shared.cs
```diff
@@ -0,0 +1,42 @@
+namespace CommunityToolkit.Maui.Core;
+
+/// <summary>
+/// Base class for output based processing for camera related activities.
+/// </summary>
+public abstract class CameraScenario : BindableObject, IDisposable
```
pictos, line 6: Before moving forward with this, would be good to see if it's possible to work with more than one scenario at once. How the Camera control would behave, and so on

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.windows.cs
```diff
@@ -0,0 +1,109 @@
...
+
+		var source = mediaCapture.FrameSources.Values.FirstOrDefault(s => s.Info.SourceKind == MediaFrameSourceKind.Color);
+		if (source is null)
+		{
+			return;
+		}
+
+		var preferredFormatName = PreferredFormat switch
+		{
+			CameraFrameFormat.Nv12 => "NV12",
```
pictos, line 25: would be better to create a Constant file and put those strings there and reuse it?

src/CommunityToolkit.Maui.Camera/Views/CameraView.shared.cs
```diff
@@ -12,6 +13,24 @@ public partial class CameraView : View, ICameraView, IDisposable
 {
 	static readonly BindablePropertyKey isAvailablePropertyKey =
 		BindableProperty.CreateReadOnly(nameof(IsAvailable), typeof(bool), typeof(CameraView), CameraViewDefaults.IsAvailable);
+	
+	internal static readonly BindablePropertyKey ScenariosPropertyKey = BindableProperty.CreateReadOnly(nameof(Scenarios), typeof(IList<CameraScenario>), typeof(CameraView), default(IList<CameraScenario>),
```
pictos, line 17: if users can't change the collection type, why not use the concrete type?

pictos (commented):
I've couple of more comments, but I'm not sure about this approach. Maybe I didn't fully understand it yet, but I feel the better would be to expose the `CameraManager` and let users customize the capabilities and handling there, like we have for `MediaManager`, this current approach adds so many levels of indirection that can be dangerous.

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.android.cs
peruchali: @pictos What better solution do you suggest? Anything more than `new byte[]` will place constraints on the usage, lifetime of the byte array - which seems strange out of the box.

src/CommunityToolkit.Maui.Camera/Primitives/FrameBasedCameraScenario.shared.cs
peruchali: Yes, and also the value of the required length of the byte[] is lost here and downstream (`bytes.Length` can be > than the required size). Having it out of the box will require adding a new member to CameraFrame containing the length, which makes it more complicated.

bijington:
> I've couple of more comments, but I'm not sure about this approach. Maybe I didn't fully understand it yet, but I feel the better would be to expose the `CameraManager` and let users customize the capabilities and handling there, like we have for `MediaManager`, this current approach adds so many levels of indirection that can be dangerous.

Thanks @pictos. I'm not stuck on this approach, just spiking it to see if it makes sense, so your feedback here is valuable! I'm also happy to go down the exposing of `CameraManager` route, as you say I've been exposing more levels here which is introducing much more to maintain now. 

My current goal is to find a way to solve the following:
- Provide platform specific frame processing
- Provide shared frame processing

I think we are closing to having those abilities but then we can evaluate and simplify how we actually expose it

src/CommunityToolkit.Maui.Camera/Primitives/CameraScenario.shared.cs
bijington: I agree, I've got this on my list to try out

peruchali:
@bijington  for: 
> * provide better memory management in the base scenario and allow developers to override the defaults

I'm not sure that the library out of the box can implement any efficient memory management (like ArrayPool) without at the same time introducing constraints on the usage and lifetime of the bytes of CameraFrame (for eg the user should be very careful with passing the bytes around rather than copying). The default should be `Allocate(int size) => new byte[size]` (we shouldn't need a `Free` method, that is up to the consumer). This way, out of the box, the user can use as they feel like without worrying about any dangling or corrupted memory. I feel like an example of good memory management using ArrayPool (or even just a single reused `byte[]` for simplicity) can belong in the samples/guidance by overriding the Scenario and usage of CameraFrame.

To "soften" the blow of the "default" out-of-the-box experience just being the humble `new byte[...]`, I think we should:
1. Add a property `public int SamplingRate` to FrameBasedCameraScenario (default - maybe `3`?)
    Add another override to the FrameBasedCameraScenario:
   ```
   private int _samplingCounter = 0;
   protected virtual bool ShouldProcessFrame(...) => (++_samplingCounter % SamplingRate) == 0;
   ```   
   and call this from `FrameAnalyzer.Analyze`. If `ShouldProcessFrame` is false, just return right away without allocating or returning a CameraFrame, reducing allocations to 1/3rd.
   For practical purposes like a bar code scanner, 30 fps is overkill and sampling will ease some of the pressure.
   A consumer can override this to provide more custom or variable sampling.

2. Add a property to FrameBasedCameraScenario to specify the `public Size Resolution`. It looks like this can be plumbed through in Android:
   ```
   var resolutionStrategy = new ResolutionStrategy(
        targetSize, 
        ResolutionStrategy.RuleClosestLowerThenHigher
    );

    var customResolutionSelector = new ResolutionSelector.Builder()
        .SetResolutionStrategy(resolutionStrategy)
        .Build();

    imageAnalysis = new ImageAnalysis.Builder()
        .SetResolutionSelector(customResolutionSelector) ....
   ```
   Setting a cheap default will ensure that the allocations don't end up in the LOH, and so should work decently out of the box.

Other requests:
* CameraFrame can have a Timestamp field (`image.ImageInfo.Timestamp` in Android). This will be useful for any kind of real-time tracking algorithm, etc.

bijington:
@peruchali thanks. Yes I don't think I had fully considered my idea last night 😂. It might be safest to just revert to something very simple and let the user decide

pictos:
So, using a `ArrayPool` will not prevent the LOH, also I don't know if that's a hot path or not, imagine that during a scan it triggers the `CameraFrame` 100 times, and the max array of a specific size allocated are 50, so it will end up creating new 50 arrays and throwing away after the usage. But still using the `ArrayPool` will mitigate the GC pressure.

That said, for using an `ArrayPool` directly the code structure may change, so the lib handles with the array lifetime and between that let users use it and handle that. That can be done easily, from what I saw, due to the usage of `abstract class` so we can use the templated method approach. Something like

```cs
abstract class FrameBasedCameraScenario
{
    protected abstract WorkWithData(ReadOnlySpan<byte> data); // Or create another model just to use `ReadOnlySpan`

    public void OnFrameReceived(CameraFrame frame)
    {
        try
        {
            var data = frame.Data.AsSpan(); //We know this one is from ArrayPool
            WorkWithData(data);
        }
        finally
        {
            ArrayPool<byte>.Shared.Return(data);
        }    
    }
}
```

Another approach is to see there's a way to convert from native platform collection into arrays directly, like a cast. For example the java word uses `Java.Nio.ByteBuffer` so the contiguous space in memory already exist, would be just a matter of shape it into an array or Span. Right now we duplicate that data.

Even another approach is to use another kind of data collection to represent that data, so instead of a gigantic array it can be represented with segmented arrays that are made of small arrays (that will not land on LOH), I know that roslyn team has this kind of collection. And create a pool to reuse those arrays.

The next steps of optimizations would be related with the approach that we choose.

Issue #3271, opened 2026-07-23
[BUG] COMException (0x8007007E): Unable to load resource dll Microsoft.WindowsAppRuntime.Insights.Resource.dll on Windows (.NET 10 / MAUI)


Transis-Felipe:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Did you read the "Reporting a bug" section on Contributing file?

- [x] I have read the "Reporting a bug" section on Contributing file: https://github.com/CommunityToolkit/Maui/blob/main/CONTRIBUTING.md#reporting-a-bug

### Current Behavior

<img width="3211" height="1396" alt="Image" src="https://github.com/user-attachments/assets/6d24dd6a-53f6-43c7-b9c3-c5da603d360f" />

### Description
An unhandled `COMException` occurs on Windows during application startup or lifecycle events (`OnLaunched`). The exception indicates that `Microsoft.WindowsAppRuntime.Insights.Resource.dll` could not be loaded (`ERROR_MOD_NOT_FOUND`).

### Exception Details
* **Exception Type:** `System.Runtime.InteropServices.COMException`
* **HResult:** `0x8007007E` (The specified module could not be found)
* **Message:** `Unable to load resource dll. Microsoft.WindowsAppRuntime.Insights.Resource.dll` / `Não foi possível encontrar o módulo especificado.`
* **Location:** `CommunityToolkit.Maui.Options.shared.cs` in `Application__LaunchActivatedEventLogs` / `Microsoft.UI.Xaml.Application.OnLaunched`

### Expected Behavior

No error

This error start with version 14.2.0

### Steps To Reproduce

Update the community toolkit to latest version and start a project with windows

### Link to public reproduction project repository

N/A

### Environment

```markdown
### Environment
- .NET MAUI CommunityToolkit:15.0.0
- .NET MAUI:10.0.90
* **Framework:** .NET 10.0.10
* **OS:** Windows 11 25H2 build
```

TheCodeTraveler:
We're seeing a similar error in our CI/CD pipeline: https://github.com/CommunityToolkit/Maui/actions/runs/30027765216/job/89276526975#step:9:666

I can not confirm at this time whether or not it is related to this Issue.

duindain:
I was just about to raise the same issue for 14.2.2 and 15.0.0
I was able to trigger the error by registering for snackbar messages on the windows platform and if i remove that line it runs successfully

[...]

PR #3276, opened 2026-08-02
Enforce Snackbar usage only in packaged Windows apps

ne0rrmatrix:
## Description of Change

Fixes a regression introduced in `CommunityToolkit.Maui` v15.0.0 where calling `SetShouldEnableSnackbarOnWindows(true)` (or otherwise initializing the toolkit's Windows `Snackbar`/`Toast` support) in an **unpackaged** Windows app throws a `System.Runtime.InteropServices.COMException (0x80670016)`:

```
Package dependency criteria could not be resolved.
```

or

```
Unable to load resource dll. Microsoft.WindowsAppRuntime.Insights.Resource.dll
```

### Root Cause

`Options.SetShouldEnableSnackbarOnWindows(true)` unconditionally registered the app with the Windows App SDK's `Microsoft.Windows.AppNotifications.AppNotificationManager`, which internally resolves a Windows App Runtime package dependency via the dynamic dependency (Bootstrap) APIs. This registration/resolution only succeeds for **packaged (MSIX)** applications. When called from an **unpackaged** app, the underlying WinRT call (`IAppNotificationManagerMethods.Register`) fails with a `COMException`, since the required package dependency and its resource DLL (`Microsoft.WindowsAppRuntime.Insights.Resource.dll`) cannot be resolved outside of a packaged identity context.

Previously (v14.0.1) this registration path was not exercised for unpackaged apps, so the issue was not present until v15.0.0.

### Fix

- `Options.SetShouldEnableSnackbarOnWindows` now detects whether the app is running as a **packaged** app (`IsPackagedApp()`, via reflection over `Windows.ApplicationModel.Package.Current`) before calling `AppNotificationManager.Default.Register()`/`Unregister()`. For unpackaged apps, registration is skipped entirely and a `Trace.WriteLine` diagnostic message is emitted instead of allowing the unmanaged `COMException` to propagate.
- `Snackbar`'s constructor now throws a clear, actionable `InvalidOperationException` (instead of letting the unmanaged/WinRT `COMException` surface) when:
  - `SetShouldEnableSnackbarOnWindows(true)` was not called, or
  - the app is not packaged (MSIX) — since `Snackbar` on Windows depends on the Windows App SDK `AppNotificationManager` API, which requires a packaged app identity.
- Both exceptions include a `HelpLink` pointing to the `Snackbar` documentation describing platform-specific initialization requirements.

### Impact

- Unpackaged Windows MAUI apps no longer crash with an opaque `COMException`/missing resource DLL error during startup or when calling `SetShouldEnableSnackbarOnWindows(true)`.
- Developers now get a clear, descriptive `InvalidOperationException` explaining that `Snackbar` requires a packaged (MSIX) Windows app, with a link to documentation.
- Packaged Windows apps are unaffected; `AppNotificationManager` registration continues to work as before.

## Additional Fix: Windows PRI Build Error (`PRI175`/`PRI277`)

While validating the above change, the Windows build of `CommunityToolkit.Maui.Sample` failed with:

```
WINAPPSDKGENERATEPROJECTPRIFILE : error PRI175: 0x80073b0f - Processing Resources failed with error: Duplicate Entry.
WINAPPSDKGENERATEPROJECTPRIFILE : error PRI277: 0x80073b0f - Conflicting values for resource 'Files/Microsoft.Maui/Platform/Windows/Styles/Resources.xbf'
```

### Root Cause

`CommunityToolkit.Maui.Sample.csproj` referenced `Microsoft.Maui.Controls` with a wildcard version:

```xml
<PackageReference Include="Microsoft.Maui.Controls" Version="*" />
```

The wildcard resolved to **10.0.90**, while every other MAUI package in the repo is pinned to **10.0.60** via `$(MauiPackageVersion)` in `Directory.Build.props` (including `Microsoft.Maui.Controls.Maps` and the MAUI assemblies pulled in transitively by the `CommunityToolkit.Maui.*` project references). The build output confirmed the conflict:

```
microsoft.maui.controls.core\10.0.90\...\Microsoft.Maui.Controls.dll
microsoft.maui.controls.maps\10.0.60\...\Microsoft.Maui.Controls.Maps.dll
```

Two different MAUI versions each ship the framework resource `Files/Microsoft.Maui/Platform/Windows/Styles/Resources.xbf` with different content. When the Windows App SDK PRI generator merges them, it finds the same resource path with conflicting values, producing `PRI175`/`PRI277`.

### Fix

Pinned `Microsoft.Maui.Controls` to the repo-wide version so all MAUI packages align on 10.0.60:

```xml
<PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiPackageVersion)" />
```

> Note: The `<WindowsSdkPackageVersion>10.0.19041.56</WindowsSdkPackageVersion>` override in the sample project was initially suspected, but removing it surfaced a different error (`MVVMTKCFG0003`) — CommunityToolkit.Mvvm 8.4.2 requires that SDK version. That line is correct and was left in place; the actual culprit was the wildcard MAUI version.

 ### Linked Issues ###

Fixes:
-  [dotnet/maui#36983](https://github.com/dotnet/maui/issues/36983)
- [CommunityToolkit#3271](https://github.com/CommunityToolkit/Maui/issues/3271)
 ### PR Checklist ###
 
 - [x] Documentation created or updated: https://github.com/MicrosoftDocs/CommunityToolkit/pull/655

Unpackaged Sample app in Windows now builds and runs in Release mode.

### DIFF
--- a/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
+++ b/samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
@@ -59,7 +59,7 @@
     <!-- Custom Fonts -->
     <MauiFont Include="Resources\Fonts\*" />
 
-    <PackageReference Include="Microsoft.Maui.Controls" Version="*" />
+    <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiPackageVersion)" />
     <PackageReference Include="CommunityToolkit.Maui.Markup" Version="7.0.1" />
     <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.2" />
     <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="10.5.0" />
--- a/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
+++ b/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
@@ -177,9 +177,7 @@ public partial class Snackbar : ISnackbar
 #if WINDOWS
 	static bool IsPackagedApp()
 	{
-		var packageType = Type.GetType("Windows.ApplicationModel.Package, Windows, ContentType=WindowsRuntime");
-		var currentProperty = packageType?.GetProperty("Current", System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Static);
-		return currentProperty?.GetValue(null) is not null;
+		return global::Windows.ApplicationModel.Package.Current is not null;
 	}
 #endif
 
--- a/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
+++ b/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
@@ -177,15 +177,9 @@ public partial class Snackbar : ISnackbar
 #if WINDOWS
 	static bool IsPackagedApp()
 	{
-		try
-		{
-			_ = global::Windows.ApplicationModel.Package.Current;
-			return true;
-		}
-		catch
-		{
-			return false;
-		}
+		var packageType = Type.GetType("Windows.ApplicationModel.Package, Windows, ContentType=WindowsRuntime");
+		var currentProperty = packageType?.GetProperty("Current", System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Static);
+		return currentProperty?.GetValue(null) is not null;
 	}
 #endif
 
--- a/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
+++ b/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
@@ -179,9 +179,8 @@ public partial class Snackbar : ISnackbar
 	{
 		try
 		{
-			var packageType = Type.GetType("Windows.ApplicationModel.Package, Windows, ContentType=WindowsRuntime");
-			var currentProperty = packageType?.GetProperty("Current", System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Static);
-			return currentProperty?.GetValue(null) is not null;
+			_ = global::Windows.ApplicationModel.Package.Current;
+			return true;
 		}
 		catch
 		{
--- a/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
+++ b/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
@@ -1,4 +1,5 @@
 using CommunityToolkit.Maui.Core;
+using CommunityToolkit.Maui.Extensions;
 
 namespace CommunityToolkit.Maui.Alerts;
 
@@ -34,14 +35,13 @@ public partial class Snackbar : ISnackbar
 
 	bool isDisposed;
 	WeakReference<IView>? weakView;
-
 	/// <summary>
 	/// Initializes a new instance of <see cref="Snackbar"/>
 	/// </summary>
 	public Snackbar()
 	{
 #if WINDOWS
-		if (!IsPackagedApp())
+		if (!AppPackageExtensions.IsPackagedApp)
 		{
 			throw new InvalidOperationException($"{nameof(Snackbar)} is not supported on unpackaged Windows apps. {nameof(Snackbar)} requires the app to be packaged (MSIX) because it uses the Windows App SDK AppNotificationManager API. See https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar for more information.")
 			{
@@ -173,19 +173,6 @@ public partial class Snackbar : ISnackbar
 
 	internal static TimeSpan GetDefaultTimeSpan() => TimeSpan.FromSeconds(3);
 
-#if WINDOWS
-	static bool IsPackagedApp()
-	{
-		try
-		{
-			return global::Windows.ApplicationModel.Package.Current is not null;
-		}
-		catch
-		{
-			return false;
-		}
-	}
-#endif
 
 	void OnShown()
 	{
--- a/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
+++ b/src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
@@ -41,17 +41,16 @@ public partial class Snackbar : ISnackbar
 	public Snackbar()
 	{
 #if WINDOWS
-		if (!Options.ShouldEnableSnackbarOnWindows)
+		if (!IsPackagedApp())
 		{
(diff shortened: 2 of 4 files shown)


src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
```diff
@@ -48,6 +48,14 @@ public Snackbar()
...
 			};
 		}
+
+		if (!IsPackagedApp())
+		{
+			throw new InvalidOperationException($"{nameof(Snackbar)} is not supported on unpackaged Windows apps. {nameof(Snackbar)} requires the app to be packaged (MSIX) because it uses the Windows App SDK AppNotificationManager API. See https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar for more information.")
+			{
+				HelpLink = "https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar"
+			};
+		}
```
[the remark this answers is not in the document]
ne0rrmatrix: Test suite only tests .Net 10.0 and does not have support for device specific code for platforms directly.

src/CommunityToolkit.Maui/Options.shared.cs
```diff
@@ -109,6 +117,19 @@ static void OnSnackbarNotificationInvoked(Microsoft.Windows.AppNotifications.App
 			{
 				Alerts.Snackbar.HandleSnackbarAction(args);
 			}
+
+			static bool IsPackagedApp()
```
TheCodeTraveler: Is it possible to refactor this so that it doesn't require a try/catch block? Using a try/catch block is an expensive operation and if we can avoid it, we should. But if we can't, that's ok too.

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
```diff
@@ -166,6 +174,21 @@ public void Dispose()
 
 	internal static TimeSpan GetDefaultTimeSpan() => TimeSpan.FromSeconds(3);
 
+#if WINDOWS
+	static bool IsPackagedApp()
```
TheCodeTraveler: Is it possible to refactor this method so that it doesn't require a try/catch block? Using a try/catch block is an expensive operation and if we can avoid it, we should. But if we can't, that's ok too.

samples/CommunityToolkit.Maui.Sample/CommunityToolkit.Maui.Sample.csproj
```diff
@@ -59,7 +59,7 @@
     <!-- Custom Fonts -->
     <MauiFont Include="Resources\Fonts\*" />
 
-    <PackageReference Include="Microsoft.Maui.Controls" Version="*" />
+    <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiPackageVersion)" />
```
TheCodeTraveler, line 62: Is this necessary?

We keep `Microsoft.Maui.Controls` version as `*` to ensure the sample app always uses the most recent MuGet release. This allows us to run + test the sample app on the latest version to avoid breaking changes in our library which runs on an older version of `Microsoft.Maui.Controls`.

ne0rrmatrix: The wildcard * caused the sample app to resolve Microsoft.Maui.Controls to the latest NuGet version (10.0.90), while every other MAUI package in the repo — including Microsoft.Maui.Controls.Maps, Microsoft.Maui.Core, and the MAUI assemblies pulled in transitively by the CommunityToolkit.Maui.* project references — was pinned to 10.0.60 via $(MauiPackageVersion).

This version mismatch broke the Windows build because both MAUI 10.0.60 and 10.0.90 ship the same framework resource file (Files/Microsoft.Maui/Platform/Windows/Styles/Resources.xbf) but with different content. When the Windows App SDK PRI generator merged resources from both versions, it found the same path with conflicting values, producing:

By pinning the sample to $(MauiPackageVersion) (10.0.60), all MAUI packages resolve to the same version, eliminating the duplicate resource conflict.

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
ne0rrmatrix: I removed try/catch and replaced it with a reflection call

src/CommunityToolkit.Maui/Options.shared.cs
ne0rrmatrix: I removed try/catch and replaced it with a reflection call

ne0rrmatrix: I changed it to not use reflection now and it returns true for non null

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
ne0rrmatrix: I changed it to not use reflection now and it returns true for non null

TheCodeTraveler:
Thanks James! Could you open up a Docs PR to note this new Snackbar/Toast behavior for unpackaged apps?

ne0rrmatrix:
> Thanks James! Could you open up a Docs PR to note this new Snackbar/Toast behavior for unpackaged apps?

Documentation created! https://github.com/MicrosoftDocs/CommunityToolkit/pull/655

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
```diff
@@ -48,6 +48,14 @@ public Snackbar()
 				HelpLink = "https://learn.microsoft.com/dotnet/communitytoolkit/maui/alerts/snackbar"
 			};
 		}
+
+		if (!IsPackagedApp())
```
VladislavAntonyuk: Probably should be the first check before configurations

ne0rmatrix:
I reverted to using try catch again. Otherwise we do not control what the throw message is.

src/CommunityToolkit.Maui/Alerts/Snackbar/Snackbar.shared.cs
ne0rrmatrix: Updated as suggested

VladislavAntonyuk:
What exception do you expect from IsPackagedApp()? When may the exception occur?

ne0rrmatrix:
> What exception do you expect from IsPackagedApp()? When may the exception occur?

When I run that function it checks to see if the app is packaged by referencing an assembly. If that assembly is present it returns true. If it is not it will 100 percent throw an exception with a reference to a missing assembly. That assembly is the Windows App SDK which is not present in an Unpackaged app. I did previously have a very neat way of avoiding this and did not need to use try catch but with internals visible gone we cannot use the previous method.

If you have a way to check if an assembly exists without throwing an error or a way to check if app is packaged that does not use MS internals I would appreciate your input.

ne0rrmatrix:
@VladislavAntonyuk and @TheCodeTraveler I updated the PR to use an already existing Extension method that I copied from `MediaElement` that detects if app is packaged. I wrote the original and Brandon updated it to be compliant a while ago. It is currently only being used in one other place. I copied a class to Extensions. I switched to using it since it already exists and has been in toolkit for a while.

src/CommunityToolkit.Maui/Extensions/AppPackageExtensions.windows.cs
```diff
@@ -0,0 +1,43 @@
+using Windows.ApplicationModel;
+
+namespace CommunityToolkit.Maui.Extensions;
+
+// Since MediaElement can't access .NET MAUI internals we have to copy this code here
```
VladislavAntonyuk, line 5: please update comment since it is not only for Media Element

TheCodeTraveler:
Closing in favor of https://github.com/CommunityToolkit/Maui/pull/3301

